PR #4, opened 2016-12-22
[ROCKETMQ-10] Null check is not needed -- can't be null when reaches this point.

nottyjay:
Validators.checkGroup(String group) will ensure group is not null

JIRA Issue: https://issues.apache.org/jira/browse/ROCKETMQ-10


vongosling:
@nottyjay  Thanks for your PR, Could you paste some unit test for these question :-) . please see our contributing checklist, https://github.com/apache/incubator-rocketmq/blob/master/CONTRIBUTING.md

nottyjay:
I see, the best unit test is `com.alibaba.rocketmq.example.quickstart.Producer` which in Model `rocketmq-example`.
Delete these codes because in `Validators.checkGroup(String group)`，it calls `UtilAll.isBlank(group)` to check group. So, in `DefaultMQProducerImpl`， null check is not needed.

vongosling:
Could you abide our PR process(modify your PR subject and mention your jira address in the description), please follow this PR. https://github.com/apache/incubator-rocketmq/pull/5

PR #18, opened 2016-12-27
[ROCKETMQ-20] Default 'consumeFromWhere' is inconsist with example's

dongeforever:
The example pull consumer has a line code like this:
consumer.setConsumeFromWhere(ConsumeFromWhere.CONSUME_FROM_FIRST_OFFSET)
This is not necessary for a minimum viable pull consumer.
If this line is important, then it should be the default value.


dongeforever:
https://issues.apache.org/jira/browse/ROCKETMQ-20?filter=-2

lollipopjin:
@dongeforever Thanks for you PR, we will add some notes in the example to eliminate confuse. :-)

PR #19, opened 2016-12-28
[ROCKETMQ-14] rocketmq-14 unit test updated

Jaskey:
A runnable unit test update.

Create a mock server before test.


vongosling:
IMO, we could close this pr without any more works @lollipopjin 

thanks @Jaskey

Jaskey:
@vongosling Is there any problem on this unit test code?

PR #20, opened 2016-12-28
[ROCKETMQ-23] MappedFileQueue#flush should return true when flushing …

shroman:
…is successful.

Jira issue: https://issues.apache.org/jira/browse/ROCKETMQ-23


shroman:
@vongosling Does `RETRY_TIMES_OVER` here means retries on failures, or the number of times to flush on success? I think the former, but need to be sure I understand the intentions of the person who wrote the code.

zhouxinyu:
Hi, shroman

Could you please resolve conflicts in this pr?

And we will remain cautious about the changes of rocketmq-store module, so this pr may be won't merge quickly.

Thanks.

shroman:
Hi, the github says "This branch has no conflicts with the base branch"
I think conflicts are resolved.

Reviewing it by several committers is already being cautious, isn't it? ;)
This issue looks much like a pretty nasty bug.

Anyway, looking forward to your reviews ;)

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
```diff
@@ -39,11 +39,15 @@
 
     private final int mappedFileSize;
 
-    private final CopyOnWriteArrayList<MappedFile> mappedFiles = new CopyOnWriteArrayList<MappedFile>();
+    private final CopyOnWriteArrayList<MappedFile> mappedFiles = new CopyOnWriteArrayList<>();
 
     private final AllocateMappedFileService allocateMappedFileService;
 
-    private long flushedWhere = 0;
```
zhouxinyu: Consider the abuse of `offset` in rocketmq-store, so we use `flushedWhere` and `committedWhere`.

`flushedWhere` means the offset of all mappedFile, while `offset` means the offset of one mappedFile in general.

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
```diff
@@ -191,7 +199,7 @@ public long howMuchFallBehind() {
         return 0;
     }
 
-    public MappedFile getLastMappedFile(final long startOffset, boolean needCreate) {
+    private MappedFile getLastMappedFile(final long startOffset, boolean needCreate) {
```
zhouxinyu: Yes,  some methods in `MappedFileQueue` may have their access modifier narrowed down in grammatical, but I think these methods should be `public`.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -964,9 +964,11 @@ public void run() {
             boolean result = false;
             for (int i = 0; i < RETRY_TIMES_OVER && !result; i++) {
                 result = CommitLog.this.mappedFileQueue.flush(0);
-                CommitLog.log.info(this.getServiceName() + " service shutdown, retry " + (i + 1) + " times " + (result ? "OK" : "Not OK"));
```
zhouxinyu: Just keep it to record the shutdown retry times, it's useful.

shroman: You are right, I'll add it to the log message.

zhouxinyu:
please @vongosling @lollipopjin review this PR.

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
shroman: I see, you need it to keep public as other `getLastMappedFile(...)` methods. Thanks for pointing to it!

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -964,9 +964,11 @@ public void run() {
             boolean result = false;
             for (int i = 0; i < RETRY_TIMES_OVER && !result; i++) {
```
zhouxinyu: Consider `result = flush(0)`, the false result means there may be more data need to flush, so retry. Otherwise, the true result means all data are flushed, the broker can exit safely.

So if you change `result`, also please change this `for` loop.

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
shroman: yes, I understand this, but `flushedWhere` doesn't sound good. File's offset is the file's offset, but this offset is the offset withing the queue of all files ;) Should we rename it to `flushedPosition`? @zhouxinyu What do you think?

zhouxinyu: Sounds great. If do rename, rename `committedWhere ` together.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
shroman: right, that's what I also thought and ask above about `RETRY_TIMES_OVER`. Thank you for clarifying this.

How about this (flushing until data is available and `MAX_FLUSH_NUM` (which can be quite large) is reached, and log warnings when cannot flush)?

```java
for (int i = 0; i < MAX_FLUSH_NUM && remainHowManyDataToFlush() > 0; i++) {
     if (!CommitLog.this.mappedFileQueue.flush(0))
          CommitLog.log.warn(this.getServiceName() + " service shutdown, failed flush retried "
                        + (i + 1) + " times");
}
CommitLog.log.info(this.getServiceName() + " service shutdown, all data is flushed successfully");
```

shroman:
@zhouxinyu I made changes according to your comments, and added more tests.
In {{CommitLog.FlushRealTimeService#run()}}, we will try to flush as long as the data is available or {{MAX_FLUSH_COMMIT_NUM}} is not reached, logging when flushing fails. It is almost same as before (checking the return value of {{flush()}}) but checking if looks neat to me. What do you think?

vongosling:
We must keep cautious about storage polish, although some minor rename. IMO, Could we optimize here in the 4.1.0  or more later version :-)

shroman:
Ok, guys, I set the fix version of the JIRA issue to 4.1.0 then 👍 
@vongosling what are your recommendations on how to treat the storage module? Should I create a JIRA issue if I find a problem, but not work on it? Or should I create a PR, but it would be merged only in 4.1.0? Just curious, why merging the fixes to it is better for later versions?

vongosling:
@shroman 

what are your recommendations on how to treat the storage module? Should I create a JIRA issue if I find a problem, but not work on it?

yeah, if you find some bugs or improvement, you can raise it in JIRA, just later work on it :-)  we could start a thread in the email list, to discuss how to optimize here in the latter version.

dongeforever:
@shroman could you please resolve the conflicts?
IMO, the return value of flush() and commit(), means whether there is data left to be flushed or committed, that is OK.
if you'd like to polish the semantic of the returned value, may you do the same polishing for commit() too, just to keep the same semantic for similar methods.

shroman:
And why is it closed?

dongeforever:
@shroman It seems that last cleaning, for too long-time no-resolved prs,  included this one.
If you feel it is necessary to reopen this PR, please feel free to do it.
you may also could create a new PR or launch new discussion in the mailing list.

PR #22, opened 2016-12-29
[ROCKETMQ-25] Possible data race while query msg by key

lizhanhui:
Refer to JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-25


lollipopjin:
@lizhanhui Could you please verify mentioned problem, no matter using unit or other ways. At last, please fix the conflict problem.

PR #31, opened 2017-01-07
[ROCKETMQ-33] CPU Occupy 100%

qinliujie:
https://issues.apache.org/jira/browse/ROCKETMQ-33

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1045,7 +1045,7 @@ public class CommitLog {
 
             while (!this.isStopped()) {
                 try {
-                    this.waitForRunning(0);
+                    this.waitForRunning(1000);
                     this.doCommit();
                 } catch (Exception e) {
                     CommitLog.log.warn(this.getServiceName() + " service has exception. ", e);
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1045,7 +1045,7 @@ public class CommitLog {
 
             while (!this.isStopped()) {
                 try {
-                    this.waitForRunning(1000);
+                    this.waitForRunning(10);
                     this.doCommit();
                 } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


qinliujie:
Sorry,I wrote the wrong issue number,it shoud be [ROCKETMQ-33] CPU Occupy 100%

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1045,7 +1045,7 @@ public void run() {
 
             while (!this.isStopped()) {
                 try {
-                    this.waitForRunning(0);
+                    this.waitForRunning(1000);
```
zhouxinyu: `GroupCommitService` will be used when broker is in `SYNC_FLUSH`, so the flushCommitLogService need always `doCommit` without a break.

And this won't cause 100% CPU usage unless your server only has one cpu core.

BTW, how about use `ASYNC_FLUSH` if you mind.

zhouxinyu: I also think `waitForRunning(0)` is too rigorous, but `waitForRunning(1000)` is adventurous while the `GroupCommitService` may miss some notify signals.

How about set a smaller wait time, like 10ms? what's your opinion ? @qinliujie

qinliujie:
When I use ASYNC_MASTER,there was the same problem,Because  HAService also has `this.waitForRunning(0)`,whether SYNC_MASTER or not.

When you said that the `GroupCommitService` may miss some notify signals,I think it's right.So I think wait time to be 10ms is acceptable.

vintagewang:
review ok.

zhouxinyu:
Hi, @qinliujie  This PR will be merged soon, but next time please send PR from branch ROCKETMQ-33 instead of master.

qinliujie:
@zhouxinyu OK.

Ah39:
waitForRunning(0) is call wait(0). wait(0)  is wait forever util somenoe notify the Thread

Jaskey:
I agreee with @Ah39  , 0ms is wait forever, which should not cause high cpu problem, any other justice reason to modify it to 10ms?

qinliujie:
old version use the Object wait() method,but it had changed to countdown latch in new version,please notice that @Jaskey

PR #34, opened 2017-01-11
[ROCKETMQ-39]  avoid duplicated codes.

wu-sheng:
ref issue: https://issues.apache.org/jira/browse/ROCKETMQ-39

### DIFF
--- a/filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
+++ b/filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
@@ -23,7 +23,6 @@ import java.io.FileInputStream;
 import java.io.InputStream;
 import java.util.Properties;
 import java.util.concurrent.Callable;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
--- a/filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
+++ b/filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
@@ -50,7 +50,8 @@ public class FiltersrvStartup {
 
         try {
             controller.start();
-        } catch (Exception e) {
+        }
+        catch (Exception e) {
             e.printStackTrace();
             System.exit(-1);
         }
@@ -122,7 +123,7 @@ public class FiltersrvStartup {
                 System.exit(-2);
             }
 
-            LoggerContext lc = (LoggerContext) LoggerFactory.getILoggerFactory();
+            LoggerContext lc = (LoggerContext)LoggerFactory.getILoggerFactory();
             JoranConfigurator configurator = new JoranConfigurator();
             configurator.setContext(lc);
             lc.reset();
@@ -146,7 +147,8 @@ public class FiltersrvStartup {
             }));
 
             return controller;
-        } catch (Throwable e) {
+        }
+        catch (Throwable e) {
             e.printStackTrace();
             System.exit(-1);
         }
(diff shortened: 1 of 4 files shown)


lizhanhui:
The extracted shutdown hook misses required license header and causes check-style failure. Please fix it and double check if there are more similar code style issues.

You may refer [here](http://rocketmq.incubator.apache.org/docs/code-guidelines/) in terms of code style.

vongosling:
@wu-sheng Could you modify your PR , following our codestyle and PR guide[1] :-)

[1]http://rocketmq.incubator.apache.org/docs/code-guidelines/

wu-sheng:
@vongosling , fixed and done. Hoping `style-check` works well next time. 😭

filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
```diff
@@ -135,27 +137,17 @@ public static FiltersrvController createController(String[] args) {
...
-                        log.info("shutdown hook was invoked, " + this.shutdownTimes.incrementAndGet());
-                        if (!this.hasShutdown) {
-                            this.hasShutdown = true;
-                            long begineTime = System.currentTimeMillis();
-                            controller.shutdown();
-                            long consumingTimeTotal = System.currentTimeMillis() - begineTime;
-                            log.info("shutdown hook over, consuming time total(ms): " + consumingTimeTotal);
-                        }
-                    }
+                public Object call() throws Exception {
```
WillemJiang: If you don't want to return any thing here , you could define the method just like this.
    public Void call() throws Exception {
    }

wu-sheng: @WillemJiang done.

zhouxinyu:
It could be better if we add some unit tests for this PR and please remove the author info.

filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
```diff
@@ -139,7 +139,7 @@ public static FiltersrvController createController(String[] args) {
 
             Runtime.getRuntime().addShutdownHook(new ShutdownHookThread(log, new Callable() {
                 @Override
-                public Object call() throws Exception {
+                public Void call() throws Exception {
                     controller.shutdown();
                     return null;
```
WillemJiang: When use the Void as the return signature, you don't need to use return in the function.  
Please remove the line of 144.

namesrv/src/main/java/org/apache/rocketmq/namesrv/NamesrvStartup.java
```diff
@@ -123,7 +123,7 @@ public static NamesrvController main0(String[] args) {
 
             Runtime.getRuntime().addShutdownHook(new ShutdownHookThread(log, new Callable() {
                 @Override
-                public Object call() throws Exception {
+                public Void call() throws Exception {
                     controller.shutdown();
                     return null;
```
WillemJiang: This line should be removed.

filtersrv/src/main/java/org/apache/rocketmq/filtersrv/FiltersrvStartup.java
wu-sheng: @WillemJiang, Are you sure about no need `return null`? I have run a `mvn package` on local, and you can see, `return statement` is required.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project rocketmq-namesrv: Compilation failure
[ERROR] /Users/wusheng/Documents/code/github/incubator-rocketmq-fork/namesrv/src/main/java/org/apache/rocketmq/namesrv/NamesrvStartup.java:[128,17] 缺少返回语句
[ERROR] -> [Help 1]
```

WillemJiang: @wu-sheng ,  oh, it's my mistake, I just wrote a code and the return statement need to be kept.
As I'm not a big fan of null,  I always want to kill it if possible.

wu-sheng: +1 on `kill return`, sadly, it's not working.

dongeforever:
@wu-sheng this PR has been merged. you may close the PR.

PR #35, opened 2017-01-11
[ROCKETMQ-42] Output is not friendly, when Main failed.

wu-sheng:
ref issue: https://issues.apache.org/jira/browse/ROCKETMQ-42

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,5 @@
-RocketMQ
-Copyright 2016 Alibaba Group.
+Apache RocketMQ (incubating)
+Copyright 2016 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
@@ -180,6 +180,7 @@ public class BrokerStartup {
             }
 
             messageStoreConfig.setHaListenPort(nettyServerConfig.getListenPort() + 1);
+
             LoggerContext lc = (LoggerContext)LoggerFactory.getILoggerFactory();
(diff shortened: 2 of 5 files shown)


wu-sheng:
Whether the pull request be merged or not, the project team should close it without anything.

Using a commit to close, only comment on "won't fix", seem not regular for outside contributors. More communications should stay on GitHub, rather than your inside team, as least have a discussion, and then have a conclusion. In **Apache Software Foundation**, `benevolent dictator` is not welcome, I think you should pay attention to this.

This is only a suggestion, but I think this is important. btw @vongosling @WillemJiang

zhouxinyu:
Hi @wu-sheng , I am sorry for this rash close~

Actually, as @shroman mentioned in this [issue](https://github.com/apache/incubator-rocketmq/pull/3), we also think avoid using `e.printStackTrace()` or `system.out` is a better way, and there are some discussions in this thread.

Finally, thanks for this suggestion, we will have more discussions in GitHub issue before closing a PR.

wu-sheng:
I have known about @shroman 's issue. And I am sure he is right about that log is better than `e.printStackTrace()` or `system.out`. And here, there is a special scene.

Let's focus on `FiltersrvStartup.java`'s change logs. This `system.out` is unavoidable.
* `try-catch` block starts at line 74
*  Field `log` is initialized at line 128

If codes occur exception between these two lines, certainly be `catch`ed, and trigger `NullPointException` in catch.

Other changes like this one, this is why I only change these classes, as so many `System.out`s are used in project.

You see, this is also why we need discuss before merge or close. 😊

wu-sheng:
Maybe, we should reopen this pull request, and discuss more details.

wu-sheng:
@vongosling, your idea?

wu-sheng:
I have already explain it before. Generally speaking, `log` is not initialized.

PR #36, opened 2017-01-11
[ROCKETMQ-43] Remove NoWhitespaceBefore module in “rmq_checkstyle.xml”

wu-sheng:
There are both NoWhitespaceBefore and WhitespaceAfter modules in checkstyle.xml, like this:
<module name="NoWhitespaceBefore"/>
<module name="WhitespaceAfter"/>

ref https://issues.apache.org/jira/browse/ROCKETMQ-43

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/MQAdminStartup.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/MQAdminStartup.java
@@ -100,7 +100,8 @@ public class MQAdminStartup {
                             if (options != null) {
                                 ServerUtil.printCommandLineHelp("mqadmin " + cmd.commandName(), options);
                             }
-                        } else {
+                        }
+                        else {
                             System.out.printf("The sub command \'" + args[1] + "\' not exist.%n");
                         }
                         break;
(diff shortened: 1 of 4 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/MQAdminStartup.java
```diff
@@ -194,7 +197,7 @@ public static void initCommand() {
     private static void initLogback() throws JoranException {
         String rocketmqHome = System.getProperty(MixAll.ROCKETMQ_HOME_PROPERTY, System.getenv(MixAll.ROCKETMQ_HOME_ENV));
 
-        LoggerContext lc = (LoggerContext) LoggerFactory.getILoggerFactory();
+        LoggerContext lc = (LoggerContext)LoggerFactory.getILoggerFactory();
```
zhouxinyu: Also please follow the RocketMQ [code-style](http://rocketmq.incubator.apache.org/docs/code-guidelines/)

wu-sheng: Already do that. Seem not work correctly.

wu-sheng: This file https://github.com/apache/incubator-rocketmq/blob/master/style/rmq_codeStyle.xml is used. 
In your website doc, just mention about import, but no style-file name.

PR #40, opened 2017-01-12
[ROCKETMQ-40]Ack mode support for consume concurrently service of push consumer

Jaskey:
according to https://docs.oracle.com/cd/E19587-01/821-0029/aeqbk/index.html.

Add ack mode suport for push consumer when consuming concurrently

jira:https://issues.apache.org/jira/browse/ROCKETMQ-40?jql=project%20%3D%20ROCKETMQ

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/listener/MessageListenerConcurrentlyV2.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/listener/MessageListenerConcurrentlyV2.java
@@ -1,0 +1,43 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


zhouxinyu:
Hi, please resolve the conflicts and follow the rocketmq [codestyle](), and IMO, we need more discuss if we intent to add new feature to rocketmq-client module.

Jaskey:
@zhouxinyu confict has been resolved. please review.

Of course we can discuss the details. Please provide a comment box or disucss post for this so that we can follow.

But the commited code should be working fine for some cases , at least for my case.

shroman:
How about discussing new features in dev mailing list before implementing them? Just my two cents ;)

Jaskey:
@shroman how do I find the dem mailing list and get involved

shroman:
@Jaskey just send an email to dev-subscribe@rocketmq.incubator.apache.org. See https://github.com/apache/incubator-rocketmq/blob/master/pom.xml

Jaskey:
@shroman email has been sent , please also review my imlementation and any suggestions are welcome

zhouxinyu:
@Jaskey , could you please add the discuss mail thread to here?

Jaskey:
@zhouxinyu

Jaskey:
@lizhanhui @vongosling @shroman @zhouxinyu 

Hi guys, would you please review this pr and let's discuss for this new feature? IMO, this is a important and useful enhancement, which actually what I am looking for.

This pr is open for more than 1 month without review or discussion though I send emails and comment for times.

dongeforever:
It seems good. And maybe the name(suffixed with V2) could be polished.
And maybe different ack mode should support orderly consuming too.

Jaskey:
@dongeforever 

what's your opinion for the new listener?  Can we just add a method to the existing interface while which will make the user suffer from compile error after updating.

If we add a new interface with the ackmode method, what will be a better name?

@lizhanhui @zhouxinyu @shroman

PR #41, opened 2017-01-12
when the client host have more than 2 ip address, the ClientConfig get the last one ipv4 always. It will make some unpredictable result.

tain198127:
When the client to start the producer, the producer will use the last one ipv4 address as default ip.but if the os have more than one ipv4 address, it will get some unexpected result.
Although ClientConfig provide a method setClientIP to avoid this issue, but in production environment with cluster ,every producer need to set a ip address, it will make some mistake.


tain198127:
When the client to start the producer, the producer will use the last one ipv4 address as default ip.but if the os have more than one ipv4 address, it will get some unexpected result.
Although ClientConfig provide a method setClientIP to avoid this issue, but in production environment with cluster ,every producer need to set a ip address, it will make some mistake.

lizhanhui:
Please follow instructs [here](http://rocketmq.incubator.apache.org/docs/pull-request/).

zhouxinyu:
IMO, this  feature is not necessary, clientIP is only used in bornHost.

zhouxinyu:
Hi @tain198127, could you please use English in the pr title and commit logs?

Also please refer to [here](http://rocketmq.apache.org/docs/code-guidelines/) and [here](http://rocketmq.apache.org/docs/pull-request/) before send a pull request to RocketMQ.

vongosling:
@tain198127 Could you follow other PR[1] style at first

[1] https://github.com/apache/incubator-rocketmq/pull/5

vongosling:
please follow us PR guide, modify your topic and ensure all testing validity

PR #43, opened 2017-01-16
[ROCKETMQ-59] Change Charset usages in RocketMQSerializable to Rocket…

iskl:
The PR is to resolve [issue-59](https://issues.apache.org/jira/browse/ROCKETMQ-59)


lizhanhui:
Minor changes, appears OK.

zhouxinyu:
@iskl It could be better if we narrow down `CHARSET_UTF8 ` access modifier to `private` and without `static`, for both `RocketMQSerializable` and `RemotingSerializable`.

iskl:
@zhouxinyu That's really a better solution! I've change the access modifiers to `private` for borh.

Since the CHARSET is used in `static` methods, it could not be modified to non-static.

Please help to review~

vongosling:
alright, thanks @iskl

PR #49, opened 2017-01-23
[ROCKETMQ-72] DefaultMessageStore cannot be properly shutdown when fa…

shroman:
…iled in the middle of start()

JIRA issue: https://issues.apache.org/jira/browse/ROCKETMQ-72

Setting `shutdown` to `false` doesn't seem semantically very correct to me, but without it, message store cannot be properly shutdown.
Please review.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -205,11 +205,17 @@ public class DefaultMessageStore implements MessageStore {
         }
         this.reputMessageService.start();
 
-        this.haService.start();
+        try {
+            this.haService.start();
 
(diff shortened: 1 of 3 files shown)


zhouxinyu:
Hi, @shroman , your worry make sense, but if there are some exceptions in `start` process, the JVM will exit.

shroman:
@zhouxinyu I guess you are talking about a shutdown hook that calls `shutdown()`. But the problem remains -- imagine `start()` fails in the middle and even if `shutdown()` is called, no proper termination is done because `this.shutdown` is `true`.

vongosling:
What's going on this pr ?

shroman:
It needs a review.

vongosling:
@shroman I close the pr, if we happened to the same problem next time, please let me know :-)

PR #51, opened 2017-01-25
[ROCKETMQ-75] RemotingCommand header decoding swallows exceptions

shroman:
JIRA issue: https://issues.apache.org/jira/browse/ROCKETMQ-75

Added logging on errors.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RemotingCommand.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RemotingCommand.java
@@ -110,7 +110,8 @@ public class RemotingCommand {
         return createResponseCommand(RemotingSysResponseCode.SYSTEM_ERROR, "not set any response code", classHeader);
     }
 
-    public static RemotingCommand createResponseCommand(int code, String remark, Class<? extends CommandCustomHeader> classHeader) {
+    public static RemotingCommand createResponseCommand(int code, String remark,
+        Class<? extends CommandCustomHeader> classHeader) {
         RemotingCommand cmd = new RemotingCommand();
         cmd.markResponseType();
         cmd.setCode(code);
@@ -230,7 +231,8 @@ public class RemotingCommand {
         this.customHeader = customHeader;
     }
 
-    public CommandCustomHeader decodeCommandCustomHeader(Class<? extends CommandCustomHeader> classHeader) throws RemotingCommandException {
+    public CommandCustomHeader decodeCommandCustomHeader(
+        Class<? extends CommandCustomHeader> classHeader) throws RemotingCommandException {
         CommandCustomHeader objectHeader;
         try {
             objectHeader = classHeader.newInstance();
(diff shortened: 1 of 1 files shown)


shroman:
Looks like there's no reason (and even may be dangerous) to keep `makeCustomHeaderToNet()` _public_. Made it _private_.

Ready for the review.

zhouxinyu:
Hi @shroman ,

IMO, it could be better that one PR/issue one thing, can we move `makeCustomHeaderToNet` to another issue, and make `makeCustomHeaderToNet ()` private will bring some reflection code, I think it's not very reasonable.

shroman:
@zhouxinyu  I agree on moving it to another issue. I will create another PR for that.
As to making the method private, using reflection for testing is safer than making a method public solely for testing, IMO. It is even called by another private method ;)

shroman:
@zhouxinyu I removed the 2nd commit from this PR. Please review it.

So, what shall we do with `makeCustomHeaderToNet()`?

vongosling:
@shroman  since 4.1.0, we will start a new branching model, more formalized, more flexible. All PRs except those hotfixs, we recommended to merge back to develop branch. This guide will publish in website sooner. So, we hope to postpone this PR merge action util 4.0.0 release :-)

zhouxinyu:
LGTM, thanks @shroman , please @vongosling @lizhanhui help review.

shroman:
@vongosling @lizhanhui are you ok with this fix? If yes, I will merge.

vongosling:
@shroman Recently, some our customer resort to us to degrade JDK version to 1.6. IMO, we can keep the pace with netty for our SDK. So, if we polish code associating with SDK, we'd better use JDK 1.6 semantic :-)

shroman:
@vongosling Is it for `rocketmq-remoting` only or for the whole project?
It doesn't sound exciting to go back to old JDKs, but ok :)

lizhanhui:
Agree, for the client related module, I think it's best to have maximum compatibility as many application developers are still working on legacy systems.

shroman:
@lizhanhui which modules in RocketMQ are 'client-related'? to be aware when changing the code.

Ideally, it would be nice to split the client-related code with what we are doing at Apache. Just my thoughts.

lizhanhui:
@shroman client/common/remoting

Frankly speaking, it's kind of dilemma. pom.xml has already defined Java language version to 1.7 and JREs under 1.7 are no longer maintained by Oracle and other vendors; But the reality is many application developers are stilling working on legacy systems which requires Java 1.5, 1.6.

shroman:
@lizhanhui @vongosling I see. I changed the code, please have a look.
Also, I will try to keep the modules in mind, but we need something not to forget they have to be treated differently from others. Can't propose any clean solution now.

vongosling:
LGTM :-) We will degrade the client SDK version to 1.6 in the developer branch

lizhanhui:
Looks good.

vongosling:
cool~

PR #53, opened 2017-02-07
[ROCKETMQ-80] Add batch feature

dongeforever:
Tests show that Kafka's million-level TPS is mainly owed to batch. When set batch size to 1, the TPS is reduced an order of magnitude. So I try to add this feature to RocketMQ.

For a minimal effort, it works as follows:
 
* Only add synchronous send functions to MQProducer interface, just like **send(final Collection<Message> msgs)**
* Use **MessageBatch** which extends **Message** and implements **Iterable\<Message\>**
* Use byte buffer instead of list of objects to avoid too much GC in Broker.
* Split the decode and encode logic from **lockForPutMessage** to avoid too many race conditions.

Tests:
 On linux with 24 Core 48G Ram and SSD, single broker, using 50 threads to send 50Byte(body) message in batch size 50, we get about 150w TPS until the disk is full.
 

Potential problems:
Although the messages can be accumulated in the Broker very quickly, it need time to dispatch to the consume queue, which is much slower than accepting messages. So the messages may not be able to be consumed immediately.

We may need to refactor the **ReputMessageService** to solve this problem.

And if guys have some ideas, please let me know or just share it in this issue.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
@@ -331,7 +331,7 @@ public class ConsumeQueue {
     public void putMessagePositionInfoWrapper(long offset, int size, long tagsCode, long storeTimestamp,
         long logicOffset) {
         final int maxRetries = 30;
-        boolean canWrite = this.defaultMessageStore.getRunningFlags().isWriteable();
+        boolean canWrite = this.defaultMessageStore.getRunningFlags().isCQWriteable();
         for (int i = 0; i < maxRetries && canWrite; i++) {
             boolean result = this.putMessagePositionInfo(offset, size, tagsCode, logicOffset);
             if (result) {
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -374,9 +374,11 @@ public class BrokerController {
 
         this.remotingServer.registerProcessor(RequestCode.SEND_MESSAGE, sendProcessor, this.sendMessageExecutor);
         this.remotingServer.registerProcessor(RequestCode.SEND_MESSAGE_V2, sendProcessor, this.sendMessageExecutor);
+        this.remotingServer.registerProcessor(RequestCode.SEND_BATCH_MESSAGE, sendProcessor, this.sendMessageExecutor);
         this.remotingServer.registerProcessor(RequestCode.CONSUMER_SEND_MSG_BACK, sendProcessor, this.sendMessageExecutor);
         this.fastRemotingServer.registerProcessor(RequestCode.SEND_MESSAGE, sendProcessor, this.sendMessageExecutor);
         this.fastRemotingServer.registerProcessor(RequestCode.SEND_MESSAGE_V2, sendProcessor, this.sendMessageExecutor);
+        this.fastRemotingServer.registerProcessor(RequestCode.SEND_BATCH_MESSAGE, sendProcessor, this.sendMessageExecutor);
         this.fastRemotingServer.registerProcessor(RequestCode.CONSUMER_SEND_MSG_BACK, sendProcessor, this.sendMessageExecutor);
         /**
          * PullMessageProcessor
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
@@ -17,11 +17,6 @@
 package org.apache.rocketmq.broker.processor;
 
 import io.netty.channel.ChannelHandlerContext;
-import java.net.InetSocketAddress;
-import java.net.SocketAddress;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.mqtrace.SendMessageContext;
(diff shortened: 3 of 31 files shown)


dongeforever:
@vongosling @zhouxinyu @WillemJiang please have a review.

vongosling:
Cool, Thanks @dongeforever providing this feature. We will have a look at this implementation. please hold your horses :-)

zhouxinyu:
I wonder that are there some compatibility issues between new client version and old server version?

dongeforever:
@zhouxinyu yeah, this is a problem. When new client send batched messages to old server, it will get no error, for the batched messages are treated as normal message internally. 

Maybe I need a new request code, therefore the old server cannot recognize it and throw error.

shroman:
Sounds good to check for the protocol version in the request header, and reject with an error when not matched.
Probably, `RemotingCommand.REMOTING_VERSION_KEY` will work, but I would create a version policy based not on the release (that is used now) but on the protocol changes, so that we don't have to introduce a new checking condition in the code with each release. Just an idea.

dongeforever:
@shroman thank you. 
Great idea to create a version policy based on protocol changes.
And we would need to control every api  separately.
So maybe we can write all these checking codes in one place for all apis.

BTW, version control is a global problem, not only for batch, maybe  we need more discussion in the mailing list.

shroman:
@dongeforever yes, I think we have to discuss it in the dev mailing list :+1: 
I think having a class/methods for wire protocol version checking and applying them (and, as a result, accepting or rejecting) when a client wants to establish a connection to a broker, etc. should work. I guess that's what you meant by having "checking codes in one place" :)
The version information can be placed into headers.

Such protocol changes shouldn't happen that often (RocketMQ v.4.0.0, v.4.1.0 may use protocol v.1, but RocketMQ v.5.0.0 may use protocol v.2), but if they happen, in this way we can guarantee that the communication between system components is done correctly.

Let's discuss it in the ml. My understanding of RocketMQ is still not complete ;) Maybe there are better ideas not to over-complicate is with versioning.

dongeforever:
@zhouxinyu Note that the batch is a new api, I add a new request code named "SEND_BATCH_MESSAGE". As the old broker cannot recognize it,  the compatible problem is gone.

Also I use the ThreadLocal to avoid race condition for MessageExtBatchEncoder, which previously was cached in the blocking queue.

Jaskey:
@dongeforever 

I have the same wishes for batch send too, but what drives me is that user may propably need a batch id for one batch of message, and these message should be success all to one single queue, which is nessecary when sneding order message. say msgA msgB and msgC should be consumed in order, they should be send to one same queue, but if we use for loop to send this, A may success and B may fail to the same queue since the queue numbers may changes at that exctly time.

Batch send could solve this problem. But we may also need to generate a uniq batch id for this in client, which will help us to optimze the performance of consumeorderlyservice in the furture. Currently, message in one single queue can only be consumed only if the previous one consumed successfully which actually is too strict. Actully we only need the message in one batch consumed in order, batch id will help us to do this.

**So in general, I suggest adding batch id  when sending batch message in all message property.**

PS: There looks like two many repeated code, any ways or plans to clean it?

dongeforever:
@Jaskey Now there is no batch id, but the messages in one batch are sent to the same queue, and they can only be sent all successfully or all unsuccessfully.
You could check the code or test it.

Jaskey:
@dongeforever I know your implementation.What I suggest is that you add batch id for this so that we can inditify them, which is actully a very minimum effort for you.

And as I sai,  in the future we can optimize consumeOrderlyService with batch id.

dongeforever:
@Jaskey Sorry. You are right. And it is worth a new PR, maybe we can add batch id along with a new optimized consumeOrderlyService.

lizhanhui:
This feature is definitely nice to have and it expands scenarios that RocketMQ fits best. For example, RocketMQ may have close, if not better, performance with Kafka in log collecting usage.

That said, this is a pretty important feature and it matters so much that we need to get it right at the beginning.  We'd better have a design document first, then discuss various impacts it brings about in the mailing list.

As for this PR, it's generally good, yet, still needs more work: code duplication, message validation logic discrepancy, excessive constraints on usages(No delay messages, messages of a batch must have same topic, etc) and previously mentioned compatibility issue with older brokers.

vongosling:
Yep agreed @lizhanhui . I will make some comment as much detail as possible.

vongosling:
@dongeforever we can continue to polish this PR, IMO. if you have any problem. please let me know. BTW, can you post your performance test result for us

zhouxinyu:
Hi, @shroman @Jaskey @lizhanhui , what's your opinion about this updated PR?

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -331,7 +331,7 @@ public long getMinOffsetInQueue() {
     public void putMessagePositionInfoWrapper(long offset, int size, long tagsCode, long storeTimestamp,
         long logicOffset) {
         final int maxRetries = 30;
-        boolean canWrite = this.defaultMessageStore.getRunningFlags().isWriteable();
+        boolean canWrite = this.defaultMessageStore.getRunningFlags().isCQWriteable();
```
Jaskey: isConsumeQWritable will be better I guess

dongeforever: @Jaskey Yeah, it is better to use the name isConsumeQWritable

dongeforever:
this PR has been merged into develop branch

vongosling:
LGTM~

PR #57, opened 2017-02-10
[ROCKETMQ-91] Reduce lock granularity for putMessage to get better performance

dongeforever:
## Motivation
CommitLog putMessage has a lock as:
```
lockForPutMessage()
try {
  .....
} finally {
releasePutMessageLock()
}
```
The logic inside the lock includes two main operations:
1 encode the message
2 write to the PageCache.
However, we can take the first operation(encode message) out from the lock to achieve better performance.

### Results
reduce before: about **14w TPS**
reduce after: about **16w TPS**
And if we annotate the store layer, only test the network performance, we got about **17W TPS**.

The bottleneck now is in the network layer.

@zhouxinyu @vongosling @shroman please have a review.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -59,6 +59,8 @@ public class CommitLog {
 
     private final AppendMessageCallback appendMessageCallback;
     private final ThreadLocal<MessageExtBatchEncoder> batchEncoderThreadLocal;
+
+    private final ThreadLocal<MessageEncoder> messageEncoderThreadLocal;
     private HashMap<String/* topic-queueid */, Long/* offset */> topicQueueTable = new HashMap<String, Long>(1024);
     private volatile long confirmOffset = -1L;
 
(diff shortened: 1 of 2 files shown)


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -567,6 +577,11 @@ public PutMessageResult putMessage(final MessageExtBrokerInner msg) {
         MappedFile unlockMappedFile = null;
         MappedFile mappedFile = this.mappedFileQueue.getLastMappedFile();
 
+        //maybe need to wrap the exception
```
shroman: When encoding fails, let's notify a user with a `PutMessageResult` about the failure. There's nothing else we can do, I think.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1154,7 +1249,7 @@ public AppendMessageResult doAppend(final long fileFromOffset, final ByteBuffer
 
             if (propertiesLength > Short.MAX_VALUE) {
                 log.warn("putMessage message properties length too long. length={}", propertiesData.length);
-                return new AppendMessageResult(AppendMessageStatus.PROPERTIES_SIZE_EXCEEDED);
+                throw new RuntimeException("PROPERTIES_SIZE_EXCEEDED");
```
shroman: By changing to this, `PROPERTIES_SIZE_EXCEEDED` won't be caught in the following switch.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
zhouxinyu: Agree with @shroman .

vongosling:
LGTM

PR #58, opened 2017-02-12
Update runbroker.sh

yuweitaocn:
when use g1 grabage, needen't  set xmn,because it's cause   full gc !

http://www.oracle.com/technetwork/articles/java/g1gc-1984535.html

Young Generation Size: Avoid explicitly setting young generation size with the -Xmn option or any or other related option such as -XX:NewRatio. Fixing the size of the young generation overrides the target pause-time goal.

### DIFF
--- a/bin/runbroker.sh
+++ b/bin/runbroker.sh
@@ -36,7 +36,7 @@ export CLASSPATH=.:${BASE_DIR}/conf:${CLASSPATH}
 #===========================================================================================
 # JVM Configuration
 #===========================================================================================
-JAVA_OPT="${JAVA_OPT} -server -Xms8g -Xmx8g -Xmn4g"
+JAVA_OPT="${JAVA_OPT} -server -Xms8g -Xmx8g"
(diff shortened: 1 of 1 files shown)


zhouxinyu:
Hi, @yuweitaocn 
Thanks for your PR, Do you have some test data to support your theory? The current GC parameters are well-tuned in our production env.

vongosling:
@yuweitaocn Performance Tuning is a Galileo thing. Apache RocketMQ all gc paramters have been proved in many company production environment. That is said. All Parameters are have been sustained in many Galileo experimentation including various JDK versions.So, could you provide some clue to expose this problem. May be, it is friendly to be accepted.

lizhanhui:
Hi @yuweitaocn , if you examine the options here carefully, you would find there is no "target pause time-goal", aka, RocketMQ does not set this option: -XX:MaxGCPauseMillis=xyz. So it's fine to set "young generation size with the -Xmn option".

As a matter of fact, if you read the following section that's closely following the part you quoted

> Therefore, when you evaluate the G1 GC for throughput, relax your pause-time target. Setting too aggressive a goal indicates that you are willing to bear an increase in garbage collection overhead, which has a direct impact on throughput. When you evaluate the G1 GC for latency, you set your desired (soft) real-time goal, and the G1 GC will try to meet it. As a side effect, throughput may suffer.

You will find the current settings make sense.

zhouxinyu:
@yuweitaocn , so, it seems that we could close this PR? what's your opinion？

vongosling:
if any update for this pr, we will close it~

PR #60, opened 2017-02-14
[ROCKETMQ-96]Rename some temp variable and field

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-96

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
@@ -46,7 +46,10 @@ public class ProcessQueue {
     private final TreeMap<Long, MessageExt> msgTreeMap = new TreeMap<Long, MessageExt>();
     private final AtomicLong msgCount = new AtomicLong();
     private final Lock lockConsume = new ReentrantLock();
-    private final TreeMap<Long, MessageExt> msgTreeMapTemp = new TreeMap<Long, MessageExt>();
+    /**
+     * A subset of msgTreeMap, will only be used when orderly consume
+     */
+    private final TreeMap<Long, MessageExt> consumingMsgOrderlyTreeMap = new TreeMap<Long, MessageExt>();
     private final AtomicLong tryUnlockTimes = new AtomicLong(0);
     private volatile long queueOffsetMax = 0L;
     private volatile boolean dropped = false;
@@ -235,8 +238,8 @@ public class ProcessQueue {
         try {
             this.lockTreeMap.writeLock().lockInterruptibly();
             try {
-                this.msgTreeMap.putAll(this.msgTreeMapTemp);
-                this.msgTreeMapTemp.clear();
+                this.msgTreeMap.putAll(this.consumingMsgOrderlyTreeMap);
+                this.consumingMsgOrderlyTreeMap.clear();
             } finally {
                 this.lockTreeMap.writeLock().unlock();
             }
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
```diff
@@ -46,7 +46,7 @@
     private final TreeMap<Long, MessageExt> msgTreeMap = new TreeMap<Long, MessageExt>();
     private final AtomicLong msgCount = new AtomicLong();
     private final Lock lockConsume = new ReentrantLock();
-    private final TreeMap<Long, MessageExt> msgTreeMapTemp = new TreeMap<Long, MessageExt>();
+    private final TreeMap<Long, MessageExt> consumingMsgOrderlyTreeMap = new TreeMap<Long, MessageExt>();//subset of msgTreeMap, used when consume orderly
```
zhouxinyu: Hi, in this PR, what's the difference between `msgTreeMap` and `consumingMsgOrderlyTreeMap`?

Jaskey: I do not change anything except the variable, this PR is to solve that it is very hard to know what's the purpose of `msgTreeMapTemp `, when I review the code, I think it is only for order service, so I rename the field.

The difference between `msgTreeMap` and `consumingMsgOrderlyTreeMap` is that , `consumingMsgOrderlyTreeMap` is the subset of `msgTreeMap` when using consumeOrderlyService , msg will only be putinto `consumingMsgOrderlyTreeMap` when they are ready to submit to thread pool.

Actually it takes me more than ten minutes to figure out that, that's why I issue this pr.

Jaskey:
@zhouxinyu @shroman @vongosling @lizhanhui 
Guys, how do you think of this pr, which does not change any functionality , I just rename some variables/fields to make code more readable.

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
shroman: It would be nice to have a Javadoc comment here. Something like,
```java
/**
* Subset of msgTreeMap used when consuming orderly.
*/
private final TreeMap<Long, MessageExt> msgOrderlyTreeMap = ...
```

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -449,9 +449,9 @@ private SendResult sendDefaultImpl(//
             String[] brokersSent = new String[timesTotal];
             for (; times < timesTotal; times++) {
                 String lastBrokerName = null == mq ? null : mq.getBrokerName();
-                MessageQueue tmpmq = this.selectOneMessageQueue(topicPublishInfo, lastBrokerName);
-                if (tmpmq != null) {
-                    mq = tmpmq;
+                MessageQueue mqAttempt = this.selectOneMessageQueue(topicPublishInfo, lastBrokerName);
```
Jaskey: IMO , any variable named like XXXtmp  is bad. Maybe we can rename it to another one but tmpMq is far from good enough

Jaskey:
@shroman @lizhanhui @zhouxinyu 

There are users asking the same question about the logic of consume orderly and they both complain that the `msgTreeMapTemp` is very hard to understand it's implication, so would you please review this pr and merge it in 4.1.

Besides, I wonder if there is the same issue in below method, the signature of the method is trying to determine if there are some tmp msg(which will be only used in consumer orderly service), while the implementation is to determine the `msgTreeMap`?

Is the implementation wrong or the is signature misleading.

    public boolean hasTempMessage() {
        try {
            this.lockTreeMap.readLock().lockInterruptibly();
            try {
                return !this.msgTreeMap.isEmpty();
            } finally {
                this.lockTreeMap.readLock().unlock();
            }
        }

        return true;
    }

lizhanhui:
+1 to merge it.

shroman:
@Jaskey No idea about `hasTempMessage()` -- no comments or unit tests...
All the rest is ok to merge.

Jaskey:
@zhouxinyu @vongosling 
Any comments on this pr and any ideas about `hasTempMessage()`?

dongeforever:
This PR has been merged

PR #61, opened 2017-02-15
[ROCKETMQ-98]Fix risk of unable to release putMessage Lock forever.

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-98

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
+++ b/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
@@ -47,6 +47,9 @@ public class BrokerConfig {
     private boolean autoCreateSubscriptionGroup = true;
     private String messageStorePlugIn = "";
 
+    /**
+     * thread numbers for send message thread pool, since spin lock will be used by default since 4.0.x, the default value is 1.
+     */
     private int sendMessageThreadPoolNums = 1; //16 + Runtime.getRuntime().availableProcessors() * 4;
     private int pullMessageThreadPoolNums = 16 + Runtime.getRuntime().availableProcessors() * 2;
     private int adminBrokerThreadPoolNums = 16;
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -23,8 +23,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.ReentrantLock;
 import org.apache.rocketmq.common.ServiceThread;
 import org.apache.rocketmq.common.UtilAll;
 import org.apache.rocketmq.common.constant.LoggerName;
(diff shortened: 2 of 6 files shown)


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -799,27 +796,18 @@ public long lockTimeMills() {
     }
 
     /**
-     * Spin util acquired the lock.
+     * acquired the lock and return
      */
-    private void lockForPutMessage() {
+    private PutMessageLock lockForPutMessage() {
         if (this.defaultMessageStore.getMessageStoreConfig().isUseReentrantLockWhenPutMessage()) {
```
zhouxinyu: It could be better if we cache the `isUseReentrantLockWhenPutMessage` and invalidate the dynamic change of this variable.

Jaskey:
@zhouxinyu , please review the updated code

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
```diff
@@ -47,6 +47,7 @@
     private boolean autoCreateSubscriptionGroup = true;
     private String messageStorePlugIn = "";
 
+    //thread counts for sending message, if large number threads are needed, we suggest setting useReentrantLockWhenPutMessage=true(in MessageStoreConfig)
```
shroman: Is this suggestion based on observations in RocketMQ production environments?

lizhanhui: IMO, this is the tradeoff between two kinds of synchronization implementations: busy-waiting and mutex-lock; The former approach indeed wastes more CPU cycles which may be utilized to finish other tasks by scheduler, but it has its benefits too--doing so actually prioritize current data flow; Even if the latter is energy-saving, it has its cons, say, more latencies are introduced.  

Another factor is the operation system being used. OS is getting so smart that things may have changed considerably.

If someone may provide raw data, comparing effects of these two choices under various conditions(# of cpu cores, # of send-message-threads, flush-disk-type, etc), it would be very informative and valuable.

shroman: > OS is getting so smart that things may have changed considerably.
Exactly. That is why I wouldn't make such a suggestion in the code. You cannot bet it would perform better when you switch to lock mutex.

Jaskey: @shroman 

This is to make hints to developer after discuss with @zhouxinyu , since if developer hopes to change the sendNums, which may means propobly very feriece race conditions which by default using spin lock is not quite good enough. 

Accoring to the current spin lock implementations, the cpu will be runnig all the time without any sleep when trying to get spin lock, which may cause cpu problem when they set sendNums to a very large number with high concurrent message writing.

The comment to pointout that.

lizhanhui: @Jaskey , would you carry out an experiment, figuring out when spin-lock is better and when the alternative mutex-lock wins out? 

You may compose an article based on that experiment, which should be added to [best practice](http://rocketmq.incubator.apache.org/docs/best-practice-broker/) section of RocketMQ official website.  It's currently vacant!

zhouxinyu:
Hi, @lizhanhui @shroman  do you have some review results about this PR?

lizhanhui:
This PR itself is OK. Abstracting the concept of`PutMessageLock` is pretty handy.

What I suggest here is presenting the advice more formally instead of a simple inline comment. Preferably, we may create a blog, backed up with real experiment data, suggesting decision-making process between spin-lock and mutex-lock.

Jaskey:
@lizhanhui 

Acutully , if thread numbers are large , the cpu problem should be obvious without doubt in a spin lock with `while(true)` without `sleep`.

But I also agree that we can create a blog for this, remove the comment for me is fine , just DO NOT merge that commit.

Actully , my very first advice is that when implementing the spin lock add `sleep(1)` when continiously fail to abatain the lock  say 100 times, which should solve most of the problems. 


@Override
 public void lock() {
 
 int tryTimes = 0;
 boolean flag;
 do {
     if (tryTimes++ % 100 ==0 ){
          Thread.sleep(1); //fierce race condition
     }
     flag = this.putMessageSpinLock.compareAndSet(true, false);
  }
     while (!flag);
 }

What's your advice 
@zhouxinyu @vongosling @lizhanhui

Jaskey:
@lizhanhui 
the comments has been updated and the suggestion in the previous pr has been removed, please help review.

lizhanhui:
+1

Jaskey:
@zhouxinyu @shroman @vongosling  please help review

dongeforever:
LGTM @zhouxinyu

zhouxinyu:
+1, thanks @Jaskey

PR #63, opened 2017-02-15
[ROCKETMQ-101]Fix possible NullPointerException when retry in send Async way

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-101

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -436,34 +436,40 @@ public class MQClientAPIImpl {
     ) {
         int tmp = curTimes.incrementAndGet();
         if (needRetry && tmp <= timesTotal) {
-            MessageQueue tmpmq = producer.selectOneMessageQueue(topicPublishInfo, brokerName);
-            String addr = instance.findBrokerAddressInPublish(tmpmq.getBrokerName());
+            String retryBrokerName = brokerName;//by default, it will send to the same broker
+            if (topicPublishInfo != null) { //select one message queue accordingly, in order to determine which broker to send
+                MessageQueue mqChosen = producer.selectOneMessageQueue(topicPublishInfo, brokerName);
+                retryBrokerName = mqChosen.getBrokerName();
+            }
+            String addr = instance.findBrokerAddressInPublish(retryBrokerName);
             log.info("async send msg by retry {} times. topic={}, brokerAddr={}, brokerName={}", tmp, msg.getTopic(), addr,
-                tmpmq.getBrokerName());
+                retryBrokerName);
             try {
                 request.setOpaque(RemotingCommand.createNewRequestId());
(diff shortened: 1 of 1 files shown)


lizhanhui:
This issue is valid and the bug will be triggered under the following conditions:
1. A target message queue is specified, which will pass null for TopicPublishInfo parameter;
2. Async sending;
3. Retry happens;
4. sendLatencyFaultEnable is enabled(On default, it's not enabled)


The PR, IMO, is not good as it simply disabled promised retry attempts without notice.  Instead of logging retry failure, retry sending should be done against the original chosen broker.

Jaskey:
@lizhanhui , # 4 is not the nessary conditions, since even it is not enble, the tpInfo's method is still used.

            try {
                int index = tpInfo.getSendWhichQueue().getAndIncrement();
                for (int i = 0; i < tpInfo.getMessageQueueList().size(); i++) {
                    int pos = Math.abs(index++) % tpInfo.getMessageQueueList().size();
                    if (pos < 0)
                        pos = 0;
                    MessageQueue mq = tpInfo.getMessageQueueList().get(pos);
                    if (latencyFaultTolerance.isAvailable(mq.getBrokerName())) {
                        if (null == lastBrokerName || mq.getBrokerName().equals(lastBrokerName))
                            return mq;
                    }
                }

                final String notBestBroker = latencyFaultTolerance.pickOneAtLeast();
                int writeQueueNums = tpInfo.getQueueIdByBroker(notBestBroker);
                if (writeQueueNums > 0) {
                    final MessageQueue mq = tpInfo.selectOneMessageQueue();
                    if (notBestBroker != null) {
                        mq.setBrokerName(notBestBroker);
                        mq.setQueueId(tpInfo.getSendWhichQueue().getAndIncrement() % writeQueueNums);
                    }
                    return mq;
                } else {
                    latencyFaultTolerance.remove(notBestBroker);//my npe is thrown here
                }
            } 


Besides, if topic route info is null which propably means user is send through my seletor method, resend should still respect user's seletor, so chosen broker is not enough, the chosen queue is needed, which may be another issue, I guess it is not a very minal effort, since the existing interface does not record any chosen queue info.

lizhanhui:
Yes, #4 is not required as MQFaultStrategy#92 may also generate NPE.
Queue ID info is contained in the request header, minimal effort suffices.

Jaskey:
@lizhanhui , please review the updated pr

lizhanhui:
Looks good.

zhouxinyu:
Looks good, please @vongosling help review.

Jaskey:
@vongosling @lizhanhui 

This bug is crucial and easy to reproduce which makes send with seletor is  not reliable, when can this pr be merged and release?

lizhanhui:
@Jaskey You are right, I'll follow up and merge this PR ASAP.

vongosling:
LGTM~

lizhanhui:
@Jaskey I'll merge this PR soon. Please close this issue.

Jaskey:
Thank you @vongosling @lizhanhui @zhouxinyu

vongosling:
@lizhanhui @Jaskey we'd better close after merge :-)

lizhanhui:
Merged

PR #64, opened 2017-02-16
[ROCKETMQ-102] When shutdown(), the persisted offet is not the latest consumed message, which may cause repeated messages.

Jaskey:
Solution: add interface for push consumer to accept await termination time to await consuming.

JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-102


lizhanhui:
Can you provide some test data, say before/after applying this patch, how many duplications are found respectively?

IMHO, we should make the API as concise as possible.

Jaskey:
@lizhanhui 

The problem exists for long, please review an old issue.
https://github.com/alibaba/RocketMQ/issues/367

The number of duplicated messages are depending on how many messages are consuming in the thread pool or in the pending queue. This should be very easy to meet when consumptions are takes long and with massive accumulation.

The problem should be addressed since rocketmq should prevend duplication as much as possible.

lizhanhui:
I know this issue has been brought up in the past. What I suggest here is providing some testing data to consolidate the rationality of this patch. By doing so, you'll find it easier to get developers convinced.

Anyway, thanks a lot for bringing this issue back to attention.

Jaskey:
@lizhanhui please refer to the test case and just switch the `shutdown(long milis)` method to `shutdown()` to find out the problem

Jaskey:
@lizhanhui @zhouxinyu @vintagewang 

What do you think my proposol of adding a field called `awaitMilisWhenShutdown` in `DefaultPushConsumer`, which will not need to add a `shutdown(long awaitMilis)` interface but one more config field.

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
```diff
@@ -93,8 +93,24 @@ public void run() {
...
+
+    @Override
+    public void shutdown(long awaitTerminateMillis) {
         this.scheduledExecutorService.shutdown();
         this.consumeExecutor.shutdown();
+        //await to consume
+        if (awaitTerminateMillis > 0) {
+            try {
+                this.consumeExecutor.awaitTermination(awaitTerminateMillis,TimeUnit.MILLISECONDS);
+                if (!this.consumeExecutor.isTerminated()) log.info("There are messages still being consumed in thread pool, but not going to await them anymore. Have awaited for {} ms",awaitTerminateMillis);
```
zhouxinyu: Did you follow our code code guidelines[1] ? Below `if` block is recommended.
```
if {
}
```
http://rocketmq.incubator.apache.org/docs/code-guidelines/

client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
```diff
@@ -461,7 +461,11 @@ public void start() throws MQClientException {
      */
     @Override
     public void shutdown() {
-        this.defaultMQPushConsumerImpl.shutdown();
+        shutdown(0);
+    }
+
+    public void shutdown(long awaitTerminateMillis) {
```
zhouxinyu: Since `MQPushConsumer` only has the method interface for `shutdown( )`, so we could consider pull `shutdown(long awaitTerminateMillis)` up to the parent interface or make `shutdown(long awaitTerminateMillis)` private and give a default input(non-zero) in `shutdown()`.

zhouxinyu:
Hi @Jaskey ,
Thanks for this PR, I added two comments for your reference.

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
Jaskey: Ok, I will update the pr accordingly.

client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
Jaskey: @zhouxinyu  

I know and agree to your concern, actually, as I my previous comment, maybe we could add an configuration for awaitMills, what do you think ? The default value could be 0 to keep the same behavior

Jaskey:
@lizhanhui @zhouxinyu 

please review the updated pr, which remains the same interface of push consumer.

client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
zhouxinyu: Good idea, it's ok for me.

client/src/test/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumerTest.java
```diff
@@ -52,6 +56,7 @@
 import org.apache.rocketmq.common.protocol.header.PullMessageRequestHeader;
 import org.apache.rocketmq.remoting.exception.RemotingException;
 import org.junit.After;
+import org.junit.Assert;
```
zhouxinyu: Hi, let's unify the assert tool and use `org.assertj.core.api.Assertions.assertThat`.

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
```diff
@@ -92,10 +92,22 @@ public void run() {
...
+    @Override
+    public void shutdown(long awaitTerminateMillis) {
         this.stopped = true;
         this.scheduledExecutorService.shutdown();
         this.consumeExecutor.shutdown();
+        //await to consume
+        if (awaitTerminateMillis > 0) {
+            try {
+                this.consumeExecutor.awaitTermination(awaitTerminateMillis,TimeUnit.MILLISECONDS);
+                if (!this.consumeExecutor.isTerminated()) log.info("There are messages still being consumed in thread pool, but not going to await them anymore. Have awaited for {} ms",awaitTerminateMillis);
```
zhouxinyu: May be we need a common method to shutdown executor gracefully, like:

```
public static void shutdownGracefully(ExecutorService executor, long timeout, TimeUnit timeUnit) {
        executor.shutdown();

        try {
            if(!executor.awaitTermination(timeout, timeUnit)) {
                executor.shutdownNow();
                if(!executor.awaitTermination(timeout, timeUnit)) {
                    LOG.warn(String.format("%s didn\'t terminate!", new Object[]{executor}));
                }
            }
        } catch (InterruptedException var5) {
            executor.shutdownNow();
            Thread.currentThread().interrupt();
        }

    }
```

Jaskey: We can, but where do we put this method in, in a Common Util?

zhouxinyu: add a new java file `ThreadUtils`?

Jaskey:
@zhouxinyu 

Please check the updated pr.

Since shutdownNow needs developer to take respond to interrupts but actually they should not care about this, so I still remain using shutdown , and after timeout , the method will return and remains the very same flow.

BTW, unit test has been updated using  `AssetThat`

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
zhouxinyu: Hi @Jaskey ,

I added a ThreadUtils, please refer to [here](https://github.com/apache/incubator-rocketmq/commit/e3f4251c91a73f4e51732bcb1690554ac5fb3096).

Jaskey: Then maybe you merge that commit first and I will merge it later.

zhouxinyu: `shutdown` is the first choice, `shutdownNow` will be called if timeout.

And `develop` will be merged to `master` in next release, please refer to our new [branching model](http://rocketmq.apache.org/docs/branching-model).

Jaskey: @zhouxinyu In my option, rocketmq should have no right to interrupt what dev's business are doing, we may be doing some time-cost job which are doing transcation or inserting database, we should leave the task running if executor is still not terminated.

Besides, since the old version has not termination millis, so 0 of termination millis is the default  behavior , shutdown now will cost task being interrupt/cancel immediately which is not proper in my opinion.

I have updated the pr , but still using my old method, please review and let's discuss more about it.

Jaskey:
Repeated messages is a big problem for a message queue, and this is a known issue. When can this pr be reviewed and merged?
@lizhanhui @zhouxinyu @vongosling

Jaskey:
I request a another issue  #2085 since I find this feature is still not supported for the latest release

PR #67, opened 2017-02-20
[ROCKETMQ-67] Consistent Hash allocate strategy

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-67

Consitent hash alogrithm to allocate message queue to prevent "thundering herd". 

Core implementation:

Define a ConsitentHashRouter to hash clientId to a hash ring. And try to find the nearest clientId for every message queue.

High scalability considered, there are several new class to support as many scenarios as possible when user needs to use consitent hash  

1. Any types of object can be routed as long as they implement `Node` interface.
2. Custom hash function can be specified
3. Numbers of virtual nodes can be specifed.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/consistenthash/ConsistentHashRouter.java
+++ b/common/src/main/java/org/apache/rocketmq/common/consistenthash/ConsistentHashRouter.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.common.consistenthash;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+/**
+ * To hash Node objects to a hash ring with a certain amount of virtual node.
+ * Method routeNode will return a Node instance which the object key should be allocated to according to consistent hash algorithm
+ *
+ * @param <T>
+ */
+public class ConsistentHashRouter<T extends Node> {
+    private final SortedMap<Long, VirtualNode<T>> ring = new TreeMap<>();
+    private final HashFunction hashFunction;
+
+    public ConsistentHashRouter(Collection<T> pNodes, int vNodeCount) {
+        this(pNodes,vNodeCount, new MD5Hash());
+    }
+
+    /**
+     *
+     * @param pNodes collections of physical nodes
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/rocketmq/common/consistenthash/Node.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.common.consistenthash;
+
+/**
+ * Created by jaskeylin on 2017/2/20.
```
zhouxinyu: Hi, please remove the author info.

Jaskey: Sorry , i didn'it notice that. Just a minute

Jaskey:
@zhouxinyu @lizhanhui @shroman @vongosling 
any advice on this pr?

Since I think consistent hash can be also applied to order message shading, the classes in this pr can be reused. I will submit a new pr for that sharding message queue selector after this pr is merged : https://issues.apache.org/jira/browse/ROCKETMQ-136?jql=project%20%3D%20ROCKETMQ

shroman:
@Jaskey Can you explain more about this feature? Having a better description on JIRA would help a lot review this PR -- especially for new functionalities. Since the description is very general, I have the following questions.

What is the advantage of introducing this strategy?
Consistent hashing could be useful if you have to move data between nodes on broker addition/removal. But since `AllocateMessageQueueStrategy` is used by consumers just to read from queues (if I get it right), on broker topology change a simple rehashing works well.
Of course, consistent hashing will glue consumers to the queues they are already consuming from as much as possible, even if topology is changed. Is this your concern?

Jaskey:
@shroman 

This is feature to give choice to users who cares more about latency stabilization and messages duplication.

As you know, the default `AllocateMessageQueueStrategy` is averaging strategy which allocate queue to consumer  as evenly as possible. But as you states whenever queues numbers or consumer numbers changed, say a new consumer starts or an old consumer shutdown, a rehashing will be triggered and almost all consumer suffered from this and they rebalance to drop old queues and get new queue.

This will cause 
1. message latency from producer to consumer increases at that moment 
2. messages will be duplicated significantly since the offset may not be persisted to broker and that queue is assigned to another consumer to pull messages from.

This is especially significant when they have tens of consumer instances and scale-up or deployment is often.

Consistent Hash strategy to allocate queue is a good choice for these users.

shroman:
@Jaskey This is a very good description, and now reviewers will understand well your intentions :) If you clearly state your intentions in JIRA, or discuss in the ml, it saves much time, and good for the record.

Sorry, it's not a code review yet, just wanted understand your intentions.

Jaskey:
@shroman 
The detail descriptions has been updated to JIRA.  I will try described as detail as possible when I create an issue in JIRA in the future.

Jaskey:
@shroman @zhouxinyu @lizhanhui @vongosling 

I heard that consistent hash strategy will be accomplished in 4.1.x , while this pr has been open for months, can we accelerate it, please help review the implementations?

common/src/main/java/org/apache/rocketmq/common/consistenthash/ConsistentHashRouter.java
```diff
@@ -0,0 +1,140 @@
...
+        public long hash(String key) {
+            instance.reset();
+            instance.update(key.getBytes());
+            byte[] digest = instance.digest();
+
+            long h = 0;
+            for (int i = 0; i < 4; i++) {
+                h <<= 8;
+                h |= ((int) digest[i]) & 0xFF;
+            }
```
dhchao11: digest is 128 bits， but the generated h only use 32 bits， other 96 bits are ignored,  will it be better use ^ method

PR #68, opened 2017-02-23
[ROCKETMQ-107] fix possible concurrency problem on ServiceState when consumer start/shutdown

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-107

No Syncrinations in start or shutdown which may result in : 

1. Start twice if two thread starts the same consumer in the same time, which leads to initizize the resources twice.
2. Shutdown may not be effective if thread A start and Thread b tries to shutdown in a very short time after A's starting.

Implemenations: 

1. Add lock in getter and setter, and replace all direct read/write operations to getter and setter
2. Add lock before `start` and `shutdown`

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPullConsumerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPullConsumerImpl.java
@@ -70,7 +70,7 @@ public class DefaultMQPullConsumerImpl implements MQConsumerInner {
     private final RPCHook rpcHook;
     private final ArrayList<ConsumeMessageHook> consumeMessageHookList = new ArrayList<ConsumeMessageHook>();
     private final ArrayList<FilterMessageHook> filterMessageHookList = new ArrayList<FilterMessageHook>();
-    private ServiceState serviceState = ServiceState.CREATE_JUST;
+    private volatile ServiceState serviceState = ServiceState.CREATE_JUST;
     private MQClientInstance mQClientFactory;
     private PullAPIWrapper pullAPIWrapper;
     private OffsetStore offsetStore;
@@ -161,7 +161,8 @@ public class DefaultMQPullConsumerImpl implements MQConsumerInner {
         return this.pullSyncImpl(mq, subExpression, offset, maxNums, false, timeout);
     }
 
-    private PullResult pullSyncImpl(MessageQueue mq, String subExpression, long offset, int maxNums, boolean block, long timeout)
+    private PullResult pullSyncImpl(MessageQueue mq, String subExpression, long offset, int maxNums, boolean block,
+        long timeout)
         throws MQClientException, RemotingException, MQBrokerException, InterruptedException {
         this.makeSureStateOK();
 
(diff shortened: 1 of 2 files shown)


Ah39:
1.AtomicReference<State> State = new AtomicReference()
state.compareAndSet

2.volatile

can fix the problem

Jaskey:
@Ah39 
The problem is not only the synchrinaztion on state's read/write, but on `shutdown` and `start` , `shutdown` and `start` should be syncronized

For example, two thread may start in the same time and the the resources will be initized twice, actually the later one should be rejected, which is the problem I issue.

Ah39:
**# apache/curator TreeCache Start** 

  public TreeCache start() throws Exception
    {
        **Preconditions.checkState(treeState.compareAndSet(TreeState.LATENT, TreeState.STARTED)**, "already started");
        if ( createParentNodes )
        {
            client.createContainers(root.path);
        }
        client.getConnectionStateListenable().addListener(connectionStateListener);
        if ( client.getZookeeperClient().isConnected() )
        {
            root.wasCreated();
        }
        return this;
    }

Jaskey:
@Ah39 
I guess you still not understand my concern or I don't get it, you may try post your suggested ways and lets discuss it , a simpler solution is certainlly better if we have.

zhouxinyu:
IMO, consider that `start/shutdown` is not a high frequency usage method, use `synchronized` may be more graceful and readable.

please @lizhanhui @shroman help review at your convenience.

shroman:
I also think that using `synchronized` fits well here.
But I would do it in the following way:
1. Make `serviceState` volatile
```java
private volatile ServiceState serviceState;
```
2. Synchronize only blocks that needs to be atomic (possibly making synchronized methods -- current huge `start` chunk of code doesn't look nice)

And, of course, provide comments to public methods ;)

lizhanhui:
As pointed out, I agree marking shutdown and start `synchronized` to keep code as concise as possible.

vongosling:
@Jaskey Could you continue to polish this PR as we have pointed out :-)

Jaskey:
@shroman @vongosling 

Do all guys consider makeing method synchrinized which is locking the consumer object is a better way instead of a more fine-grained lock.

If so, I will do the following changes:

1. Making state as votilte 
2. synchronize `start` and `shutdown`  and `setServiceState` method

shroman:
Probably you don't want to synchronize `setServiceState`. For other methods, it's only large chunks like that of `CREATE_JUST` the has to be synchronized ;)

Jaskey:
@shroman 
Actully setServiceState should be synchroinzed in my opnion, say if I am trying to shutdown, and some other thread B is trying to setServiceState ,which should be waiting for `shutdown` , or shutdown may not be completed since only `RUNNIG` state can be really shutdowned .

If thread B is setting the state as SHUTDOWN_AREADY afterwards, shutdown will actually do nothing. So with the setServiceState, I do think synchrinization is needed, but I do NOT think `setServiceState` should be public which is dangerous

shroman:
@Jaskey I agree that `setServiceState` should be deprecated. I overlooked it was _public_.

Jaskey:
@lizhanhui @zhouxinyu @shroman 
Any more advice on this pr? and when can this be merged?

zhouxinyu:
Merged, thanks @Jaskey , please help close this PR.

Jaskey:
Thank you @zhouxinyu

PR #72, opened 2017-03-08
[ROCKETMQ-134]the offset of message filter by tags may be not commit to broker

StyleTang:
JIRA:https://issues.apache.org/jira/browse/ROCKETMQ-134
when different string has a same hash code.the message commit offset of filtered message may be not commit to broker.
for example:
1.consumer pull message from broker, broker return status FOUND and messages filter by tags hash code
2.consumer client get the messages and than processPullResult will filter message by tags.
3.PullCallback may get a pullResult which status is FOUND but messageList is empty.(filter by tags)
but only NO_MATCHED_MSG and NO_NEW_MSG will correctTagsOffset
we can't commit the right with status of FOUND(for messageList is empty).
Is that so?

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImpl.java
@@ -280,6 +280,12 @@ public class DefaultMQPushConsumerImpl implements MQConsumerInner {
 
                     switch (pullResult.getPullStatus()) {
                         case FOUND:
+                            if (pullResult.getMsgFoundList().size() == 0) {
+                                pullRequest.setNextOffset(pullResult.getNextBeginOffset());
+                                DefaultMQPushConsumerImpl.this.correctTagsOffset(pullRequest);
+                                DefaultMQPushConsumerImpl.this.executePullRequestImmediately(pullRequest);
+                                break;
+                            }
                             long prevRequestOffset = pullRequest.getNextOffset();
                             pullRequest.setNextOffset(pullResult.getNextBeginOffset());
                             long pullRT = System.currentTimeMillis() - beginTimestamp;


Jaskey:
You are right , but this is actually not a very big problem since though the offset this round is not committed, the pull request for next round is still continue so next time the offset will still committed.

If the consumer crashes at that moment, an unnecessary pull request will be submitted, which is the what the problem actually is , but it is very minor.

Please firstly relate the JIRA in your descriptions.

For this PR, I don't suggest doing correctTags when pull list is empty because it is always empty as long as there is no message.

If this issue should be solved, I suggest modify the pull result to NO_MATCHED when pulling some message but filter by client by tag in method `processPullResult`

            //filter by client again and no messages left
            if(msgListFilterAgain.isEmpty() && !msgList.isEmpty()) {
                pullResultExt.setPullStatus(PullStatus.NO_MATCHED_MSG);
            }

StyleTang:
@Jaskey 
I found pullStatus is final, and I don't want to redefine it, so I judge the msgList size.
may be change the value of pullStatus will be better.

As you say, the problem occurred when your consumer client get messages every pull request, and the message can be filter by tag.

The probability is very low.If everyone think it doesn't matter,I will close this PR.

But I think have a check is better.

client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImpl.java
```diff
@@ -280,6 +280,12 @@ public void onSuccess(PullResult pullResult) {
 
                     switch (pullResult.getPullStatus()) {
                         case FOUND:
+                            if (pullResult.getMsgFoundList().size() == 0) {
```
dongeforever: It maybe confusing without comments.
For if the status is FOUND, it is expected that the msgList size will be greater than 0.

zhouxinyu:
@vsair Please help check this PR.

vsair:
Hi, @StyleTang  I think the broker should keep the logic that status FOUND indicate there are messages in pull result. Do you have some tests show the wrong logic?

dongeforever:
@vsair IMO, this situation will appear when the hash collision happened.
The broker respond the status FOUND means some messages in pull result, but these messages may be filtered by the client.

vongosling:

PR #75, opened 2017-03-10
[ROCKETMQ-138]Remove hard coded Aliyun authentication code

Jaskey:
JIRA:https://issues.apache.org/jira/browse/ROCKETMQ-138

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -126,11 +126,8 @@ public abstract class NettyRemotingAbstract {
                             }
                         }
                     } catch (Throwable e) {
-                        if (!"com.aliyun.openservices.ons.api.impl.authority.exception.AuthenticationException"
-                            .equals(e.getClass().getCanonicalName())) {
-                            PLOG.error("process request exception", e);
-                            PLOG.error(cmd.toString());
-                        }
+                        PLOG.error("process request exception", e);
+                        PLOG.error(cmd.toString());
 
                         if (!cmd.isOnewayRPC()) {
                             final RemotingCommand response = RemotingCommand.createResponseCommand(RemotingSysResponseCode.SYSTEM_ERROR, //


remoting/src/main/java/org/apache/rocketmq/remoting/exception/AuthenticationException.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.remoting.exception;
+
+/**
+ * Used when authorization failure , when Remoting component catches this exception, will ignore it.
```
shroman: Let's not have "when Remoting component catches this exception, will ignore it." here, because it describes how it is used in NettyRemotingAbstract, not the class itself or its properties.
The description of what this class is and its role is sufficient.

shroman:
:+1: to this fix.

remoting/src/main/java/org/apache/rocketmq/remoting/exception/AuthenticationException.java
Jaskey: OK, I will modify it later. 

Besides, just a small questions about the commit process, should I squash the commits and recommit or just append a commit to let the admin to squash them?OK, I will modify it later. 

Besides, just a small questions about the commit process, should I squash the commits and recommit or just append a commit to let the admin to squash them?

shroman: Normally, a committer can easily squash it (I don't exclude request to you to do it in some cases -- too many modifications, etc.) -- no problem, go ahead and just make your modifications.

vongosling:
please notice your PR topic

Jaskey:
@vongosling @shroman
 please review the updated  pr

zhouxinyu:
@lollipopjin Please help review this PR.

vongosling:
LGTM

lollipopjin:
@Jaskey In my opinion, it' better to just remove the    

> if (!"com.aliyun.openservices.ons.api.impl.authority.exception.AuthenticationException" .equals(e.getClass().getCanonicalName())) 

and record the log in any case throwable.

vongosling:
Yep agreed. We can remove the instanceof check and just log it.

Jaskey:
Do you guys really consider that is good enough?? @vongosling @lollipopjin 

IMO, the origin design is OK, since the AuthenticationException is thrown from the hook.

The devs will log their own Authentication fail log already (may be print periodically or somehow ), but rocketmq remoting should only care about the real error.

If we remove it, all authentication fail will have one log, and this could be possibly  very often and fill up with the remoting log, which I do not think it is good enough. 

User should care about their custom authentication log, and remoting component should care about real error thrown from the processor.

vongosling:
If we remove it, all authentication fail will have one log, and this could be possibly very often and fill up with the remoting log, which I do not think it is good enough.

We can just keep it as the previous. We will introduce a new remoting module in the 4.1 with seriously consideration about exception mechanism~

Jaskey:
@vongosling OK, look forward to that! 

So when will this pr be merged?

vongosling:
Could you remove the code snippet “ (!(e instanceof AuthenticationException)) {” and class AuthenticationException

Jaskey:
So do you mean we just remove the hard code snippet in the pr and just make it log for that? Better solution will be introduced in 4.1.x in another patches?

shroman:
If no special consideration for authentication errors is needed (which was obviously the intention in the original code), +1 for just removing the hard-coded snippet. No `AuthenticationException` is needed then.
I see no problem logging on every `Throwable`.

vongosling:
@Jaskey Thanks for your elaborative consideration about exception. Let it go as you have changed. we will merge this PR.

Jaskey:
@vongosling 

Now nothing special has been done unless remove the hard corded aliyun code snippet.

I will close this pr after merge.

Jaskey:
It seems this pr is not merged yet, so I reopen it.

zhouxinyu:
Done, please help close this PR. @Jaskey

PR #78, opened 2017-03-13
[ROCKETMQ-141]Make producer client connect to new-joined brokers eagerly

lizhanhui:
When new brokers joins the cluster, MQ producer clients polls name server and add the new-joined brokers to topicPublishInfo in case the new-joined broker bears the topic. Later on, the producers may send messages to the new-joined brokers.
This achieves part of scalable goals and works fine for most cases. However, we ran an issue in this process. The problem is when new broker joins the cluster, the producer clients blocks for quite a while even if when asynchronous send method is employed, which is very miserable for latency sensitive scenarios.
After analyzing the root cause, it turns out that producer clients need to establish a connection to the new-joined brokers the first time it sends a message, as is blocking.
This issue is to establish a connection to new-joined brokers immediately after polling name server and make them available to send methods thereafter. Thus, when send is called against new-joined brokers, there is a writable channel readily available.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
+++ b/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
@@ -32,7 +32,7 @@ public class ClientConfig {
     /**
      * Pulling topic information interval from the named server
      */
-    private int pollNameServerInteval = 1000 * 30;
+    private int pollNameServerInterval = 1000 * 30;
     /**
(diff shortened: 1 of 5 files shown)


lizhanhui:
Guys, any idea on this PR?

vongosling:
@lizhanhui I will close the pr, if you happened to the same question, please let me know.

PR #86, opened 2017-04-01
[ROCKETMQ-160]SendHeartBeat may not be logged in the expected period

Jaskey:
JIRA:https://issues.apache.org/jira/browse/ROCKETMQ-160

A issue will be found that  every time  the times will increase even though acatully no heartbeat is sent, which will cause the very first heartbeat is not logged since the first time it is very possible to have no heartbeat data.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -112,7 +112,7 @@ public class MQClientInstance {
     private final RebalanceService rebalanceService;
     private final DefaultMQProducer defaultMQProducer;
     private final ConsumerStatsManager consumerStatsManager;
-    private final AtomicLong storeTimesTotal = new AtomicLong(0);
+    private final AtomicLong sendHeartbeatTimesTotal = new AtomicLong(0);
     private ServiceState serviceState = ServiceState.CREATE_JUST;
     private DatagramSocket datagramSocket;
     private Random random = new Random();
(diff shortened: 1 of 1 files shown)


Jaskey:
@lizhanhui @zhouxinyu 
Any guys come to review this trival changes?

Jaskey:
@shroman 

For your first point, I would like to remain the existing flow? Because the heart beat sent may be failed and if we only mark the success one, we maybe miss the log for long turn when send heartbeat is failed for many times and then become success.  But anyway this is a minor issue , I will follow your advice if you guys think it is better.

For your second point, I agree and I will proceed.

zhouxinyu:
IMO, log SendHeartBeat records of `all brokers` every 20 times is reasonable.

Jaskey:
@zhouxinyu , hmm, it depends on what we need.

The issue I post is that , the log is not doing as expected when the addr table is empty. Even though log for all broker is reasonable, a empty check will be needed. 

I will follow your advice after more guys review.

shroman:
@Jaskey Incrementing and logging heartbeats on success seems reasonable to me. All failures are logged in `catch` clause.

Jaskey:
@lizhanhui  yes it seems reasonable to send heartbeat for all broker as one time.

Please review the updated pr which only
1. check if the beartbeat table is empty, only log and calculate times when it is not empty
2. rename the heartbeat times variable

vsair:
IMO, it's good and reasonable that counter should record the heartbeat's real request.

Jaskey:
@dongeforever 
hiw does this pr go, is it been merged？

dongeforever:
@Jaskey Yeah. it has been merged

PR #87, opened 2017-04-01
[ROCKETMQ-161] Update runbroker.sh and runserver.sh to support user defined jvm mem flag

dongeforever:
https://issues.apache.org/jira/browse/ROCKETMQ-161?filter=-1

### DIFF
--- a/distribution/bin/runbroker.sh
+++ b/distribution/bin/runbroker.sh
@@ -46,6 +46,7 @@ JAVA_OPT="${JAVA_OPT} -XX:MaxDirectMemorySize=15g"
 JAVA_OPT="${JAVA_OPT} -XX:-UseLargePages -XX:-UseBiasedLocking"
 JAVA_OPT="${JAVA_OPT} -Djava.ext.dirs=${BASE_DIR}/lib"
 #JAVA_OPT="${JAVA_OPT} -Xdebug -Xrunjdwp:transport=dt_socket,address=9555,server=y,suspend=n"
+JAVA_OPT="${JAVA_OPT} ${JAVA_OPT_EXT}"
 JAVA_OPT="${JAVA_OPT} -cp ${CLASSPATH}"
 
 numactl --interleave=all pwd > /dev/null 2>&1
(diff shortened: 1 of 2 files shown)


dongeforever:
@shroman please help review and check it.

vongosling:
@shroman @lizhanhui Could you help us to review this long-time delayed PR

vesense:
LGTM +1

shroman:
@dongeforever Sorry, I don't think this modification is needed. A user will modify this file anyway in order to match his/her system requirements.

Also, there are even more flags, and if you take this approach, you may have to add more variables for each in future, which will make the script tricky.

vsair:
When running in docker,  this would be useful.  Users just want to run a docker demo, but the computer could not allocate 8g memory, and there is no way to define the JVM flags in docker's script. OMG!

shroman:
Ok, but let's set it through `JVM_OPTS` instead of introducing a variable for that specific setting.

netroby:
+1 JVM_OPTS, Do not try to introducing new variable.

netroby:
Or just check JAVA_OPTS have already defined Xmx

netroby:
http://stackoverflow.com/questions/229551/string-contains-in-bash

vsair:
+1, maybe it's better to merge JAVA_OPTS

dongeforever:
The intention is to make user-defined JVM flags to replace the initial ones. 
So just put JAVA_OPT_EXT after JAVA_OPT is enough.
The later defined JVM parameters will replace the previous ones. I have already tested it.

@vsair @netroby @shroman  how do you think

shroman:
@dongeforever Hmm, sorry I don't understand yet why `JAVA_OPTS` won't work. Can you please explain how/when it happens? :)

dongeforever:
@shroman In runbroker.sh and runserver.sh, some default options, such as -Xmx8g, are placed after the JVM_OPTS, so user-defined -Xmx will not take effect.

And if we put JAVA_OPT_EXT after all default options, it will replace the default ones.

shroman:
@dongeforever I see, thank you!
How about just changing the code to
```
if [ -z "$JVM_OPTS" ] ; then
  #default_options
  JAVA_OPT="${JAVA_OPT} -server -Xms8g -Xmx8g -Xmn4g"
fi
```
?
But if you guys think we need `JAVA_OPT_EXT`, please go ahead ;)

dongeforever:
@shroman I insist on using JAVA_OPT_EXT, for that we may not only want to override -Xms -Xmx -Xmn, but also other jvm flags.
We could use JAVA_OPT_EXT once and for all.

lizhanhui:
It's OK to introduce an extra variable to override default Java VM options.  

+1

PR #88, opened 2017-04-06
[ROCKETMQ-165]Maximum pull batch size hard-coded restricted

lizhanhui:


lizhanhui:
Merged as this is really trivial.

PR #89, opened 2017-04-06
[ROCKETMQ-166] onException callback may capture compressed message body

lizhanhui:
If message body size exceeds specified threshold, client would try to compress the message body.
Here there are two issues: 1) current implementation changes message body directly, which is not a good practice; 2) if asynchronous send method is employed to deliver message, when onException is invoked, the callback may capture the compressed message body before the finally block restores it.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.client.impl;
 
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


lizhanhui:
Comment to bump up this PR.

vongosling:
@shroman @stevenschew

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -306,6 +310,19 @@ public SendResult sendMessage(//
...
     ) throws RemotingException, MQBrokerException, InterruptedException {
+
+        byte[] msgBody = msg.getBody();
+        Integer sysFlag = requestHeader.getSysFlag();
+        if (null != sysFlag && ((sysFlag & MessageSysFlag.COMPRESSED_FLAG) == MessageSysFlag.COMPRESSED_FLAG)) {
+            try {
+                msgBody = UtilAll.compress(msgBody, zipCompressLevel);
+            } catch (IOException e) {
+                requestHeader.setSysFlag(MessageSysFlag.clearCompressedFlag(sysFlag));
+                log.error("tryToCompressMessage exception", e);
```
shroman: What is the reason to have ERROR and WARN here?

lizhanhui: You mean why we log error on failing to compress message body?

lizhanhui:
@stevenschew Any idea?

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
shroman: I just wonder why both WARN and ERROR are needed.

lizhanhui: Got your point.  

1. The code is moved here, and same log level should be applied IMO. @vongosling @vintagewang  any particular reason behind this?

vongosling:
@lizhanhui I will close the pr, if you happened to the same question, please let me know.

PR #91, opened 2017-04-12
[ROCKETMQ-174]Fix spelling errors

lizhanhui:
This is to fix large number of spelling errors.


vongosling:
@shroman @zhouxinyu Could you help review this PR :-)

CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ To submit a change for inclusion, please do the following:
 
 #### If the change is non-trivial please include some unit tests that cover the new functionality.
 #### If you are introducing a completely new feature or API it is a good idea to start a wiki and get consensus on the basic design first.
-#### It is our job to follow up on patches in a timely fashion. Nag us if we aren't doing our job (sometimes we drop things).
+#### It is our job to follow up on patches in a timely fashion. Nag us if we arn't doing our job timely enough (sometimes we may miss things).
```
vesense: arn't -> aren't ?

vesense:
Overall looks good to me. Nice job!

CONTRIBUTING.md
lizhanhui: Yes, aren't is right.

zhouxinyu:
The JIRA issue is here: https://issues.apache.org/jira/browse/ROCKETMQ-174
And I have added a time tracker for this issue.

Jaskey:
@zhouxinyu @vesense 
for some public class or enum, I don't agree we rename it directly since it will introduce the compatibility problem for the user who upgrade from the older versions.

client/src/main/java/org/apache/rocketmq/client/producer/LocalTransactionExecutor.java
```diff
@@ -18,6 +18,6 @@
 
 import org.apache.rocketmq.common.message.Message;
 
-public interface LocalTransactionExecuter {
-    public LocalTransactionState executeLocalTransactionBranch(final Message msg, final Object arg);
+public interface LocalTransactionExecutor {
```
lizhanhui: 1) This project is still at incubation stage, if we do not correct things now, we may never ever get the chance to fix them afterwards.
2) Transaction related feature is not currently implemented, IMO, we may fix this interface before it's widely used.

Jaskey: So in the next version, we must announce that they may face the compatible problem when they use XXXX in the older version.

Before they update, they must know about this and know how to fix them , and then upgrade.

lizhanhui: This is why I make these changes thorough PR process instead of regarding them trivial and merging directly.

We can discuss to figure out an acceptable way.

You may also notice that I do not change configuration name spelling issues yet and I am expecting opinions too.

Jaskey: For cinfiguration issue, we can fix them by pointing them to the same right spelling config.

    private int rightConfig;
       @Deprecated    
    public int getTypoConfig(){
         return rightConfig;
    }

    @Deprecated
    public void setTypoConfig(int val) {
        this.rightConfig = val;
    }

    getter/setter for right config...


But since client config and server config is working in tatolly different ways that server config uses reflection to load configurations so this should be only work for client configuration typo issue. 

I suggest we post another two typo issues for client configuration and server configuration.

common/src/main/java/org/apache/rocketmq/common/protocol/header/GetEarliestMsgStoreTimeRequestHeader.java
```diff
@@ -16,15 +16,15 @@
  */
 
 /**
- * $Id: GetEarliestMsgStoretimeRequestHeader.java 1835 2013-05-16 02:00:50Z vintagewang@apache.org $
+ * $Id: GetEarliestMsgStoreTimeRequestHeader.java 1835 2013-05-16 02:00:50Z vintagewang@apache.org $
```
shroman: Why do we need this line?

common/src/main/java/org/apache/rocketmq/common/protocol/header/GetEarliestMsgStoreTimeResponseHeader.java
```diff
@@ -16,15 +16,15 @@
  */
 
 /**
- * $Id: GetEarliestMsgStoretimeResponseHeader.java 1835 2013-05-16 02:00:50Z vintagewang@apache.org $
+ * $Id: GetEarliestMsgStoreTimeResponseHeader.java 1835 2013-05-16 02:00:50Z vintagewang@apache.org $
```
shroman: Why do we need this line?

lizhanhui: It should not be here.

shroman: There are many files with such lines. If you don't mind, let's have them removed with this commit ;)

shroman:
Guys, let's merge this fix after the conflicts are resolved.
I also don't like breaks in minor versions, but there are too many spelling mistakes and writing workarounds for each is hard. Let's do it while we are in incubating phase, and make an announcement in release notes.
Any other ideas?

vongosling:
Agreed. Any other opinion for this PR? @lizhanhui please resolve the conflicts, we could merge this PR later~

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -907,10 +907,10 @@ public void queryMessage(
         final QueryMessageRequestHeader requestHeader,
         final long timeoutMillis,
         final InvokeCallback invokeCallback,
-        final Boolean isUnqiueKey
+        final Boolean isUnqueKey
```
vsair: is it isUniqueKey ?

common/src/main/java/org/apache/rocketmq/common/protocol/route/TopicRouteData.java
```diff
@@ -27,23 +27,23 @@
 
 public class TopicRouteData extends RemotingSerializable {
     private String orderTopicConf;
-    private List<QueueData> queueDatas;
-    private List<BrokerData> brokerDatas;
+    private List<QueueData> queueDataList;
+    private List<BrokerData> brokerDataList;
```
vsair: It is modification of protocol, IMO, it's improper for upgrade of server or client.

common/src/main/java/org/apache/rocketmq/common/protocol/header/QueryMessageResponseHeader.java
```diff
@@ -28,7 +28,7 @@
     @CFNotNull
     private Long indexLastUpdateTimestamp;
     @CFNotNull
-    private Long indexLastUpdatePhyoffset;
+    private Long indexLastUpdatePhyOffset;
```
vsair: Protocol modification.

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
lizhanhui: Thanks for point it out, unique key.

common/src/main/java/org/apache/rocketmq/common/protocol/route/TopicRouteData.java
lizhanhui: Right. looks we have to keep it intact.

lizhanhui:
> please resolve the conflicts firstly, we have to be cautious about this PR.

Agree, this PR introduces too many trivial changes and I'd like abandon this PR and create several small changes, applying  divide-and-conquer strategy

PR #92, opened 2017-04-14
[ROCKETMQ-175] Consumer may miss messages because of inconsistent sub…

vsair:
For push mode, update subscription version when rebalance result changed, then send heart beat to broker immediately.

Maybe subscriptions management mechanism could be re-designed to meet more complicated scenarios.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
@@ -30,6 +30,7 @@ import org.apache.rocketmq.common.consumer.ConsumeFromWhere;
 import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.common.protocol.heartbeat.ConsumeType;
 import org.apache.rocketmq.common.protocol.heartbeat.MessageModel;
+import org.apache.rocketmq.common.protocol.heartbeat.SubscriptionData;
 
 public class RebalancePushImpl extends RebalanceImpl {
     private final static long UNLOCK_DELAY_TIME_MILLS = Long.parseLong(System.getProperty("rocketmq.client.unlockDelayTimeMills", "20000"));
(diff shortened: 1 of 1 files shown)


Jaskey:
Actually missing message can not be avoided when the diffrerent tags are subscribed in the same consumer groups.

Say for two queues q1 q2, two consumers c1 c2.
c1->TAG1
c2->TAGB

when allocating, c1->q1, c2->q2 100 messages are sent to this topic with TAG1.

C2 will recive nothing and c1 will recieve 50 messages. but the consumer offset of q2 is till moving on. Then 50 messages are missing.

vsair:
This PR does't break the restrict that all consumers should have the same subscription. It only focus on broker should hold the online consumer's subscription. Miss messages is just one result caused by inconsistency subscriptions.

Jaskey:
@vsair 
What I mean is that the scene this pr is going to fix is breaking the restrict that the same consumer group has the same subscriptions.

Though this may help improve the refresh the subscriptions, but the same problem is still there remains  no solution in the existing design.

IMO, we should solve this problem directly to solve all the related problem. Which is issued in https://issues.apache.org/jira/browse/ROCKETMQ-94

vsair:
Yeah, you're right. There is no way to solve the problem in existing design.
But the result is : 
when user stop the consumer has diff subscription, broker could hold the correct info.
I also look forward to the solution in https://issues.apache.org/jira/browse/ROCKETMQ-94.

dongeforever:
@Jaskey It may take a long time to fix the issue ROCKETMQ-94.
And we could improve it step by step.
Now one consumer startup will refresh the subscriptions, but the consumer shutdown does not.
This maybe confusing.
IMO, it is reasonable to refresh when rebalance occurs.

dongeforever:
@zhouxinyu @vongosling

dongeforever:
This PR is merged  @vsair

vsair:
OK

CoderLan0668:
so what's the solution for this problem? it's 2019 now, the problem is still on.

PR #95, opened 2017-04-20
[ROCKETMQ-184]-It takes too long(3-33 seconds) to switch to read from slave when master crashes

Jaskey:
JIRA:https://issues.apache.org/jira/browse/ROCKETMQ-184?jql=project%20%3D%20ROCKETMQ

Problem, no listener is triggered when Chanel is close. 

When async command sent to the server, and the server is crash before sending response to client, the callback can not be invoked in time. Instead, the callback can only be triggered by the timeout scan service. 

This is obvious for pulling message since the timeout is by default 30 seconds. So if master crashes before process response to the client, the client can not repull until scan service tell it, which takes at most 30 seconds. And repull will have 3 seconds delay, so the HA to read from slave has to take 3-33 seconds when this problem occurs.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -357,7 +357,7 @@ public abstract class NettyRemotingAbstract {
         final int opaque = request.getOpaque();
 
         try {
-            final ResponseFuture responseFuture = new ResponseFuture(opaque, timeoutMillis, null, null);
+            final ResponseFuture responseFuture = new ResponseFuture(channel, opaque, timeoutMillis, null, null);
             this.responseTable.put(opaque, responseFuture);
             final SocketAddress addr = channel.remoteAddress();
             channel.writeAndFlush(request).addListener(new ChannelFutureListener() {
@@ -400,8 +400,7 @@ public abstract class NettyRemotingAbstract {
         boolean acquired = this.semaphoreAsync.tryAcquire(timeoutMillis, TimeUnit.MILLISECONDS);
         if (acquired) {
             final SemaphoreReleaseOnlyOnce once = new SemaphoreReleaseOnlyOnce(this.semaphoreAsync);
-
-            final ResponseFuture responseFuture = new ResponseFuture(opaque, timeoutMillis, invokeCallback, once);
+            final ResponseFuture responseFuture = new ResponseFuture(channel, opaque, timeoutMillis, invokeCallback, once);
             this.responseTable.put(opaque, responseFuture);
             try {
                 channel.writeAndFlush(request).addListener(new ChannelFutureListener() {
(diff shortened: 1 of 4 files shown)


remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -306,29 +307,35 @@ public void invokeAsyncImpl(final Channel channel, final RemotingCommand request
         if (acquired) {
             final SemaphoreReleaseOnlyOnce once = new SemaphoreReleaseOnlyOnce(this.semaphoreAsync);
 
-            final ResponseFuture responseFuture = new ResponseFuture(opaque, timeoutMillis, invokeCallback, once);
+            final GenericFutureListener<ChannelFuture> chanelCloseListener = new ChannelFutureListener() {
```
lizhanhui: chanel-->channel

lizhanhui:
invokeOneway should also be updated if so, otherwise, it's likely to leak semaphore.

lizhanhui:
Thanks @Jaskey, this is indeed a good place to improve. 

For the implementation, I suggest an alternative generic way, instead of add a close future for each request, we add the opaque integer into a collection per channel. Remove the opaque integer on response or invalidate all of them in NettyConnectManageHandler. Suggested approach has fewer memory footprint and we may also easily cover the sync request scenario -- respond earlier before `timeoutMillis` amount of time elapsed in case channel experiences problems.

Jaskey:
@lizhanhui 

I have considered that, which will takes more efforts to achieve the same goal. Since I need to change some structure to make the connect manager to get access to the responseFuture map.

For my first implementation, I just want to issue this problem and involve you guys to discuss. 

If you think that is indeed a better approach, I will submit an updated implementations for that, and then let all guys to choose.

lizhanhui:
You are right that more changes are required for my suggested approach. But, IMO, the suggested way is more unified in design and may also save a few memory footprint in case we have very large semaphore initial capacity.

Indeed, this is the place we need to enhance.

Let's bring more guys into discussion before you implement the suggested approach.  They should easily conceive what's going on here via checking changes made in your PR. Any opinion on this issue? @zhouxinyu @shroman @vongosling

zhouxinyu:
Thanks @Jaskey , agree with @lizhanhui , use NettyConnectManageHandler to handle this close event is a better way, no need to import two mechanisms.

Jaskey:
@zhouxinyu @lizhanhui 
please review the updated pr

lizhanhui:
> please review the updated pr

Looks this PR is not updated. Do you forget to push your changes?

Jaskey:
@lizhanhui 

It has been refactored using the close event handler mechanism according to your advice, please review the pr from `close` callback in  `NettyRemotingClient.java`

Jaskey:
@lizhanhui @vsair @zhouxinyu @shroman 

any ideas for this updated solution?

lizhanhui:
+1

dongeforever:
LGTM @zhouxinyu

Jaskey:
@zhouxinyu @vongosling @shroman what's your advice, can this pr be merged?

Jaskey:
This pr is not updated from the source for long, so I updated just now , please review.

I think this is a good improvement for HA

PR #96, opened 2017-04-21
[ROCKETMQ-187]Measure the code coverage for Integration Tests, and add sonar-apache profile

dongeforever:
https://issues.apache.org/jira/browse/ROCKETMQ-187

Now we could browse the Unit Tests and IT Tests at 
https://builds.apache.org/analysis/component_measures/?id=org.apache.rocketmq%3Arocketmq-all
But the IT Test coverage is not correct. It should cover the original sources instead of the the classes in test module.
As for as I known, the coverage report is generated by matching the collected data(often using java agent) against a set of classes (the module classes compiled from src/main/). you could refer to: http://olafsblog.sysbsb.de/measuring-test-coverage-of-integration-tests-for-separated-modules-with-jacoco/
So we could match the jacoco-it.exec to each module's source classes to get the correct IT coverage report.
By the way, we'd better exclude the classes in the test module.

we may use sonar locally.  
So we'd better use profile to handle different occasions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ script:
 
 after_success:
   - mvn clean install -Pit-test
-  - mvn sonar:sonar
+  - mvn sonar:sonar -Psonar-apache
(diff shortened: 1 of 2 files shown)


dongeforever:
@lizhanhui @zhouxinyu @vongosling please have a review

vongosling:
LGTM

dongeforever:
The PR is better not to be merged by the original author.
@lizhanhui @zhouxinyu @shroman please help merge this PR.

lizhanhui:
It's good and looks it's already been merged.

PR #100, opened 2017-04-28
[ROCKETMQ-191] Fix socket options

lizhanhui:
As we are setting socket options(sndbuf/recvbuf size) for the future SocketChannel, we need to use childOption for ServerBootstrap class.

Semantics differences can be found [Netty docs](http://netty.io/wiki/user-guide-for-4.x.html)

Actually, this is an easy-to-get-confused issue, [Spark had a similar bug too](https://issues.apache.org/jira/browse/SPARK-3502)

[Vertx has similar settings that we can refer to](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/net/impl/NetServerBase.java) See the `applyConnectionOptions` method.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
@@ -26,7 +26,9 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.channel.epoll.Epoll;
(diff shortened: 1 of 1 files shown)


vongosling:
That's a good polish, but I must point out the 2 point for this optimization. 1. please comply with PR best practice for title as you have corrected before. 2. you can consider the order when many logic and operator together, such as 0&&1&&1(similar 1||0||0, nor 0||0||1), making your logic more effective. For this context, linux platform and netty epoll judge, which one happen to 0 more easily.

lizhanhui:
> please comply with PR best practice for title as you have corrected before. 
Yes. Fixed.
> 2. you can consider the order when many logic and operator together
What you said is definitely reasonable, but this method is actually a single shot run, rendering optimization sort of unnecessary.

lizhanhui:
Merged.

PR #104, opened 2017-05-17
[ROCKETMQ-201]Remove spaces before and after the properties

huangyiminghappy:
Remove spaces before and after the properties,Avoid errors due to spaces before and after the properties of data,here is the JIAR link:https://issues.apache.org/jira/browse/ROCKETMQ-201 ,thank you!

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/MixAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/MixAll.java
@@ -52,7 +52,8 @@ public class MixAll {
     public static final String NAMESRV_ADDR_ENV = "NAMESRV_ADDR";
     public static final String NAMESRV_ADDR_PROPERTY = "rocketmq.namesrv.addr";
     public static final String MESSAGE_COMPRESS_LEVEL = "rocketmq.message.compressLevel";
-    public static final String WS_DOMAIN_NAME = System.getProperty("rocketmq.namesrv.domain", "jmenv.tbsite.net");
+    public static final String DEFAULT_NAMESRV_ADDR_LOOKUP = "jmenv.tbsite.net";
+    public static final String WS_DOMAIN_NAME = System.getProperty("rocketmq.namesrv.domain", DEFAULT_NAMESRV_ADDR_LOOKUP);
     public static final String WS_DOMAIN_SUBGROUP = System.getProperty("rocketmq.namesrv.domain.subgroup", "nsaddr");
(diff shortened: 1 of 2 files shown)


huangyiminghappy:
if i use the file like properties 
image
image

the properties's property has any space,it will get an Not expected value, Numeric possible type conversion error，but The string may get a bug that is not expected,Very difficult to troubleshoot

Jaskey:
Thank you. Please notice your topic and also relate the issue to the jira link https://issues.apache.org/jira/browse/ROCKETMQ.

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -339,6 +339,8 @@ public static void properties2Object(final Properties p, final Object object) {
                     String key = first.toLowerCase() + tmp;
                     String property = p.getProperty(key);
                     if (property != null) {
+                        // remove spaces before and after
```
Jaskey: I think the purpose is clean enough so I don't think the comment is needed here for `trim()`

huangyiminghappy: ok,i remove the comment

Jaskey:
+1

common/src/main/java/org/apache/rocketmq/common/MixAll.java
Jaskey: what about we also trim the method `public static String properties2String(final Properties properties)` ?

dongeforever:
+1

dongeforever:
IMO. if using trim, it is better to trim in all related places

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -279,7 +279,7 @@ public static String properties2String(final Properties properties) {
         StringBuilder sb = new StringBuilder();
         for (Map.Entry<Object, Object> entry : properties.entrySet()) {
             if (entry.getValue() != null) {
-                sb.append(entry.getKey().toString() + "=" + entry.getValue().toString() + "\n");
+                sb.append(entry.getKey().toString().trim() + "=" + entry.getValue().toString().trim() + "\n");
```
huangyiminghappy: ok，I try it, thanks

huangyiminghappy: @zhouxinyu   i have remove  the above change ,only change the properties's trim(), thank you review it

PR #117, opened 2017-06-07
[ROCKETMQ-215]-Use java 7 syntax to refator code base

dzyangxi:
JIRA:https://issues.apache.org/jira/browse/ROCKETMQ-215?jql=project%20%3D%20ROCKETMQ

### DIFF
--- a/logappender/src/main/java/org/apache/rocketmq/logappender/common/ProducerInstance.java
+++ b/logappender/src/main/java/org/apache/rocketmq/logappender/common/ProducerInstance.java
@@ -39,7 +39,7 @@ public class ProducerInstance {
 
     public static final String DEFAULT_GROUP = "rocketmq_appender";
 
-    private static ConcurrentHashMap<String, MQProducer> producerMap = new ConcurrentHashMap<String, MQProducer>();
+    private static ConcurrentHashMap<String, MQProducer> producerMap = new ConcurrentHashMap<>();
 
     private static String genKey(String nameServerAddress, String group) {
         return nameServerAddress + "_" + group;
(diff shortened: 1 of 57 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/cluster/ClusterListSubCommand.java
```diff
@@ -113,7 +113,7 @@ private void printClusterMoreStats(final DefaultMQAdminExt defaultMQAdminExt) th
         while (itCluster.hasNext()) {
             Map.Entry<String, Set<String>> next = itCluster.next();
             String clusterName = next.getKey();
-            TreeSet<String> brokerNameSet = new TreeSet<String>();
```
Jaskey: @dongeforever  

Do we still need to make rocketmq-tools be compatible  with jdk1.6? If yes, we can not do the changes in rocketmq-tools module.

logappender/src/main/java/org/apache/rocketmq/logappender/common/ProducerInstance.java
```diff
@@ -39,7 +39,7 @@
 
     public static final String DEFAULT_GROUP = "rocketmq_appender";
 
-    private static ConcurrentHashMap<String, MQProducer> producerMap = new ConcurrentHashMap<String, MQProducer>();
```
Jaskey: @dongeforever  Does logappender needs to be compatible with jdk 1.6 ?

tools/src/main/java/org/apache/rocketmq/tools/command/cluster/ClusterListSubCommand.java
lizhanhui: I think tools module should be consistent with broker module. There is no need to maintain low version compatibility.

dongeforever:
the client module and its dependencies should be compatible with 1.6

logappender/src/main/java/org/apache/rocketmq/logappender/common/ProducerInstance.java
lindzh: There is log4j2 implementation in logappender.Since log4j2 not supported jdk 1.6,logappender do not need to be compatible.

Jaskey:
@dongeforever  it looks no changes have been made to the modules which rocketmq-client depends on.

But what I am concerned is that should rocketmq-tools needs to be compatable with 1.6 since they may possibly integrate them with their application.

zhouxinyu:
Hi, @dzyangxi any update?

PR #118, opened 2017-06-13
[ROCKETMQ-28] Encrypt transmission layer.

lizhanhui:
JIRA issue is [here](https://issues.apache.org/jira/browse/ROCKETMQ-28)


client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
```diff
@@ -173,12 +177,20 @@ public void setVipChannelEnabled(final boolean vipChannelEnabled) {
...
+    }
+
     @Override
     public String toString() {
         return "ClientConfig [namesrvAddr=" + namesrvAddr + ", clientIP=" + clientIP + ", instanceName=" + instanceName
             + ", clientCallbackExecutorThreads=" + clientCallbackExecutorThreads + ", pollNameServerInterval=" + pollNameServerInterval
             + ", heartbeatBrokerInterval=" + heartbeatBrokerInterval + ", persistConsumerOffsetInterval="
             + persistConsumerOffsetInterval + ", unitMode=" + unitMode + ", unitName=" + unitName + ", vipChannelEnabled="
-            + vipChannelEnabled + "]";
+            + vipChannelEnabled + ", useTLS=" + useTLS + "]";
```
zhouxinyu: `ToStringBuilder` is a better choice for this long `toString` method.

common/src/main/java/org/apache/rocketmq/common/protocol/RequestCode.java
```diff
@@ -17,7 +17,9 @@
 
 package org.apache.rocketmq.common.protocol;
 
-public class RequestCode {
+import org.apache.rocketmq.remoting.protocol.RemotingSysRequestCode;
+
```
zhouxinyu: I am not sure it's a good idea that `RequestCode` in `common` module has a dependency `RemotingSysRequestCode` in `remoting` module.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/FileRegionEncoder.java
```diff
@@ -0,0 +1,76 @@
...
+import java.nio.channels.WritableByteChannel;
+
+/**
+ * <p>
+ *     By default, file region are directly transferred to socket channel which is known as zero copy. In case we need
+ *     to encrypt transmission, data being sent should go through the {@link SslHandler}. This encoder ensures this
+ *     process.
+ * </p>
+ */
+public class FileRegionEncoder extends MessageToByteEncoder<FileRegion> {
```
zhouxinyu: Actually, I didn't figure out any sense of `FileRegionEncoder` in SSL transmission, so could you please give some hints?

lizhanhui: FileRegionEncoder here is to disable Netty from applying zero-copy so that SSL handler in the pipeline may encrypt the payload.

common/src/main/java/org/apache/rocketmq/common/protocol/RequestCode.java
lizhanhui: ResponseCode in common module has already had remoting as its dependency as it extends RemotingSysResponseCode, which is a class of remoting module.

lizhanhui: In fact, we should remove RemotingSysRequestCode which is part of earlier StartTLS implementation and is no longer in use.

client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
lizhanhui: Number of fields here, IMO, is not that long. Optimization of none-hot method is largely unnecessary.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/FileRegionEncoder.java
zhouxinyu: You mean that SSL handler can't work properly without `FileRegionEncoder`?

client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
zhouxinyu: It's not optimization, just make the toString method more beautiful.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/FileRegionEncoder.java
lizhanhui: > You mean that SSL handler can't work properly without FileRegionEncoder?

It depends. 
If the data is in the user space memory, SSL handler works fine. If the code directly transfers data from file to network, similar to sendFile, SSL handler won't work as expected.

client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
lizhanhui: I do not get your point. You mean change "+" to a chain of append() method invocations?

dongeforever:
@lindzh It seems lack IT tests for TLS.  IMO,  it needs two IT Test at least.  one for old client which dose not support TLS, one for new.

client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
vsair: it looks good to use the  ToStringBuilder  of  commons-lang.

dongeforever:
OK. IT tests have been added.

vongosling:
LGTM

dongeforever:
@lizhanhui I want to merge this PR and release it in 4.2. please help to resolve the conflicts

dongeforever:
This PR has been merged

PR #119, opened 2017-06-13
[ROCKETMQ-223]-Rename DEFAULT_TOPIC

Jaskey:
JIRA: https://issues.apache.org/jira/browse/ROCKETMQ-223

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
@@ -66,9 +66,9 @@ public class TopicConfigManager extends ConfigManager {
             this.topicConfigTable.put(topicConfig.getTopicName(), topicConfig);
         }
         {
-            // MixAll.DEFAULT_TOPIC
+            // MixAll.AUTO_CREATE_TOPIC_KEY_TOPIC
             if (this.brokerController.getBrokerConfig().isAutoCreateTopicEnable()) {
-                String topic = MixAll.DEFAULT_TOPIC;
+                String topic = MixAll.AUTO_CREATE_TOPIC_KEY_TOPIC;
                 TopicConfig topicConfig = new TopicConfig(topic);
                 this.systemTopicList.add(topic);
                 topicConfig.setReadQueueNums(this.brokerController.getBrokerConfig()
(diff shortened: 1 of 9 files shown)


Jaskey:
@lizhanhui @zhouxinyu @shroman @dongeforever 

What's your opinions on this?

lizhanhui:
I am +1 to rename the default topic name.

In addition,  we'd better resolve all known issues relating the auto-create-topic mechanism.

1.  Can we rename the name to something to void potential topic name collision?
1. Currently, once auto create was enabled, the auto_create topic is persisted, even if we later disable auto-creation, the topic is still there which does not make sense. IMO, we may exclude this topic from persistence.

Jaskey:
@lizhanhui 

In my option for your consern # 2, I even consider that we separate the configuration logic from the default topic, since the clients may use different default key.

In my option, the below flow may be much more clearer:

1. Default topic is always created , just like the other preserved topic.
2. auto-creation configuration has no impact on this default topic, while it is a switch which enable the broker to support auto topic creation.

    if (autoCreateTopicEnable) { //broker allow to auto create topic
        if (topicKeyExist){ // exist the specified topic key
           if (isInherite(topic)) {  // topic has the right perm
                 log("topic key does not has the sufficient perm to auto create topic ")
           }
        }
       else { 
          log("The topic key does not exist")
       }
    }  else {
        log("broker does not allow to auto create topic")
    }

lizhanhui:
Looks good to me. Actually some of the client code relies on presence of the topic which would log exceptions otherwise.

IMO, we'd better rename the variable name only to maintain best backward compatibility.

shroman:
Another thing is -- do we need the default topic at all?

Jaskey:
@lizhanhui .
I think we'd better rename the topic name also since it has impact many users that they don't actually know what is this topic and hot to maintain it.

In the client code, the error of topic `defaultTopic` not exits will be ignored so it has no impact for the current users, unless that they will not be able to auto create topic if they only upgrade the broker but using the old client and they do not specify another topic key.

If needed , we can refactor the code flow on topic auto creation on another issue. For now, we just rename it to make it clear.

@shroman The reason I rename the variable is that the name `defaultTopic ` is confusing since it does not explain what it is used for and actually it is not a defaultTopic since it may not exist if auto creation is disabled.

shroman:
@Jaskey Ok, it makes sense.

lizhanhui:
All right.

Jaskey:
@lizhanhui @dongeforever 
any opinons? Can this pr be merged? please review

lizhanhui:
I am OK with this change.

PR #121, opened 2017-06-16
[ROCKETMQ-226]Remove the code that is not useful in the loop

huangyiminghappy:
Remove the code that is not useful in the loop

### DIFF
--- a/1
+++ b/1
@@ -1,0 +1,105 @@
+# This is a combination of 3 commits.
+# The first commit's message is:
+remove spaces before and after the properties
+
+# This is the 2nd commit message:
+
+remove spaces before and after the properties
+
+remove spaces before and after the properties
+
+remove spaces before and after the properties
+
+remove spaces before and after the properties
+
+Remove spaces before and after the properties
+
+Remove spaces before and after the properties
+
+Remove spaces before and after the properties
+
+Remove spaces before and after the properties
+
+Remove the code that is not useful in the loop
+
+Remove the code that is not useful in the loop
+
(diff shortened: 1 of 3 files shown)


huangyiminghappy:
https://issues.apache.org/jira/browse/ROCKETMQ-226

vongosling:
Could you follow standard PR（notice topic style） ? and squash your many commits ?

huangyiminghappy:
the  code ## FOR_SEG   is not useful,It is not multiple loop,You do not need to jump to the specified loop,so i suggest remove the FOR_SEQ code
` public FindBrokerResult findBrokerAddressInAdmin(final String brokerName) {
        String brokerAddr = null;
        boolean slave = false;
        boolean found = false;

        HashMap<Long/* brokerId */, String/* address */> map = this.brokerAddrTable.get(brokerName);
        if (map != null && !map.isEmpty()) {
            FOR_SEG:
            for (Map.Entry<Long, String> entry : map.entrySet()) {
                Long id = entry.getKey();
                brokerAddr = entry.getValue();
                if (brokerAddr != null) {
                    found = true;
                    if (MixAll.MASTER_ID == id) {
                        slave = false;
                        break FOR_SEG;
                    } else {
                        slave = true;
                    }
                    break;

                }
            } // end of for
        }

        if (found) {
            return new FindBrokerResult(brokerAddr, slave);
        }

        return null;
    }`

dongeforever:
LGTM @huangyiminghappy Thanks for your contribution.
@vongosling @zhouxinyu please help a review

PR #122, opened 2017-06-20
ROCKETMQ-214

lqjack:
wrapper the AllocateMessageQueueStrategy with ThreadLocal


lqjack:
later , i will append junit test to verify the function. 
if you have better solution , please reply.thanks .

Jaskey:
What do you want to fix by setting the allocate strategy as thread local?

lqjack:
yes

vongosling:
if any update for this pr, we will close it~

PR #125, opened 2017-06-20
ROCKETMQ-205 :

lqjack:
change the constant to variable


dongeforever:
@lqjack Thanks for you contribution. You may want to refactor the PR name and attach the jira issue in the comment.
It is better to add some detailed words to the PR name.

vongosling:
@lqjack what's your opinion

PR #126, opened 2017-06-23
[ROCKETMQ-231] fix pull consumer pull result size

lindzh:
When using PullConsumer pull message by default result size is 32,and messages is more than 32 in a queue,but broker always returns 31.


test/src/test/java/org/apache/rocketmq/test/client/consumer/pull/PullSizeTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.rocketmq.test.base.IntegrationTestBase;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.nio.channels.Pipe;
+import java.util.*;
+
+public class PullSizeTest extends BaseConf {
+
+    private static Logger logger = Logger.getLogger(PullSizeTest.class);
```
vongosling: Could we remove logger in unit test ?

test/src/test/java/org/apache/rocketmq/test/client/consumer/pull/PullSizeTest.java
```diff
@@ -0,0 +1,135 @@
...
+                if (sendResult.getSendStatus().equals(SendStatus.SEND_OK)) {
+                    successCount++;
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
+                Thread.sleep(500);
+            }
+        }
+        producer.shutdown();
+        if (successCount > 800) {
```
vongosling: why use this magic number 800 in your if statement

test/src/test/java/org/apache/rocketmq/test/client/consumer/pull/PullSizeTest.java
```diff
@@ -0,0 +1,135 @@
...
+                    default:
+                        break;
+                }
+            } catch (Exception e) {
+                logger.error("[SEND] send failed", e);
+                e.printStackTrace();
+            }
+        }
+        consumer.shutdown();
+        Assert.assertTrue(result);
```
vongosling: Could we assert the pull size is what you want, while not the simple true or false ?

test/src/test/java/org/apache/rocketmq/test/client/consumer/pull/PullSizeTest.java
zhouxinyu: IMO, we don't need a separate it test file to cover this point, use unit test.

Jaskey: @zhouxinyu aggree

test/src/test/java/org/apache/rocketmq/test/client/consumer/pull/PullSizeTest.java
lindzh: This test case has removed and add new unit test in store.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -1110,7 +1110,7 @@ private boolean isTheBatchFull(int sizePy, int maxMsgNums, int bufferTotal, int
             return false;
         }
 
-        if ((messageTotal + 1) >= maxMsgNums) {
```
Jaskey: I will suggest  replace all the condition of `messageTotal +1 > xxx` into 
    
    messageTotal >= xxx

lindzh: Yet,That a good idea to make code clean

vongosling: why change to -1 in your new change:-)

Jaskey: I will still suggest using `>=`

lindzh: Yes,this problem has been fixed to maxMsgNums <= messageTotal

vongosling:
Now, LGTM

lollipopjin:
+1

store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
```diff
@@ -45,19 +47,22 @@ public void init() throws Exception {
...
     }
 
+    public MessageStore buildMessageStore() throws Exception {
+        MessageStoreConfig messageStoreConfig = new MessageStoreConfig();
+        messageStoreConfig.setMapedFileSizeCommitLog(1024 * 1024 * 10);
+        messageStoreConfig.setMapedFileSizeConsumeQueue(1024 * 1024 * 10);
+        messageStoreConfig.setMaxHashSlotNum(10000);
+        messageStoreConfig.setMaxIndexNum(100 * 100);
+        messageStoreConfig.setFlushDiskType(FlushDiskType.ASYNC_FLUSH);
+        return new DefaultMessageStore(messageStoreConfig, new BrokerStatsManager("simpleTest"), new MyMessageArrivingListener(), new BrokerConfig());
```
lindzh: Indeed this is a must after testing,and this bug has been fix in PR https://github.com/apache/incubator-rocketmq/pull/141

vongosling:
Thanks @lindzh, i have merged it. Please close this PR :-)

PR #128, opened 2017-06-27
[ROCKETMQ-234] bugfix - broker will write response twice in batch scenario

evthoriz:
Bug in handle batch response. It will write response twice to client in batch scenario.
Jira: [ROCKETMQ-234](https://issues.apache.org/jira/browse/ROCKETMQ-234)

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
@@ -532,8 +532,7 @@ public class SendMessageProcessor extends AbstractSendMessageProcessor implement
 
         PutMessageResult putMessageResult = this.brokerController.getMessageStore().putMessages(messageExtBatch);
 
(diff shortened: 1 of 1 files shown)


zhouxinyu:
@evthoriz thanks, It's a bug indeed, could you please attach a JIRA issue link here and polish the PR title with more details?

evthoriz:
@zhouxinyu JIRA issue attached.

dongeforever:
@evthoriz Great catch.

lizhanhui:
+1 
Nice catch.

Jaskey:
+1.
This is is indeed buggy.

Would you please update the description that how the client behave for this problem so that we make our users clear about this.

IMO, this two results should be exactly the same and only logs can be observed but no factual influence

evthoriz:
@Jaskey From the aspect of client, every batch request would cause printing 2 warning logs. But the further influence is it may damage the performance in high throughput scenario where batch interface would more likely be used, because it wasted netty's IO thread time doing meaningless encoding, decoding and printing logs things.

Jaskey:
@evthoriz please update the jira's discriptions so that the user knows this bug and determine whether they should upgrade

lindzh:
+1
Great! This is indeed buggy.

PR #129, opened 2017-07-06
[ROCKETMQ-239]fix query message by time and fix get queue offset by time

lindzh:
When using queryMessageByTime or searchOffsetByTime or resetOffsetByTime , DefaultMessageStore return offset is not so accurate as we want.


store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -151,7 +151,7 @@ public void recover() {
         }
     }
 
-    public long getOffsetInQueueByTime(final long timestamp) {
+    public long getOffsetInQueueByTime(final long timestamp, boolean isGetTimeLast) {
         MappedFile mappedFile = this.mappedFileQueue.getMappedFileByTime(timestamp);
```
dongeforever: It is not reliable to get mapped file by modification time, as if the dispatch service fall behind too much, it may get the wrong mapped file.
How about checking each file's first message and last message it pointed to?
This may cause performance loss, but it is acceptable, for such api is used only for some specific diagnoses. It is rarely used, but when using it, it'd better to return the exactly correct result.

dongeforever:
@vongosling @zhouxinyu @Jaskey please help a review.

store/src/main/java/org/apache/rocketmq/store/MessageStore.java
```diff
@@ -118,7 +118,7 @@ GetMessageResult getMessage(final String group, final String topic, final int qu
      * @param timestamp Timestamp to look up.
```
zhouxinyu: Missing the JavaDoc of new param `isGetTimeLast` here. BTW, `isGetTimeLast` is confused~

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
dongeforever: @lindzh please consider such problem

lindzh: Thanks for you advice,I will check it later.If exist i will change it in the next commit.

store/src/main/java/org/apache/rocketmq/store/MessageStore.java
lindzh: Yes,and the JavaDoc if param isGetTimeLast has been added

lindzh:
Please help a review,Accuracy message offset search by store time has been fixed.

lizhanhui:
I scrutinized the code changes. Corner cases are handled properly and things look good except variables/method naming.

vongosling:
@lindzh  I will close the pr, if you happened to the same question, please let me know.

PR #130, opened 2017-07-10
[ROCKETMQ-243] BrokerData#selectBrokerAddr() picks the 1st element in the list of addresses

shroman:
https://issues.apache.org/jira/browse/ROCKETMQ-243

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
+++ b/common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
@@ -42,7 +42,8 @@ public class BrokerData implements Comparable<BrokerData> {
     }
 
     /**
-     * Randomly selects a broker address from the registered list.
+     * Selects a (preferably master) broker address from the registered list.
+     * If the master's address cannot be found, a slave broker address is selected in a random manner.
      *
      * @return Broker address.
      */
--- a/common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
+++ b/common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
@@ -18,10 +18,8 @@
 package org.apache.rocketmq.common.protocol.route;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Random;
 import org.apache.rocketmq.common.MixAll;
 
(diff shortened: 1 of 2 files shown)


Jaskey:
Broker contention should be safe but random pick can balance the network call for to all masters.

but we have to be cautions about this since for the current implementations, the broker addr will be very stable since it always return the same broker addr to the caller.

There are more than ten caller methods on this `selectBrokerAddr`, we have to make sure the random pick will not cause any side effect.

I believe it has not @lizhanhui @vongosling @zhouxinyu what's your opions.

@shroman , and since this method only returns the master addrs, I will suggest rename it to `pickMasterAddr`. Since the method is only called by some inner impl caller, I think it is safe to rename.

shroman:
@Jaskey It returns a random broker address if the one with MASTER_ID is not found. Why do you suggest renaming?

I hope the execution flow doesn't depend on the current method implementation and broker address change is ok -- otherwise it is a huge flaw for a distributed system like this. Guys, please share your thoughts.

Jaskey:
@shroman  Sorry I misunderstand this method that I though it returns the master address, since it is prefered to return the master address then slave, renaming seems to be not necessary while comment should be modified since it is only randomly pick while master is not found, thus only slave will be randomly pick.

shroman:
@Jaskey I modified the comment, thanks!

common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
```diff
@@ -37,15 +41,21 @@ public BrokerData(String cluster, String brokerName, HashMap<Long, String> broke
...
-        String value = this.brokerAddrs.get(MixAll.MASTER_ID);
-        if (null == value) {
-            for (Map.Entry<Long, String> entry : this.brokerAddrs.entrySet()) {
-                return entry.getValue();
-            }
+        String addr = this.brokerAddrs.get(MixAll.MASTER_ID);
+
+        if (addr == null) {
+            List<Long> keys = new ArrayList<Long>(brokerAddrs.keySet());
+            return brokerAddrs.get(keys.get(random.nextInt(keys.size())));
```
vsair: why not use values to construct list directly?

shroman: what exactly is the problem with the current approach?

Jaskey: @shroman use `values()` will benefit from one less method call of `map.get(i)`

shroman: Ok, changed to the list.

dongeforever:
LGTM

shroman:
Merged to develop branch.

PR #133, opened 2017-07-24
[ROCKETMQ-249] Do not attempt to clear disk even if disk store path i…

shroman:
…s found

https://issues.apache.org/jira/browse/ROCKETMQ-249

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
@@ -184,15 +184,27 @@ public class UtilAll {
             cal.get(Calendar.SECOND));
     }
 
+    /**
+     * Estimates the disk usage percentage by the specified path.
+     *
+     * @param path Path.
+     * @return Disk usage percentage by path.
+     */
     public static double getDiskPartitionSpaceUsedPercent(final String path) {
-        if (null == path || path.isEmpty())
+        if (null == path || path.isEmpty()) {
+            log.warn("Can't estimate used disk space percentage by an empty path");
+
             return -1;
+        }
 
         try {
             File file = new File(path);
 
-            if (!file.exists())
+            if (!file.exists()) {
+                log.warn("No file to estimate used disk space percentage");
+
                 return -1;
+            }
 
             long totalSpace = file.getTotalSpace();
 
@@ -203,6 +215,8 @@ public class UtilAll {
                 return usedSpace / (double) totalSpace;
             }
         } catch (Exception e) {
+            log.warn("Failed to estimate used disk space percentage", e);
+
             return -1;
         }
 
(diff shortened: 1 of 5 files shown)


shroman:
@vongosling @zhouxinyu @lizhanhui @vsair Anyone's willing to review the PR?

zhouxinyu:
It seems that broker can't work properly if disk store path is not found?

shroman:
@zhouxinyu The problem is described in the JIRA issue -- "If disk usage estimates fail because the disk store path is not found (doesn't exist), message store falsely decides the disk is full and tries to delete."
No failure, just a wasteful operation. I corrected the execution path, plus wrote the unit tests for it.

common/src/main/java/org/apache/rocketmq/common/UtilAll.java
```diff
@@ -184,15 +184,27 @@ public static String timeMillisToHumanString3(final long t) {
...
 
+    /**
+     * Estimates the disk usage percentage by the specified path.
+     *
+     * @param path Path.
+     * @return Disk usage percentage by path.
+     */
     public static double getDiskPartitionSpaceUsedPercent(final String path) {
-        if (null == path || path.isEmpty())
+        if (null == path || path.isEmpty()) {
```
solosky: we should use StringUtils.isBlank(..) instead as a basic practice.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -1509,65 +1514,62 @@ private boolean isTimeToDelete() {
...
-                    }
+        /**
+         * Checks if cleaning on the disk is needed.
+         *
+         * @param usageRatio Usage ratio.
+         * @param allowedRatio Allowed ratio.
+         * @return <code>True</code> if cleaning is needed, otherwise <code>false</code>.
+         */
+        private boolean needCleaning(double usageRatio, double allowedRatio) {
+            if (usageRatio == -1)
```
dongeforever: It is not safe to use == for float value. In this case, usageRatio < 0 is OK

store/src/test/java/org/apache/rocketmq/store/CleanCommitLogServiceTest.java
```diff
@@ -0,0 +1,115 @@
...
+        try {
+            messageStore.start();
+            messageStore.load();
+
+            Method m = clazz.getDeclaredMethod("isSpaceToDelete");
+            m.setAccessible(true);
+
+            Assert.assertFalse((boolean) m.invoke(cleanCommitLogService));
+        } finally {
+            messageStore.shutdown();
```
dongeforever: How about putting the finally block into teardown() ?
The test framework will handle it.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
shroman: Thanks, fixed.

store/src/test/java/org/apache/rocketmq/store/CleanCommitLogServiceTest.java
shroman: Good point. thanks.

shroman:
@dongeforever Modified as you suggested, thanks!

vongosling:
@shroman I will close the pr, if you happened to the same question, please let me know.

PR #138, opened 2017-07-28
[ROCKETMQ-138]add implement SelectMessageQueueByConsistentHash for MessageQueueSele…

lyy4j:
…ctor

SelectMessageQueueByConsistentHash  user  the  consistent hash  algorithm  to  select msg queue, producer client  can reuse  the instance  for the same topic


shroman:
Also you'll need a JIRA issue opened for this, and specified as the description of this pull request.
Please see other accepted pull requests.

lyy4j:
wanting  to  send   a   message with the same  key  combine with  orderly consumer,   user  can  use   the implements of the MessageQueueSelector, which  has one optimization compared with  SelectMessageQueueByHash when  a broker  crash   ,  it can   reduce the message Arriving out  of the order

lyy4j:
I hava  oopen an issue for this,what should do for next

shroman:
Please see this pull request description https://github.com/apache/incubator-rocketmq/pull/130, for example.

shroman:
`wanting to send a message with...`
yes, this should be the top comment of your class

```java
/**
wanting to send a message with ...
*/
public class SelectMessageQueueByConsistentHash implements MessageQueueSelector {
...
```

lyy4j:
I hava  add the  comment of this class

zhouxinyu:
Hi, @Jaskey could you please have a look at this PR?

Jaskey:
I will suggest reuse the util class of ConsistentHashRouter which provide a simple interface to implement consistent hashing

Jaskey:
@lyy4j , also, would you please modify your title and also link jira issue?

lyy4j:
@Jaskey ,there are two question ,firstly, what the meaning of havaing try on  ConsistentHashRouter ,I can not find the class;Secondly,  how should I modify my title and link jira issue

Jaskey:
@lyy4j  please see https://github.com/apache/incubator-rocketmq/blob/0c5e53db6f4d0ed9f25747379a8b679e2da5392d/common/src/main/java/org/apache/rocketmq/common/consistenthash/ConsistentHashRouter.java


As to modify the title,just click the edit on the pr, and put [ROCKETMQ-XXX]where XXX the jira number, and, link the jira to the pr description so that the reviewer and visitor know what feature you are going to implement.

lyy4j:
I have look at the class which has a good abstract of Consistent Hash ,but I think the class SelectMessageQueueByConsistentHash  is more simple and have the unified style with the original code

Jaskey:
@lyy4j  

1. I will still suggest resuing ConsistentHashRouter since it has a good enough abstraction which will make your selector much cleanner and easier as well as extensible(say specified virtual node nums, hash funtions)

2. I will suggest we expose a more friendly interface for user to use, for example , just let user to implement a shardingKey(MessageQueue q) method, the `select` method should be hidden since it is not friendly enough in this case IMO

lyy4j:
I agree with you  that ConsistentHashRouter has a good enough abstraction， making the code more  cleanner and  extensible. But I think SelectMessageQueueByConsistentHash will not  increase the  difficult of the using and more  easier  to  understand  and friendly to read comparing with using ConsistentHashRouter to implement

client/src/main/java/org/apache/rocketmq/client/producer/selector/SelectMessageQueueByConsistentHash.java
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+            i = subMap.firstKey();
+            virtualNode = subMap.get(i);
+        }
+
+        String result = virtualNode.substring(0, virtualNode.indexOf("&&"));
+        return result;
+    }
+
+    private int getHash(String str) {
```
dongeforever: Why not just use str.hashCode?

client/src/main/java/org/apache/rocketmq/client/producer/selector/SelectMessageQueueByConsistentHash.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.TreeMap;
+
+/**
+* wanting to send a message with the same key combine with orderly consumer, user can 
+* use the implements of the MessageQueueSelector, which has one optimization compared with SelectMessageQueueByHash
+* when a broker crash , it can reduce the message Arriving out of the order
+*/
+public class SelectMessageQueueByConsistentHash implements MessageQueueSelector {
+
+    private volatile SortedMap<Integer, String> virtualNodes =
```
dongeforever: how about  just using <Integer, MessageQueue>?

dongeforever:
@lyy4j It is important to put the code of same logic in one place. For easy maintenance, it is suggested to reuse ConsistentHashRouter. If you think that ConsistentHashRouter is not clear and extensible enough, please feel free to polish it.

vongosling:

PR #139, opened 2017-07-28
[ROCKETMQ-242] mqclient can not fetch nameSrvAddr periodly

qqeasonchen:


qqeasonchen:
@Jaskey @shroman @zhouxinyu 
create a pull request again due to previous wrong commit ,pls help to review, thx.

qqeasonchen:
@lizhanhui pls help to review

client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -255,19 +255,17 @@ public void start() throws MQClientException {
     }
 
     private void startScheduledTask() {
-        if (null == this.clientConfig.getNamesrvAddr()) {
```
lizhanhui: If name server addresses have been manually specified, why should we still periodically fetch name server addresses from web endpoint?

qqeasonchen: in case namesrv changed and client no need to restart

qqeasonchen: in fact, even though we specified the web endpoint, client will not fetch name server addresses any more beacase of below code in start():

`// If not specified,looking address from name server
                    if (null == this.clientConfig.getNamesrvAddr()) {
                        this.mQClientAPIImpl.fetchNameServerAddr();
                    }`

shroman: I think what @qqeasonchen says makes sense, but we have discuss what behavior we want to have.

To me, if addresses have been manually specified,
it makes sense to fetch name servers only if a call to manually specified servers fail. No need to do periodic calls. What do you think?

qqeasonchen: @shroman fetch name servers task must be running when client started so that client can update the failed namesrv by itself, we can  make some conditions to trigger it such as if wsAddr is not default

Jaskey: @shroman 

while to me, I will consider that if the nameserver address is specified manually, no dramatically fetch should be triggered since we should respect what user's choice. Or they may be confusing about the inconsistent behavor.

So, I would think the existing design for:

1. when specified nameserver address, no need to fetch nameserver address dymatically.
2. Only fetch nameserver address when user has not specified manully.

qqeasonchen: @Jaskey agree with you , and my question is if user does not specify an namesrv and its purpose is fetch from wsAddr , but client can not fetch namesrvs dymatically , see 

// If not specified,looking address from name server 
if (null == this.clientConfig.getNamesrvAddr()) { this.mQClientAPIImpl.fetchNameServerAddr(); }
 
private void startScheduledTask() {
      if (null == this.clientConfig.getNamesrvAddr()) {.......}

shroman: @qqeasonchen If you agree with @Jaskey, then the current implementation is correct, isn't it? :)

shroman: @Jaskey yes, this (current) design is surely ok. But it makes sense to discuss it in the dev mailing list, because everyone has a different idea what the best one is.

For instance, I would change what you describe a little bit like this
```
    1. when specified nameserver address, no need to fetch nameserver address dynamically.
    2. Only fetch nameserver address when the user-defined nameserver fails.
       I don't think any inconsistencies should happen, because it's a user's choice to enable dynamic fetching (We can implement such a switch to enable/disable it).
```

Jaskey: @shroman  

Whether It has a  problem or not which depends on how user understand its purpose. In my point of view, my understanding would be:  user specified is always prior to dymatically fetched, which means if I specify, JUST DO NOT change it anymore since I have totolly understand the which nameserver I am using, I don't want to risk of changing my nameservers without my notice.

Sure, somebody's understanding would be: user specified is prior to dymatically fetch, however, this only happens when my specified nameserver address is available. 

I would say it is also reasonable. Maybe we can provide a policy for them to choose.

qqeasonchen: @shroman @Jaskey 
sure the design is ok, but i think you guys didn't catch my question, there is a bug here : 
if i didn't specified nameserver address and choose to fetch dynamically, but infact
when client start ,if nameserver address is null ,it will fetch it first
// If not specified,looking address from name server
if (null == this.clientConfig.getNamesrvAddr()) { this.mQClientAPIImpl.fetchNameServerAddr(); }

and then in startScheduledTask(), the schedule task will not be started because just now fetch it
so the client will not be able to fetch nameserver dynamically.
hope make it clear about my question.

private void startScheduledTask() {
        if (null == this.clientConfig.getNamesrvAddr()) {
            this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {

                @Override
                public void run() {
                    try {
                        MQClientInstance.this.mQClientAPIImpl.fetchNameServerAddr();
                    } catch (Exception e) {
                        log.error("ScheduledTask fetchNameServerAddr exception", e);
                    }
                }
            }, 1000 * 10, 1000 * 60 * 2, TimeUnit.MILLISECONDS);
        }

qqeasonchen: @lizhanhui @zhouxinyu  @shroman @Jaskey 
we can only keep one between two of them, am i right?

if (null == this.clientConfig.getNamesrvAddr()) { this.mQClientAPIImpl.fetchNameServerAddr(); }

and

private void startScheduledTask() {
if (null == this.clientConfig.getNamesrvAddr()) {.......}

client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -255,19 +255,17 @@ public void start() throws MQClientException {
...
-            this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
+        this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
```
qqeasonchen: @Jaskey exactly, i did not update the code

PR #140, opened 2017-07-30
[ROCKETMQ-253]Compress RegisterBrokerBody

lizhanhui:
Refer to the description part of [this issue](https://issues.apache.org/jira/browse/ROCKETMQ-253) please.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
@@ -148,12 +148,13 @@ public class BrokerOuterAPI {
         requestHeader.setBrokerName(brokerName);
         requestHeader.setClusterName(clusterName);
         requestHeader.setHaServerAddr(haServerAddr);
+        requestHeader.setCompressed(true);
         RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.REGISTER_BROKER, requestHeader);
 
         RegisterBrokerBody requestBody = new RegisterBrokerBody();
         requestBody.setTopicConfigSerializeWrapper(topicConfigWrapper);
         requestBody.setFilterServerList(filterServerList);
-        request.setBody(requestBody.encode());
+        request.setBody(requestBody.encode(true));
 
         if (oneway) {
             try {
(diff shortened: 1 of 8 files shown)


lindzh:
Wonderfull,it's a good idea to make package smaller.

lizhanhui:
@lindzh Welcome to review this PR.

broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
```diff
@@ -148,12 +148,13 @@ private RegisterBrokerResult registerBroker(
...
         requestHeader.setClusterName(clusterName);
         requestHeader.setHaServerAddr(haServerAddr);
+        requestHeader.setCompressed(true);
         RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.REGISTER_BROKER, requestHeader);
 
         RegisterBrokerBody requestBody = new RegisterBrokerBody();
         requestBody.setTopicConfigSerializeWrapper(topicConfigWrapper);
         requestBody.setFilterServerList(filterServerList);
-        request.setBody(requestBody.encode());
+        request.setBody(requestBody.encode(true));
```
shroman: Great idea!
How about setting a threshold for compression being triggered, because if the data you compress is small, it overhead is higher than the benefits you get from the compression?

common/src/main/java/org/apache/rocketmq/common/protocol/body/RegisterBrokerBody.java
```diff
@@ -40,4 +57,114 @@ public void setTopicConfigSerializeWrapper(TopicConfigSerializeWrapper topicConf
...
     }
+
+    public byte[] encode(boolean compress) {
+
+        if (!compress) {
+            return encode();
+        }
+
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        DeflaterOutputStream outputStream = new DeflaterOutputStream(byteArrayOutputStream, new Deflater(Deflater.BEST_COMPRESSION));
```
shroman: I think `Deflater.BEST_COMPRESSION` may be even slower than GZIP. Do we care about compression speed?

vongosling:

PR #144, opened 2017-08-09
[ROCKETMQ-257] name server address and web server address should be specified at least one

qqeasonchen:
…pecified at least one


client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -225,10 +225,13 @@ public void start() throws MQClientException {
             switch (this.serviceState) {
                 case CREATE_JUST:
                     this.serviceState = ServiceState.START_FAILED;
-                    // If not specified,looking address from name server
-                    if (null == this.clientConfig.getNamesrvAddr()) {
+                    // name server address and web server address should be specified at least one
+                    if (null == this.clientConfig.getNamesrvAddr() && MixAll.getWSAddr().equals(MixAll.WS_ADDR)) {
+                        throw new MQClientException("name server address and web server address should be specified at least one.", null);
```
lindzh: What if someone want to use default wsaddr?

qqeasonchen: default wsaddr is not reachable， he need to set his own address

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -59,7 +59,7 @@
     public static final String WS_DOMAIN_NAME = System.getProperty("rocketmq.namesrv.domain", DEFAULT_NAMESRV_ADDR_LOOKUP);
     public static final String WS_DOMAIN_SUBGROUP = System.getProperty("rocketmq.namesrv.domain.subgroup", "nsaddr");
 //    // http://jmenv.tbsite.net:8080/rocketmq/nsaddr
-//    public static final String WS_ADDR = "http://" + WS_DOMAIN_NAME + ":8080/rocketmq/" + WS_DOMAIN_SUBGROUP;
+    public static final String WS_ADDR = "http://" + WS_DOMAIN_NAME + ":8080/rocketmq/" + WS_DOMAIN_SUBGROUP;
```
shroman: Is it legal to bring `WS_ADDR`? It was commented out for a reason.

Jaskey:
Actually, the existing design is that  :  we have a default name server service address, which will be specified in our doc(though it is indeed not demonstrated well for now), the default address is for the user's convenience that they just need to modify their hosts file and map the domain to the right service address.

But of course, some user may consider the a strict limit will be better. It depends. @shroman @vongosling @zhouxinyu If so, we need to make wsAddr to default to null.

Or , remains the design but demonstrated it better in our user guide.

Jaskey:
@qqeasonchen also, please notice your tittle to make it complete. And link the jira issue in the desciption

PR #146, opened 2017-08-14
[ROCKETMQ-265] fix consume queue’s data maybe repeat bug

fuyou001:
[more info](https://issues.apache.org/jira/browse/ROCKETMQ-265)


vongosling:
@fuyou001 Could you verify your polish using unit-test ?

lizhanhui:
Please add a link to the JIRA issue.

fuyou001:
@vongosling unit test add

broker/src/test/java/org/apache/rocketmq/broker/filter/MessageStoreWithFilterTest.java
```diff
@@ -238,7 +238,7 @@ public void testGetMessage_withFilterBitMapAndConsumerChanged() {
 
             // sleep to wait for consume queue has been constructed.
             try {
-                Thread.sleep(1000);
+                Thread.sleep(2001);
```
zhouxinyu: Do we have any method to ensure CQ has been constructed? Sleep is not a good practice~

store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
```diff
@@ -148,6 +148,52 @@ protected void deleteFile(File file) {
...
+            assertThat(cq).isNotNull();
+
+            Object dispatchResult = method.invoke(cq, dispatchRequest.getCommitLogOffset(),
+                dispatchRequest.getMsgSize(), dispatchRequest.getTagsCode(), dispatchRequest.getConsumeQueueOffset());
+
+            assertThat(Boolean.parseBoolean(dispatchResult.toString())).isTrue();
+
+        } catch (Exception e) {
+            e.printStackTrace();
+            assertThat(Boolean.FALSE).isTrue();
```
zhouxinyu: `assertThat(Boolean.FALSE).isTrue()`, it's really a trick, there is no need to catch exception, we can just add exception to method signature.

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -446,6 +446,13 @@ private boolean putMessagePositionInfo(final long offset, final int size, final
 
             if (cqOffset != 0) {
                 long currentLogicOffset = mappedFile.getWrotePosition() + mappedFile.getFileFromOffset();
+
+                if (expectLogicOffset < currentLogicOffset) {
+                    log.warn("build consume queue idempotent, expectLogicOffset: {} currentLogicOffset: {} Topic: {} QID: {} Diff: {}",
```
vongosling: How to understand the meaning of building consume queue idempotent？

broker/src/test/java/org/apache/rocketmq/broker/filter/MessageStoreWithFilterTest.java
```diff
@@ -238,7 +238,7 @@ public void testGetMessage_withFilterBitMapAndConsumerChanged() {
...
             } catch (InterruptedException e) {
                 e.printStackTrace();
```
vongosling: I would like to recommend to remove it in test module

broker/src/test/java/org/apache/rocketmq/broker/filter/MessageStoreWithFilterTest.java
lizhanhui: Looks there is no such interface, maybe keep polling till it finds something for a period of time and claiming failure if timeout.

store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
lizhanhui: Should directly fail it.

dongeforever: It is better to remove try....catch  from unit test, let the test framework handle it.
Another suggestion:
Use a  list to hold all the generated MessageStores, and release them inside "@After"  function. This will make the test more clear and look simple without the finally block.

vongosling:
LGTM, now~

vsair:
LGTM, but be careful to upgrade. If the situation already happened at a server, then upgrade to this feature, it will miss some messages. It would be better to have some mechanism to keep consistency between consume queue and commitlog

store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
```diff
@@ -131,6 +130,65 @@ protected void putMsg(DefaultMessageStore master) throws Exception {
...
+
+            assertThat(cq).isNotNull();
+
+            Object dispatchResult = method.invoke(cq, dispatchRequest.getCommitLogOffset(),
+                dispatchRequest.getMsgSize(), dispatchRequest.getTagsCode(), dispatchRequest.getConsumeQueueOffset());
+
+            assertThat(Boolean.parseBoolean(dispatchResult.toString())).isTrue();
+
+        } finally {
+            if (messageStore != null) {
```
dongeforever: @fuyou001 It would look clear to remove try finally block from unit test. Better to do it in "@after".

dongeforever:
@fuyou001 I want to merge this PR now. Could you please polish the unit tests?

zhouxinyu:
Waiting the response from @fuyou001 ~

vongosling:
@fuyou001 Thanks for your PR, we will consider a better way to fix this problem :-)

PR #147, opened 2017-08-14
[ROCKETMQ-266] Add a specific Exception message for comparing consumerThreadMax and consumerThreadMin

mark800:
When the client set the consumerThreadMax to a value less than the default value(20), 
we get the “consumeThreadMin Out of range [1, 1000] “ Exception,  which is hard to understand.

JIRA Issue: https://issues.apache.org/jira/browse/ROCKETMQ-266


vsair:
Thanks for your contributions. But,
1. Please refer to http://rocketmq.apache.org/docs/pull-request/, then modify your pr title.
2. Please change the merge target to develop.

mark800:
@vsair 
Thanks for your advice!

lindzh:
While consumeThreadMin is specified by a range,why not add consumeThreadMax to a range too?

mark800:
@lindzh 
My opinion is：consumeThreadMax and consumeThreadMin are two separate configuration, 
if we only configure one parameter, it should work, or at least give a reasonable exception message. 

If the two configurations have to be used at the same time, we should combine them to one configure option.

client/src/test/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImplTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.rocketmq.client.exception.MQClientException;
+import org.apache.rocketmq.common.message.MessageExt;
+import org.junit.Test;
+
+public class DefaultMQPushConsumerImplTest {
+    @Test(expected = MQClientException.class)
+    public void checkConfigTest() throws Exception {
+        DefaultMQPushConsumer consumer = new DefaultMQPushConsumer("test_consumer_group");
+
+        consumer.setConsumeThreadMax(10);
```
Jaskey: In the unit test , we should specify the min explicitly as well

mark800: The test case’s purpose is: If the consumeThreadMax is less than consumeThreadMin, we should get a proper exception message, the one line is enough.

The DefaultMQPushConsumerImpl.java class needs a lot more test case, a lot of condition should be checked, but that’s not this PR’s purpose.

Jaskey: @mark800 , I think we should make it more explict since the default max core size may be changed in the future, make it explicit will be clearer.

If you unit test, you are only verifying an expcetion is thrown which is not enough to verify this pr since an exception has been thrown even though you don't submit the changes, you should also very the error message is as expected.

mark800:
@Jaskey 
Thanks for your advice, I've updated the code.

mark800:
The travis-ci's error log is :
$ mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java
The command "eval mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V " failed.

It seems it is not the code's problem

dongeforever:
LGTM @zhouxinyu @vongosling  any view?

PR #149, opened 2017-08-18
[ROCKETMQ-136]Provide a handy message queue producer for order message sharding

Jaskey:
jira: https://issues.apache.org/jira/browse/ROCKETMQ-136

implementation brief: provide a OrderMessageProducer which need to implement an abstract method `sharingKey` and use a consistent hash selector to select a mq to send.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/producer/OrderMQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/OrderMQProducer.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.client.producer;
+
+import org.apache.rocketmq.client.exception.MQBrokerException;
+import org.apache.rocketmq.client.exception.MQClientException;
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/apache/rocketmq/client/producer/selector/SelectMessageQueueByConsistentHash.java
```diff
@@ -0,0 +1,94 @@
...
+        if (arg == null) {
+            throw new NullPointerException();
+        }
+        synchronized (consistentHashRouter) {
+            rehash(mqs);
+            return consistentHashRouter.routeNode(arg.toString()).mq;
+        }
+    }
+
+    private void rehash(List<MessageQueue> mqs) {
```
dongeforever: How about do a check before allocating two array lists?
Check if mqs is the same as lastMqs.
For in production environment, the mqs is rarely changed.

client/src/main/java/org/apache/rocketmq/client/producer/OrderMQProducer.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Callback method to return a unique sharding key for each messages. The sharding key will be used to pick a
+     * message queue. The messages with the same message queue will be sent to the same message queues.
+     *
+     * @param msg the message which is ready to send
+     * @return the sharding key which will be used to select message queue
+     */
+    abstract String shardingKey(Message msg);
```
dongeforever: Let user implement something is expensive.
How about add a method send(msg, shardingkey) or  just set one specified property "SHARDING_KEY" to msg?
@Jaskey

dongeforever:
@zhouxinyu @vongosling any idea about this?

client/src/main/java/org/apache/rocketmq/client/producer/OrderMQProducer.java
Jaskey: hmm, this should be a way. But when user want to send order messages, which should be in a common scene, say they want to send the msgs for ordering and paying. They must have a order number for every batch of messages, actually it is easy for them to provide a instance for each kind of messages needs ordering. Say, for payment messages , they may have one instance, for another kind of messages , they may have another one.

But of course, we may discuss more to make this api more friendly.

client/src/main/java/org/apache/rocketmq/client/producer/selector/SelectMessageQueueByConsistentHash.java
Jaskey: Yes we can, but I don't think it will have explicit better performance since , which just reduce two inner for loop in memory operation. However, maybe somebody may consider it is more easy to read. 

What do you think @zhouxinyu @shroman

PR #153, opened 2017-08-24
[ROCKETMQ-272] Fix sync slave timeout when using SYNC_MASTER

evthoriz:
Jira: https://issues.apache.org/jira/browse/ROCKETMQ-272

The timeout logic doesn't work correctly.
Thread waiting in GroupTransferService may frequently waked up by ReadSocketService in HAConnection.
So the transfer logic may return soon and wake up the thread waiting for the HA handling, which will make the timeout value in HA handling useless.

This patch repairs the timeout logic in syncing, and also introduces an option `syncSlaveTimeout` in `MessageStoreConfig` to distinguish from the disk flush timeout option.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -667,7 +667,7 @@ public class CommitLog {
                     service.putRequest(request);
                     service.getWaitNotifyObject().wakeupAll();
                     boolean flushOK =
-                        request.waitForFlush(this.defaultMessageStore.getMessageStoreConfig().getSyncFlushTimeout());
+                        request.waitForFlush(this.defaultMessageStore.getMessageStoreConfig().getSyncSlaveTimeout());
                     if (!flushOK) {
                         log.error("do sync transfer other node, wait return, but failed, topic: " + messageExt.getTopic() + " tags: "
                             + messageExt.getTags() + " client address: " + messageExt.getBornHostNameString());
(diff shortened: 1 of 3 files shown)


evthoriz:
@vongosling @zhouxinyu  The CI environment is not correctly set. Would you guys have a loot at that?

Jaskey:
@evthoriz 

Would you please also point out that what kind of scenario  will this issue  be reproduced?

evthoriz:
@Jaskey This issue can be easily reproduced with the config I provided in Jira. Please take a look at the description there.

Jaskey:
@evthoriz , so the scenario is for huge message. 
Please modify the jira title to make it clearer.
But actually, I sent huge message before, but not able to reproduce it, so huge messages scenario 
 is only the necessary condition, it may also depends on other factor.

 Let's get more people involved. @vongosling @shroman @lizhanhui

evthoriz:
@Jaskey Another key condition is the MASTER-SLAVE replicating mode, which is SYNC_MASTER in the case.
However, the option `syncFlushTimeout` is useless not only for big messages, but also those small messages. Small Messages just not trigger the bug, but it still exists.

evthoriz:
@vongosling @zhouxinyu @shroman @lizhanhui Anybody willing to review this pr？

dongeforever:
@evthoriz may you mock a test for this scenario?

evthoriz:
@dongeforever thanks for your reply. I tried, but it seems that there's no easy way to mock a test for this, since the ha mechanism involves lots of salve replicating behavior. Do you have any good ideas?

vongosling:

PR #157, opened 2017-08-28
[Rocketmq-277] fix get localhost throw java.net.UnknownHostException,when server hostname  not in /etc/hosts

fuyou001:
Motivation:
 fix bug [ROCKETMQ-277](https://issues.apache.org/jira/browse/ROCKETMQ-277)

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/MixAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/MixAll.java
@@ -26,6 +26,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.NetworkInterface;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -407,12 +408,50 @@ private static String localhost() {
...
         }
     }
 
+    private static String getLocalhostByNetworkInterface() throws SocketException {
+        List<String> candidatesHost = new ArrayList<String>();
+        Enumeration<NetworkInterface> enumeration = NetworkInterface.getNetworkInterfaces();
+
+        while (enumeration.hasMoreElements()) {
+            NetworkInterface networkInterface = enumeration.nextElement();
+            if ("docker0".equals(networkInterface.getName()) || !networkInterface.isUp()) {
```
zhouxinyu: Do we have a test in docker container for this method ?

vongosling:
Thank you everyone. I will make some polish and merge this PR :-)

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -407,12 +408,50 @@ private static String localhost() {
...
+            } catch (Exception ignored) {
+            }
+
             throw new RuntimeException("InetAddress java.net.InetAddress.getLocalHost() throws UnknownHostException"
                 + FAQUrl.suggestTodo(FAQUrl.UNKNOWN_HOST_EXCEPTION),
                 e);
         }
     }
 
+    private static String getLocalhostByNetworkInterface() throws SocketException {
```
lizhanhui: RemotingUtil has already had a similar method, aka, getLocalAddress, is there any rational reason to create a duplicated one?

vongosling: I have marked it using FIXME label. reverse logic comparing to RemotingUtil method, we could consider to refactor it in RocketMQ 5.0.

PR #159, opened 2017-08-28
[ROCKETMQ-279]  commit log data and consume queue data logic consistent

fuyou001:
To avoid commit log  and consume queue logic inconsistent .
Motivation:
 add a check when mq start,if commit log  and consume queue logic inconsistent ,start failed.

fast failed if check failed.

(cherry picked from commit 6e77be0)


store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -84,6 +84,59 @@ public boolean load() {
         return result;
     }
 
+    public void checkCommitLogAndConsumeQueueConsistent() {
+
+        String queueDir = this.storePath
+            + File.separator + topic
+            + File.separator + queueId + File.separator;
+
+        long maxOffsetInQueue = getMaxOffsetInQueue();
```
zhouxinyu: I would like to use `long lastRecordOffset = getMaxOffsetInQueue();` instead, to avoid using too many `+1 or -1` in below code block. Actually, `getMaxOffsetInQueue()` returns the next write position, not the position of last consume queue record.

store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
```diff
@@ -207,4 +226,34 @@ public void dispatch(DispatchRequest request) {
...
+
+                private int runCount = 0;
+            });
+
+            putMsg(master);
+            // wait build consume queue
+            Thread.sleep(2000);
+
+            ConsumeQueue cq = master.getConsumeQueueTable().get(topic).get(queueId);
+            cq.checkCommitLogAndConsumeQueueConsistent();
```
zhouxinyu: It seems that this test only covers the consistent condition?

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
fuyou001: refactor

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -84,6 +84,58 @@ public boolean load() {
         return result;
     }
 
+    public void checkCommitLogAndConsumeQueueConsistent() {
+
+        String queueDir = this.storePath
```
zhouxinyu: There is a helper to get queue dir: `StorePathConfigHelper#getStorePathConsumeQueue`

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -84,6 +84,58 @@ public boolean load() {
...
+        if (consumeQueueBuffer == null) {
+            String errorMsg = String.format("Can't find consume queue mappedBuffer by offset %s,queueDir %s", lastRecordOffset, queueDir);
+            log.error(errorMsg);
+            throw new IllegalStateException(errorMsg);
+        }
+
+        //position latest consume queue offset,
+        long offsetPy = consumeQueueBuffer.getByteBuffer().getLong();
+        int sizePy = consumeQueueBuffer.getByteBuffer().getInt();
+        consumeQueueBuffer.getByteBuffer().getLong();
```
zhouxinyu: Do we need the side effect of `getLong`?

fuyou001: fixed

PR #161, opened 2017-08-31
注释

liucyu:


vsair:
Thanks for your contribution, but please read the contributor's guide first. Then, you would,

1. Create an jira issue, and rename the pr's title and description.
2. Chinese is not allowed in code.

vongosling:
@liucyu If no update for this PR, we will close it.

lindzh:
As RocketMQ is widely use all over the world,I think there is no need to add chinese comment.I suggest add a blog for source code reading.

goodchax:
是否支持事务消息？

PR #163, opened 2017-09-13
[ROCKETMQ-289] getBrokerRuntimeInfo method output fileReservedTime value

fuyou001:
Motivation:
add fileReservedTime value for getBrokerRuntimeInfo return   `runtimeInfo`

Modification:
getBrokerRuntimeInfo method add a key `fileReservedTimeByHour` mean fileReservedTime in `runtimeInfo`


zhouxinyu:
Hi, thanks for your contribution. We could get FileReservedTime through `GetBrokerConfigCommand`, there is no need to add it to `runtimeInfo `.

PR #164, opened 2017-09-13
Fix tool's parameter

zd-double:


zhouxinyu:
Hi, thanks for your contribution, but this command is to send messages to a specified broker and test the response time. So these descriptions and variables make sense, any thoughts?

dongeforever:
@zd-double Thanks for your contribution. Just as @zhouxinyu said, it is test for sending to specified broker. Could you please close this PR and try another?

PR #165, opened 2017-09-16
[ROCKETMQ-296] fix DefaultMessageStoreTest bug: wait more time for consume queue build

wenweihu86:
when run DefaultMessageStoreTest::testPullSize, it should wait  more time to build consume queue.
The sleep time should be greater than commit log flush interval plus consume queue flush interval.

### DIFF
--- a/store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
@@ -73,6 +73,7 @@ public class DefaultMessageStoreTest {
         messageStoreConfig.setMaxHashSlotNum(10000);
         messageStoreConfig.setMaxIndexNum(100 * 100);
         messageStoreConfig.setFlushDiskType(FlushDiskType.SYNC_FLUSH);
+        messageStoreConfig.setFlushIntervalConsumeQueue(1);
         return new DefaultMessageStore(messageStoreConfig, new BrokerStatsManager("simpleTest"), new MyMessageArrivingListener(), new BrokerConfig());
     }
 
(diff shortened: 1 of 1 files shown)


dongeforever:
@wenweihu86 By default, the commitlog will flush in real-time. If enable flushCommitLogTimed, then it will flush at the interval.
So there is no need to sleep 100 ms, 10 is enough, just for the building time of consume queue.

wenweihu86:
@dongeforever The unit test of DefaultMessageStoreTest::testPullSize run failed for 10ms in my environment. Could you run successful?

dongeforever:
@wenweihu86  Yeah. I have run it on Mac many times, and all results are successful.  What environment do you run it on ?

wenweihu86:
@dongeforever My environment is the following:
java version "1.8.0_144"
Apache Maven 3.3.9

dongeforever:
@wenweihu86 how about sleep 50ms?

store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
```diff
@@ -148,8 +149,10 @@ public void testPullSize() throws Exception {
             messageExtBrokerInner.setQueueId(0);
             messageStore.putMessage(messageExtBrokerInner);
         }
-        //wait for consume queue build
-        Thread.sleep(10);
+        // wait for consume queue build
+        // the sleep time should be great than
+        // commit log flush interval plus consume queue flush interval
```
dongeforever: Such comment is not true, for the disk flush type is SYNC_FLUSH. It only need time for build consume queue.
Normally 10 ms is enough, but for safe, waiting 100ms is OK.

dongeforever:
@wenweihu86 Thanks for your contribution. It will be nice if you add checklist for this PR. You could refer to https://rocketmq.apache.org/docs/pull-request/#check-list-for-a-pr.

store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
wenweihu86: Yes, it only needs time for building consume queue, not include commit log flush interval

dongeforever:
LGTM @zhouxinyu @Jaskey @vsair @shroman please help a review.

dongeforever:
@wenweihu86 please change the target branch to develop, as current branch model merges the PR to develop firstly.

zhouxinyu:
Hi, thanks to your contribution sincerely, Maybe add a notify/wait mechanism to wait util the consume queue is built is a better solution, What do you think? @wenweihu86

wenweihu86:
@zhouxinyu Yes, it is better to use condition variable, but it is just a unit case, so just sleep looks good to me.

PR #190, opened 2017-11-02
[ROCKETMQ-309] Add new consumer implement

lyy4j:
When application  use consume client in Broadcast model  which runs in the docker,we can pull message only from fist  offset, because docker  allocate the apps  at random and  cause  the store file do not reuse. However, we  only want to pull message from last offset sometimes  even though the apps runs in the docker, and we can use  MQBroadcastFromLastConsumer in this case.


zhouxinyu:
Did you mean `CONSUME_FROM_LAST_OFFSET` don't take effect for the broadcast model?

BTW, could you please report a JIRA issue before submitting a pull request?

lyy4j:
@zhouxinyu ,sorry I have modify my commented again .

lyy4j:
I have report  JIRA ISSUE(309), would you have check  ? @zhouxinyu

zhouxinyu:
Thanks, let's discuss in ROCKETMQ-309 first~

lyy4j:
even if the apps do not run in the docker, it also has the problem.
for exmaple , 
the client has consume offset 3, and flush the store file , and the consumer shutdown .
in the mean while ,the broker offset has increan from 3 to 6;
when the client reboot , it only can consum offset from 3 ,but we want consum offset 6 only.

lyy4j:
what about the result of  this pull request？ @zhouxinyu

zhouxinyu:
Yes, It's exactly expected, `CONSUME_FROM_LAST_OFFSET` only take effect at the first subscription action of the consumer, you can reset the offset of stale consumers by `mqadmin` tool.

Consume from the last offset for each restart isn't common.

lindzh:
@lyy4j When using Broadcast Model , offset is stored in local. If you are using docker,please map volume offset store directory to host.

francisoliverlee:
@zhouxinyu 
when CONSUME_FROM_LAST_OFFSET is set ,
for old consumers,  they would use last offset when every restart, right?
for new consumers, they would use offset=0. it means new consumers will consume all messages the broker have got since it start, am i right?

vsair:
---for old consumers, they would use last offset when every restart, right?
it's right.
---for new consumers, they would use offset=0. it means new consumers will
consume all messages the broker have got since it start, am i right?
Should from max offset, not 0.

PR #220, opened 2018-01-16
[ROCKETMQ-332] MappedFileQueue#findMappedFileByOffset is not thread safe, which will cause message loss.

zhouxinyu:

Fix concurrent bug in MappedFileQueue#findMappedFileByOffset, which may cause message loss

## Brief changelog

The `findmappedfilebyoffset` method may return the wrong mapped file or null if the delete behavior occurs concurrently.

So change the find process:

1. Find the target mapped file by calculated index.
2. Verify the target mapped file is right.
3. Iterate the whole list and find the right mapped file.

In most cases, it's a O(1) method and will be degrade to O(n) if the concurrent issue occurs.

## Verifying this change

Review the source code and run the unit/it tests

- [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/projects/ROCKETMQ/issues/) filed for the change (usually before you start working on it). Trivial changes like typos do not require a JIRA issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[ROCKETMQ-XXX] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
@@ -461,26 +461,38 @@ public class MappedFileQueue {
      */
     public MappedFile findMappedFileByOffset(final long offset, final boolean returnFirstOnNotFound) {
         try {
-            MappedFile mappedFile = this.getFirstMappedFile();
-            if (mappedFile != null) {
-                int index = (int) ((offset / this.mappedFileSize) - (mappedFile.getFileFromOffset() / this.mappedFileSize));
+            MappedFile firstMappedFile = this.getFirstMappedFile();
+            if (firstMappedFile != null) {
+                int index = (int) ((offset / this.mappedFileSize) - (firstMappedFile.getFileFromOffset() / this.mappedFileSize));
                 if (index < 0 || index >= this.mappedFiles.size()) {
-                    LOG_ERROR.warn("Offset for {} not matched. Request offset: {}, index: {}, " +
-                            "mappedFileSize: {}, mappedFiles count: {}",
-                        mappedFile,
+                    LOG_ERROR.warn("Offset for {} not matched. Request offset: {}, index: {}, mappedFileSize: {}, mappedFiles count: {}",
+                        firstMappedFile,
                         offset,
                         index,
                         this.mappedFileSize,
                         this.mappedFiles.size());
+                } else {
+                    MappedFile targetFile = null;
+                    try {
+                        targetFile = this.mappedFiles.get(index);
+                    } catch (Exception ignored) {
+                    }
+
+                    if (targetFile != null && offset >= targetFile.getFileFromOffset()
+                        && offset < targetFile.getFileFromOffset() + this.mappedFileSize) {
+                        return targetFile;
+                    }
+
+                    for (MappedFile tmpMappedFile : this.mappedFiles) {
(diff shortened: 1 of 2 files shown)


zhouxinyu:
@dongeforever please help review this  PR

dongeforever:
+1 LGTM

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
```diff
@@ -461,26 +461,38 @@ public boolean commit(final int commitLeastPages) {
      */
     public MappedFile findMappedFileByOffset(final long offset, final boolean returnFirstOnNotFound) {
         try {
-            MappedFile mappedFile = this.getFirstMappedFile();
-            if (mappedFile != null) {
-                int index = (int) ((offset / this.mappedFileSize) - (mappedFile.getFileFromOffset() / this.mappedFileSize));
+            MappedFile firstMappedFile = this.getFirstMappedFile();
```
Jason918: If the first mappedFile is deleted between line 464 and line 467, this method still returns null when the requested offset is in the last mappedFile( mappedFiles.size() is less than index).
@zhouxinyu @dongeforever

zhouxinyu: Thanks, I missed this case, please help review again -:)

zhouxinyu:
Please, @Jason918 @lollipopjin @vongosling  help review this PR~

vongosling:
when we verified the target mapped file is right, thus, the delete action occurs, but we have gone into the Iterate step, how do we assure we find the right the index?

lollipopjin:
+1

Jason918:
@vongosling As mappedFiles is an instance of CopyOnWriteArrayList, a snapshot of this array is used in the iterate step.

Jason918:
@zhouxinyu This seems OK. But IMHO, doing retries maybe better than iterating through the whole array. : )

zhouxinyu:
@Jason918  We must consider the below cases if we adopt the retry strategy:
1. How many times? 
2. What's the next step if we only retry a specific number and failed again?
3. It may increase the CPU load and block time if we retry indefinitely.
4. How to avoid the stack overflow risk if we retry recursively?

While there is another method to fix the bug: Hold the first mapped file to hang up the delete process. But this method couldn't handle the forcible deletion case.

So iteration is a tradeoff method, consider that the probability of iterating through the array is very small and the size of the array is fewer than one thousand in most cases.

Jason918:
@zhouxinyu 
All of your question applies to all the optimistic locking implementation, right? 
For most cases, the deletion of  one `mappedFileQueue` does not happen concurrently, either in CleanConsumeQueueService or CleanCommitLogService, so I think the probability of multi-retries is very small.

So in my fix of this issue, I did retries up to 3 times in a for loop. If it still fails after 3 times, I would do the iteration.  After several days of test running, it all succeed in the second time (first retries).

I will upload my code in the following days. :)

zhouxinyu:
Your test meets expectation, consider the smaller probability and little costs of iteration, I chose the clear and simple way. But retry 3 times then iterate also is a good idea ~

PR #226, opened 2018-02-12
[ROCKETMQ-346] One message could only has one tag，“tags” is confusing

mark800:

in org.apache.rocketmq.common.message, the class Message's constructor parameter

“tags” is plural，but one message could only has one tag，“tags” is confusing

## Brief changelog

Change “tags” to “tag” in the Constructor’s parameter name


vongosling:
@mark800 Thanks, I will fix the typo before the 5.0.0 release.

PR #312, opened 2018-05-22
[ISSUE#311] Improve broker register topicrouter info performance

fuyou001:
Please do not create a Pull Request without creating an issue first. 

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

- [ ] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).
- [ ] Run `mvn -B clean apache-rat:check findbugs:findbugs checkstyle:checkstyle` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -741,7 +741,7 @@ public class BrokerController {
                     log.error("registerBrokerAll Exception", e);
                 }
             }
-        }, 1000 * 10, brokerConfig.getRegisterNameServerPeriod(), TimeUnit.MILLISECONDS);
+        }, 1000 * 10, Math.max(10000, Math.min(brokerConfig.getRegisterNameServerPeriod(), 60000)), TimeUnit.MILLISECONDS);
 
         if (this.brokerStatsManager != null) {
             this.brokerStatsManager.start();
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -741,7 +741,7 @@ public class BrokerController {
                     log.error("registerBrokerAll Exception", e);
                 }
             }
-        }, 1000 * 10, 1000 * 30, TimeUnit.MILLISECONDS);
+        }, 1000 * 10, brokerConfig.getRegisterNameServerPeriod(), TimeUnit.MILLISECONDS);
 
         if (this.brokerStatsManager != null) {
             this.brokerStatsManager.start();
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -62,6 +63,7 @@ import org.apache.rocketmq.broker.subscription.SubscriptionGroupManager;
 import org.apache.rocketmq.broker.topic.TopicConfigManager;
 import org.apache.rocketmq.common.BrokerConfig;
 import org.apache.rocketmq.common.Configuration;
+import org.apache.rocketmq.common.DataVersion;
 import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.TopicConfig;
 import org.apache.rocketmq.common.UtilAll;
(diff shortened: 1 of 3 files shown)


lizhanhui:
I understand the purpose of this pull request is to combine topic changes and report them all to nameserver in one registration action. The problem is, each time a new topic is registered, data version of topic config increases and `needRegister` method always returns true, causing active registration immediately, regardless of `forceRegister` being true or false for registerBrokerAll.

lizhanhui:
It makes sense now.  Good!

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
```diff
@@ -136,6 +136,16 @@
...
 
+    /**
+     *
+     * This configurable item defines interval of topics registration of broker to name server. Allowing values are
+     * between 10, 000 and 60, 000 milliseconds.
+     *
+     * If set to 0, newly created topics will be immediately reported to name servers and interval of periodical
+     * registration defaults to 10, 000 in milliseconds.
+     */
+    private int registerNameServerPeriod = 1000 * 30;
```
vongosling: registerNameServerPeriod is not a good name, maybe aliveICheckInterval is better. Another, do we need to polish here if no changes in a long time?

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -741,7 +741,7 @@ public void run() {
                     log.error("registerBrokerAll Exception", e);
                 }
             }
-        }, 1000 * 10, 1000 * 30, TimeUnit.MILLISECONDS);
+        }, 1000 * 10, Math.max(10000, Math.min(brokerConfig.getRegisterNameServerPeriod(), 60000)), TimeUnit.MILLISECONDS);
```
vongosling: How do we prove this is a better value, why not other?

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
fuyou001: origin default value

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
lizhanhui: Name servers will mark the broker as dead and remove routes to this broker if they do not receive registration heartbeat within 2T period. Empirically, brokers should send heartbeats every T moments at most to maintain their aliveness in case of occasional network failure, tolerating one-time heartbeat loss. 

As lite heartbeat mechanism has already been implemented, it's safe to heartbeat more frequently. This is why we allow interval of 10 seconds.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -245,7 +245,10 @@ private RemotingCommand updateAndCreateTopic(ChannelHandlerContext ctx,
         topicConfig.setTopicSysFlag(requestHeader.getTopicSysFlag() == null ? 0 : requestHeader.getTopicSysFlag());
 
         this.brokerController.getTopicConfigManager().updateTopicConfig(topicConfig);
-        this.brokerController.registerBrokerAll(false, true, true);
+
+        if (brokerController.getBrokerConfig().getRegisterNameServerPeriod() == 0) {
+            this.brokerController.registerBrokerAll(false, true, true);
```
dongeforever: If could not get the topic route info immediately after the topic is created, this may fail some unit tests and confuse some users.

It may be better to detect if the topic route  info is changed, and then decide whether it is necessary to register or not.

lizhanhui: We may set the interval to 0 to fix unit tests.  Purpose of this pull requests is to group changes and register them all periodically. Users are unlikely to administratively create a topic and send messages immediately. IMO, this change won't bring about much confusion. I won't say no if we keep the default value 0 to maintain behavioral consistency between before and after this pull requests.

dongeforever: If creating topic is frequenct, it is better to register the just created topic only. In fact, we do not need to change the register protocol, just do something in preparing heartbeat data.

lizhanhui: Following your suggestions, name server has to be updated to accept route data delta type of heartbeat. This would force name server upgrade in deployment prior to deployment of brokers having this patch.

lizhanhui: This results in extra complexities to existing users and deployments.

vongosling:
well, Please update your pr template and make your pr description clearly.

lizhanhui:
This issue is partially merged. In case incremental yet immediate topic registration is required, let's discuss it in a separate thread.

Issue #292, opened 2018-04-23
RocketMQ Feature Request: Transaction-Aware Message


lizhanhui:
Messaging is a major approach to decouple distributed systems. However, when trying to decouple system modules whose boundaries are associated with transactions, it is subtle to determine when to send messages. For example, when module A completes a database transaction, it needs to send a message to notify module B, C, and D to perform subsequent actions. The problem is what the best time point to send the message is in comparison with database transaction. If we sent the message before database transaction, what if the application or database server crashed out. The same counts if the transaction rolls back due to any potential reason. If we sent the message after the transaction commits, there will be chances that application server may suffer crashes before message delivery is made. Thus, to maintain consistency of the whole system, we need to either implement an extra module to periodically check messages of each transaction is sent given messages are designed to send after the transaction commits and supplementally send those missed ones to achieve eventual consistency or make our messaging system transaction aware.

The first option is viable, but it excessively complicates business application systems and brings about noticeable overhead as the extra module has to repeatedly check committed transactions and messages sent are in accordance after a period of time. The latter choice, however, confines complexities to the messaging system itself and once achieved, all application developers may enjoy simplicity and consistency.

To achieve the goal described above, we define a new type of message: transaction aware message. A transaction-aware message is composed of a pair of deliverable messages: prepare-message and commit/rollback message. Prepare messages are supposed to be sent before local transaction commits/roll-back, which are not visible to consumers. Commit/rollback messages are sent after transaction commits/rolls back. Brokers are responsible to inquire application servers, which share the same role with those prepare-message senders, to handle pending prepare messages after the defined period of time. On receiving of the commit message, the transaction aware message is made visible to consumers; Similarly, on receiving of rollback message, pending status of the corresponding prepare message is canceled quietly and consumers are completely unaware in the whole process.

duhengforever:
Thanks for @lizhanhui's feature request，and as mentioned above, with the rise of micro service architecture，developers are eager to use a mechanism to ensure the eventual consistency of transaction, especially in distributed systems, and we also found transaction aware message can achieve this goal gracefully, so we intend to implement this feature in rocketMQ, and this process is as follows：


![transaction](https://user-images.githubusercontent.com/39583243/40928949-a6e0b728-6855-11e8-863e-f69faa80e4f5.png)




[...]

PR #358, opened 2018-07-09
[ISSUE #292] Add support of transactional message feature 

duhengforever:
Please do not create a Pull Request without creating an issue first. 

In order to implement  eventual consistency in distributed system, we define a new type of message named transactional message, it can be thought of as a two-phase commit message implementation to ensure eventual consistency in distributed system. Transactional message ensures that the execution of local transaction and the sending of message can be performed atomically.

## Brief changelog

In our implementation, we made full use of the mechanism of rocketMQ itself to avoid external dependencies, and we defined a pair of message: named half message and commit/rollback message.  Half message means prepared message, which are not visible to consumers，and it will be sent before local transaction is executed. After the local transaction ends，producer will send a commit or rollback message to broker based on local transaction execution result

## Verifying this change

This is a trivial change.

- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).


duhengforever:
Related [Issue:](https://github.com/apache/rocketmq/issues/292)
@lizhanhui @vongosling @dongeforever Have any comments and review about this PR？

example/src/main/java/org/apache/rocketmq/example/benchmark/TransactionProducer.java
```diff
@@ -153,35 +150,37 @@ private static Message buildMessage(final int messageSize) throws UnsupportedEnc
...
-    public LocalTransactionState executeLocalTransactionBranch(final Message msg, final Object arg) {
-        if (ischeck) {
-            return LocalTransactionState.UNKNOW;
-        }
-        return LocalTransactionState.COMMIT_MESSAGE;
-    }
-}
-
-class TransactionCheckListenerBImpl implements TransactionCheckListener {
+//class TransactionExecuterBImpl implements LocalTransactionExecuter {
```
vongosling: ？

example/src/main/java/org/apache/rocketmq/example/benchmark/TransactionProducer.java
```diff
@@ -153,35 +150,37 @@ private static Message buildMessage(final int messageSize) throws UnsupportedEnc
...
+//}
+
+class TransactionListenerImpl implements TransactionListener {
     private boolean ischeckffalse;
     private StatsBenchmarkTProducer statsBenchmarkTProducer;
+    private boolean ischeckLocal;
 
-    public TransactionCheckListenerBImpl(boolean ischeckffalse,
-        StatsBenchmarkTProducer statsBenchmarkTProducer) {
+    public TransactionListenerImpl(boolean ischeckffalse, boolean isCheckLocal,
```
vongosling: ischeckffalse ？

common/src/main/java/org/apache/rocketmq/common/protocol/header/EndTransactionRequestHeader.java
```diff
@@ -121,9 +118,14 @@ public void setTransactionId(String transactionId) {
 
     @Override
     public String toString() {
-        return "EndTransactionRequestHeader [producerGroup=" + producerGroup + ", tranStateTableOffset="
-            + tranStateTableOffset + ", commitLogOffset=" + commitLogOffset + ", commitOrRollback="
-            + commitOrRollback + ", fromTransactionCheck=" + fromTransactionCheck + ", msgId=" + msgId
-            + "]";
+        return "EndTransactionRequestHeader{" +
```
vongosling: ToStringBuilder is easier way to reflect all field in one class

client/src/main/java/org/apache/rocketmq/client/producer/LocalTransactionExecuter.java
```diff
@@ -17,7 +17,7 @@
 package org.apache.rocketmq.client.producer;
 
 import org.apache.rocketmq.common.message.Message;
-
+@Deprecated
```
vongosling: ?

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionStore.java
```diff
@@ -18,7 +18,7 @@
 package org.apache.rocketmq.broker.transaction;
 
 import java.util.List;
-
+@Deprecated
```
vongosling: ？

broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
```diff
@@ -62,34 +60,22 @@ public Broker2Client(BrokerController brokerController) {
     }
 
     public void checkProducerTransactionState(
+        final String group,
         final Channel channel,
         final CheckTransactionStateRequestHeader requestHeader,
-        final SelectMappedBufferResult selectMappedBufferResult) {
+        final MessageExt messageExt) throws Exception {
```
Jaskey: Can this Exception be specific?

example/src/main/java/org/apache/rocketmq/example/benchmark/TransactionProducer.java
duhengforever: The extra commented code has been removed in the latest commit，thanks for your reminding.

example/src/main/java/org/apache/rocketmq/example/benchmark/TransactionProducer.java
duhengforever: Sorry, the last commit did not modify the original code, including variable naming, but the latest commit has been changed.

common/src/main/java/org/apache/rocketmq/common/protocol/header/EndTransactionRequestHeader.java
duhengforever: I used IDEA's automatic generation tool to override the toString method.

broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
duhengforever: Thanks for your review，in fact，actually I don't want to throw any exceptions here，but considering that the encode method may cause exception，so in this method just throws exception following with the encode method.

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionStore.java
duhengforever: This interface has been deleted in the latest commit.

broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
```diff
@@ -185,4 +190,33 @@ public void unregisterProducer(final String group, final ClientChannelInfo clien
...
+            if (0 == size) {
+                log.warn("Channel list is empty. groupId={}", groupId);
+                return null;
+            }
+
+            int index = positiveAtomicCounter.incrementAndGet() % size;
+            Channel channel = channelList.get(index);
+            int count = 0;
+            boolean isOk = channel.isActive() && channel.isWritable();
+            while (isOk && count++ < 3) {
```
vongosling: 3 is a magic number

broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
```diff
@@ -62,34 +60,22 @@ public Broker2Client(BrokerController brokerController) {
...
-                    selectMappedBufferResult.release();
-                    if (!future.isSuccess()) {
-                        log.error("invokeProducer failed,", future.cause());
-                    }
-                }
-            });
-        } catch (Throwable e) {
-            log.error("invokeProducer exception", e);
-            selectMappedBufferResult.release();
+            this.brokerController.getRemotingServer().invokeOneway(channel, request, 10);
```
vongosling: 10 ?

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionRecord.java
```diff
@@ -14,9 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.rocketmq.broker.transaction;
-
+@Deprecated
```
vongosling: deprecated only but missing the deadline and replacement is not a good practice

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -0,0 +1,496 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class TransactionalMessageServiceImpl implements TransactionalMessageService {
+    private static final InternalLogger log = InternalLoggerFactory.getLogger(LoggerName.TRANSACTION_LOGGER_NAME);
+
+    private TransactionalMessageBridge transactionBridge;
+
+    private static final int TRY_PULL_MSG_NUMBER = 1;
+
+    private final int queueTime = 60000;
```
vongosling: ?

vongosling:
LGTM if a rational response to my point out. This is an awesome feature, @lizhanhui @Jaskey @shroman Could we speed up the feedback for my kindly proposer, I would like to integrate into our develop branch :-)

vongosling:
@duhengforever Forget to comment, What is the coverage of the new code, over 80%?

broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
duhengforever: I will use a constant to represent this value, which represents the number of retries for fetch an available channel.

broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
duhengforever: This is an empirical value when using the invokeOneWay method. In the check scenario , we don't need to wait for the producer‘s response, and in order to prevent too many threads waiting，we adopted this configuration，but  we also considered that whether it is necessary to make this parameter configurable，what’s your opinion？

duhengforever:
For transactional messages, I added some documentation, including examples and design [documentation.](https://github.com/apache/rocketmq-site/pull/26), I hope these documents can help you to review this PR, thanks!

xingfudeshi:
great job!!

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionalMessageCheckService.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public TransactionalMessageCheckService(BrokerController brokerController) {
+        this.brokerController = brokerController;
+    }
+
+    @Override
+    public void start() {
+        if (started.compareAndSet(false, true)) {
+            super.start();
+            this.brokerController.getTransactionalMessageService().open();
```
lizhanhui: why open?

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionalMessageCheckService.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    @Override
+    public String getServiceName() {
+        return TransactionalMessageCheckService.class.getSimpleName();
+    }
+
+    @Override
+    public void run() {
```
lizhanhui: Should take advantage of what ServiceThread have already provided. No 
```
while(true)  {
...
sleep(xxx);
...
}
```

Use onWaitEnd() instead.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -0,0 +1,498 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class TransactionalMessageServiceImpl implements TransactionalMessageService {
+    private static final InternalLogger log = InternalLoggerFactory.getLogger(LoggerName.TRANSACTION_LOGGER_NAME);
+
+    private TransactionalMessageBridge transactionBridge;
```
lizhanhui: Having TransactionalMessageService --> TransactionalMessageServiceImpl abstraction, is it necesary to add another proxy?

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -0,0 +1,498 @@
...
+    }
+
+    private ConcurrentHashMap<MessageQueue, MessageQueue> opQueueMap = new ConcurrentHashMap<>();
+
+    @Override
+    public PutMessageResult prepareMessage(MessageExtBrokerInner messageInner) {
+        return transactionBridge.putHalfMessage(messageInner);
+    }
+
+    private boolean isNeedDiscard(MessageExt msgExt, int transactionCheckMax) {
```
lizhanhui: `isNeedXXX` naming is weird and does not follow grammar. Using needXXX suffices.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -0,0 +1,498 @@
...
+     * @param messageExt Message will be write back to queue
+     * @return Put result can used to determine the specific results of storage.
+     */
+    private PutMessageResult putBackToHalfQueueReturnResult(MessageExt messageExt) {
+        PutMessageResult putMessageResult = null;
+        try {
+            MessageExtBrokerInner msgInner = transactionBridge.renewHalfMessageInner(messageExt);
+            putMessageResult = transactionBridge.putMessageReturnResult(msgInner);
+        } catch (Exception e) {
+            e.printStackTrace();
```
lizhanhui: Just printStackTrace?
We need to recover from the error and mark tx service unavailable and propagate the unavailable state to producer clients if recovering is not possible.

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionalMessageCheckService.java
duhengforever: Considering that in other storage media implementations, maybe need to open some related settings, so provided this method.

lizhanhui:
1. Newly added classes, member variables & methods should be well documented, explaining their role in your solution.
2. Corner cases are not yet properly handled. For example, if pushing back half message experiences failure, what is the recovering procedure?  
3. How is the HA guaranteed in case the master is down?
4. We prefer to use full-word in the identifier, it would be better if this pull request may follow this convention. 
5. I am wondering whether the service locator and tx service bridge is an excessive abstraction.  IMO,  simplicity takes precedence. 
6. A lot of the tx related configurations are hard-coded, making them configurable is preferable.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
duhengforever: Thanks for helping to point out the problem，output to the log will be a good choice, as in this place, The outer method uses this method to decide whether to send back the request ，based on the status of putMessageResult, so there's no need to propagate the unavailable state to producer.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
duhengforever: good suggestion

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
duhengforever: In fact, in the development process, I have considered this problem, I personally think it is not needed, and we provide the Serviceloader to load other implementations. and if add another proxy，it will make the call hierarchy deeper and deeper, and need to add a strategy to ensure loading to the correct implementation，not convenient for later code maintenance.

broker/src/main/java/org/apache/rocketmq/broker/transaction/TransactionalMessageCheckService.java
duhengforever: Do you mean waitForRunning() and onWaitEnd() methods? yeah, I used these two methods to transform this implementation in the latest commit.

xingfudeshi:
congrats!

broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
```diff
@@ -185,4 +191,33 @@ public void unregisterProducer(final String group, final ClientChannelInfo clien
...
+
+            int index = positiveAtomicCounter.incrementAndGet() % size;
+            Channel channel = channelList.get(index);
+            int count = 0;
+            boolean isOk = channel.isActive() && channel.isWritable();
+            while (isOk && count++ < GET_AVALIABLE_CHANNEL_RETRY_COUNT) {
+                index = (++index) % size;
+                channel = channelList.get(index);
+                return channel;
+            }
```
dominicqi: here if isOk= true   will return next channel.
i think  it should be like 
`while ( count++ < GET_AVALIABLE_CHANNEL_RETRY_COUNT) {
               if(isOk){
                      return channel;
                }
                index = (++index) % size;
               channel = channelList.get(index);
               isOk = channel.isActive() && channel.isWritable();
            }`

duhengforever:
@klaus-pd In order to avoid external dependencies, such as database, we used the mechanism of rocketMQ itself to implement the check logic. and if the producer down, broker will send check request to other producers with same produceId.

broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
duhengforever: Thanks for your careful review，and I will fix this problem ASAP.

klaus-pd:
@duhengforever I know, but if only one producer of the same produceId, how to solve the problem of checking back?

duhengforever:
@klaus-pd broker will resend the check request when the producer restart util reached a limit.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -0,0 +1,498 @@
...
+        }
+        List<MessageExt> opMsg = pullResult.getMsgFoundList();
+        if (opMsg == null) {
+            log.warn("The miss op offset={} in queue={} is empty, pullResult={}", pullOffsetOfOp, opQueue, pullResult);
+            return pullResult;
+        }
+        for (MessageExt opMessageExt : opMsg) {
+            Long queueOffset = getLong(new String(opMessageExt.getBody(), TransactionalMessageUtil.charset));
+            log.info("Topic: {} tags: {}, OpOffset: {}, HalfOffset: {}", opMessageExt.getTopic(),
+                opMessageExt.getTags(), opMessageExt.getQueueOffset(), queueOffset);
```
1282009188: queueOffset  = getLong(new String(opMessageExt.getBody(), TransactionalMessageUtil.charset)); 
why this  can get the value of HalfOffset？

duhengforever: Details can be found in TransactionalMessageService#deletePrepareMessage method.

PR #373, opened 2018-07-19
there is no need to use ConcurrentHashMap at synchronized method. Usi…

hzdavid:

improve performance 
 

there is no need to use ConcurrentHashMap at synchronized method. Using HashMap  instead can be helpful for the performance.


vongosling:
ping @hzdavid

PR #382, opened 2018-07-27
reduce same method invocation in one method block to improve performance.

hzdavid:

reduce same method invocation in one method block to improve performance.


vongosling:
Please follow the community pr template, It would be helpful let the community guys know your idea.

PR #527, opened 2018-11-04
There is a bug in transaction delay messages

Cicizz:
When transaction message's delayLevel > 0, the prepare message write to the topic SCHEDULE_TOPIC_XXXX so commit or rollback cannot be confirmed,the transaction message should be written to topic at the time of commit.
Fix the sysFlag when parseHalfMessageInner()


duhengforever:
@Cicizz Thanks for your PR，Transaction does not support delayed messages for the time being.

Cicizz:
> @Cicizz Thanks for your PR，Transaction does not support delayed messages for the time being.
@duhengforever  Thank you,here's another problem. When using transaction messages and set delayLevel > 0, the message will be lost. maybe the transaction producer should check it, and if the prepare message is stored to half topic,the message's sysFlag is 0,shouldn't it be 4?

Issue #522, opened 2018-11-03
org.apache.rocketmq.store.MappedFile#warmMappedFile() method has some Ineffective code and result double io in the sync flush


zongtanghu:
1.What is the purpose of the change?

In the part of RocketMQ store,the goal of the warm-up mappedFile is that prewrite some random value into memory page corresponding to the mappedFile,so the target memory will swap to the physical memory in advance.But in fact,this action will result in double IO to the operation system.First IO is in the warm-up MappedFile,and secode IO is in the process of write message data into pagecache.

2.Optimization change action
Because of the above reason,we can remove the part of write some random value,prevent-gc and force flush codes in the warm-up method.In the mmapOperation,we can call mlock() method directly.

PR #567, opened 2018-11-29
[ISSUE #522]Fix warm-up method is Ineffective code and result double io in the sync flush.

zongtanghu:
1.What is the purpose of the change?

In the part of RocketMQ store,the goal of the warm-up mappedFile is that prewrite some random value into memory page corresponding to the mappedFile,so the target memory will swap to the physical memory in advance.But in fact,this action will result in double IO to the operation system.First IO is in the warm-up MappedFile,and secode IO is in the process of write message data into pagecache.

2.Optimization change action
Because of the above reason,we can remove the part of write some random value,prevent-gc and force flush codes in the warm-up method.In the mmapOperation,we can call mlock() method directly.


RongtongJin:
Because the allocated page is CopyOnWrite, only using mlock  can't lock the physical memory , you must write some value, you can read this [article](https://blog.csdn.net/fjt19900921/article/details/8074541)

PR #602, opened 2018-12-13
Selector parse EOF

zhangyanc:

There are some minor errors in SelectorParser.jj, and it should work like ActiveMQ using javacc plugin

## Brief changelog

- An incorrect import of LRUCache was deleted from SelectorParser.jj
- Add EOF token to match the end of selector sql, so selector like "a < 10bbbbb" will fail to parse, rather than parse out (a < 10)
- Changed the way quotes are compared so that there are no compile error with '\u005c' in generated SelectorParser.java
- add javacc plugin to automatically generate code, so we don't have to manual generate code with javacc command line when each time of .jj updated

## Verifying this change
No Java code was modified and all the exist tests passed


ShannonDing:
close due to time too long and wrong to merge to master, welcome to create a new PR to the apache: develop branch.

Issue #705, opened 2019-01-20
WaitNotifyObject class has a wake up bug


chicc999:
As described in #287 ，

https://github.com/apache/rocketmq/blob/3d56841fcc0cb9666307396d549343379a56b45c/store/src/main/java/org/apache/rocketmq/store/ha/WaitNotifyObject.java#L64-L77

In line 70 , “value = true” can't change the value in the entry。

hdchen:
LGTM：）

PR #706, opened 2019-01-20
[ISSUE #705]WaitNotifyObject class has a wake up bug

chicc999:
fix issue #705


imaffe:
Seems like this one could be merged.

PR #723, opened 2019-01-25
Concurrent risk in brokerLiveTable

ZhangShenao:
brokerLiveTable使用普通的HashMap，在scanNotActiveBroker()方法中，如果直接调用remove()删除非活跃的Broker信息，可能会有并发问题。


namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -52,14 +53,14 @@
...
     private final HashMap<String/* clusterName */, Set<String/* brokerName */>> clusterAddrTable;
-    private final HashMap<String/* brokerAddr */, BrokerLiveInfo> brokerLiveTable;
+    private final Map<String/* brokerAddr */, BrokerLiveInfo> brokerLiveTable;
     private final HashMap<String/* brokerAddr */, List<String>/* Filter Server */> filterServerTable;
 
     public RouteInfoManager() {
         this.topicQueueTable = new HashMap<String, List<QueueData>>(1024);
         this.brokerAddrTable = new HashMap<String, BrokerData>(128);
         this.clusterAddrTable = new HashMap<String, Set<String>>(32);
-        this.brokerLiveTable = new HashMap<String, BrokerLiveInfo>(256);
```
duhenglucky: In the most situation, this hashmap was protected by a ReentrantReadWriteLock, so if we changed this map to ConcurrentHashmMap will cause less consistent with other codes, could we also use the lock to ensure thread safety？

ZhangShenao: > In the most situation, this hashmap was protected by a ReentrantReadWriteLock, so if we changed this map to ConcurrentHashmMap will cause less consistent with other codes, could we also use the lock to ensure thread safety？

Change to use lock

ZhangShenao:
@duhenglucky 
Is there any problem?

ZhangShenao:
@ShannonDing 有什么问题吗？麻烦指导下

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -422,9 +422,18 @@ public void scanNotActiveBroker() {
             long last = next.getValue().getLastUpdateTimestamp();
             if ((last + BROKER_CHANNEL_EXPIRED_TIME) < System.currentTimeMillis()) {
                 RemotingUtil.closeChannel(next.getValue().getChannel());
-                it.remove();
-                log.warn("The broker channel expired, {} {}ms", next.getKey(), BROKER_CHANNEL_EXPIRED_TIME);
-                this.onChannelDestroy(next.getKey(), next.getValue().getChannel());
+                try {
```
RongtongJin: Why not write like this
```java
               try {
                    this.lock.writeLock().lockInterruptibly();
                    it.remove();
                    log.warn("The broker channel expired, {} {}ms", next.getKey(), BROKER_CHANNEL_EXPIRED_TIME);
                    this.onChannelDestroy(next.getKey(), next.getValue().getChannel());
                } catch (Exception e) {
                    log.error("Delete Not Active Broker Exception", e);
                } finally {
                    this.lock.writeLock().unlock();
                }
```

RongtongJin:
Could you please link a issue to this pr ?

ZhangShenao:
@RongtongJin I've modified my pr , please review

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -422,9 +424,20 @@ public void scanNotActiveBroker() {
...
+                        log.warn("The broker channel expired, {} {}ms", next.getKey(), BROKER_CHANNEL_EXPIRED_TIME);
+                        this.onChannelDestroy(next.getKey(), next.getValue().getChannel());
+                    } catch (Exception e) {
+                        log.error("Delete Not Active Broker Exception", e);
+                    } finally {
+                        this.lock.writeLock().unlock();
+                    }
+                } catch (Exception e) {
+                    log.error("Delete Not Active Broker Exception", e);
+                }
```
RongtongJin: Nesting of try-catch still exists. 
IMO, maybe moving the writelock out of the loop is a good choice.

ZhangShenao:
@RongtongJin  Maybe just like this?

ZhangShenao:
@RongtongJin Any problems?

xiangwangcheng:
I can't see the necessity of adding write lock. Could you explain that under which circumstance could lead to concurrent problem? @ZhangShenao

ZhangShenao:
@xiangwangcheng 当NameServer剔除非活跃的broker时，如果恰好有一个新的broker注册进来，会导致多个线程并发修改brokerLiveTable

xiangwangcheng:
@ZhangShenao  OK, I got you. An optimization is needed.
A new broker registry while scanning not active brokers won't cause a concurrent problem, but putting the same key whiling removing it will.

PR #736, opened 2019-01-30
No usages found in all places with sleep(long sleepMs) method.

lujiang-wed:
I didn't found usages and I didn't think that is necessary.


RongtongJin:
I'll close this PR first  because it doesn't have any impact on existing features.

PR #780, opened 2019-02-16
[RIP-9] Add the English description document for Deployment.md

SeekerWu667:

add the English description document for Deployment.md

## Brief changelog

Added the Deployment.md file to the rocketmq/docs/en directory

## Verifying this change
Check the corresponding Deployment.md file in the rocketmq/docs/en directory to verify the change.

### DIFF
--- a/docs/en/Deployment.md
+++ b/docs/en/Deployment.md
@@ -1,10 +1,10 @@
-# Operation and maintenance management
+# Deployment Architectures and Setup Steps
 
 ## Cluster Setup
 
 ### 1 Single Master mode
 
-This is the simplest, but also the riskiest, mode that makes the entire service unavailable once the Broker restarts or goes down. Production environments are not recommended, but can be used for local testing and development. Here are the steps to build.
+This is the simplest, but also the riskiest, mode that makes the entire service unavailable once the broker restarts or goes down. Production environments are not recommended, but can be used for local testing and development. Here are the steps to build.
 
 **1）Start NameServer**
 
@@ -22,15 +22,15 @@ We can see 'The Name Server boot success.. ' in namesrv.log that indicates the N
 **2）Start Broker**
 
 ```shell
-### Also start Broker first
+### Also start broker first
 $ nohup sh bin/mqbroker -n localhost:9876 &
 
-### Then verify that the Broker is started successfully, for example, the IP of Broker is 192.168.1.2 and the name is broker-a
+### Then verify that the broker is started successfully, for example, the IP of broker is 192.168.1.2 and the name is broker-a
 $ tail -f ~/logs/rocketmqlogs/Broker.log 
 The broker[broker-a,192.169.1.2:10911] boot success...
 ```
 
-We can see 'The broker[brokerName,ip:port] boot success..' in Broker.log that indicates the Broker has been started successfully.
+We can see 'The broker[brokerName,ip:port] boot success..' in Broker.log that indicates the broker has been started successfully.
 
 ### 2 Multiple Master mode
 
@@ -44,7 +44,7 @@ Multiple master mode means a mode with all master nodes(such as 2 or 3 master no
 - Disadvantages:
   1. During a single machine outage, messages that are not consumed on this machine are not subscribed to until the machine recovers, and message real-time is affected.
 
-The starting steps for multiple Master mode are as follows:
+The starting steps for multiple master mode are as follows:
 
 **1）Start NameServer**
 
@@ -69,20 +69,20 @@ $ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-noslave/broker
 ...
 ```
 
-The boot command shown above is used in the case of a single NameServer.For clusters of multiple NameServer, the address list after the -n argument in the Broker boot command is separated by semicolons, for example, 192.168.1.1: 9876;192.161.2: 9876.
+The boot command shown above is used in the case of a single NameServer.For clusters of multiple NameServer, the address list after the -n argument in the broker boot command is separated by semicolons, for example, 192.168.1.1: 9876;192.161.2: 9876.
 
 ### 3 Multiple Master And Multiple Slave Mode-Asynchronous replication
 
-Each Master configures a Slave, with multiple pairs of Master-Slave.HA uses asynchronous replication, with a short message delay (millisecond) between Master and Slave.The advantages and disadvantages of this mode are as follows:
+Each master node configures more thran one slave nodes, with multiple pairs of master-slave.HA uses asynchronous replication, with a short message delay (millisecond) between master node and slave node.The advantages and disadvantages of this mode are as follows:
 
 - Advantages: 
   1. Even if the disk is corrupted, very few messages will be lost and the real-time performance of the message will not be affected.
-  2. At the same time, when Master is down, consumers can still consume messages from Slave, and the process is transparent to the application itself and does not require human intervention.
-  3. Performance is almost as high as multiple Master mode.
+  2. At the same time, when master node is down, consumers can still consume messages from slave node, and the process is transparent to the application itself and does not require human intervention.
+  3. Performance is almost as high as multiple master mode.
 - Disadvantages:
-  1. A small number of messages will be lost when Master is down and the disk is corrupted.
+  1. A small number of messages will be lost when master node is down and the disk is corrupted.
 
-The starting steps for multiple Master and multiple Slave mode are as follows:
+The starting steps for multiple master and multiple slave mode are as follows:
 
 **1）Start NameServer**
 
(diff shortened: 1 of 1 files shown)


docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
+# Operation and maintenance management
+
+## Cluster building
+
+### 1 Single Master mode
+
+This is the simplest, but also the riskiest, mode that makes the entire service unavailable once the Broker restarts or goes down.Online environments are not recommended, but can be used for local testing and development.Here are the steps to build。
```
ShannonDing: Please change the full stop with a half-width character ".".

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+The broker[broker-a, 192.169.1.2:10911] boot success...
+```
+
+We can see 'The broker[brokerName,ip:port] boot success.. ' in Broker.log That indicates that the start Broker has been successful.
+
+### 2 Multiple Master mode
+
+The advantages and disadvantages of a cluster without Slave, being all Master, such as 2 Master or 3 Master, are as follows:
+
+- Advantages: simple configuration and a single Master outage or restart maintenance in the cluster has no impact on the application.When the disk is configured as RAID10, messages are not lost because the RAID10 disk is very reliable, even if the machine is not recoverable (In the case of asynchronous brush disk mode of the message, a small number of messages are lost;If the brush mode of a message is synchronous, a single message is not lost).In this mode, the performance is the highest.
```
wlliqipeng: Is it better to use 'flush disk' replace  'brush disk'

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
+# Operation and maintenance management
+
+## Cluster building
```
xiangwangcheng: “Cluster Setup” will be nicer in my opinion.

docs/en/Deployment.md
xiangwangcheng: Better use "Production enviroment" than "Online environment"

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+```shell
+### Start Name Server first
+$ nohup sh mqnamesrv &
+ 
+### Then verify that the Name Server starts successfully
+$ tail -f ~/logs/rocketmqlogs/namesrv.log
+The Name Server boot success...
+```
+
+We can see 'The Name Server boot success.. ' in namesrv.log That indicates that the start NameServer has been successful.
```
xiangwangcheng: "that indicates the NameServer has been started successfully." will be more smoothly.

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+### Then verify that the Broker is started successfully, for example, the IP of Broker is 192.168.1.2 and the name is broker-a
+$ tail -f ~/logs/rocketmqlogs/Broker.log 
+The broker[broker-a, 192.169.1.2:10911] boot success...
+```
+
+We can see 'The broker[brokerName,ip:port] boot success.. ' in Broker.log That indicates that the start Broker has been successful.
+
+### 2 Multiple Master mode
+
+The advantages and disadvantages of a cluster without Slave, being all Master, such as 2 Master or 3 Master, are as follows:
```
xiangwangcheng: It will be easier to understand if we split this sentence into two，something like this:
Multiple master mode means a mode with all master nodes(such as 2 or 3 master nodes) and no slave node. The advantages and disadvantages of this mode are as follows:

docs/en/Deployment.md
xiangwangcheng: numbered list is recommended here, as below:
1. simple configuration.
2. Outage or restart(for maintenance) of just one master node has no impact on the application.
3. XXXXXX
4. XXXXXX

and a better expression for "a single message is not lost" is "no message will be lost".

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+ 
+### Then verify that the Name Server starts successfully
+$ tail -f ~/logs/rocketmqlogs/namesrv.log
+The Name Server boot success...
+```
+
+**2）Start the Broker cluster**
+
+```shell
+### For example, starting the first Master, on machine A assumes that the IP for configuring NameServer is: 192.168.1.1.
```
xiangwangcheng: starting the first Master, on machine A assumes that the IP for configuring NameServer is: 192.168.1.1.
=>
starting the first Master node on machine A,  assuming that the configured NameServer IP is: 192.168.1.1.
will be better.

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+ 
+### Last start the second Slave,on machine D assumes that the IP for configuring NameServer is:192.168.1.1 and port is 9876.
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-2s-async/broker-b-s.properties &
+```
+
+The above shows a startup command for 2M-2S-Async mode, similar to other nM-nS-Async modes.
+
+### 4 Multiple Master And Multiple Slave Mode-Synchronous dual write
+
+In this mode, one Slave is also configured for each Master and there are multiple pairs of Master-Slave.HA uses synchronous double-write, that is, only Master and Slaver write successfully to return success to the application.
```
xiangwangcheng: "that is, the success response will be returned to the application only when the message is successfully written into the master node and replicated to more than one slave node."  seems better.

docs/en/Deployment.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+The boot command shown above is used in the case of a single NameServer.For clusters of multiple NameServer, the address list after the -n argument in the Broker boot command is separated by semicolons, for example, 192.168.1.1: 9876;192.161.2: 9876.
+
+### 3 Multiple Master And Multiple Slave Mode-Asynchronous replication
+
+Each Master configures a Slave, with multiple pairs of Master-Slave.HA uses asynchronous replication, with a short message delay (millisecond) between Master and Slave.The advantages and disadvantages of this mode are as follows:
+
+- Advantages: even if the disk is corrupted, very few messages will be lost and the real-time performance of the message will not be affected;At the same time, when Master is down, consumers can still consume messages from Slave, and the process is transparent to the application itself and does not require human intervention;Performance is almost as high as multiple Master mode;
+- Disadvantages:a small number of messages will be lost when Master is down or the disk is corrupted.
```
xiangwangcheng: when the master node is down **AND** the disk is corrupted. Not **or**.

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
...
+ 
+### Last starting the second Slave on machine D, assuming that the configured NameServer IP is: 192.168.1.1 and port is 9876.
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-2s-async/broker-b-s.properties &
+```
+
+The above shows a startup command for 2M-2S-Async mode, similar to other nM-nS-Async modes.
+
+### 4 Multiple Master And Multiple Slave Mode-Synchronous dual write
+
+In this mode, one Slave is also configured for each Master and there are multiple pairs of Master-Slave.HA uses synchronous double-write, that is, the success response will be returned to the application only when the message is successfully written into the master node and replicated to more than one slave node.
```
xiangwangcheng: I think in Multiple Master and Multiple Slave Mode, more than one slave node is configured for each master. 
The specification in Chinese docs seems imprecise.

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
...
+
+The advantages and disadvantages of this model are as follows:
+
+- Advantages: 
+  1. Neither the data nor the service has a single point of failure. 
+  2. In the case of Master downtime, the message is also undelayed.
```
xiangwangcheng: "During the master node downtime"  or "In the case of master node shutdown " seems better.

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
...
+
+The advantages and disadvantages of this model are as follows:
+
+- Advantages: 
+  1. Neither the data nor the service has a single point of failure. 
+  2. In the case of Master downtime, the message is also undelayed. 
+  3. Service availability and data availability are very high;
+- Disadvantages:
+  1. The performance in this mode is slightly lower than in asynchronous replication mode (about 10% lower).
+  2. The RT sending a single message is slightly higher, and the current version cannot automatically switch to the Master after the primary node is down.
```
xiangwangcheng: "and under the current version, the slave node cannot automatically switch to the Master after the master node is down"

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
...
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-2s-sync/broker-b.properties &
+ 
+### Then starting the first Slave on machine C, assuming that the configured NameServer IP is: 192.168.1.1 and port is 9876.
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-2s-sync/broker-a-s.properties &
+ 
+### Last starting the second Slave on machine D, assuming that the configured NameServer IP is: 192.168.1.1 and port is 9876.
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-2s-sync/broker-b-s.properties &
+```
+
+The above Master and Slave pairing is paired by specifying the same BrokerName parameter, the BrokerId of the Master must be 0, and the BrokerId of the Slave must be greater than 0.
```
xiangwangcheng: The above Master and Slave are paired by specifying the same config named "brokerName", the "brokerId" of ....

brokerName and brokerId should be case sensitive. Seems the specification of Chinese docs is wrong. You can open another PR to correct it if you have time.

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
...
+$ nohup sh mqbroker -n 192.168.1.1:9876 -c $ROCKETMQ_HOME/conf/2m-noslave/broker-b.properties &
+
+...
+```
+
+The boot command shown above is used in the case of a single NameServer.For clusters of multiple NameServer, the address list after the -n argument in the Broker boot command is separated by semicolons, for example, 192.168.1.1: 9876;192.161.2: 9876.
+
+### 3 Multiple Master And Multiple Slave Mode-Asynchronous replication
+
+Each Master configures a Slave, with multiple pairs of Master-Slave.HA uses asynchronous replication, with a short message delay (millisecond) between Master and Slave.The advantages and disadvantages of this mode are as follows:
```
xiangwangcheng: 1. Each Master configures a Slave  => Each Master configures more than one Slave.
2. "node" is recommended to be added after "master" and "slave", this makes more precise. Of course, except proper nouns like "Multiple Master And Multiple Slave".
3. Both capitalized "Master/Slave" and uncapitalized  "master/slave" can be found. This makes a little confusing. I prefer uncapitalized ones except for titles.

docs/en/Deployment.md
```diff
@@ -0,0 +1,159 @@
+# Operation and maintenance management
```
xiangwangcheng: Will "Deployment Architectures and Setup Steps" be better?
@ShannonDing @wlliqipeng

xiangwangcheng:
LGTM @ShannonDing

PR #820, opened 2019-02-20
[RIP-10]  Add test cases for   TopicStatsTable.java

radishlee:
add a test case for  TopicStatsTable.java
增加测试用例 TopicStatsTable.java


radishlee:
some files problems , close it

PR #904, opened 2019-02-26
[rip-10] add test case for PrintMessageByQueueCommand

pigpdong:
add test case for PrintMessageByQueueCommand


pigpdong:
remove bugfix just add test case

PR #962, opened 2019-03-02
[RIP-10]Add test cases for common/stats

bluerstar:
[RIP-10]Add test cases for common/stats


xiangwangcheng:
The test cases themselves look good to me. But could u rerun the travis CI again since it's been a little while since last time.
@zongtanghu @wlliqipeng

PR #1196, opened 2019-05-05
Client should retry send when broker return SYSTEM_BUSY.

aftersss:
When we send message to broker, We meet the following problem: sometimes the broker return ResponseCode.SYSTEM_BUSY, then send will fail immediately.

I think in this situation, the client should try to send again(to any of the other brokers in the cluster, the same broker is also OK), because quite possibly, the other broker is not busy. 

We are using it in production, it works good for us.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -592,6 +592,7 @@ public class DefaultMQProducerImpl implements MQProducerInner {
                             case ResponseCode.TOPIC_NOT_EXIST:
                             case ResponseCode.SERVICE_NOT_AVAILABLE:
                             case ResponseCode.SYSTEM_ERROR:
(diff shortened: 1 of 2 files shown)


duhenglucky:
@aftersss it means that the Server is very busy to resolve the stacked requests, so retrying at this time will increase the pressure on the server, so fast failure is a good choice.

ShannonDing:
> @aftersss it means that the Server is very busy to resolve the stacked requests, so retrying at this time will increase the pressure on the server, so fast failure is a good choice.

I agree with you.

adamswanglin:
> @aftersss it means that the Server is very busy to resolve the stacked requests, so retrying at this time will increase the pressure on the server, so fast failure is a good choice.

image

In our production environment, queues of a topic discarded on at least two brokers. Sudden IO may cause broker SYSTEM_BUSY(actually it happened). Isn't that normal for client try another broker to send the message ?(The code above shows that the default strategy of retried queue is from another broker)

wz2cool:
it think it's a good idea if rocketmq can provide custom strategy to retry send message  if broker busy

PR #1249, opened 2019-06-06
Add unit tests for org.apache.rocketmq.common.UtilAll

EricHetti:
I've analysed your codebase and noticed that `org.apache.rocketmq.common.UtilAll` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important.

### DIFF
--- a/common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
+++ b/common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
@@ -1,0 +1,381 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.common;
+
+import java.io.File;
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Enumeration;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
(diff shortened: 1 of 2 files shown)


common/pom.xml
```diff
@@ -37,5 +37,16 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>rocketmq-remoting</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+            <version>1.6.6</version>
```
vongosling: Have you noticed the dependency mockito in the parent pom? powermock is a good extension for mockito. I strongly recommend that you could move the dependency into parent pom and scope it in the test.

common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
```diff
@@ -0,0 +1,332 @@
...
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+public class UtilAllTests {
+
+    @PrepareForTest({RuntimeMXBean.class,
+            UtilAll.class, ManagementFactory.class})
+    @Test
+    public void testGetPid1() {
```
vongosling: testMethodwithCondtion is a recommended naming not simple 1, 2,3...

common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
```diff
@@ -0,0 +1,332 @@
...
+        Assert.assertEquals(-1,
+                UtilAll.getDiskPartitionSpaceUsedPercent("foo"), 0);
+    }
+
+    @Rule
+    private TemporaryFolder folder = new TemporaryFolder();
+
+    @Test
+    public void testGetDiskPartitionSpaceUsedPercent1() throws IOException {
+        File file = folder.newFile("foo.txt");
```
vongosling: Do you consider the data clearance when test stateful function?

vongosling:
Anyway, make the class utils to be testable is a valuable thing. I am very glad to help you to polish your pr and merge to branch finally if you could follow up my comments:-)

common/pom.xml
Braavos96: PowerMock has been moved to the parent pom and version has been changed

common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
Braavos96: Test method names have been changed

common/src/test/java/org/apache/rocketmq/common/UtilAllTests.java
Braavos96: @vongosling I have added a finally block that deletes the file and TemporaryFolder. Is that ok? I am not entirely sure about that

Braavos96:
Hi @vongosling just following up and wondering if you would like me to make any changes to the PR?

pom.xml
```diff
@@ -471,6 +471,17 @@
             <version>2.6.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
liuruiyiyang: Can we implement these tests using existing test-utils dependencies?   It is required to check the license of the newly added dependencies and keep our dependency tree clean and each item is necessary.

Braavos96: Hi, Yes I can do it with `mockito`, however `powermock` is a bit more powerful and you will be losing coverage if I used `mockito` instead

Issue #1188, opened 2019-04-29
Tracer start failed when starting multiple MqProducer  in one  process


kangjiabang:
i  want to use trace feature in 4.4.0.
when i  init  two  defaultMQProducer ， use  new DefaultMQProducer(producerId,true);
then i  start to send  message to test trace， trace not  work well。

i find  when producer start with trace on，it will  produce new  AsyncTraceDispatcher，in  AsyncTraceDispatcher, it will invoke getAndCreateTraceProducer method.  

`private DefaultMQProducer getAndCreateTraceProducer(RPCHook rpcHook) {
        DefaultMQProducer traceProducerInstance = this.traceProducer;
        if (traceProducerInstance == null) {
            traceProducerInstance = new DefaultMQProducer(rpcHook);
            traceProducerInstance.setProducerGroup(TraceConstants.GROUP_NAME);
            traceProducerInstance.setSendMsgTimeout(5000);
            traceProducerInstance.setVipChannelEnabled(false);
            // The max size of message is 128K
            traceProducerInstance.setMaxMessageSize(maxMsgSize - 10 * 1000);
        }
        return traceProducerInstance;
    }`
but the ProducerGroup  is same for all trace Producer . when i start two MQProducer  ,i will report error  like following .
`throw new MQClientException("The producer group[" + this.defaultMQProducer.getProducerGroup()
                        + "] has been created before, specify another name please." + FAQUrl.suggestTodo(FAQUrl.GROUP_NAME_DUPLICATE_URL),
                        null);`

ChaosYjh:
I think it is a bug too, but it seems that no one cares about this problem.@duhenglucky ,can you look at this question? Thanks~~

duhenglucky:
@ChaosYjh try to setInstanceName when create producer

zongtanghu:
@ChaosYjh @kangjiabang  We will check this issue in our local enviroment.

ChaosYjh:
> polish

It has noting to do with instanceName,because producerGroup always be TraceConstants.GROUP_NAME.
<img width="535" alt="image" src="https://user-images.githubusercontent.com/27771595/59734474-22e0b180-9284-11e9-81f8-5b514fe0342e.png">

And if you ensure this is a bug,I will commit a PR later.@zongtanghu @duhenglucky.
Thanks !

ChaosYjh:
[...]

PR #1275, opened 2019-06-19
[ISSSUE 1188]Fix the problem when more than one producer or consumer in the same process can trace only one

ahuazhu:

Fix the bug when there are more than one producer (or consumer) , the Tracer can not show all trace infomation.

[ISSUE 1188](https://github.com/apache/rocketmq/issues/1188)

## Brief changelog

Set different ProducerGroup for different TraceProducer, the ProducerGroup will like `_INNER_TRACE_PRODUCER-user-specified-producer-group`

## Verifying this change

Produce and consume message for a same topic in the same process, the console show both produce and consume information.
image

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
+++ b/client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
@@ -74,8 +74,9 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
     private AtomicBoolean isStarted = new AtomicBoolean(false);
     private AccessChannel accessChannel = AccessChannel.LOCAL;
     private String group;
+    private Type type;
 
-    public AsyncTraceDispatcher(String group, String traceTopicName, RPCHook rpcHook) {
+    public AsyncTraceDispatcher(String group, Type type,String traceTopicName, RPCHook rpcHook) {
         // queueSize is greater than or equal to the n power of 2 of value
         this.queueSize = 2048;
         this.batchSize = 100;
@@ -83,6 +84,8 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
         this.discardCount = new AtomicLong(0L);
         this.traceContextQueue = new ArrayBlockingQueue<TraceContext>(1024);
         this.group = group;
+        this.type = type;
+
         this.appenderQueue = new ArrayBlockingQueue<Runnable>(queueSize);
         if (!UtilAll.isBlank(traceTopicName)) {
             this.traceTopicName = traceTopicName;
@@ -162,7 +165,7 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
     }
 
     private String genGroupNameForTrace() {
-        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.group;
+        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.group + "-" + this.type ;
     }
 
     @Override
--- a/client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
+++ b/client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
@@ -73,16 +73,16 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
     private String traceTopicName;
     private AtomicBoolean isStarted = new AtomicBoolean(false);
     private AccessChannel accessChannel = AccessChannel.LOCAL;
-    private String producerOrConsumerGroup;
+    private String group;
 
-    public AsyncTraceDispatcher(String producerOrConsumerGroup, String traceTopicName, RPCHook rpcHook) {
+    public AsyncTraceDispatcher(String group, String traceTopicName, RPCHook rpcHook) {
         // queueSize is greater than or equal to the n power of 2 of value
         this.queueSize = 2048;
         this.batchSize = 100;
         this.maxMsgSize = 128000;
         this.discardCount = new AtomicLong(0L);
         this.traceContextQueue = new ArrayBlockingQueue<TraceContext>(1024);
-        this.producerOrConsumerGroup = producerOrConsumerGroup;
+        this.group = group;
         this.appenderQueue = new ArrayBlockingQueue<Runnable>(queueSize);
         if (!UtilAll.isBlank(traceTopicName)) {
             this.traceTopicName = traceTopicName;
@@ -152,7 +152,7 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
         DefaultMQProducer traceProducerInstance = this.traceProducer;
         if (traceProducerInstance == null) {
             traceProducerInstance = new DefaultMQProducer(rpcHook);
-            traceProducerInstance.setProducerGroup(makeGroupNameForTrace());
+            traceProducerInstance.setProducerGroup(genGroupNameForTrace());
             traceProducerInstance.setSendMsgTimeout(5000);
             traceProducerInstance.setVipChannelEnabled(false);
             // The max size of message is 128K
@@ -161,8 +161,8 @@ public class AsyncTraceDispatcher implements TraceDispatcher {
         return traceProducerInstance;
     }
 
-    private String makeGroupNameForTrace() {
-        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.producerOrConsumerGroup;
+    private String genGroupNameForTrace() {
+        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.group;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


zongtanghu:
Hi @duhenglucky @ShannonDing @jonnxu  please help to review this pr,together.

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
```diff
@@ -73,14 +73,16 @@
     private String traceTopicName;
     private AtomicBoolean isStarted = new AtomicBoolean(false);
     private AccessChannel accessChannel = AccessChannel.LOCAL;
+    private String producerOrConsumerGroup;
 
-    public AsyncTraceDispatcher(String traceTopicName, RPCHook rpcHook) {
+    public AsyncTraceDispatcher(String producerOrConsumerGroup, String traceTopicName, RPCHook rpcHook) {
```
zongtanghu: I think it will be better if the variable "producerOrConsumerGroup" change to be "group" name.

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
```diff
@@ -73,14 +73,16 @@
...
         // queueSize is greater than or equal to the n power of 2 of value
         this.queueSize = 2048;
         this.batchSize = 100;
         this.maxMsgSize = 128000;
         this.discardCount = new AtomicLong(0L);
         this.traceContextQueue = new ArrayBlockingQueue<TraceContext>(1024);
+        this.producerOrConsumerGroup = producerOrConsumerGroup;
```
zongtanghu: The same as above mentioned.

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
```diff
@@ -159,6 +161,10 @@ private DefaultMQProducer getAndCreateTraceProducer(RPCHook rpcHook) {
         return traceProducerInstance;
     }
 
+    private String makeGroupNameForTrace() {
+        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.producerOrConsumerGroup;
```
zongtanghu: The same as above mentioned.

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
```diff
@@ -159,6 +161,10 @@ private DefaultMQProducer getAndCreateTraceProducer(RPCHook rpcHook) {
         return traceProducerInstance;
     }
 
+    private String makeGroupNameForTrace() {
```
zongtanghu: It will be better if makeGroupNameForTrace method name change to be genGroupNameForTrace

zongtanghu:
If users start the mutiple producer or consumer instances in the same process,the instance name is the same one.You can set different instance name in the start method when the traceProducer is started. @ahuazhu

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
ahuazhu: All variables were renamed as your advice.

superheizai:
LGTM

ahuazhu:
> If users start the mutiple producer or consumer instances in the same process,the instance name is the same one.You can set different instance name in the start method when the traceProducer is started. @ahuazhu

If set different instance name for trace producer, multiple client instances will be created.

client/src/main/java/org/apache/rocketmq/client/trace/AsyncTraceDispatcher.java
```diff
@@ -159,6 +161,10 @@ private DefaultMQProducer getAndCreateTraceProducer(RPCHook rpcHook) {
         return traceProducerInstance;
     }
 
+    private String genGroupNameForTrace() {
+        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.group;
```
jonnxu: if it can trace when the multi consumers or producers with the same group name?

ChaosYjh: > if it can trace when the multi consumers or producers with the same group name?
Only the first traceProducer can trace, and the other will be start failed. If specify a instanceName for each group, it will create multi client instance.

ahuazhu: @jonnxu 
Thank you for the reminder. As far as I know, in RocketMQ two producers for different topics can't share the same ProduceGroup, and consumers also. But a consumer and a producer can have the same group. Unfortunately，the trace will failure in such scenes. To fix this problem, client type(consume or produce) will be appended to the trace-group to distinguish the consumer and producer. 

```
 private String genGroupNameForTrace() {
        return TraceConstants.GROUP_NAME_PREFIX + "-" + this.group + "-" + this.type ;
    }
```
When construct` AsyncTraceDispatcher.type` will be set as "PRODUCE" or "CONSUME". 

This PR has been updated.

And my test code as below:

```
public class SameProcessSameTopicTest {
    private static String topic1 = "TopicTest1";
    private static String topic2 = "TopicTest2";

//    private static final String CONSUMER_GROUP1 = "test_c_group1";
//    private static final String CONSUMER_GROUP2 = "test_c_group2";
//
//    private static final String PRODUCER_GROUP1 = "test_p_group1";
//    private static final String PRODUCER_GROUP2 = "test_p_group2";

    private static final String CONSUMER_GROUP1 = "test__group1";
    private static final String CONSUMER_GROUP2 = "test__group2";

    private static final String PRODUCER_GROUP1 = "test__group1";
    private static final String PRODUCER_GROUP2 = "test__group2";

    private static final String NAMR_SRV = "";

    public static void startConsume(final String id, final String group, final String topic) throws MQClientException {
        DefaultMQPushConsumer consumer = new DefaultMQPushConsumer(group, true);
        consumer.setConsumeFromWhere(ConsumeFromWhere.CONSUME_FROM_LAST_OFFSET);
        consumer.subscribe(topic, "*");
        consumer.setNamesrvAddr(NAMR_SRV);
        consumer.setInstanceName(id);
        final AtomicInteger i = new AtomicInteger(0);
        consumer.registerMessageListener(new MessageListenerConcurrently() {
            public ConsumeConcurrentlyStatus consumeMessage(List<MessageExt> msgs, ConsumeConcurrentlyContext context) {
                for (MessageExt msg : msgs) {
                    System.out.println(new Date() + ":" + new String(msg.getBody()) + " consume success! " + msg.getMsgId() + "  " + topic);
                }
                i.getAndAdd(msgs.size());

                System.err.println(id + " Consumed " + i.get() + " messages for " + topic);
            }
        });

        consumer.start();
    }

    public static void startProduce(String id, String group, String topic) throws MQClientException {
        DefaultMQProducer producer = new DefaultMQProducer(group,true);
        producer.setNamesrvAddr(NAMR_SRV);
        producer.start();
        producer.setInstanceName(id);
        AtomicInteger c = new AtomicInteger(0);
        for (int i = 0; i < 10 ;) {
            try {
                Message msg = new Message(topic /* Topic */, null /* Tag */,
                        ("trace message " + i).getBytes(RemotingHelper.DEFAULT_CHARSET) /* Message body */
                );
                producer.send(msg);
                Thread.sleep(50);
                i++;
                System.out.println(id + " send " + c.getAndIncrement() + " messages to " + topic);
            } catch (Exception e) {
//                e.printStackTrace();
            }
        }

    }


    public static void main(String[] args) throws InterruptedException, MQClientException {

        Thread t1 = new Thread() {
            @Override
            public void run() {
                try {
                    startConsume("C1", CONSUMER_GROUP1, topic1);
                } catch (MQClientException e) {
                    e.printStackTrace();
                }
            }
        };

        Thread t2 = new Thread() {
            @Override
            public void run() {
                try {
                    startConsume("C2", CONSUMER_GROUP2, topic2);
                } catch (MQClientException e) {
                    e.printStackTrace();
                }
            }
        };


        Thread t3 = new Thread() {
            @Override
            public void run() {
                try {
                    startProduce("P1", PRODUCER_GROUP1, topic1);
                } catch (MQClientException e) {
                    e.printStackTrace();
                }
            }
        };
//
        Thread t4 = new Thread() {
            @Override
            public void run() {
                try {
                    startProduce("P2", PRODUCER_GROUP2, topic2);
                } catch (MQClientException e) {
                    e.printStackTrace();
                }
            }
        };


        t1.start();
        t2.start();
        t3.start();
        t4.start();

        t1.join();
        t2.join();
        t3.join();
        t4.join();

        System.out.printf("Consumer Started.%n");
    }
} 

```

duhenglucky:
LGTM

yufeng629:
@duhenglucky  when release？Thanks！

PR #1313, opened 2019-07-11
add missing log

77yang:

add missing log

## Brief changelog

add missing log

## Verifying this change

XXXX

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
@@ -71,10 +71,9 @@ public class BrokerStartup {
             }
 
             log.info(tip);
-            System.out.printf("%s%n", tip);
             return controller;
         } catch (Throwable e) {
-            e.printStackTrace();
+            log.error("BrokerStartup fail to boot.", e);
             System.exit(-1);
         }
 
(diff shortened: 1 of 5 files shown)


broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
```diff
@@ -71,10 +71,9 @@ public static BrokerController start(BrokerController controller) {
             }
 
             log.info(tip);
-            System.out.printf("%s%n", tip);
             return controller;
         } catch (Throwable e) {
-            e.printStackTrace();
+            log.error("BrokerStartup fail to boot.", e);
```
ShannonDing: IMO, you can print the stack trace to the error log, BUT, it is better to keep the standard output to show the start error directly to the users, in this case, the users can find the error as soon as possible and deal with the error instead of monitor the log file.

77yang: @ShannonDing  The users will not know what happened when the program runs in the background if we use `e.printStackTrace();`  right?   And  I think it's a more standardized way of writing as an open source middleware. what do you think? 😀

xiangwangcheng: @77yang  When the application booted successfully, generally, we redirect the log to log file. And when it comes to the process of starting the application, I agree with what @ShannonDing said. It is friendly to users if they can know when and how the application failed to start QUICKLY(Of course, log them into file is also necessary).

77yang: @xiangwangcheng  So do we need to write both of them at the same time?  😅

xiangwangcheng: Yeah, that's what I mean.

77yang: @xiangwangcheng I have written both of them now, you can merge this PR.

PR #1329, opened 2019-07-18
Fix the package of TopicRouteDataTest

lebin:

The package of TopicRouteDataTest is wrong

## Brief changelog

Change the package of TopicRouteDataTest  to 'org.apache.rocketmq.common.protocol.route'.


xiangwangcheng:
Thanks for fixing that. And could u please change the destination branch to develop instead of master?

vongosling:
@lebin Thanks for your pr. we'll fix the problem in the nearing release version.

PR #1350, opened 2019-07-30
Remove no need code

skyline1688:

Removed the no need code:

I have removed the "datagramSocket" definition in MQClientInstance.java since I think this filed was not useful in project, when I read the source code I found nobody refer to it.


duhenglucky:
@skyline1688 DatagramSocket is reserved for future expansion support for UDP protocol, so please don't remove it for a while. and if you are interested in extending this protocol, please feel free to propose a RIP or PR. and I will close this pr first, but please feel free to reopen it if you have other questions.

skyline1688:
Okay I got it. thanks for your reply, and if there are some code comments there will be better, it's a good way to avoid confuse.

PR #1427, opened 2019-08-29
Fix offset maybe lost problem.

aftersss:
If two clients commit offset at the same time(the same topic and group, but different queueId), one of the queue's offset maybe lost.


xiangwangcheng:
Thanks for your contribution!
The code LGTM. 
But could u please change the destination branch to develop add some unit test to increase the coverage?

aftersss:
> Thanks for your contribution!
> The code LGTM.
> But could u please change the destination branch to develop add some unit test to increase the coverage?

Done.

Issue #1457, opened 2019-09-16
Updating broker config of namesrvAddr is not effective


liuruiyiyang:
**BUG REPORT**

1. Please describe the issue you observed:
When use ```mqadmin updateBrokerConfig -c clusterName``` to update namesrvAddr config.  The new parameter is not effective.

- What did you do (The steps to reproduce)?
Use ```mqadmin updateBrokerConfig -c clusterName -n oldNameServerList -k namesrvAddr -v newNameServerList ``` to update namesrvAddr config of a cluster

- What did you expect to see?
All brokers in the cluster changed their namesrvAddr and also the new namesrvAddr is effective by registering to new name server. namesrv log showing like ```register broker[0]to name server newAddress:9876 OK```

- What did you see instead?
The new namesrvAddr is not effective.

2. Please tell us about your environment:
MacOS and Linux

CodeIngL:
Mq can't recognize whether your behavior should let us register, here is a special key:brokerPermission to trigger。 try it

zhaoyansheng163:
This bug is also existed.
I  test version 4.7.1 and 4.9.1。
Use mqadmin updateBrokerConfig -c clusterName -n oldNameServerList -k namesrvAddr -v newNameServerList to update namesrvAddr config of a cluster
is  not effective

BrightHao:
> This bug is also existed. I test version 4.7.1 and 4.9.1。 Use mqadmin updateBrokerConfig -c clusterName -n oldNameServerList -k namesrvAddr -v newNameServerList to update namesrvAddr config of a cluster is not effective

I test version 4.7.0 on linux.
Use mqadmin updateBrokerConfig -c clusterName -n oldNameServerList -k namesrvAddr -v newNameServerList to update namesrvAddr config of a cluster
is also not effective

PR #1458, opened 2019-09-16
[ISSUE #1457] Dynamically effective namesrvAddr for name server auto-scaling

liuruiyiyang:
Fix issue #1457

## Brief changelog

1. Update all broker config when using -c param in ```mqadmin updateBrokerConfig``` subcommand
2. Make ```namesrvAddr``` parameter effective dynamically for name server auto-scaling triggered by ```mqadmin updateBrokerConfig``` command.

## Verifying this change

Use ```mqadmin updateBrokerConfig -c clusterName -n oldNameServerList -k namesrvAddr -v newNameServerList``` to see if all brokers in the cluster has log ```register broker[0]to name server newAddress:9876 OK```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -388,6 +388,20 @@ public class BrokerController {
                 }
             }, 1000 * 10, 1000 * 60, TimeUnit.MILLISECONDS);
 
+            if (this.brokerConfig.isWatchNamesrvAddr()) {
+                this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
+
+                    @Override
+                    public void run() {
+                        try {
+                            BrokerController.this.brokerOuterAPI.updateNameServerAddressList(brokerConfig.getNamesrvAddr());
+                        } catch (Throwable e) {
+                            log.error("ScheduledTask updateNameServerAddressList exception", e);
+                        }
+                    }
+                }, 1, 1, TimeUnit.MINUTES);
+            }
+
             if (this.brokerConfig.getNamesrvAddr() != null) {
                 this.brokerOuterAPI.updateNameServerAddressList(this.brokerConfig.getNamesrvAddr());
                 log.info("Set user specified name server address: {}", this.brokerConfig.getNamesrvAddr());
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -399,7 +399,7 @@ public class BrokerController {
                             log.error("ScheduledTask updateNameServerAddressList exception", e);
                         }
                     }
-                }, 1, 1, TimeUnit.MINUTES);
+                }, 1000 * 10, 1000 * 60 * 2, TimeUnit.MILLISECONDS);
             }
 
             if (this.brokerConfig.getNamesrvAddr() != null) {
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -388,20 +388,6 @@ public class BrokerController {
                 }
             }, 1000 * 10, 1000 * 60, TimeUnit.MILLISECONDS);
 
-            if (this.brokerConfig.isWatchNamesrvAddr()) {
-                this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
-
-                    @Override
-                    public void run() {
-                        try {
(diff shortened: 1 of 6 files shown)


broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -388,6 +388,20 @@ public void run() {
...
+
+                    @Override
+                    public void run() {
+                        try {
+                            BrokerController.this.brokerOuterAPI.updateNameServerAddressList(brokerConfig.getNamesrvAddr());
+                        } catch (Throwable e) {
+                            log.error("ScheduledTask updateNameServerAddressList exception", e);
+                        }
+                    }
+                }, 1, 1, TimeUnit.MINUTES);
```
duhenglucky: It would be better to keep consistent with the scheduled tasks in org.apache.rocketmq.broker.BrokerController#initialize(fetchNameServerAddr task)

liuruiyiyang: Frequency changed to keep consistent.

RongtongJin (dismissed):

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -388,6 +388,20 @@ public void run() {
                 }
             }, 1000 * 10, 1000 * 60, TimeUnit.MILLISECONDS);
 
+            if (this.brokerConfig.isWatchNamesrvAddr()) {
+                this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
+
+                    @Override
+                    public void run() {
+                        try {
+                            BrokerController.this.brokerOuterAPI.updateNameServerAddressList(brokerConfig.getNamesrvAddr());
```
vongosling: scheduledExecutorService is a global SingleThreadScheduledExecutor, how do we make sure that we have a reasonable schedule, and that if the schedule fails, it really doesn't affect the other logic?

vongosling: Do we need to filter out unnecessary update operation?

wqliang: Should it be under consideration that` fetchNamesrvAddrByAddressServer` is true when you configure atchNamesrvAddr to true and update the `namesrvAddr` as you mentioned ? Maybe this two configurations are contradictory.

wqliang:
Should it be under consideration that `fetchNamesrvAddrByAddressServer `is true when you configure `watchNamesrvAddr `to true and update the `namesrvAddras ` as you mentioned ? Maybe this two configurations are contradictory.

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
liuruiyiyang: This scheduled thread has been removed.

liuruiyiyang:
@wqliang Thanks for your question. The scheduled thread has been removed. The updated ```namesrvAddr``` applying process is triggered by the ```mqadmin updateBrokerConfig``` command only once now.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -493,6 +490,17 @@ private synchronized RemotingCommand updateBrokerConfig(ChannelHandlerContext ct
         return response;
     }
 
+    private void applyUpdatedConfig(Properties properties) {
+        if (properties.containsKey(MixAll.DUP_NAMESRV_ADDR)) {
+            String updatedNamesrvAddr = this.brokerController.getBrokerConfig().getNamesrvAddr();
+            this.brokerController.getBrokerOuterAPI().updateNameServerAddressList(updatedNamesrvAddr);
```
wqliang: [QUESTION]  If `fetchNamesrvAddrByAddressServer` is true and namesrAddr is null when broker startup, broker will update namesrv address from address server periodically.  After you update it , the scheduled task will overwrite it at next period. Maybe you should cancle the scheduled task if it exist. Or you can return some tips if such update schedule task exist. I think remind user and let him decide whether or not update and cancle the task might be better.

liuruiyiyang: In current design, the priority of ```mqadmin updateBrokerConfig namesrvAddr``` is lower than ```fetchNamesrvAddrByAddressServer```, which means the scheduled task will overwrite it. The purpose of this PR is to enable dynamically effective namesrvAddr for name server auto-scaling when ```fetchNamesrvAddrByAddressServer``` scheduled task is not enabled. If the ```fetchNamesrvAddrByAddressServer``` scheduled task is enabled, it already have the ability of  dynamically effective.

Therefore, I don't think we should cancel the scheduled task.

Git-Yang (approved):

PR #1462, opened 2019-09-19
skip encode null value property as string "null"

carlvine500:
skip encode null value property as string "null"

send msg property(x=null), but receive msg property(x="null")


vongosling:
@carlvine500  It would be helpful to verify this problem if you could add test case :-) I am willing to help you merge this pr.

carlvine500:
@vongosling there is a new test case for MessageDecoder now

ShannonDing:
link #1899

ShannonDing:
closed due to #1899

Issue #1477, opened 2019-09-23
loginfo is ambiguity when visiting namesrv


keranbingaa:
**BUG REPORT**

1. Please describe the issue you observed:
When client wants to get some info such as topicinfo from namesrv,  it call invokeSync in NettyRemotingClient.java, and the parameter addr is null. When some exception occur, the log which contains the parameter addr will be "the channel [null]", it's ambiguity and misleading.

2. Other information (e.g. detailed explanation, logs, related issues, suggestions how to fix, etc):
don't use the parameter addr in log.

duhenglucky:
@keranbingaa address seems more clear than the channel, IMHO, print all channel info isn't a good choice in here.

PR #1478, opened 2019-09-23
modify loginfo when visiting namesrv

keranbingaa:
work for this [issue](https://github.com/apache/rocketmq/issues/1477), the loginfo is ambiguity when visiting namesrv, so we modify the loginfo to make it clear.

## Brief changelog
use the variable channel to replace addr in the log statement.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
@@ -374,15 +374,15 @@ public class NettyRemotingClient extends NettyRemotingAbstract implements Remoti
                 doAfterRpcHooks(RemotingHelper.parseChannelRemoteAddr(channel), request, response);
                 return response;
             } catch (RemotingSendRequestException e) {
-                log.warn("invokeSync: send request exception, so close the channel[{}]", channel);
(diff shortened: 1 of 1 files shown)


remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
```diff
@@ -374,15 +374,15 @@ public RemotingCommand invokeSync(String addr, final RemotingCommand request, lo
                 doAfterRpcHooks(RemotingHelper.parseChannelRemoteAddr(channel), request, response);
                 return response;
             } catch (RemotingSendRequestException e) {
-                log.warn("invokeSync: send request exception, so close the channel[{}]", addr);
+                log.warn("invokeSync: send request exception, so close the channel[{}]", channel);
```
wqliang: I think `RemotingUtil.socketAddress2String(channel.remoteAddress())` would be better

duhenglucky:
@keranbingaa address seems more clear than the channel, IMHO, print all channel info isn't a good choice in here.

qqeasonchen:
@duhenglucky I modify what print out， and now it just print the address, and when the variable address is null, it will also print the remote address.

keranbingaa:
@duhenglucky  I modify what print out， and now it just print the address, and when the variable address is null, it will also print the remote address.

ChaosYjh:
I think printing address is more clear too.And just printing channel.remoteAddress() will be ok.

Issue #1519, opened 2019-10-11
Duplicated transaction check when queue of AbstractTransactionalMessageCheckListener.excuetorService is full


areyouok:
```
master1        | java.util.concurrent.RejectedExecutionException: Task org.apache.rocketmq.broker.transaction.AbstractTransactionalMessageCheckListener$2@1e179a79 rejected from java.util.concurrent.ThreadPoolExecutor@bb1ccc7[Running, pool size = 5, active threads = 4, queued tasks = 2000, completed tasks = 30247]
master1        | 	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
master1        | 	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
master1        | 	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379)
master1        | 	at org.apache.rocketmq.broker.transaction.AbstractTransactionalMessageCheckListener.resolveHalfMsg(AbstractTransactionalMessageCheckListener.java:80)
master1        | 	at org.apache.rocketmq.broker.transaction.queue.TransactionalMessageServiceImpl.check(TransactionalMessageServiceImpl.java:223)
master1        | 	at org.apache.rocketmq.broker.transaction.TransactionalMessageCheckService.onWaitEnd(TransactionalMessageCheckService.java:55)
master1        | 	at org.apache.rocketmq.common.ServiceThread.waitForRunning(ServiceThread.java:144)
master1        | 	at org.apache.rocketmq.broker.transaction.TransactionalMessageCheckService.run(TransactionalMessageCheckService.java:44)
master1        | 	at java.lang.Thread.run(Thread.java:748)
```

Because the queue of ExcuetorService (created in AbstractTransactionalMessageCheckListener) is full :
```java
private static ExecutorService executorService = new ThreadPoolExecutor(2, 5, 100, TimeUnit.SECONDS, new ArrayBlockingQueue<Runnable>(2000), new ThreadFactory() {
        @Override
        public Thread newThread(Runnable r) {
            Thread thread = new Thread(r);
            thread.setName("Transaction-msg-check-thread");
            return thread;
        }
    });
```

When the exception occurs, the newOffset of TransactionalMessageServiceImpl.check is not update to ConsumeQueue, so lots of duplicated check occurs.

duhenglucky:
[...]

PR #1521, opened 2019-10-12
[Issue #1519] Fix problem when queue of transaction check ExecutorService is full

areyouok:
1. catch RejectedExecutionException and retry
2. add shutdown status detect in TransactionalMessageServiceImpl.check


areyouok:
force push 了一下，这个测试是通过了，不知道为什么这里还是显示为未通过

PR #1544, opened 2019-10-18
[ISSUE #1519]Optimise performance/stability of  transaction message

areyouok:

Optimise performance/stability of  transaction message

## Brief changelog

1. Add -n (nameServer) option to benchmark consumer; Enable options(-n/-t/w/-s/-c/-r) to TransactionProducer
1. Fix bug that tx message will be checked even it has been committed or rollback.(Half message with unknown status may cause a lots of unnecessary check to half messages which send in 6 seconds before check startTime)
1. Fix problem when queue of transaction check ExecutorService is full
1. Optimise transaction check performance: fix problem that processed opOffset not add to doneOpOffset
1. Change log level, optimise performance of endTransaction and txCheckTask.(This may cause rate limit for endTransaction operation, but send half message is not limited so lots of commit/rollback operations lost)

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/transaction/AbstractTransactionalMessageCheckListener.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/transaction/AbstractTransactionalMessageCheckListener.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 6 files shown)


broker/src/main/java/org/apache/rocketmq/broker/transaction/AbstractTransactionalMessageCheckListener.java
```diff
@@ -48,7 +49,7 @@ public Thread newThread(Runnable r) {
             thread.setName("Transaction-msg-check-thread");
             return thread;
         }
-    });
+    }, new CallerRunsPolicy());
```
duhenglucky: Much better than the loop retry submission in the last commit.

Issue #1561, opened 2019-10-31
Support specify max reconsume times in message level for concurrently consume service


Jaskey:
**FEATURE REQUEST**

In some case, user may need to decide different max reconsume times  for different kind of messages(for example, messages from topic A should be reconsume 10 times , messages from topicB should be reconsumed for at most 3 times, the rest messages will be use the default max reconsume times say 16) 

In current implementations, consumer uses the same configuration of max reconsume times in cousumer level, though use set property  MAX_RECONSUME_TIMES, it has no effect. The related code is as below:

```
    public void sendMessageBack(MessageExt msg, int delayLevel, final String brokerName)
        throws RemotingException, MQBrokerException, InterruptedException, MQClientException {
        try {
            String brokerAddr = (null != brokerName) ? this.mQClientFactory.findBrokerAddressInPublish(brokerName)
                : RemotingHelper.parseSocketAddressAddr(msg.getStoreHost());
            this.mQClientFactory.getMQClientAPIImpl().consumerSendMessageBack(brokerAddr, msg,
                this.defaultMQPushConsumer.getConsumerGroup(), delayLevel, 5000, getMaxReconsumeTimes());// HERE! the max reconsumeTimes is always in consumer level
        } .......
```

YuShanMuGong:
Good Suggestion。 I Need

YuShanMuGong:
![image](https://user-images.githubusercontent.com/13248312/68103059-4bee9f80-ff10-11e9-897d-b4ef3d499eaf.png)
![image](https://user-images.githubusercontent.com/13248312/68103064-53ae4400-ff10-11e9-8d47-c7a1d0ae3991.png)

Hi ，I find code in `ConsumeMessageOrderlyService#processConsumeResult`.If do not set `defaultMQPushConsumer.maxReconsumeTimes`,the message will be reconsume `Integer.MAX_VALUE` times.
Why Not is 16 times ? same times as`ConsumeMessageConcurrentlyService`.

Jaskey:
@YuShanMuGong becaurce for orderly service, it is totally different. If we discard the message, order consuming will be broken since the next message may dependent on the previous message successfully consumed. Thus, in the case of order consumer service, the default value will be max value.

a83313886:
[...]

PR #1562, opened 2019-10-31
[ISSUE #1561]Support specify max reconsume times in message level

Jaskey:

setting MAX_RECONSUME_TIMES in message has no effect to control max reconsume times, this pr is intending to fix this issue.

Since the retry strategy between consuming concurrently and  consuming orderly, this pr is going to enhance concurrently only

## Brief changelog

When user need to control max reconsume times for a specific message, user may set MAX_RECONSUME_TIMES in the message property, it should be effected. But currently, it is not, only max reconsume times in consumer level will be effected.

        consumer.registerMessageListener(new MessageListenerConcurrently() {

            @Override
            public ConsumeConcurrentlyStatus consumeMessage(List<MessageExt> msgs,
                ConsumeConcurrentlyContext context) {
                System.out.printf("%s Receive New Messages: %s %n", Thread.currentThread().getName(), msgs);
                
                for (MessageExt msg : msgs) {
                    if (msg.getTopic().equals("TOPICA")) {
                        MessageAccessor.setMaxReconsumeTimes(msg, "5");//reconsume to 5 times if consome message from topicA fail 
                    }
                }
                
                //do consume now 
                if (consumefail){
                    return ConsumeConcurrentlyStatus.RECONSUME_LATER;
                }

            }
        });

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultMQPushConsumerImpl.java
@@ -510,11 +510,21 @@ public class DefaultMQPushConsumerImpl implements MQConsumerInner {
 
     public void sendMessageBack(MessageExt msg, int delayLevel, final String brokerName)
         throws RemotingException, MQBrokerException, InterruptedException, MQClientException {
+        int maxReconsumeTimes = getMaxReconsumeTimes();
         try {
             String brokerAddr = (null != brokerName) ? this.mQClientFactory.findBrokerAddressInPublish(brokerName)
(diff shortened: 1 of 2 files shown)


YuShanMuGong:
你好，你这个修改，只修改了并行消费模式的最大重试次数。。我觉得既然是个消息级别的设置，那么应该在各种消费模式下都能起作用。

我尝试写了一版功能，你们帮我Review一下吗？[PR1570](https://github.com/apache/rocketmq/pull/1570)

Jaskey:
@YuShanMuGong  Yes, this is only effective for concurrently service, since it is most popular. For orderly consume service, it may be different since it does not redeliever so it is not included in this pr

Jaskey:
@duhenglucky  would you please help to review this pr?

PR #1577, opened 2019-11-06
[ISSUE #1576] Fix logic in select MessageQueue

### LINKED ISSUE #1576: Bug in select MessageQueue
When SendLatencyFaultEnable is enabled and the currently selection is available, you need to make sure that the currently selected broker is not the last selected.
But this judgment is wrong for now:
```java
// MQFaultStrategy#selectOneMessageQueue
MessageQueue mq = tpInfo.getMessageQueueList().get(pos);
if (latencyFaultTolerance.isAvailable(mq.getBrokerName())) {
    // this judgment shuold be change to if(!mq.getBrokerName().equals(lastBrokerName))
    if (null == lastBrokerName || mq.getBrokerName().equals(lastBrokerName))
        return mq;
}
```

ShadowySpirits:
fix #1576


xiangwangcheng:
LGTM.

xiangwangcheng:
@zongtanghu @duhenglucky 
I also have some confusion about this code. What do u think?

ShannonDing:
link #1593

ShannonDing:
It seems duplicate with #1593

PR #1606, opened 2019-11-21
[ISSUE #1605] use ThreadLocalRandom instead of Random

### LINKED ISSUE #1605: [improvement] use ThreadLocalRandom instead of Random
** Performance improvement **

Random is thread safe for use by multiple threads. But if multiple threads use the same instance of Random, the same seed is shared by multiple threads. It leads to contention between multiple threads and so to performance degradation.

ThreadLocalRandom is solution to above problem. ThreadLocalRandom has a Random instance per thread and safeguards against contention.

According to jmh benchmark
Benchmark                                               Mode  Cnt     Score     Error  Units
RandomBenchmark.random                   avgt   10  2616.198 ± 316.004  ns/op
RandomBenchmark.threadRandom      avgt   10     3.764 ±   0.095  ns/op

binglind:

this commit resolves #1605  use ThreadLocalRandom instead of Random

## Brief changelog

1. remove the 1.6 maven compiler of common and log and client and remoting moudle
2. use ThreadLocalRandom instead of Random

## Verifying this change

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -27,10 +27,6 @@
     <artifactId>rocketmq-client</artifactId>
     <name>rocketmq-client ${project.version}</name>
 
-    <properties>
-        <maven.compiler.source>1.6</maven.compiler.source>
-        <maven.compiler.target>1.6</maven.compiler.target>
-    </properties>
 
     <dependencies>
         <dependency>
--- a/client/src/main/java/org/apache/rocketmq/client/common/ThreadLocalIndex.java
+++ b/client/src/main/java/org/apache/rocketmq/client/common/ThreadLocalIndex.java
@@ -17,16 +17,15 @@
 
 package org.apache.rocketmq.client.common;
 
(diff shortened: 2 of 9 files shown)


client/pom.xml
```diff
@@ -27,10 +27,6 @@
     <artifactId>rocketmq-client</artifactId>
     <name>rocketmq-client ${project.version}</name>
 
-    <properties>
```
RongtongJin: Do not modify compiler level. we need to keep to be compatible with JDK 1.6 in SDK

RongtongJin:
If Random method is not called frequently, using ThreadLocalRandom instead of Random will not substantially improve performance.

karthick-me:
Hi! I noticed this issue was previously closed, but the fix was never merged.
Is it okay if I take a fresh look at it and try to submit a working PR?
I’ve already set up the project and can start right away.

Thanks!

PR #1669, opened 2019-12-19
clean code

rushsky518:
org.apache.rocketmq.broker.client.ProducerManager#getAvaliableChannel 
is too redundant

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
@@ -206,31 +206,28 @@ public class ProducerManager {
     }
 
     public Channel getAvaliableChannel(String groupId) {
-        Channel checkChannel = null;
+        Channel channel = null;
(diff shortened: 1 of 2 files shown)


broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
```diff
@@ -206,35 +206,31 @@ public void unregisterProducer(final String group, final ClientChannelInfo clien
...
+        channelList = new ArrayList<Channel>(channelClientChannelInfoHashMap.keySet());
+        int index = positiveAtomicCounter.incrementAndGet() % channelList.size();
+        Channel channel = null;
+        boolean isOk = false;
+        int count = 0;
+        do {
+            channel = channelList.get(index);
+            isOk = channel.isActive() && channel.isWritable();
+            if (isOk) {
+                checkChannel = channel;
```
xiangwangcheng: `checkCannel` and `channel` local variable seem a little bit redundant. Is it cleaner to use one?

PR #1748, opened 2020-01-27
Object's equals method is prone to null pointer exceptions

WellsYuu:
the equals method of Object is prone to null pointer exceptions and should be invoked using constants or objects that determine values


RongtongJin:
IMO most of the places you modify have done NPE prevention.

thenx-wei:
@RongtongJin but i think the should be more standardized

PR #1749, opened 2020-01-28
ResetOffsetByTimeCommand supports specified queue

rushsky518:
ResetOffsetByTimeCommand supports specified queue:
`mqadmin.cmd resetOffsetByTime -n localhost:9876 -b localhost:10911 -t TopicTest -q "0,1" -g group-name -s 2020-01-28#12:10:00:000`

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
@@ -17,6 +17,9 @@
 
 package org.apache.rocketmq.tools.command.offset;
 
+import java.util.Iterator;
+import java.util.Map;
+
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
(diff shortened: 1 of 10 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeOldCommand.java
```diff
@@ -119,7 +123,7 @@ public void execute(CommandLine commandLine, Options options, RPCHook rpcHook) t
                 }
 
                 defaultMQAdminExt.start();
-                resetOffset(defaultMQAdminExt, consumerGroup, topic, timestamp, force, timeStampStr);
+                resetOffset(defaultMQAdminExt, consumerGroup, topic, null, timestamp, force, timeStampStr);
```
xujianhai666: null? maybe mistake. I thought u should convey paramater queue, but u not get queue from commandLine

tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
```diff
@@ -49,10 +50,18 @@ public Options buildCommandlineOptions(Options options) {
...
 
+        opt = new Option("b", "broker", true, "set broker address");
+        opt.setRequired(false);
+        options.addOption(opt);
+
         opt = new Option("t", "topic", true, "set the topic");
         opt.setRequired(true);
         options.addOption(opt);
 
+        opt = new Option("i", "queue", true, "set the queue, eg: 0,1");
```
xujianhai666: use parameter q instead of i, which is understandable.

Issue #1655, opened 2019-12-16
ClientRemotingProcessor execute receiveReplyMessage method timeout


hdp-piplus:
main reasons are as follows
```
if (requestHeader.getBornHost() != null) {
                msg.setBornHost(RemotingUtil.string2SocketAddress(requestHeader.getBornHost()));
            }

  if (requestHeader.getStoreHost() != null) {
                msg.setStoreHost(RemotingUtil.string2SocketAddress(requestHeader.getStoreHost()));            }
```
the RemotingUtil.string2SocketAddress method  create new InetSocketAddress instance , when run code ``addr = InetAddress.getByName(hostname);``  it cost a long time,  and   the ``RequestResponseFuture``   instance was  removed from ``RequestFutureTable``   map。

env : 
os: centos linux 7 64bit
jdk: openjdk 8 
ip:  172.18.5.188 eth0

the mechine is one of the kubernates node ， and there are many vethxxxx network device

PR #1773, opened 2020-02-14
[ISSUE #1655] bornhost and storehost use string 

rushsky518:
bornhost and storehost use string rather than SocketAddress


xujianhai666:
@rushsky518  more ci error info: `error: Unused import - org.apache.rocketmq.remoting.common.RemotingUtil.`

PR #1792, opened 2020-02-26
Fix get wrong client ip when client running in linux with container t…

nic-6443:
Fix get wrong client ip when client running in linux with container tech, like docker will create docker0 network interface, etc..

Some linux container tech will create many network interface, like docker/k8s cni, when rocketmq client run within those machine will get wrong client ip to generate clientId.

## Brief changelog

Skip network inteface as clientIP candidate when it's status is not up


CoderLan0668:
it works  in the container for aliyun eci, related to: #1791

xiangwangcheng:
Could you please change the destination branch to develop?

ProNic-QY:
> Could you please change the destination branch to develop?

done

ShannonDing:
link #1877

PR #1797, opened 2020-02-27
[ISSUE #1796]feat(admin): filter deleted group for getConsumeStats

### LINKED ISSUE #1796: [Broker] useless code
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
AdminBrokerProcessor#getConsumeStats 里面

image

这里，如果 group 已经没有连接了，868行就是 null 了，但是 这个时候 872 行 结果肯定是 0 > 0 的判断，这个逻辑很拗口。感觉应该把 872 行的判断去掉
image

image

2. Provide any additional detail on your proposed use case for this feature.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

4. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:

- [sub-task1-issue-number](example_sub_issue1_link_here): sub-task1 description here, 
- [sub-task2-issue-number](example_sub_issue2_link_here): sub-task2 description here,
- ...

xujianhai666:
Closes #1796

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -419,7 +419,7 @@ public class AdminBrokerProcessor extends AsyncNettyRequestProcessor implements 
 
         final RemotingCommand response = RemotingCommand.createResponseCommand(GetBrokerAclConfigResponseHeader.class);
 
-        final GetBrokerAclConfigResponseHeader responseHeader = (GetBrokerAclConfigResponseHeader)response.readCustomHeader();
+        final GetBrokerAclConfigResponseHeader responseHeader = (GetBrokerAclConfigResponseHeader) response.readCustomHeader();
 
         try {
             AccessValidator accessValidator = this.brokerController.getAccessValidatorMap().get(PlainAccessValidator.class);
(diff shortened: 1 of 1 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -864,17 +864,6 @@ private RemotingCommand getConsumeStats(ChannelHandlerContext ctx,
                 continue;
             }
 
-            {
-                SubscriptionData findSubscriptionData =
-                    this.brokerController.getConsumerManager().findSubscriptionData(requestHeader.getConsumerGroup(), topic);
-
-                if (null == findSubscriptionData
```
duhenglucky: findSubscriptionData = null not means the subscription count is 0 because one consumer group can subscribe more than one topic.

xujianhai666: got it

xujianhai666: @duhenglucky no as ur expected, as code above, when consumerGroupInfo==null, then findSubscriptionDataCount will return 0.

xujianhai666: @duhenglucky I know ur idea, but this code should check the case: group is not used, then the findSubscriptionData return null and findSubscriptionDataCount return 0. As the code run, the not used group will return back to user. so maybe we should distinguish two case, one as u say above, and the second as I refer. Do u agree with that?

duhenglucky: yep, if the topic was set in the request, these codes look a bit messy,  and we need to deal with this in an elegant way

xujianhai666: @duhenglucky According to discuss above, I update code to distinguish group deleted and sub many topic.

PR #1799, opened 2020-02-27
[ISSUE #1798]Remove group of deleted sub info when queryTopicConsumeB…

### LINKED ISSUE #1798: [Admin] remove group of deleted sub info when queryTopicConsumeByWho
**BUG REPORT**

使用命令 `mqadmin deleteSubGroup`， 还能通过监控发现 有流量监控，发现原来是 queryTopicConsumeByWho 方法没有过滤掉 group
除此之外, 在同步group信息的时候，已经被删除的group 是没有同步价值的, 也不应该返回给 业务方.

- What did you expect to see?

- What did you see instead?

3. Other information (e.g. detailed explanation, logs, related issues, suggestions how to fix, etc):

xujianhai666:
when use `mqadmin deleteSubGroup` to delte subInfo, broker will
remove subInfo from SubscriptionGroupManager. but when admin use
`queryTopicConsumeByWho` to get consumer info, the delted group
of sub will back, which is not expected.

- use SubscriptionGroupManager to filter groups

Closes #1798

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -419,7 +419,7 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
 
         final RemotingCommand response = RemotingCommand.createResponseCommand(GetBrokerAclConfigResponseHeader.class);
 
-        final GetBrokerAclConfigResponseHeader responseHeader = (GetBrokerAclConfigResponseHeader)response.readCustomHeader();
+        final GetBrokerAclConfigResponseHeader responseHeader = (GetBrokerAclConfigResponseHeader) response.readCustomHeader();
 
         try {
             AccessValidator accessValidator = this.brokerController.getAccessValidatorMap().get(PlainAccessValidator.class);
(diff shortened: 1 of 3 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -1025,8 +1026,15 @@ private RemotingCommand queryTopicConsumeByWho(ChannelHandlerContext ctx,
             groups.addAll(groupInOffset);
         }
 
+        HashSet<String> filteredGroups = new HashSet<String>();
```
duhenglucky: It would be better to make an optional configuration here, because some users also want to show the consumption progress of topics that they have subscribed to, and it seems that users in the community are already used to this usage, so forward compatibility is necessary

xujianhai666: @duhenglucky so maybe I should add request flag for client, then broker will check flag bool indicator to dicide to filter ?

duhenglucky: @ IMHO, it seems that add a flag in the client will be a better choice

wz2cool:
Any update about this issue. we meet the same problem.

PR #1801, opened 2020-02-27
add client custom nameserver TopAddressing

Cczzzz:
I change TopAddressing to interface.
```
public interface TopAddressing {

    String fetchNSAddr();

    void registerChangeCallBack(Consumer<String> changeCallBack);
}
```
Uesr can ues ServiceLoader to loading  custom  TopAddressing implementation. When there is a custom implementation,Custom implementations are preferred.
```
   private TopAddressing findCustomTopAddressing() {
        ServiceLoader<TopAddressing> serviceLoader = ServiceLoader.load(TopAddressing.class);
        Iterator<TopAddressing> iterator = serviceLoader.iterator();
        if (iterator.hasNext()) {
            return iterator.next();
        }
        return null;
    }
```
I also provide a way to call back.this is a methods in interface TopAddressing. MQClientAPIImpl will register  call back  function to perception changes 
```
  this.topAddressing.registerChangeCallBack(this::fetchNameServerAddr);
```


client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -212,20 +213,25 @@ public RemotingClient getRemotingClient() {
...
-                }
-            }
+            return fetchNameServerAddr(addrs);
         } catch (Exception e) {
             log.error("fetchNameServerAddr Exception", e);
         }
         return nameSrvAddr;
     }
 
+    public String fetchNameServerAddr(String addrs){
```
Cczzzz: I'm going to split up the method so it's easy to call

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -174,16 +174,17 @@
     }
 
     private final RemotingClient remotingClient;
-    private final TopAddressing topAddressing;
+    private final TopAddressingImpl topAddressing;
     private final ClientRemotingProcessor clientRemotingProcessor;
-    private String nameSrvAddr = null;
+    private volatile String nameSrvAddr = null;
```
Cczzzz: It could be another thread in the callback function

xujianhai666:
@Cczzzz Is there any issue about this pr?

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -215,20 +216,25 @@ public RemotingClient getRemotingClient() {
...
+            return fetchNameServerAddr(addrs);
         } catch (Exception e) {
             log.error("fetchNameServerAddr Exception", e);
         }
         return nameSrvAddr;
     }
 
+    public String fetchNameServerAddr(String addrs){
+        if (addrs != null) {
+            if (!addrs.equals(this.nameSrvAddr)) {
```
xujianhai666: maybe we could combine two statements info one.

xujianhai666: why not rename method `onNameServerChange` ? because is just deal with nameserver change and then update if necessary

lizhanhui:
A few issues in this pull request.
1.  Please read [how to contribue](http://rocketmq.apache.org/docs/how-to-contribute/) page. In particular, PR should be made on top of "develop" branch; Code style should be fixed; Client related code modules are assumed to be Java 1.6 compatible. 
2. TopAddressing.java misses copyright info;

xujianhai666:
@Cczzzz something wrong about ci

lizhanhui:
> @Cczzzz something wrong about ci `/home/travis/build/apache/rocketmq/common/src/main/java/org/apache/rocketmq/common/namesrv/TopAddressing.java:1: error: Line does not match expected header line of '/\*'.`

See issue #2  of my previous comment.

lizhanhui:
By the way, I prefer to rename TopAddressingImpl  to DefaultTopAddressing because it provides a default implementation if no customized ones are loaded.

lizhanhui:
#findCustomTopAddressing, Should we keep a chain of implementations instead of the first one? Sort them according to some criteria, say an annotation @Order(index).
In method #fetchNSAddr, call the impls in order until one impl  returns address list that makes sense.

Cczzzz:
This is my first time to submit the request, thank you, he, it has a lot of problems and I will modify it.
@lizhanhui  Your suggestion is very good

ShannonDing:
Could ypu pls move the PR to the develop branch? @Cczzzz

PR #1808, opened 2020-02-28
Add a index query unit test

rushsky518:

Add a index query unit test


xujianhai666:
@rushsky518 ci failed

PR #1814, opened 2020-03-02
Filter message key on server side when query message

rushsky518:

Now, message key is filtered on client side when query message.
It would be better to filter key on server side.


client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
```diff
@@ -381,42 +380,12 @@ public void operationComplete(ResponseFuture responseFuture) {
                     for (MessageExt msgExt : qr.getMessageList()) {
                         if (isUniqKey) {
                             if (msgExt.getMsgId().equals(key)) {
-
-                                if (messageList.size() > 0) {
```
xujianhai666: why delte this logic?

rushsky518: i don't understand this logic is for what.
In unexpected situation, producer send msgs to broker(s) with the same uniqKey, when query msg with the key, users may want to see all the msgs.

Issue #1770, opened 2020-02-12
Add a query message trace command in mqadmin


zhangjidi2016:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
Add a query message trace command in mqadmin

2. Provide any additional detail on your proposed use case for this feature.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

4. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:

- [sub-task1-issue-number](example_sub_issue1_link_here): sub-task1 description here, 
- [sub-task2-issue-number](example_sub_issue2_link_here): sub-task2 description here,
- ...

zongtanghu:
I think this feature for message tracing is critical. @duhenglucky @zhangjidi2016 Users can use this command to query messaging tracing details.

zhangjidi2016:
Yes， I will submit a pr to add this command later.

zongtanghu:
This issue related pr has already been merge, so I close this issue. @duhenglucky @zhangjidi2016 @RongtongJin

PR #1824, opened 2020-03-05
[ISSUE #1770] Add a query message trace command in mqadmin.

zhangjidi2016:

Add a query message trace command in mqadmin.

## Brief changelog

Add a query message trace command in mqadmin.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/trace/TraceDataEncoder.java
+++ b/client/src/main/java/org/apache/rocketmq/client/trace/TraceDataEncoder.java
@@ -52,15 +52,16 @@ public class TraceDataEncoder {
                 bean.setTags(line[6]);
                 bean.setKeys(line[7]);
                 bean.setStoreHost(line[8]);
-                bean.setBodyLength(Integer.parseInt(line[9]));
-                pubContext.setCostTime(Integer.parseInt(line[10]));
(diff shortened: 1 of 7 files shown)


zhangjidi2016:
The command result in console，@zongtanghu @duhenglucky ，please help to review it，thanks！
image

client/src/main/java/org/apache/rocketmq/client/trace/TraceDataEncoder.java
```diff
@@ -52,15 +52,16 @@
...
 
-                if (line.length == 13) {
-                    pubContext.setSuccess(Boolean.parseBoolean(line[12]));
-                } else if (line.length == 14) {
-                    bean.setOffsetMsgId(line[12]);
+                if (line.length == 14) {
                     pubContext.setSuccess(Boolean.parseBoolean(line[13]));
+                } else if (line.length == 15) {
+                    bean.setOffsetMsgId(line[13]);
+                    pubContext.setSuccess(Boolean.parseBoolean(line[14]));
```
zongtanghu: Here, you must think of a critical compatibility problem that users depend on the older version of rocketmq client ,such as 4.4.0 or 4.5.0, then it send older format's message trace data, so you can add this command can't resolve this problem.

PR #1835, opened 2020-03-11
[ISSUE #1857] Refactor the duplicated code

Rookiexu:

Extract the code block copied three times into a separate method


vongosling:
@Rookiexu Thanks for your refactoring. It would be better to open an issue before you summit a pr.

Rookiexu:
> @Rookiexu Thanks for your refactoring. It would be better to open an issue before you summit a pr.

OK,i agree with you.This is my issue [issue1857](https://github.com/apache/rocketmq/issues/1857)

PR #1841, opened 2020-03-13
[ISSUE #1839] UpdateAccessConfigSubCommand will broke the consistency of master and slave 

### LINKED ISSUE #1839: [tools]UpdateAccessConfigSubCommand will broke the consistency of master and slave
**BUG REPORT**
the UpdateAccessConfigSubCommand in tools only update acl config in master broker,but if master broker offline, consumer will continue consume msgs from slave.it failed because the acl config is not update when use this command.

coder-zzzz:
fix #1839


yuz10:
#3184

PR #1897, opened 2020-03-31
Correct the Java doc for MessageListenerOrderly.java

Crearns:
Correct the java doc


vongosling:
Thanks for your rectification. It would be better to create issue before your request the pr, just like other folks do.

PR #1899, opened 2020-03-31
[ISSUE #1463] Skip encode the null value property as string "null"

carlvine500:

@ShannonDing  【Could you please change the merge dest to apache: develop branch? 】

OK


carlvine500:
@ShannonDing  can I close this issue now ?

PR #1903, opened 2020-04-01
[ISSUE #1904] Print log when flush timeout

rushsky518:

Print log when flush timeout

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -660,14 +660,20 @@ public class CommitLog {
         storeStatsService.getSinglePutMessageTopicTimesTotal(msg.getTopic()).incrementAndGet();
         storeStatsService.getSinglePutMessageTopicSizeTotal(topic).addAndGet(result.getWroteBytes());
 
-        CompletableFuture<PutMessageStatus> flushResultFuture = submitFlushRequest(result, putMessageResult, msg);
-        CompletableFuture<PutMessageStatus> replicaResultFuture = submitReplicaRequest(result, putMessageResult, msg);
+        CompletableFuture<PutMessageStatus> flushResultFuture = submitFlushRequest(result, msg);
+        CompletableFuture<PutMessageStatus> replicaResultFuture = submitReplicaRequest(result, msg);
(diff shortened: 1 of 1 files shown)


rushsky518:
#1904

ShannonDing:
It seems to conflict with #1894, could you please check again?

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -761,15 +767,20 @@ public long getBeginTimeInLock() {
...
 
-        CompletableFuture<PutMessageStatus> flushOKFuture = submitFlushRequest(result, putMessageResult, messageExtBatch);
-        CompletableFuture<PutMessageStatus> replicaOKFuture = submitReplicaRequest(result, putMessageResult, messageExtBatch);
+        CompletableFuture<PutMessageStatus> flushOKFuture = submitFlushRequest(result, messageExtBatch);
+        CompletableFuture<PutMessageStatus> replicaOKFuture = submitReplicaRequest(result, messageExtBatch);
         return flushOKFuture.thenCombine(replicaOKFuture, (flushStatus, replicaStatus) -> {
             if (flushStatus != PutMessageStatus.PUT_OK) {
-                putMessageResult.setPutMessageStatus(PutMessageStatus.FLUSH_DISK_TIMEOUT);
+                putMessageResult.setPutMessageStatus(flushStatus);
+                log.error("do groupcommit, wait for flush failed, topic: {} tags: {} client address: {}",
```
ShannonDing, line 775: If the disk flashing has failed, and continue to log here, it will increase the system load and increase disk IO.
IMO, as the status will be sent to the client, it is better to remove the error log here.

rushsky518: Indeed so.
I still have a question about [#1894 pr]
````java
req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
````
flushOK is false means FLUSH_DISK_TIMEOUT?

RongtongJin (approved):

vongosling:
Need to associate with milestone :-) @RongtongJin

PR #1920, opened 2020-04-08
[ISSUE #1558] Make ACL module compatible with 1.6

### LINKED ISSUE #1558: Whether module `acl` can compatible with 1.6 as same as module `client`
when use client with acl to connect broker, it will throw compatibility exception，as follows：
`java.lang.UnsupportedClassVersionError: org/apache/rocketmq/acl/common/AclClientRPCHook : Unsupported major.minor version 52.0`.
I read the latest(4.5.2) client version's pom.xml,it compile with 1.6 . but module ‘acl’ inherit parent java version is 1.8.  

Whether module `acl` can compatible with 1.6 as same as module `client`?

geektao:

fix #1558 module acl compatible with 1.6

### DIFF
--- a/acl/pom.xml
+++ b/acl/pom.xml
@@ -21,6 +21,9 @@
     <url>http://maven.apache.org</url>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <maven.compiler.source>1.6</maven.compiler.source>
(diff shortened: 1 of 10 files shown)


duhenglucky:
@geektao could you fix this PR based on the CI result?

zongtanghu:
Good job! And please check the CI result.If the CI result is ok, I think we can merge this pr! @duhenglucky @geektao

PR #1958, opened 2020-04-22
[ISSUE 1950] fix DefaultMQProducer didn't shutdown completely

### LINKED ISSUE #1950: when DefaultMQProducerImpl shutdown didn't shutdown some demon thread.
**BUG REPORT**

   start a DefaultMQProducer producer , after send message , then shutdown it, but the main thread is not finish.

- What did you expect to see?
    I want all the MQ producer relative threads is finished. 
- What did you see instead?
  there is still some demon threads left there .
     RequestHouseKeepingService , and AsyncAppender-Dispatcher-Thread-3,4,6

     windows 64bit  eclipse 
3. Other information (e.g. detailed explanation, logs, related issues, suggestions how to fix, etc):
     for the RequestHouseKeepingService , there is because timer is not shutdown
     for the AsyncAppender-Dispatcher-Thread , I didn't found it now

### LINKED ISSUE #1957: DefaultMQProducer didn't shutdown completely
**BUG REPORT**
class DefaultMQProducerImpl 
when new a DefaultMQProducerImpl instance, it will create a timer to scanExpireRequest at the same time.But didn't cancle the timer when this instance is shutdown.
If somebody create a lot of producers then shutdown them, there are a lot RequestHouseKeepingService threads status TIMED_WAITING.It will cause CPU usage and other resources.
image

coder-zzzz:
fix #1957 DefaultMQProducer didn't shutdown completely
fix #1950

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -853,7 +853,6 @@ public class MQClientInstance {
                     this.scheduledExecutorService.shutdown();
                     this.mQClientAPIImpl.shutdown();
                     this.rebalanceService.shutdown();
-                    this.defaultMQProducer.shutdown();
 
                     if (this.datagramSocket != null) {
                         this.datagramSocket.close();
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -853,6 +853,7 @@ public void shutdown() {
                     this.scheduledExecutorService.shutdown();
                     this.mQClientAPIImpl.shutdown();
                     this.rebalanceService.shutdown();
+                    this.defaultMQProducer.shutdown();
```
RongtongJin, line 856: `this.defaultMQProducer.shutdown();` and `this.defaultMQProducer.getDefaultMQProducerImpl().shutdown(false);`  conflict?

coder-zzzz: oh, it is conflict ,I delete the last line

PR #1985, opened 2020-05-12
[ISSUE #1976]System topic should add permission checking globally

### LINKED ISSUE #1976: System Topic like SCHEDULE_TOPIC_XXXX should not be create or delete by user
**FEATURE REQUEST**
If I create a topic named SCHEDULE_TOPIC_XXXX , then the delay msg will be put into this topic in cycle~~
I think it is must to forbid create topic named SCHEDULE_TOPIC_XXXX

coder-zzzz:
fix #1976 system topic should not be create or delete by user


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -285,7 +273,19 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
...
 
-        return null;
+        response.setCode(ResponseCode.SUCCESS);
+        return response;
+    }
+
+    private boolean isSystemTopic(RemotingCommand response, String topic) {
+        if (this.brokerController.getTopicConfigManager().isSystemTopic(topic)) {
+            String errorMsg = "the topic[" + topic + "] is conflict with system reserved words.";
+            log.warn(errorMsg);
```
RongtongJin, line 283: You can combine two code into one and use placeholder like `log.warn("xxxxxxx { } xxxx", topic)`

coder-zzzz:
> Do you need to check in the msgCheck method? There will create topic automatically.

@RongtongJin you are right, but the better way is to check topic in TopicConfigManager's createTopicInSendMessageMethod and createTopicInSendMessageBackMethod
I'm do it  later

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
coder-zzzz: errorMsg used in log and response's remark, so it is better to use a local var to store it

RongtongJin: > errorMsg used in log and response's remark, so it is better to use a local var to store it

You are right

coder-zzzz:
> Do you need to check in the msgCheck method? There will create topic automatically.

@RongtongJin hi, I check the code. broker will auto create topic only when the topic is not exists in the broker ,but the system topic is init when broker startup. So it will not auto create the system when send msg to broker.
But there is a necessary to forbid message send to SCHEDULE_TOPIC_XXXX，because it will cause endless loop

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -252,29 +253,16 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
             (CreateTopicRequestHeader) request.decodeCommandCustomHeader(CreateTopicRequestHeader.class);
         log.info("updateAndCreateTopic called by {}", RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
 
-        if (requestHeader.getTopic().equals(this.brokerController.getBrokerConfig().getBrokerClusterName())) {
-            String errorMsg = "the topic[" + requestHeader.getTopic() + "] is conflict with system reserved words.";
```
duhenglucky, line 256: the -> The

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -285,7 +273,19 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
 
         this.brokerController.registerIncrementBrokerData(topicConfig, this.brokerController.getTopicConfigManager().getDataVersion());
 
-        return null;
+        response.setCode(ResponseCode.SUCCESS);
+        return response;
+    }
+
+    private boolean isSystemTopic(RemotingCommand response, String topic) {
+        if (this.brokerController.getTopicConfigManager().isSystemTopic(topic)) {
```
duhenglucky, line 281: It would be better to integrate the two methods: org.apache.rocketmq.common.MixAll#isSystemTopic

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
coder-zzzz: ok

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
coder-zzzz: The MixAll#isSystemTopic is used in NamespaceUtil, but NamespaceUtil is use in both broker and client. Some topic can be mixed like AUTO_CREATE_TOPIC_KEY_TOPIC, but some can't, such as the topic of broker' clusterName.
I try to use a new method in nameserver which offer the ability to judge if the topic will be created is system topic before this pr.But I found there always will be mistake when some broker register itself to nameserver after the judge request. 
So I just promise broker itself will not create topics which it thought is system topic.

Do u have any ideas to fix it?

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -285,7 +273,19 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
...
+            String errorMsg = "The topic[" + topic + "] is conflict with system reserved words.";
```
vongosling, line 282: String errorMsg = "The topic[" + topic + "] is conflict with system reserved words.";

The reminder info(reserved words should update system topic )is not uniform with your modification.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
vongosling: It is a good time to arrange the optimization for system topic or special inner topic globally in this pr.

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
```diff
@@ -124,6 +124,14 @@ public TopicConfigManager(BrokerController brokerController) {
             topicConfig.setWriteQueueNums(1);
             this.topicConfigTable.put(topicConfig.getTopicName(), topicConfig);
         }
+        {
+            String topic = MixAll.SCHEDULE_TOPIC;
+            TopicConfig topicConfig = new TopicConfig(topic);
```
vongosling, line 129: SCHEDULE_TOPIC is a special system topic? IMO, we should refactor this naming logic here.

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
```diff
@@ -124,6 +124,14 @@ public TopicConfigManager(BrokerController brokerController) {
...
+            this.systemTopicList.add(topic);
+            topicConfig.setReadQueueNums(18);
+            topicConfig.setWriteQueueNums(18);
```
vongosling, line 132: Should use static number while not a special number 18

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -56,6 +56,7 @@
     //http://jmenv.tbsite.net:8080/rocketmq/nsaddr
     //public static final String WS_ADDR = "http://" + WS_DOMAIN_NAME + ":8080/rocketmq/" + WS_DOMAIN_SUBGROUP;
     public static final String AUTO_CREATE_TOPIC_KEY_TOPIC = "TBW102"; // Will be created at broker when isAutoCreateTopicEnable
+    public static final String SCHEDULE_TOPIC = "SCHEDULE_TOPIC_XXXX";
```
vongosling, line 59: MixAll class should not inflate continue. could you help to move the system topic related logic to a new class, such as TopicValidator or TopicChecker?

store/src/main/java/org/apache/rocketmq/store/dledger/DLedgerCommitLog.java
```diff
@@ -383,7 +384,7 @@ public PutMessageResult putMessage(final MessageExtBrokerInner msg) {
                     msg.setDelayTimeLevel(this.defaultMessageStore.getScheduleMessageService().getMaxDelayLevel());
                 }
 
-                topic = ScheduleMessageService.SCHEDULE_TOPIC;
+                topic = MixAll.SCHEDULE_TOPIC;
```
vongosling, line 387: I don't like to see many MixAll scatter in so many places:-)

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
coder-zzzz: ok

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
coder-zzzz: Yeah, it is a special system topic used in delay msg. When producer send a delay  msg, the msg will be send to SCHEDULE_TOPIC, then send to the real topic when the delay time comes.
Do u meaning  add prefix to system topic , or remove system topics to a independent class?

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
coder-zzzz: I will do it.

common/src/main/java/org/apache/rocketmq/common/MixAll.java
coder-zzzz: It's a good way to make code cleaner.I'will try my best to do it

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
vongosling: system topic should be a List, also, we could judge it from the naming, a prefix is a good way.

common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
```diff
@@ -57,13 +85,24 @@ public static boolean validateTopic(String topic, RemotingCommand response) {
...
         }
 
-        //whether the same with system reserved keyword
-        if (topic.equals(MixAll.AUTO_CREATE_TOPIC_KEY_TOPIC)) {
+        if (isSystemTopic(topic)) {
             response.setCode(ResponseCode.SYSTEM_ERROR);
-            response.setRemark("The specified topic is conflict with AUTO_CREATE_TOPIC_KEY_TOPIC.");
+            response.setRemark("The topic[" + topic + "] is conflict with system topic.");
             return false;
         }
```
RongtongJin, line 92: [Important] There is a serious problem. msgCheck call this method and check all system topics, this will cause the client to fail to send messages to **all system topics**, but it is necessary for the client to send messages to certain system topics, such as BenchmarkTest, OFFSET_MOVED_EVENT.  IMO, We need a blacklist to prohibit sending to certain system topics, but not all system topics.

coder-zzzz: This is really a deadly problem. Schedule Topic is the only one should prevent send by user,I’ll check the method TopicValidator#validateTopic everywhere invoked.

broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
```diff
@@ -173,13 +173,10 @@ protected RemotingCommand msgCheck(final ChannelHandlerContext ctx,
...
 
-        if (TopicValidator.RMQ_SYS_SCHEDULE_TOPIC.equals(requestHeader.getTopic())) {
-            response.setCode(ResponseCode.NO_PERMISSION);
-            response.setRemark("sending message to topic[" + requestHeader.getTopic() + "] is forbidden");
+        if (!TopicValidator.validateTopic(requestHeader.getTopic(), response)) {
             return response;
         }
-
-        if (!TopicValidator.validateTopic(requestHeader.getTopic(), response)) {
+        if (!TopicValidator.validateBlacklistTopic(requestHeader.getTopic(), response)) {
```
vongosling, line 179: what is a blacklist? It's just not allowed to send a message to this topic. It would be better to name it notAllowedSending.

broker/src/test/java/org/apache/rocketmq/broker/topic/TopicValidatorTest.java
```diff
@@ -70,6 +64,62 @@ public void testAddSystemTopic() {
         assertThat(TopicValidator.getSystemTopicSet()).contains(topic);
     }
 
+    @Test
+    public void testValidateSystemTopic() {
+        RemotingCommand response = RemotingCommand.createResponseCommand(-1, "");
+
+        String topic = TopicValidator.RMQ_SYS_SCHEDULE_TOPIC;
+        Boolean res = TopicValidator.validateSystemTopic(topic, response);
+        assertThat(res).isFalse();
```
vongosling, line 73: 1. validateSystemTopic -> isSystemTopic ?
2. RMQ_SYS_SCHEDULE_TOPIC is not system topic? why return false here.

broker/src/test/java/org/apache/rocketmq/broker/topic/TopicValidatorTest.java
```diff
@@ -70,6 +64,62 @@ public void testAddSystemTopic() {
...
+        res = TopicValidator.validateSystemTopic(topic, response);
+        assertThat(res).isTrue();
+    }
+
+    @Test
+    public void testValidateBlacklistTopic() {
+        RemotingCommand response = RemotingCommand.createResponseCommand(-1, "");
+
+        String topic = TopicValidator.RMQ_SYS_SCHEDULE_TOPIC;
+        Boolean res = TopicValidator.validateBlacklistTopic(topic, response);
```
vongosling, line 87: it still needs to polish when you change your blacklist method in validator class.

broker/src/test/java/org/apache/rocketmq/broker/topic/TopicValidatorTest.java
```diff
@@ -70,6 +64,62 @@ public void testAddSystemTopic() {
...
+        assertThat(response.getRemark()).isEqualTo("Sending message to topic[" + topic + "] is forbidden.");
+
+        topic = "test_topic";
+        res = TopicValidator.validateBlacklistTopic(topic, response);
+        assertThat(res).isTrue();
+    }
+
+    @Test
+    public void testIsSystemTopic() {
+        String topic = TopicValidator.RMQ_SYS_SCHEDULE_TOPIC;
```
vongosling, line 99: true ?

broker/src/test/java/org/apache/rocketmq/broker/topic/TopicValidatorTest.java
```diff
@@ -70,6 +64,62 @@ public void testAddSystemTopic() {
...
+        assertThat(res).isTrue();
+
+        topic = "test_topic";
+        res = TopicValidator.isSystemTopic(topic);
+        assertThat(res).isFalse();
+    }
+
+    @Test
+    public void testIsBlacklistTopic() {
+        String topic = TopicValidator.RMQ_SYS_SCHEDULE_TOPIC;
```
vongosling, line 114: where is IsBlacklistTopic in here

common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        return CLIENT_BLACKLIST_TOPIC_SET.contains(topic);
+    }
+
+    public static boolean validateBlacklistTopic(String topic, RemotingCommand response) {
+        if (isBlacklistTopic(topic)) {
+            response.setCode(ResponseCode.NO_PERMISSION);
+            response.setRemark("Sending message to topic[" + topic + "] is forbidden.");
+            return false;
+        }
+        return true;
```
RongtongJin, line 121: It would be better to reconsider the naming of these methods. The difference between `validateSystemTopic` and `isSystemTopic` are quite confusing, same for  `isBlacklistTopic` and `validateBlacklistTopic`. For example, when a topic is system topic, but validateSystemTopic return false.

broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
coder-zzzz: ok，I'll change it

PR #2084, opened 2020-06-10
[ISSUE#2085]support graceful shutdown push consumer

Jaskey:

Issue :  #2085 

For now, RocketMQ does not support a graceful shutdown for push consumer, which means that when a user shutdown a push consumer, the threadpool will shutdown but rocketmq client will not wait for the existing task to be completed, which may cause 
1. Error is found when shutdown. Since the application since application may shutdown some external source like datasource when application shutdowns, so the existing task may fail.
2. Message duplication. Even though the task can be completed, since rocketmq will update consumer offset after the shutdown method returns, so the latest consumer offset should be updated after the remaining tasks are completed

## Brief changelog

Add a parameter awaitTerminationMillisWhenShutdown(default is 0), to determine how long should the rocketmq client to await for the consumer threadpool to terminate before update consumer offset

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
@@ -45,6 +45,7 @@ import org.apache.rocketmq.common.message.MessageExt;
 import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.common.protocol.body.CMResult;
 import org.apache.rocketmq.common.protocol.body.ConsumeMessageDirectlyResult;
+import org.apache.rocketmq.common.utils.ThreadUtils;
 import org.apache.rocketmq.logging.InternalLogger;
 import org.apache.rocketmq.remoting.common.RemotingHelper;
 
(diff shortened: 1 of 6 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
```diff
@@ -92,9 +93,9 @@ public void run() {
         }, this.defaultMQPushConsumer.getConsumeTimeout(), this.defaultMQPushConsumer.getConsumeTimeout(), TimeUnit.MINUTES);
     }
 
-    public void shutdown() {
```
duhenglucky, line 95: It would be better to keep this method forward compatible.

Jaskey: fine!

Jaskey:
@dongeforever @RongtongJin would you please help to review this pr ?

Jaskey:
@dongeforever @duhenglucky one more discussion. Graceful shutdown should be a basic feature, what do you think of the default await millis? Should we give a non-zero default value rather than 0

Jaskey:
@ShannonDing which version will be this pr included ?

Issue #2102, opened 2020-06-17
send message with empty Body cause NullPointerException


wangshaojie4039:
**BUG REPORT**

1. when client receive empty body message from broker,consumeService throw NullPointerException

![image](https://user-images.githubusercontent.com/22910636/84847630-2029bf80-b084-11ea-983b-66ae88b62c6f.png)

imaffe:
Hi, can you please provide more information on how to reproduce this problem, including

- which version rocketmq-client/rocketmq you are using (because we need to locate which line caused the problem, but different version has different line numbers)
- ...

Thanks so much and we will follow up~

wangshaojie4039:
> Hi, can you please provide more information on how to reproduce this problem, including
> 
> * which version rocketmq-client/rocketmq you are using (because we need to locate which line caused the problem, but different version has different line numbers)
> * ...
> 
> Thanks so much and we will follow up~

rocketmq version is 4.5.1

zongtanghu:
I think it's necessary to add the empty Body limination logical codes for messages on broker. @ShannonDing @wangshaojie4039 @RongtongJin

coder-zzzz:
IMO,empty body should be allowed, sometimes mq just used to notify clients about sth change with no message.
I think should fix the client's problem instead of add limit on broker

Jaskey:
@wangshaojie4039  how do you produce this message to producer? Since the producer api should check if the body is null

coder-zzzz:
@Jaskey maybe he use other client instead of java client

zongtanghu:
@Jaskey  In java sdk of rocketmq, it check the message body exactly.But our members use other language rocektmq sdk, such as rocektmq-go sdk. I also put forward a issue. https://github.com/apache/rocketmq-client-go/issues/501

Jaskey:
@zongtanghu Thx, would you please help to review the bug fix pr https://github.com/apache/rocketmq/pull/2104 ?

PR #2104, opened 2020-06-17
[ISSUE #2102] Fix NPE when receive empty body message.

Jaskey:

ISSUE #2102

## Brief changelog

add null check for message body when calculate message size.


zongtanghu:
You just fix this issue in rocketmq-client.But this need users to update rocketmq-client.

Jaskey:
@zongtanghu better than nothing.

RongtongJin:
Plz name the pull request in the correct format next time.

PR #2155, opened 2020-07-13
[ISSUE #XXX] Add more tip for acl

zhangjin-007:
When broker does not configure acl=true,the following error will appear
image
After adding hints，the following error will appear
image
The broker's log will also report the correct error
image


RongtongJin:
I will close the Issue first, but please feel free to reopen it if you have any other issues.

PR #2185, opened 2020-07-22
[ISSUE #1416] Consumer group offset auto delete

guixincui:
auto delete consumer group offset  when delete subscriptionGroup  or topic


wz2cool:
any update about this issue ?

Issue #2194, opened 2020-07-23
the judgement logic worng in  org.apache.rocketmq.common.MixAll#string2FileNotSafe


horizonzy:
**BUG REPORT**

1. Please describe the issue you observed:
the judgement logic of  fileParent.mkdirs() is wrong.
in org.apache.rocketmq.common.MixAll#string2FileNotSafe.
the code is: 
```
 if (fileParent != null) {
            fileParent.mkdirs();
     }
```

maybe it should change to 
```
  if (fileParent == null) {
            fileParent.mkdirs();
        }
```

PR #2195, opened 2020-07-23
[ISSUE #2194] Change the judgement logic of file creation

horizonzy:
fix [ISSUE #2194]


RongtongJin:
> fileParent is null, call fileParent.mkdirs() will occur NPE. Is the logic right?

Yes, you are right. The original logic is correct. I will close the PR and the related issue.

horizonzy:
> fileParent is null, call fileParent.mkdirs() will occur NPE. Is the logic right?

ye.it's my mistake.sry,

Issue #2241, opened 2020-08-12
[RocketMQ broker hook refactor] add AbortProcessException to support flow control


imaffe:
**FEATURE REQUEST**

1. Currently the broker hooks resides in mqtrace package, which doesn't describe hook's general usage. We need to refactor them to a different subpackage
2. We need to make hooks be able to interrupt the sending/consume process, by return early with error messages.

zongtanghu:
Any progress ? @imaffe

PR #2243, opened 2020-08-12
[ISSUE #2241] Reinforcement hook implementation to support abort processing 

imaffe:
FEATURE REQUEST

Currently the broker hooks resides in mqtrace package, which doesn't describe hook's general usage. We need to refactor them to a different subpackage
We need to make hooks be able to interrupt the sending/consume process, by return early with error messages.


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -628,6 +628,8 @@ private SendResult sendDefaultImpl(
                             case ResponseCode.NO_BUYER_ID:
                             case ResponseCode.NOT_IN_CURRENT_UNIT:
                                 continue;
+                            case ResponseCode.FLOW_CONTROL:
+                                break;
```
RongtongJin, line 632: This break can only jump out of the switch but not out of the for loop

broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
```diff
@@ -273,8 +274,13 @@ public void executeSendMessageHookBefore(final ChannelHandlerContext ctx, final
...
                     }
-                } catch (Throwable e) {
-                    // Ignore
+                }
+                catch (AbortProcessException e) {
+                    throw e;
+                } catch (RemotingCommandException e) {
+                    // ignore
+                } catch (RuntimeException e) {
+                    // ignore
```
RongtongJin, line 283: Is it necessary to print the log ？

RongtongJin:
Plz pass the CI

PR #2249, opened 2020-08-13
[ISSUE# 2248] Config client worker threads

guixincui:

Config client worker threads

## Brief changelog

Config client worker threads

## Verifying this change

Config client worker threads


imaffe:
We want to distinguish between consumer/producer configs and netty client configs as mentioned in the related issue, would you be interested in refactor your commit ? Thanks ~

vongosling:
Do you consider any protection logic for some evil input?

guixincui:
> Do you consider any protection logic for some evil input?

how about add input argument range check ?

Issue #2289, opened 2020-09-04
[New Feature]Flexible timer and delay messages


Aaron-He:
At present, RocketMQ only supports delay messages with fixed delay levels(1s 5s 10s 30s 1m 2m 3m 4m 5m 6m 7m 8m 9m 10m 20m 30m 1h 2h), which is not flexible enough. Current delay message process diagram:

![开源RMQ延时消息实现](https://user-images.githubusercontent.com/21126659/92218363-94c60a80-eecb-11ea-8059-b16a23af33a6.png)
The current delay message design is realized by using new separate Topic(SCHEDULE_TOPIC_XXXX) and Queue, if we want to achieve any time delay  of seconds within 40 days, based on the current solution, 40 * 24 * 60 * 60 = 345600 Queues are needed, and the cost is very high. Also, the message cannot be delayed too long because the CommitLog will be deleted after a fixed amount of hours (default 48 hours).

In order to meet the requirements of flexible time delay messages, we can redesign the architecture of delay messages(different from the current delay message, the delay message here  refers to the message delivered at a specific time or delayed a period time) and introduce [Time Wheel](http://www.cs.columbia.edu/~nahum/w6998/papers/sosp87-timing-wheels.pdf).

vongosling:
https://github.com/apache/rocketmq/pull/2290

PR #2290, opened 2020-09-07
[ISSUE #2289] Serverless schedule support builtin core

Aaron-He:

In order to meet the requirements of flexible time delay messages

## Brief changelog

redesign the architecture of delay messages(different from the current delay message, the delay message here refers to the message delivered at a specific time or delayed a period time) and introduce Time Wheel.
自定义定时延时消息实现

## Verifying this change

Yes

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -41,6 +41,8 @@ import org.apache.rocketmq.broker.client.DefaultConsumerIdsChangeListener;
 import org.apache.rocketmq.broker.client.ProducerManager;
 import org.apache.rocketmq.broker.client.net.Broker2Client;
 import org.apache.rocketmq.broker.client.rebalance.RebalanceLockManager;
+import org.apache.rocketmq.broker.delaymsg.service.LocalReplayDelayMsgService;
+import org.apache.rocketmq.broker.delaymsg.timewheel.TimeWheelManager;
 import org.apache.rocketmq.broker.dledger.DLedgerRoleChangeHandler;
 import org.apache.rocketmq.broker.filter.CommitLogDispatcherCalcBitMap;
 import org.apache.rocketmq.broker.filter.ConsumerFilterManager;
@@ -103,6 +105,7 @@ import org.apache.rocketmq.store.MessageArrivingListener;
 import org.apache.rocketmq.store.MessageStore;
 import org.apache.rocketmq.store.config.BrokerRole;
 import org.apache.rocketmq.store.config.MessageStoreConfig;
+import org.apache.rocketmq.store.delaymsg.AbstractReplayDelayMsgService;
 import org.apache.rocketmq.store.dledger.DLedgerCommitLog;
 import org.apache.rocketmq.store.stats.BrokerStats;
 import org.apache.rocketmq.store.stats.BrokerStatsManager;
@@ -167,6 +170,7 @@ public class BrokerController {
     private AbstractTransactionalMessageCheckListener transactionalMessageCheckListener;
     private Future<?> slaveSyncFuture;
     private Map<Class,AccessValidator> accessValidatorMap = new HashMap<Class, AccessValidator>();
+    private TimeWheelManager timeWheelManager;
 
     public BrokerController(
         final BrokerConfig brokerConfig,
@@ -482,8 +486,23 @@ public class BrokerController {
             initialTransaction();
             initialAcl();
             initialRpcHooks();
+            initialDelayMsg();
         }
         return result;
+    }
+
+    private void initialDelayMsg() {
+        if (messageStoreConfig.getBrokerRole() != BrokerRole.SLAVE) {
+            // only support MessageStore implementation is DefaultMessageStore currently
+            if (messageStore instanceof DefaultMessageStore) {
+                AbstractReplayDelayMsgService replayDelayMsgService = ServiceProvider.loadClass(ServiceProvider.REPLAYDELAYMSG_SERVICE_ID, AbstractReplayDelayMsgService.class);
+                if (replayDelayMsgService == null) {
+                    replayDelayMsgService = new LocalReplayDelayMsgService(messageStore, ((DefaultMessageStore) messageStore).getDelayMsgCheckPoint());
+                    log.warn("Load default Replay DelayMsg Service: {}", LocalReplayDelayMsgService.class.getSimpleName());
+                }
+                this.timeWheelManager = new TimeWheelManager(replayDelayMsgService, ((DefaultMessageStore) messageStore).getDelayMsgIndexService(), ((DefaultMessageStore) messageStore).getDelayMsgCheckPoint());
+            }
+        }
     }
 
     private void initialTransaction() {
@@ -742,6 +761,10 @@ public class BrokerController {
     }
 
     public void shutdown() {
+        if (this.timeWheelManager != null) {
+            this.timeWheelManager.shutdown();
+        }
+
         if (this.brokerStatsManager != null) {
             this.brokerStatsManager.shutdown();
         }
@@ -907,6 +930,9 @@ public class BrokerController {
             this.brokerFastFailure.start();
         }
 
+        if (this.timeWheelManager != null) {
+            this.timeWheelManager.start();
+        }
 
     }
 
--- a/broker/src/main/java/org/apache/rocketmq/broker/delaymsg/service/LocalReplayDelayMsgService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/delaymsg/service/LocalReplayDelayMsgService.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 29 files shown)


zongtanghu:
Please help to review this pr, thanks. @RongtongJin @duhengforever @vongosling

vongosling:
Great, I would like to follow it.

zongtanghu:
Can you write some test cases and their results which you have already tested in this pr's link? @Aaron-He

shudo1219:
The approach is obvious and vivid,  here are some suggestions(maybe I don't have a thorough graspt of code):
1) Besides DelayMsgCheckPoint file is persisted on master node, the checkpoint should also be synchronized to slave node periodly; otherewise, if master node crashes down and fails to restart, slave node does not know the timestamp the message can be delivered from;

2) In method putKey() of file DelayMsgIndexFile.java, if the method is used in multi threads, we will need to lock before operating the index file; otherwise, we might not need to release the file lock since we never use it;

3) For the message delivered, we'd better keep its physical commitlog offset to a log let's call it deliveryLog,
if master node restarts, we need to find out all the the undelivered messages between deliveryLog and index file, it can try to reduce the number of duplicated messages on the server side.
Whether or not to sync the deliveryLog to slave node? Maybe we need to discuss it later.

4)  If a large mount of delay messages fall into lots of commitlog files randomly,  the speed of delivering delay messages will be slow because of random reading. From our internal tests, the speed is about 200-300 tps on normal disk(15K Rounds Per Minute). Shall we need to limit produce rate?

Let's work together to build open source community better.

shudo1219:
I found another problem needed to discuss:
if timewheel deliveries the delay messages  a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever. 

Sometimes it is useless to deliver the delay message much later than excepted,  maybe we need to rebalance how late the message we can remove directly.

Aaron-He:
> The approach is obvious and vivid, here are some suggestions(maybe I don't have a thorough graspt of code):
> 
> 1. Besides DelayMsgCheckPoint file is persisted on master node, the checkpoint should also be synchronized to slave node periodly; otherewise, if master node crashes down and fails to restart, slave node does not know the timestamp the message can be delivered from;
> 2. In method putKey() of file DelayMsgIndexFile.java, if the method is used in multi threads, we will need to lock before operating the index file; otherwise, we might not need to release the file lock since we never use it;
> 3. For the message delivered, we'd better keep its physical commitlog offset to a log let's call it deliveryLog,
>    if master node restarts, we need to find out all the the undelivered messages between deliveryLog and index file, it can try to reduce the number of duplicated messages on the server side.
>    Whether or not to sync the deliveryLog to slave node? Maybe we need to discuss it later.
> 4. If a large mount of delay messages fall into lots of commitlog files randomly,  the speed of delivering delay messages will be slow because of random reading. From our internal tests, the speed is about 200-300 tps on normal disk(15K Rounds Per Minute). Shall we need to limit produce rate?
> 
> Let's work together to build open source community better.

1. Good suggestion
2. putKey() is used by reputService, which is single thread. yeah, we can remove the release code block.
3. Good suggestion. As current implementation, user identifies duplicate messages. 
4. It may not work to limit produce rate, because the delayMsgs delivery is up to its startDeliverTime. We can use more memory to make performance better. There may be other better solutions.

Let's work together to build open source community better. ^_^

Aaron-He:
> I found another problem needed to discuss:
> if timewheel deliveries the delay messages a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever.
> 
> Sometimes it is useless to deliver the delay message much later than excepted, maybe we need to rebalance how late the message we can remove directly.

We can separate the commitlog of delayMsg from the normal messages commitlog in the next version.  we can work together.

vongosling:
Any process for this pr?

zongtanghu:
@vongosling We have already tested by ourselves, and deploy in my production enviroments, but no one review this codes in commnunity.

vongosling:
There are many versions in the community. we could figure out how to integrate them. I would like to help your process.

shudo1219:
> > I found another problem needed to discuss:
> > if timewheel deliveries the delay messages a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever.
> > Sometimes it is useless to deliver the delay message much later than excepted, maybe we need to rebalance how late the message we can remove directly.
> 
> We can separate the commitlog of delayMsg from the normal messages commitlog in the next version. we can work together.

Looking forward to working together.
In our internal test cases, if there are a large number of delay messages located in one second, it will cause io utils much higher than usual and take a long time to finish deliver these messages. We'd better suggest max delay messages per second and monitor them in backgroud manage system.

Aaron-He:
> > > I found another problem needed to discuss:
> > > if timewheel deliveries the delay messages a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever.
> > > Sometimes it is useless to deliver the delay message much later than excepted, maybe we need to rebalance how late the message we can remove directly.
> > 
> > 
> > We can separate the commitlog of delayMsg from the normal messages commitlog in the next version. we can work together.
> 
> Looking forward to working together.
> In our internal test cases, if there are a large number of delay messages located in one second, it will cause io utils much higher than usual and take a long time to finish deliver these messages. We'd better suggest max delay messages per second and monitor them in backgroud manage system.

Good suggestion. Setting the max number of delay messages per second ensures that the broker can provide normal messaging services. Also, the user and topic should be considered to ensure fair delivery.

vongosling:
https://github.com/airbnb/dynein, These guys have open-source its delay service on top of the SQS. That is the real pattern I've always think about. Messaging is only used to improve the scalability of queues, scheduling, and concrete workers as an external component. I'd like to see what you're thinking about. There are many restrictions on the current implementation,

cserwen:
This way of implementation is intuitive and can achieve this feature without relying on other components. But as @shudo1219  said, it has some limitations. I also have another problem. If there is someone who want to support messages with a large time span, he must extend the save time of commitLog because the delayed messages are saved together with the normal messages. But this means that we will save normal messages for a long time. It seems to be useless to do this.

cserwen:
@vongosling 
Based on this implementation, we have a better optimization plan. We can use [RocksDB](https://github.com/facebook/rocksdb) to save delayed messages. 

delay-msg


**This has the following advantages:**
- Delay messages can be stored separately from normal messages
- Save the message with the delivery time as the key, which can serve as an index
- When the message is delivered, the commitLog will not be read randomly
- It is also necessary to record the messages that have been delivered to avoid the problem of repeated message delivery caused by master-slave switching

**Of course it has the following disadvantages:**
- RocksDB will be added to RocketMQ, resulting in an increase in RocketMQ dependency

maixiaohai:
Any process？

Jason918:
> We can use [RocksDB](https://github.com/facebook/rocksdb) to save delayed messages.

@cserwen we have done this in https://github.com/didi/DDMQ/tree/master/carrera-chronos.
One of the problems in this way is that Rocksdb needs heavy maintenance when instance get large. Compaction will consume quite a mount of IO and interfere with message writing. So we have to make it an external service, and causing more maintenance work.

Git-Yang:
> > I found another problem needed to discuss:
> > if timewheel deliveries the delay messages a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever.
> > Sometimes it is useless to deliver the delay message much later than excepted, maybe we need to rebalance how late the message we can remove directly.
> 
> We can separate the commitlog of delayMsg from the normal messages commitlog in the next version. we can work together.

Is the separate storage of delayed messages and normal messages realized now?

Aaron-He:
> > > I found another problem needed to discuss:
> > > if timewheel deliveries the delay messages a little later than excepted, assuming now is 9:00 but the timewheel is deleverying the message at 8:30, meanwhile, the system begins to remove the commitlog, the message between 8:30 and 9:00 in the commitlog(going to be removed soon) might be lost forever.
> > > Sometimes it is useless to deliver the delay message much later than excepted, maybe we need to rebalance how late the message we can remove directly.
> > 
> > 
> > We can separate the commitlog of delayMsg from the normal messages commitlog in the next version. we can work together.
> 
> Is the separate storage of delayed messages and normal messages realized now?


 A new architectural implementation (still based on time wheels) will be released in the future, making timing delay message configuration more flexible, more efficient to send and receive, and less impact on “normal messages”.

Jeff-Lv (approved):
nice

rushsky518:
If 2 different schedule-times have the same hash value, there may be a problem in `org.apache.rocketmq.store.index.DelayMsgIndexFile#selectPhyOffset` method

rushsky518:
It will be better to partition DelayMsgIndex files by time range

PR #2343, opened 2020-10-09
[ISSUE# 2330] Store the properties of MessageBatch

AVP42:

store the properties of MessageBatch

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
@@ -577,6 +577,7 @@ public class SendMessageProcessor extends AbstractSendMessageProcessor implement
         messageExtBatch.setFlag(requestHeader.getFlag());
         MessageAccessor.setProperties(messageExtBatch, MessageDecoder.string2messageProperties(requestHeader.getProperties()));
         messageExtBatch.setBody(request.getBody());
+        messageExtBatch.setPropertiesString(requestHeader.getProperties());
         messageExtBatch.setBornTimestamp(requestHeader.getBornTimestamp());
         messageExtBatch.setBornHost(ctx.channel().remoteAddress());
         messageExtBatch.setStoreHost(this.getStoreHost());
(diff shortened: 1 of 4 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
```diff
@@ -577,6 +577,7 @@ private RemotingCommand handlePutMessageResult(PutMessageResult putMessageResult
         messageExtBatch.setFlag(requestHeader.getFlag());
         MessageAccessor.setProperties(messageExtBatch, MessageDecoder.string2messageProperties(requestHeader.getProperties()));
         messageExtBatch.setBody(request.getBody());
+        messageExtBatch.setPropertiesString(requestHeader.getProperties());
```
RongtongJin, line 580: IMO, set properties has been done in this code `MessageAccessor.setProperties(messageExtBatch, MessageDecoder.string2messageProperties(requestHeader.getProperties()));`, so we don't need add a new variable named propertiesString to MessageExtBatch to do this thing.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1871,9 +1878,13 @@ public ByteBuffer encode(final MessageExtBatch messageExtBatch) {
...
                 // 17 PROPERTIES
-                this.msgBatchMemory.putShort(propertiesLen);
-                if (propertiesLen > 0)
+                this.msgBatchMemory.putShort((short) (propertiesLen + batchPropLen));
+                if (propertiesLen > 0) {
                     this.msgBatchMemory.put(messagesByteBuff.array(), propertiesPos, propertiesLen);
+                }
+                if (batchPropLen > 0) {
+                    this.msgBatchMemory.put(batchPropData, 0, batchPropLen);
+                }
```
RongtongJin, line 1887: Just put properties of MessageExtBatch into msgBatchMemory. The original code takes the properties of a single message, so it does not exist.

broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
AVP42: Yes, I agree with you. My Intention was to avoid another transformation since there was already a transformed string.  I thought that MessageExtBrokerInner is out of the same consideration.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
AVP42: Sure.

AVP42:
I don't think the travis error is caused by these commits. Perhaps need to rebuild?

RongtongJin:
> I don't think the travis error is caused by these commits. Perhaps need to rebuild?

The code is OK, I rebuild the CI.

PR #2370, opened 2020-10-21
[ISSUE #XXX] Set benchmark isSuffixEnable variable default false

caorong:
fix benchmark isSuffixEnable default false

help is different with actual behavior


RongtongJin:
Plz create an issue first then link and rename this pull request.

PR #2562, opened 2021-01-06
Using utils classes makes the code more readable

jessonzh:

Using utils classes makes the code more readable

## Brief changelog

Using commons-collections CollectionUtils method, makes the code more readable

## Verifying this change

Using commons-collections CollectionUtils method, makes the code more readable


RongtongJin:
IMO, this PR is not very helpful, I will close it first

PR #2566, opened 2021-01-07
Update RocketMQ_Example.md

xiguaxiansheng:
Make examples more reasonable.


xiguaxiansheng:
Make examples more reasonable.

xiguaxiansheng:
Make the examples more reasonable and avoid readers' doubts

PR #2568, opened 2021-01-07
Update RocketMQ_Example.md

xiguaxiansheng:
Make the examples more reasonable and avoid readers' doubts


RongtongJin:
IMO, this PR is not very helpful, I will close it first

Issue #2571, opened 2021-01-08
The judgement in org.apache.rocketmq.store.ConsumeQueue#truncateDirtyLogicFiles maybe exist problem.


horizonzy:
**BUG REPORT**

1. Please describe the issue you observed:
the method `truncateDirtyLogicFiles` is to delete the consumeQueue data which ahead of commit log.
 
The judgemnt which check consume data is more than commitlog exist as follow:
https://github.com/apache/rocketmq/blob/39bb9386f10d5d8dfe81183c172a3a86f6d313bd/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java#L263

In fact, use `if (offset + size > phyOffet)` will be better.

PR #2573, opened 2021-01-08
[ISSUE #2571] Fix consumeQueue's maxPhysicOffset may greater than commitLog's phyOffset problem.

horizonzy:

For-#2571.

Fix consumeQueue's maxPhysicOffset may greater than commitLog's phyOffset problem.


RongtongJin (approved):

RongtongJin:
I think this is indeed better, what do you think?  @zongtanghu @duhenglucky

Issue #2842, opened 2021-04-28
HAService#dispatchReadRequest optimization point when processing the body


BurningCN:
We should deal with this logic at bodySize>0

![image](https://user-images.githubusercontent.com/43363120/116338182-f0078d00-a80d-11eb-822c-decf28aba3f6.png)

PR #2843, opened 2021-04-28
[ISSUE #2842] Optimize the code that HAClient handles the Bod

BurningCN:

Optimize the code that HAClient handles the Body

for issue #2842

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ha/HAService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ha/HAService.java
@@ -452,11 +452,13 @@ public class HAService {
                     }
 
                     if (diff >= (msgHeaderSize + bodySize)) {
-                        byte[] bodyData = new byte[bodySize];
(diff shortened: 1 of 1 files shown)


store/src/main/java/org/apache/rocketmq/store/ha/HAService.java
```diff
@@ -452,11 +452,13 @@ private boolean dispatchReadRequest() {
                     }
 
                     if (diff >= (msgHeaderSize + bodySize)) {
-                        byte[] bodyData = new byte[bodySize];
-                        this.byteBufferRead.position(this.dispatchPosition + msgHeaderSize);
-                        this.byteBufferRead.get(bodyData);
+                        if (bodySize > 0) {
+                            byte[] bodyData = new byte[bodySize];
+                            this.byteBufferRead.position(this.dispatchPosition + msgHeaderSize);
```
dongeforever, line 457: Here will cause an unnecessary copy of the data.

If boySize == 0, just ignore the appendToCommitlog.

Issue #2853, opened 2021-04-30
Remove code duplication in CommitLog putMessage ayncPutMessage methods


andsel:
The methods `asycPutMessage` and `putMessage` in class `CommitLog` shares a lot of code:

https://github.com/apache/rocketmq/blob/b4240d5cea8d001c21b9c0d73f5aa700fcd0d568/store/src/main/java/org/apache/rocketmq/store/CommitLog.java#L559-L662

https://github.com/apache/rocketmq/blob/b4240d5cea8d001c21b9c0d73f5aa700fcd0d568/store/src/main/java/org/apache/rocketmq/store/CommitLog.java#L788-L902

removing duplication simplify the code base.

duhenglucky:
@andsel would you like to submit a PR to polish it?

andsel:
@duhenglucky the PR #2854 solves this, missed to link it here

PR #2854, opened 2021-04-30
[Issue #2853] Refactoring remove duplicated code in CommitLog putMessage/asyncPutMessage

andsel:
This PR move common code used in two methods under the same path

## Brief changelog
- Introduce some exceptions for `org.apache.rocketmq.store.PutMessageStatus` to avoid the switch-return pattern moving to switch-throw
- move common code under the new method `appendPutMessageAndTrackStats`

## Verifying this change
`mvn verify`

- [ ] ~~Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).~~


andsel:
Hi @duhenglucky do you think this PR is missing something or is it good enough?

andsel:
@vongosling do you think this PR is good or should I fix something in it? Is it blocked by missing CLA (I think this is a small code change at the end)

vongosling:
@andsel you really reach the core codes in storage engine. Do you have a better way to verify that such refactorings are trustworthy, such as some unit tests, integration tests? I'm just thinking about optimizing this, It'll take some time. Please be patient. Also, you're welcome to participate :-)

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -785,41 +786,24 @@ public long getBeginTimeInLock() {
...
+        handleHA(result, putMessageResult, msg);
 
-                msg.setTopic(topic);
-                msg.setQueueId(queueId);
-            }
-        }
+        return putMessageResult;
+    }
 
+    private void setIPV6Flags(MessageExtBrokerInner msg) {
```
yuz10, line 806: setIPV6Flags can also extract lines in putMessages, because MessageExtBatch and MessageExtBrokerInner both extends  MessageExt

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -623,27 +646,22 @@ public long getBeginTimeInLock() {
...
+                        log.error("create mapped file error, topic: " + msg.getTopic() + " clientAddr: " + msg.getBornHostString());
+                        throw new Exceptions.CreateMappedfileFailedException(result);
                     }
                     result = mappedFile.appendMessage(msg, this.appendMessageCallback);
                     break;
                 case MESSAGE_SIZE_EXCEEDED:
                 case PROPERTIES_SIZE_EXCEEDED:
-                    beginTimeInLock = 0;
-                    return CompletableFuture.completedFuture(new PutMessageResult(PutMessageStatus.MESSAGE_ILLEGAL, result));
+                    throw new Exceptions.MessageIllegalException(result);
```
yuz10, line 656: Is it possible that throw an Exception will affect performance?

andsel: It requires to create an extra object (the exception itself) and bring the stack trace of current thread, I've not any measure about performance loss. Do you think this is critical?

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
andsel: DO you think is good to move the method `setIPV6Flags` into `MessageExt` class?

andsel:
Hi @vongosling thank you a lot for  your reply and sorry for my pushing, I've got the fear this was out of radar.
I haven't done any performance test, I think a JMH test for `asyncPutMessage` and `putMessage` could be help. I'll try to put it down something in this pr

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
andsel: from https://www.baeldung.com/java-exceptions-performance throwing exception is more costly than returning an error, but how many times do you expect the exception is raised? 
The condition under which is raised are:
- message size too big
- the properties are too big
- mapped file doesn't exists.

Do you think they happen so often to generate a performance problem in throwing exception instead of returning error code?

andsel: Hi, I moved from throwing exceptions to returning a couple with error or result

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
yuz10: I dont think so, I just mean that another place could also use the function instead of writing the same code.

yuz10 (approved):

andsel:
HI @vongosling @yuz10  this PR was approved a couple of months ago, now it's conflicting, should resolve conflicts to get it merged or the changes this PR is proposing has already been merged with other PRs?

yuz10:
@andsel maybe its included in #2889

andsel:
Maybe @areyouok knows is #2889 contains all this changes?

areyouok:
> Maybe @areyouok knows is #2889 contains all this changes?

I remove some duplicated code before my refactor in #2889. And it is include in the 4.9.1 version.

There is still many code are duplicated or unused.

andsel:
Branch is older then one year and conflict with `main`. Parts of this are already been integrated in some other PRs

Issue #2898, opened 2021-05-17
Resolve rocketmq-example project failed during checkstyle execution.


zhangjidi2016:
When packaging RocketMQ， the OpenTracingTransactionProducer class in Example does not pass CheckStyle.

![image](https://user-images.githubusercontent.com/18254437/118445797-79312600-b721-11eb-8af3-5fb84580edf9.png)

PR #2899, opened 2021-05-17
[ISSUE #2898] Resolve rocketmq-example project failed during checkstyle execution.

zhangjidi2016:

#2898

### DIFF
--- a/example/src/main/java/org/apache/rocketmq/example/tracemessage/OpenTracingTransactionProducer.java
+++ b/example/src/main/java/org/apache/rocketmq/example/tracemessage/OpenTracingTransactionProducer.java
@@ -57,7 +57,7 @@ public class OpenTracingTransactionProducer {
 
         try {
             Message msg = new Message("TopicTest", "Tag", "KEY",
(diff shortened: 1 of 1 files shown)


francisoliverlee (changes requested):
as i build with lastest codes, all is fine. 
check your local IDE config

zhangjidi2016:
> as i build with lastest codes, all is fine.
> check your local IDE config

zhangjidi2016:
When I use the CheckStyle plugin to check the code following the checkstyle.xml specification in RocketMQ, the check does not pass
image
image

vongosling:
Yes, you are correct. Note that we have bound checkstyle to verify the phase now. If bound to the validate phase, it would check the code prior to compiling the code. If the code is invalid, the parsing errors reported by checkstyle may be different than what would be expected from the javac compiler. However, it's guaranteed to run. But for verify phase which would run much later (and allow the javac compiler to flag invalid code prior to checkstyle). However, if developers generally just use "mvn test" prior to pushing changes, checkstyle would not run as verify occurs after the test phase.

Issue #2883, opened 2021-05-13
Improve produce performance in M/S mode


areyouok:
I have some commit to Improve produce performance in M/S mode:

1. Change log level to debug: "Half offset {} has been committed/rolled back"
2. Optimise lock in WaitNotifyObject
3. Remove lock in HAService
4. Remove lock in GroupCommitService
5. Eliminate array copy in HA
6. Remove putMessage/putMessages method in CommitLog which has too many duplicated code.
7. Change default value of some parameters: sendMessageThreadPoolNums/useReentrantLockWhenPutMessage/flushCommitLogTimed/endTransactionThreadPoolNums
8. Optimise performance of asyncPutMessage (extract some code out of putMessage lock)
9. extract generation of msgId out of lock in CommitLog (now only for single message processor)
10. extract generation of topicQueueTable key out of sync code
11. extract generation of msgId out of lock in CommitLog (for batch)
12. fix ipv6 problem introduced in commit "Optimise performance of asyncPutMessage (extract some code out of putMessage lock)"
13. Remove an duplicate MessageDecoder.string2messageProperties for each message, and prevent store "WAIT=true" property (in most case) to save 9 bytes for each message.
14. Improve performance of string2messageProperties/messageProperties2String, and save 1 byte for each message.
15. Optimise parse performance for SendMessageRequestHeaderV2

duhenglucky:
merged

PR #2961, opened 2021-06-01
[ISSUE #2883] [Part F] Improve produce performance in M/S mode.

areyouok:
1. Change log level to debug: "Half offset {} has been committed/rolled back"
2. Optimise lock in WaitNotifyObject
3. Remove lock in HAService
4. Remove lock in GroupCommitService
5. Eliminate array copy in HA
6. Remove putMessage/putMessages method in CommitLog which has too many duplicated code.
7. Change default value of some parameters: sendMessageThreadPoolNums/useReentrantLockWhenPutMessage/flushCommitLogTimed/endTransactionThreadPoolNums
8. Optimise performance of asyncPutMessage (extract some code out of putMessage lock)
9. extract generation of msgId out of lock in CommitLog (now only for single message processor)
10. extract generation of topicQueueTable key out of sync code
11. extract generation of msgId out of lock in CommitLog (for batch)
12. fix ipv6 problem introduced in commit "Optimise performance of asyncPutMessage (extract some code out of putMessage lock)"
13. **Remove an duplicate MessageDecoder.string2messageProperties for each message, and prevent store "WAIT=true" property (in most case) to save 9 bytes for each message.**
14. **Improve performance of string2messageProperties/messageProperties2String, and save 1 byte for each message.**

```
Benchmark                               Mode  Cnt     Score   Error  Units(10000 loop in each op)
TempTest.messageProperties2String      thrpt    2  2257.276          ops/s
TempTest.messageProperties2String_old  thrpt    2  1464.342          ops/s
TempTest.string2messageProperties      thrpt    2  1590.499          ops/s
TempTest.string2messageProperties_old  thrpt    2   605.118          ops/s
```

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
@@ -409,7 +409,23 @@ public class MessageDecoder {
     }
 
     public static String messageProperties2String(Map<String, String> properties) {
-        StringBuilder sb = new StringBuilder();
+        if (properties == null) {
+            return "";
+        }
+        int len = 0;
+        for (final Map.Entry<String, String> entry : properties.entrySet()) {
+            final String name = entry.getKey();
+            final String value = entry.getValue();
+            if (value == null) {
+                continue;
+            }
+            if (name != null) {
+                len += name.length();
+            }
+            len += value.length();
+            len += 2; // separator
+        }
+        StringBuilder sb = new StringBuilder(len);
         if (properties != null) {
             for (final Map.Entry<String, String> entry : properties.entrySet()) {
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
```diff
@@ -409,7 +409,20 @@ public static MessageExt decode(
     }
 
     public static String messageProperties2String(Map<String, String> properties) {
-        StringBuilder sb = new StringBuilder();
+        int len = 0;
```
caigy, line 412: Extract a method for calculating size

caigy (approved):

duhenglucky (approved):

PR #2999, opened 2021-06-15
[ISSUE #3000] Optimize the code and fix a comment

IntroDrop:

Optimize and simplify the code

## Brief changelog

1. Use java built-in functions to simplify the code

2. Fixed wrong words in comments

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -279,12 +279,7 @@ public class DefaultMQProducerImpl implements MQProducerInner {
 
     @Override
     public Set<String> getPublishTopicList() {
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -750,7 +745,7 @@ private SendResult sendKernelImpl(final Message msg,
                 }
 
                 final String tranMsg = msg.getProperty(MessageConst.PROPERTY_TRANSACTION_PREPARED);
-                if (tranMsg != null && Boolean.parseBoolean(tranMsg)) {
+                if (Boolean.parseBoolean(tranMsg)) {
```
caigy, line 748: `tranMsg != null` can be viewed as double check here although `Boolean.parseBoolean` will check it, so it seems that removing this null check won't gain much profit.

PR #3106, opened 2021-06-29
[ISSUE #3105] [ISSUE #3108]fix the method of getting 'now' timestamp and fix the wrong word in comment.

AshleyLiuuuuu:

[ISSUES #3105]
[ISSUES #3108]

 1. When the set time is equal to string now, the offset of rollback should be consistent with the current maximum consumption offset.
 
 2. So if the time set by user is now, the timestamp should be directly set to -1 to call the method getMaxOffsetInQueue() in the Broker2Clitent.class as shown below,and if when the timestamp is "now" with calling getOffsetInQueueByTime(), because of the algorithm of binary search in method, the returned offset will not equal to the maximum offset and will only returns the offset of the second-to-last message that is not up to expectations.
 
3.Fixed wrong words in comment.[ISSUES #3108]
image

## Brief changelog
1. Added comment content of the reason why the timestamp is set to -1 when it's equal to string "now".

2. When the set time is equal to "now", in order to call method getMaxOffsetInQueue(), it will assigned a value of -1.
## Verifying this change

When the timestamp is set to now, the offset returned by the console will be less than the actual maximum offset that is not up to expectations.

As shown below:
image

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
@@ -75,7 +75,8 @@ public class ResetOffsetByTimeCommand implements SubCommand {
             String group = commandLine.getOptionValue("g").trim();
             String topic = commandLine.getOptionValue("t").trim();
             String timeStampStr = commandLine.getOptionValue("s").trim();
-            long timestamp = timeStampStr.equals("now") ? System.currentTimeMillis() : 0;
+            //when the param "timestamp" is set to now,it should return the max offset of this queue
+            long timestamp = timeStampStr.equals("now") ? -1 : 0;
 
             try {
                 if (timestamp == 0) {
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
@@ -33,6 +33,8 @@ import org.apache.rocketmq.tools.command.SubCommandException;
 
 public class ResetOffsetByTimeCommand implements SubCommand {
 
+    private static int TIMESTAMP_BY_NOW = -1;
+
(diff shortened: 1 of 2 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
```diff
@@ -75,7 +75,8 @@ public void execute(CommandLine commandLine, Options options, RPCHook rpcHook) t
             String group = commandLine.getOptionValue("g").trim();
             String topic = commandLine.getOptionValue("t").trim();
             String timeStampStr = commandLine.getOptionValue("s").trim();
-            long timestamp = timeStampStr.equals("now") ? System.currentTimeMillis() : 0;
+            //when the param "timestamp" is set to now,it should return the max offset of this queue
+            long timestamp = timeStampStr.equals("now") ? -1 : 0;
```
odbozhou, line 79: How to change to -1 is still in line with the semantics of now？

AshleyLiuuuuu: We can also set a static variable "TIMESTAMP_BY_NOW" with a value of -1 to indicate more clearly that the timestamp entered by the user is "now", and you can also see that if the timestamp is not "now" in the code, the initial value is 0, but it will be assigned to the specific timestamp entered by the user afterwards, which means that the user does not need to pay attention to the details that will be encapsulated. So in terms of business,as long as the timestamp entered is "now", the process of getting the current maximum timestamp will be fine.Of course ,add some comments will be relatively more friendly :)

ltamber:
IMO, the option `now` is ambiguous, it may be the current `operator time` in the operator machine or the current `broker time`, should we modify this option by an more clear word or just remove this option ?

AshleyLiuuuuu:
> IMO, the option `now` is ambiguous, it may be the current `operator time` in the operator machine or the current `broker time`, should we modify this option by an more clear word or just remove this option ?

#2661 you can see this issue

PR #3134, opened 2021-07-07
[ISSUE 2883] [Part G] Optimise parse performance for SendMessageRequestHeaderV2

areyouok:
Optimise parse performance for SendMessageRequestHeaderV2


caigy (changes requested):
The new implementation indeed improves performance, but may introduce the risk of missing converting new fields in the future, which should be prevented.
Please add unit tests to assure that the new implementation can produce the same result as reflection does.

caigy (approved):

duhenglucky (approved):

Issue #3136, opened 2021-07-10
TraceDataEncoder add SubAfter trace bean timestamp


StyleTang:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

**Add SubAfter trace bean timestamp field.**

2. Provide any additional detail on your proposed use case for this feature.

**org.apache.rocketmq.client.trace.TraceDataEncoder#encoderFromContextBean add SubAfter trace bean timestamp**

3. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

**should-have.** 
**Mentioned in [PR-744](https://github.com/apache/rocketmq-externals/pull/744#discussion_r667088448) , the timeStamp and consumeGroup have been deleted by this commit.**

**(It is fine for deleting SubAfter consumeGroup because we can find consume group from SubBefore traceContext by messageId and requestId)**

**I use subBefore timestamp+costTime instead of subAfter timestamp, but it is not reasonable.**

4. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:

**None**

PR #3137, opened 2021-07-10
[ISSUE #3136] TraceDataEncoder add SubAfter trace bean timestamp

StyleTang:

TraceDataEncoder add SubAfter trace bean timestamp
Mentioned in [PR-744](https://github.com/apache/rocketmq-externals/pull/744#discussion_r667088448) , the timeStamp and consumeGroup have been deleted by this commit.

(It is fine for deleting SubAfter consumeGroup because we can find consume group from SubBefore traceContext by messageId and requestId)

I use subBefore timestamp+costTime instead of subAfter timestamp, but it is not reasonable.

## Brief changelog

TraceDataEncoder add SubAfter trace bean timestamp

## Verifying this change

Unit Tests succeed


vongosling (approved):

client/src/main/java/org/apache/rocketmq/client/trace/TraceDataEncoder.java
```diff
@@ -104,9 +104,8 @@
                     subAfterContext.setContextCode(Integer.parseInt(line[6]));
                 }
                 // compatible with the old version
-                if (line.length >= 9) {
+                if (line.length >= 8) {
                     subAfterContext.setTimeStamp(Long.parseLong(line[7]));
-                    subAfterContext.setGroupName(line[8]);
```
yuz10, line 109: The code here is used for compatible reason, if the message is produced by old versions of producer, there will be group name.

duhenglucky (commented):
It is not recommended to modify this format in the current state. The previous version upgrade brought many compatibility issues.

client/src/main/java/org/apache/rocketmq/client/trace/TraceDataEncoder.java
StyleTang: Get it. 
I removed it because I thought this value is useless, maybe add the branch line.length >= 9 back is a better way. 
```
        // compatible with the old version
        if (line.length >= 8) {
            subAfterContext.setTimeStamp(Long.parseLong(line[7]));
        }
        if (line.length >= 9) {
            subAfterContext.setGroupName(Long.parseLong(line[8]));
        }
```

StyleTang:
@yuz10 @duhenglucky Thanks for your comment.

For subAfterContext, in the previous version, it was already defined line[7] is TimeStamp and line[8] is GroupName, it means for compatible reason in new revision we should still follow this protocol.

The subAfter GroupName is duplicated with subBeofre, so based on performance considerations(I guess), it was removed it by https://github.com/apache/rocketmq/pull/3005/files


We have several ways to make it compatible.

### Option1 ( Prefer )

Just follow the previous version subAfter protocol.

( line[7] is TimeStamp and line[8] is GroupName )

##### Props

* The protocol remains consistent, easy to maintain

##### Cons

* In the future revision should add line[8] group name back, even if it is useless

### Option2

If line[8] GroupName was never used, just remove it as if it not exist.

##### Props

* Performance friendly (can remove the duplicated groupName)

##### Cons

* The protocol is changed (Not a good way)
* Compatible logic is complicated, old version line[8] is group name while new revision may be not. (The new revision content length should not be 9, if it is 9, we can’t tell what the value it is)

Do you have any suggestions for this?

yuz10:
@StyleTang I prefer just add 2 back and keep the same with 4.9.0.

vongosling:
Go ahead~ Designed well is what we have advocated. The previous disturbing is the obvious lack of this point.

vongosling:
For this problem, Who said before and after must be atomic exit?

StyleTang:
@yuz10 
Already add them back.

@vongosling 
You are right. There is a small probability before and after trace not atomic exist.
If consumeMessageBefore has already been executed and the trace message has been sent by localDispatcher, then the system crashed, subAfter trace may be lost.

org.apache.rocketmq.client.impl.consumer.ConsumeMessageConcurrentlyService.ConsumeRequest#run
```java
      if (ConsumeMessageConcurrentlyService.this.defaultMQPushConsumerImpl.hasHook()) {
          consumeMessageContext = new ConsumeMessageContext();
  ConsumeMessageConcurrentlyService.this.defaultMQPushConsumerImpl.executeHookBefore(consumeMessageContext);
      }
     try {
          status = listener.consumeMessage(Collections.unmodifiableList(msgs), context);
      } catch (Throwable e) {
      }
  
      // System crash happen 
      if (ConsumeMessageConcurrentlyService.this.defaultMQPushConsumerImpl.hasHook()) {           ConsumeMessageConcurrentlyService.this.defaultMQPushConsumerImpl.executeHookAfter(consumeMessageContext);
      }
```

yuz10 (approved):

vongosling:
@duhengforever Pls help to look at this issue, especially compatibility backword:-0

StyleTang:
I think it is necessary for us to add Timestamp and GroupName back.
The reason is that send SubBefore and SubAfter message trace is not a atomic operation, we may only get SubBefore or SubAfter trace, If SubAfter trace don't have GroupName we don't know which subscription group consumed it.

vongosling:
Yes, I would like to hear any compatible issues that resolved from the opinion of the other @duhengforever @zongtanghu

PR #3226, opened 2021-08-07
[ISSUE #3229] check running status when schedule new DeliveryMessageTimerTask.

WJL3333:
when ScheduleMessageService stoped. the timer canceled. and new schedule task will cause exception.
then show "ScheduleMessageService, messageTimeup execute error, drop it. msgExt" in log.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
@@ -260,6 +260,16 @@ public class ScheduleMessageService extends ConfigManager {
             return result;
         }
 
+        private void scheduleNextDeliverTask(long nextOffset, long countdown) {
+            if (ScheduleMessageService.this.isStarted()) {
+                ScheduleMessageService.this.timer.schedule(
+                        new DeliverDelayedMessageTimerTask(this.delayLevel, nextOffset),
+                        countdown);
+            } else {
+                log.info("ScheduleMessageService stopped. skip schedule Delivery task.");
+            }
+        }
+
         public void executeOnTimeup() {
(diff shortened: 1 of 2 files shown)


WJL3333:
@panzhi33 hello, can you take a look ?

maixiaohai:
@WJL3333 can u create a related issue

store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
```diff
@@ -260,6 +260,16 @@ private long correctDeliverTimestamp(final long now, final long deliverTimestamp
             return result;
         }
 
+        private void scheduleNextDeliverTask(long nextOffset, long countdown) {
+            if (ScheduleMessageService.this.isStarted()) {
```
panzhi33, line 264: DeliverDelayedMessageTimerTask has been judged here, I want to know what is the basis for adding judgment here。

WJL3333:
@maixiaohai thanks for reply. I create a ISSUE #3229

store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
WJL3333: thanks for reply. please check issue #3229 which provide some stack trace log.

if one thread call shutdown ScheduleMessageService. the timer will be canceled.

and at the same time when the timer running in `executeOnTimeup` method. (which may cost a long time)
the latter thread may also schedule a new DeliverDelayedMessageTimerTask to run another round.
but the current code not check the timer status. and we got an IllegalStateException here.

WJL3333:
close this pr and reopen in #3230

Issue #3229, opened 2021-08-07
ScheduleMessageService sometimes throw IllegalStateException when shutdown.


WJL3333:
**BUG REPORT**

1. Please describe the issue you observed:

- send a lot of schedule msg.
- shutdown broker process or trigger DLedger change.
- got Exception sometimes 

2021-08-07 20:18:20 WARN ScheduleMessageTimerThread - message store has shutdown, so putMessage is forbidden
2021-08-07 20:18:21 ERROR ScheduleMessageTimerThread - ScheduleMessageService, a message time up, but reput it failed, topic: SCHEDULE_TOPIC_XXXX msgId C0A81F0A00002A9F000000002E369798
2021-08-07 20:18:21 ERROR ScheduleMessageTimerThread - ScheduleMessageService, messageTimeup execute error, drop it. msgExt=MessageExt [brokerName=null, queueId=0, storeSize=262, queueOffset=657588, sysFlag=0, bornTimestamp=1628337462474, bornHost=/192.168.31.10:50221, storeTimestamp=1628337462474, storeHost=/192.168.31.10:10911, msgId=C0A81F0A00002A9F000000002E369798, commitLogOffset=775329688, bodyCRC=198614610, reconsumeTimes=0, preparedTransactionOffset=0, toString()=Message{topic='SCHEDULE_TOPIC_XXXX', flag=0, properties={REAL_TOPIC=TopicTest, KEYS=OrderID188, UNIQ_KEY=7F0000016FC818B4AAC2232EB0CADAE6, CLUSTER=DefaultCluster, WAIT=false, TAGS=TagA, REAL_QID=2}, body=[72, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100], transactionId='null'}], nextOffset=657588,offsetPy=775329688,sizePy=262
java.lang.IllegalStateException: Timer already cancelled.
	at java.util.Timer.sched(Timer.java:397) ~[na:1.8.0_301]
	at java.util.Timer.schedule(Timer.java:193) ~[na:1.8.0_301]
	at org.apache.rocketmq.store.schedule.ScheduleMessageService$DeliverDelayedMessageTimerTask.executeOnTimeup(ScheduleMessageService.java:333) [rocketmq-store-4.9.1-SNAPSHOT.jar:4.9.1-SNAPSHOT]
[...]

PR #3230, opened 2021-08-07
[ISSUE #3229] check running status when schedule new DeliveryMessageTimerTask.

WJL3333:
when ScheduleMessageService stoped. the timer canceled. and new schedule task will cause exception.
then show "ScheduleMessageService, messageTimeup execute error, drop it. msgExt" in log.


panzhi33 (approved):

ltamber:
+1

Git-Yang:
Can you resolve the conflict？

ShannonDing:
file conflicted now, it is better to resolve it first.

ShannonDing (commented):
The optimization here is very good, but the code conflict problem needs to be solved first.

PR #3263, opened 2021-08-13
The thread pool is created by ThreadPoolExecutor. This method allows students to be more clear about the running rules of the thread pool and avoid the risk of resource exhaustion.

zhaohai666:
Thread pools are not allowed to be created using Executors, but through ThreadPoolExecutor. This processing method allows students to write more clearly the running rules of the thread pool and avoid the risk of resource exhaustion.


acl/pom.xml
```diff
@@ -13,7 +13,7 @@
     <parent>
         <groupId>org.apache.rocketmq</groupId>
         <artifactId>rocketmq-all</artifactId>
-        <version>4.9.1-SNAPSHOT</version>
+        <version>4.9.2-SNAPSHOT</version>
```
ltamber, line 16: why should we bomb up project version ?

chho93:
111

PR #3273, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
Does travis_retry really solve the build issues? According to the data in paper [An empirical study of the long duration of continuous integration builds](https://dl.acm.org/doi/10.1007/s10664-019-09695-9), travis_retry can only solve 3% of the build failures. And it may cause unstable build and increase build time.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,8 +43,8 @@ before_script:
   - ulimit -c unlimited
 
 script:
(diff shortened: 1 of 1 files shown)


francisoliverlee (changes requested):
pls merge into branch develop, not master

PR #3289, opened 2021-08-25
Modify rocketmq topic routing protocol  for pulsar / Kafk is compatible with rocketmq

ronyhzhang:
pulsar or kafka‘s partitionId is gobal[0..N] , whereever topic partition be on which broker，
using the fields of topic and partitionId, we can find the partition's address;
but in RocketMQ original routing protocols, the routing info is calculated by client's logic;
where the topic partition be shift to other broker, In pulsar or Kafka system, 
use compatible protocols, it's difficult to find correct topic address

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -162,6 +162,14 @@ public class MQClientInstance {
     public static TopicPublishInfo topicRouteData2TopicPublishInfo(final String topic, final TopicRouteData route) {
         TopicPublishInfo info = new TopicPublishInfo();
         info.setTopicRouteData(route);
+        if (route.getPartitionRouteInfos() != null && !route.getPartitionRouteInfos().isEmpty()) {
+            for (Integer partition : route.getPartitionRouteInfos().keySet()) {
+                MessageQueue mq = new MessageQueue(topic, route.getPartitionRouteInfos().get(partition), partition);
+                info.getMessageQueueList().add(mq);
+            }
+            info.setOrderTopic(false);
+            return info;
+        }
         if (route.getOrderTopicConf() != null && route.getOrderTopicConf().length() > 0) {
             String[] brokers = route.getOrderTopicConf().split(";");
             for (String broker : brokers) {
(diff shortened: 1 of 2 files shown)


odbozhou:
> pulsar or kafka‘s partitionId is gobal[0..N] , whereever topic partition be on which broker，
> using the fields of topic and partitionId, we can find the partition's address;
> but in RocketMQ original routing protocols, the routing info is calculated by client's logic;
> where the topic partition be shift to other broker, In pulsar or Kafka system,
> use compatible protocols, it's difficult to find correct topic address


1. Rocketmq and other mqs are different mqs.
2. Why should rocketmq be compatible with other mq protocols?
3. Please don't change the protocol of rocketmq randomly.

odbozhou:
> > pulsar or kafka‘s partitionId is gobal[0..N] , whereever topic partition be on which broker，
> > using the fields of topic and partitionId, we can find the partition's address;
> > but in RocketMQ original routing protocols, the routing info is calculated by client's logic;
> > where the topic partition be shift to other broker, In pulsar or Kafka system,
> > use compatible protocols, it's difficult to find correct topic address
> 
> 1. Rocketmq and other mqs are different mqs.
> 2. Why should rocketmq be compatible with other mq protocols?
> 3. Please don't change the protocol of rocketmq randomly.

The protocol change is a very big improvement. At least one issue or proposal is needed to describe clearly. Direct modification is not a good way.

wolfstudy:
> The protocol change is a very big improvement. At least one issue or proposal is needed to describe clearly. Direct modification is not a good way.

Yes, agree with you. RocketMQ considers whether to abstract an interface for external exposure to users or use other more elegant ways to support the changes here. I think this is a good start. For RocketMQ, we can integrate and support more protocols

wolfstudy:
@odbozhou If we can have a reasonable abstraction of the RocketMQ protocol itself, then we can open up the interface for multi-protocol support. This will not damage the native interface protocol of Rocketmq, and it will be compatible with other protocols.

wenfengwang:
> > pulsar or kafka‘s partitionId is gobal[0..N] , whereever topic partition be on which broker，
> > using the fields of topic and partitionId, we can find the partition's address;
> > but in RocketMQ original routing protocols, the routing info is calculated by client's logic;
> > where the topic partition be shift to other broker, In pulsar or Kafka system,
> > use compatible protocols, it's difficult to find correct topic address
> 
> The connector component is the preferred solution for data communication between different message products, and the protocol benchmarking should not be modified directly.

The RocketMQ as an independent MQ product,  it don't and shouldn't provide any efforts on compatibility by other MQ product. I think use Level API is a good way to achieve this goal, the OpenMessaging spec is a good choice, https://github.com/openmessaging/openmessaging-java, https://github.com/openmessaging/openmessaging-benchmark

PR #3311, opened 2021-09-02
Fixed starup script runserver.sh runbroker.sh problems for zulu jdk and jdk version >= 9

ColinZou:

Fixed startup problems for name node  and broker for JDK > 8 and MS ZULU JDK

## Brief changelog
1. Fixed wrong classpath: added paths of jar files under ${BASE_DIR}/lib
2. Removed GC logging for ZULU JDK (TODO: figure out proper cmd arguments for zulu jdk)
3. Removed -Djava.ext.dirs=xxx for ZULU JDK and JDK > 8

## Verifying this change
Just run these commands: bin/mqnamesrv and bin/mqbroker. Or you can print JAVA_OPT before the last line "$JAVA ${JAVA_OPT} $@" for more investigation.

### DIFF
--- a/distribution/bin/runbroker.sh
+++ b/distribution/bin/runbroker.sh
@@ -33,6 +33,13 @@ export JAVA="$JAVA_HOME/bin/java"
 export BASE_DIR=$(dirname $0)/..
 export CLASSPATH=.:${BASE_DIR}/conf:${CLASSPATH}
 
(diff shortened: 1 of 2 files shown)


Jason918:
Since there is no issue related to this pr, I think it's better you can describe more details about "startup problem" with detail logs or behaviors.

PR #3354, opened 2021-09-14
[ISSUE #3355] Implement DefaultMQAdminExt::examineSubscriptionGroupConfig method

haozhijie9527:

implement examineSubscriptionGroupConfig function on the client, avoid problems that may occur when the broker and client versions are inconsistent


haozhijie9527:
#3355

areyouok (approved):

francisoliverlee (changes requested):
there would be not that OK reading ALL consumer group configs but for one.

haozhijie9527:
> there would be not that OK reading ALL consumer group configs but for one.

All right, I will close this pr, and I think pr #3039 has same problem

PR #3365, opened 2021-09-22
[ISSUE #3366] Allocate MessageQueue to client first by same Machine IP.

ni-ze:
add a new allocate MessageQueue strategy: Allocated by ip, if messageQueue localed in broker has same ip with client, Allocate messageQueue to this client firstly.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
@@ -220,13 +220,13 @@ public class QueryAssignmentProcessor implements NettyRequestProcessor {
                         } else {
                             if (cidAll.size() <= mqAll.size()) {
                                 //consumer working in pop mode could share the MessageQueues assigned to the N (N = popWorkGroupSize) consumer following it in the cid list
-                                allocateResult = allocateMessageQueueStrategy.allocate(consumerGroup, clientId, mqAll, cidAll);
+                                allocateResult = allocateMessageQueueStrategy.allocate(null, consumerGroup, clientId, mqAll, cidAll);
                                 int index = cidAll.indexOf(clientId);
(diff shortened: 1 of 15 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
```diff
@@ -220,13 +220,13 @@ private RemotingCommand queryAssignment(ChannelHandlerContext ctx, RemotingComma
                         } else {
                             if (cidAll.size() <= mqAll.size()) {
                                 //consumer working in pop mode could share the MessageQueues assigned to the N (N = popWorkGroupSize) consumer following it in the cid list
-                                allocateResult = allocateMessageQueueStrategy.allocate(consumerGroup, clientId, mqAll, cidAll);
+                                allocateResult = allocateMessageQueueStrategy.allocate(null, consumerGroup, clientId, mqAll, cidAll);
```
ShannonDing, line 223: Is it compatible with the previous version?

ShannonDing:
Why not add an interface, but directly modify the original interface?

broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
ni-ze: QueryAssignmentProcessor and AllocateMessageQueueStrategy always  has same version, new version of QueryAssignmentProcessor will not depend on old AllocateMessageQueueStrategy interface.

ni-ze:
> Why not add an interface, but directly modify the original interface?

when allocate MessageQueue, I think it is a natural thing to tall allocator which broker messageQueues local in.

RongtongJin:
Hi @ni-ze  Could you resubmit your PR to the 5.0.0-beta branch because 5.0.0-alpha already release?

PR #3530, opened 2021-11-23
Fix some request header setting errors and naming issues

dingshuangxi888:

The header of request code DELETE_TOPIC_IN_NAMESRV(216) is not same in client and the server, It will not cause problems, but should be fixed.

The naming of DeleteTopicInNamesrvRequestHeader should be DeleteTopicFromNamesrvRequestHeader, delete from , not delete in


lizhanhui:
Nice catch

lizhanhui:
Looks good. Considering the changes is relatively small. I would merge it directly.

Issue #3533, opened 2021-11-24
It is better to specify the initialCapacity when initializing a HashMap as a local variable that will not add elements  in the future


XiaoyiPeng:
**BUG REPORT**

1. Please describe the issue you observed:
  
   When initializing a HashMap as a local variable that will not add elements  in the future, it is better to specify the `initialCapacity`.  

    The code in class  `org.apache.rocketmq.broker.filter.MessageEvaluationContext#keyValues` shown below is better.

```java
public Map<String, Object> keyValues() {
        if (properties == null) {
            return null;
        }

        Map<String, Object> copy = new HashMap<String, Object>(properties.size(), 1);

        for (String key : properties.keySet()) {
            copy.put(key, properties.get(key));
        }

        return copy;
    }
```

2. Please tell us about your environment:

    Windows10

vicya:
some advices.  If there are for four elements ,the capacity should be larger than four to prevent expansion

XiaoyiPeng:
> some advices. If there are for four elements ,the capacity should be larger than four to prevent expansion

Use the style of the previous code(like` org.apache.rocketmq.broker.filter.MessageEvaluationContext#keyValues()` ), the `loadFactor` of `HashMap` has been set to 1 and will not be expanded.

PR #3534, opened 2021-11-24
[ISSUE #3533] optimize: It is better to specify the initialCapacity when initializing a HashMap as a local variable that will not add elements  in the future.

XiaoyiPeng:

for issue #3533 

## Brief changelog

Specify the initialCapacity when initializing a `HashMap` as a local variable that will not add elements  in the future, 
just like the code in class `org.apache.rocketmq.broker.filter.MessageEvaluationContext#keyValues` as shown below , it is better. 

```java
public Map<String, Object> keyValues() {
        if (properties == null) {
            return null;
        }

        Map<String, Object> copy = new HashMap<String, Object>(properties.size(), 1);

        for (String key : properties.keySet()) {
            copy.put(key, properties.get(key));
        }

        return copy;
    }
```

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/schedule/ScheduleMessageService.java
@@ -234,7 +234,7 @@ public class ScheduleMessageService extends ConfigManager {
     }
 
     public boolean parseDelayLevel() {
-        HashMap<String, Long> timeUnitTable = new HashMap<String, Long>();
+        HashMap<String, Long> timeUnitTable = new HashMap<String, Long>(4, 1);
         timeUnitTable.put("s", 1000L);
         timeUnitTable.put("m", 1000L * 60);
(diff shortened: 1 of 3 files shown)


odbozhou (commented):
In my opinion, setting loadFactor=1 is not a good way, the default value can satisfy most scenarios.

XiaoyiPeng:
> In my opinion, setting loadFactor=1 is not a good way, the default value can satisfy most scenarios.

Here is using the style of the previous version code(like `org.apache.rocketmq.broker.filter.MessageEvaluationContext#keyValues()` ),as shown below:
```java
public Map<String, Object> keyValues() {
        if (properties == null) {
            return null;
        }

        Map<String, Object> copy = new HashMap<String, Object>(properties.size(), 1);

        for (String key : properties.keySet()) {
            copy.put(key, properties.get(key));
        }

        return copy;
    }
```

 because we know exactly how many elements this map is going to put in, and we don't want  the map be expansioned(`resize`).

Issue #3543, opened 2021-11-29
fix PullConsumer notes wrong


mrhbj:
it is a minor question,I  advise fix it.

            executors.execute(new Runnable() {

                public void doSomething(List<MessageExt> msgs) {
                    //do you business

                }

duhenglucky:
@mrhbj would you like submit a PR to fix it?

mrhbj:
@duhenglucky yes,I have push a pr to fix it #3544

PR #3544, opened 2021-11-29
fix notes wrong (#3543)

mrhbj:
it is a minor question,I advise fix it.

        executors.execute(new Runnable() {

            public void doSomething(List<MessageExt> msgs) {
                //do you business

            }


vongosling:
...

PR #3588, opened 2021-12-06
[ISSUE 3585] [Part B] Improve encode/decode performance

areyouok:
1. [Part B] Improve header encode/decode performance
2. [Part B] Improve RocketMQSerializable performance with zero-copy


remoting/src/main/java/org/apache/rocketmq/remoting/protocol/FastCodesHeader.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.rocketmq.remoting.exception.RemotingCommandException;
+
+import io.netty.buffer.ByteBuf;
+
+public interface FastCodesHeader {
+
+    default String getAndCheckNotNull(HashMap<String, String> fields, String field) {
+        String value = fields.get(field);
+        if (value == null) {
+            String headerClass = this.getClass().getSimpleName();
```
xujianhai666, line 31: maybe cached headerClasss is beffer

areyouok: Here is an exception branch, there is no need to cache the headerCalss string.

common/src/main/java/org/apache/rocketmq/common/protocol/header/SendMessageResponseHeader.java
```diff
@@ -37,6 +42,37 @@
...
+    public void encode(ByteBuf out) {
+        writeIfNotNull(out, "msgId", msgId);
+        writeIfNotNull(out, "queueId", queueId);
+        writeIfNotNull(out, "queueOffset", queueOffset);
+        writeIfNotNull(out, "transactionId", transactionId);
+    }
+
+    @Override
+    public void decode(HashMap<String, String> fields) throws RemotingCommandException {
+        String str = getAndCheckNotNull(fields, "msgId");
```
yuz10, line 55: It would be better to add a test to compare the old decode result and fast decode result in case we add a new field later.

common/src/test/java/org/apache/rocketmq/common/protocol/header/FastCodesHeaderTest.java
```diff
@@ -65,9 +75,11 @@ public void testDecodeSendMessageHeaderV2() throws Exception {
         return extFields;
     }
 
-    private void check(RemotingCommand command, List<Field> fields) throws Exception {
-        SendMessageRequestHeaderV2 o1 = (SendMessageRequestHeaderV2) command.decodeCommandCustomHeader(SendMessageRequestHeaderV2.class);
-        SendMessageRequestHeaderV2 o2 = AbstractSendMessageProcessor.decodeSendMessageHeaderV2(command);
+    private void check(RemotingCommand command, List<Field> fields,
+            Class<? extends CommandCustomHeader> classHeader) throws Exception {
+        CommandCustomHeader o1 = command.decodeCommandCustomHeader(classHeader);
```
yuz10, line 80: If the classHeader is instance of FastCodesHeader, command.decodeCommandCustomHeader(classHeader) also calls fast decode, the result will always be the same

common/src/main/java/org/apache/rocketmq/common/protocol/header/SendMessageResponseHeader.java
yuz10: I found the tests, please ignore it.

yuz10 (approved):

caigy (commented):
In request or response header, some `Integer` fields(queueId, sysFlag, etc) are assigned by `java.lang.Integer#parseInt(java.lang.String)`. Is it more appropriate to use `java.lang.Integer#valueOf(java.lang.String)` to prevent auto-boxing? `valueOf()` will use `IntegerCache`, but the behavior of auto-boxing depends on the implementation of compiler.

areyouok:
> In request or response header, some `Integer` fields(queueId, sysFlag, etc) are assigned by `java.lang.Integer#parseInt(java.lang.String)`. Is it more appropriate to use `java.lang.Integer#valueOf(java.lang.String)` to prevent auto-boxing? `valueOf()` will use `IntegerCache`, but the behavior of auto-boxing depends on the implementation of compiler.

The java 8 compiler use Integer.valueOf(Integer) for auto-boxing, frequently requested values is caching.

By the way, the rocketmq header objects should use primitive types, but I do not change this time.

vongosling (approved):

vongosling:
Support go ahead.

HScarb (approved):

PR #3594, opened 2021-12-07
[ISSUE 3585] [Part F] eliminate regex match in topic/group name check

shuangchengsun:
use bit map to replace regex.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
+++ b/common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
@@ -22,8 +22,6 @@ import org.apache.rocketmq.remoting.protocol.RemotingCommand;
 
 import java.util.HashSet;
 import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public class TopicValidator {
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
```diff
@@ -62,14 +58,39 @@
...
 
         NOT_ALLOWED_SEND_TOPIC_SET.add(RMQ_SYS_SCHEDULE_TOPIC);
+
+        // regex: ^[%|a-zA-Z0-9_-]+$
+        // %
+        VALID_CHAR_BIT_MAP['%'] = true;
+        // -
+        VALID_CHAR_BIT_MAP['-'] = true;
+        // _
+        VALID_CHAR_BIT_MAP['_'] = true;
```
yuz10, line 68: is the '|' missing?  the original regex contains '|'.

shuangchengsun: oh it's my fault, I used to think it was a separator. I have fixed it.

yuz10 (approved):

areyouok (approved):

Issue #3613, opened 2021-12-10
query message by key, expect to see null, but receive one message


xijiu:
**BUG REPORT**

issue describe:

- create two topic:`AaTopic`and`BBTopic `, send a message to topic-`AaTopic` with message key `Aa`, and send a message to topic-`BBTopic` with message key `BB`. Then query message by key, topic and key are `AaTopic`-`BB`, expect to see null, but i receive **ONE** message

![image](https://user-images.githubusercontent.com/19780771/145506070-67a21f4b-f0c7-41e2-8759-8637b60c1f9c.png)

environment:

- version 4.9.2

xijiu:
I've fixed it

[PR](https://github.com/apache/rocketmq/pull/3614)

HScarb:
Did you figure out what's the reason of this problem?

odbozhou:
merged

PR #3616, opened 2021-12-10
[ISSUE 3613] bug fix, solve message hash conflict in index file

xijiu:

bug fix, solve message hash conflict in index file

[Github issue](https://github.com/apache/rocketmq/issues/3613)

## Brief changelog

fix org.apache.rocketmq.store.DefaultMessageStore

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -406,12 +407,14 @@ public class MQAdminImpl {
                             }
                         } else {
                             String keys = msgExt.getKeys();
+                            String msgTopic = msgExt.getTopic();
                             if (keys != null) {
                                 boolean matched = false;
                                 String[] keyArray = keys.split(MessageConst.KEY_SEPARATOR);
                                 if (keyArray != null) {
                                     for (String k : keyArray) {
-                                        if (key.equals(k)) {
+                                        // both topic and key must be equal at the same time
+                                        if (Objects.equals(key, k) && Objects.equals(topic, msgTopic)) {
                                             matched = true;
                                             break;
                                         }
--- a/test/src/test/java/org/apache/rocketmq/test/base/BaseConf.java
+++ b/test/src/test/java/org/apache/rocketmq/test/base/BaseConf.java
@@ -70,9 +70,12 @@ public class BaseConf {
 
     public static String initTopic() {
         String topic = MQRandomUtils.getRandomTopic();
-        IntegrationTestBase.initTopic(topic, nsAddr, clusterName);
+        return initTopicWithName(topic);
+    }
 
-        return topic;
+    public static String initTopicWithName(String topicName) {
+        IntegrationTestBase.initTopic(topicName, nsAddr, clusterName);
+        return topicName;
     }
 
     public static String initConsumerGroup() {
--- a/test/src/test/java/org/apache/rocketmq/test/client/producer/querymsg/QueryMsgByKeyIT.java
+++ b/test/src/test/java/org/apache/rocketmq/test/client/producer/querymsg/QueryMsgByKeyIT.java
@@ -18,7 +18,9 @@
 package org.apache.rocketmq.test.client.producer.querymsg;
 
 import java.util.List;
+
 import org.apache.log4j.Logger;
+import org.apache.rocketmq.client.exception.MQClientException;
 import org.apache.rocketmq.common.message.MessageExt;
 import org.apache.rocketmq.test.base.BaseConf;
 import org.apache.rocketmq.test.client.rmq.RMQNormalProducer;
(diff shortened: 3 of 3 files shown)


store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -993,6 +985,29 @@ public QueryMessageResult queryMessage(String topic, String key, int maxNum, lon
...
+     * check target msg match
+     *
+     * @param msg   message
+     * @param topic topic
+     * @param key   message key
+     * @return  is match
+     */
+    private boolean isMessageMatch(MessageExt msg, String topic, String key) {
+        String keys = msg.getKeys();
+        if (Objects.equals(topic, msg.getTopic()) && StringUtils.isNotEmpty(keys)) {
```
Git-Yang, line 998: Filtering here may reduce the number of messages. It is recommended to filter in the selectPhyOffset() method.

panzhi33:
org.apache.rocketmq.store.index.IndexService#queryOffset

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
xijiu: filter in selectPhyOffset() may increase read commitlog times, if this is acceptable, I'll modify it

HScarb: IMO, it's correct to filtering here, as we only have physical offsets in `selectPhyOffset()`, so we can't check the topic and key of message in `selectPhyOffset()`

Git-Yang: Yes, I agree with you.

HScarb:
There are some commented old code implementing topic and key check.
Anyone know Why are these code commented？

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
xijiu: can we put check method in class `DefaultMessageStore` like this ? :

```
QueryOffsetResult queryOffsetResult = this.indexService.queryOffset(topic, key, maxNum, begin, lastQueryMsgTime, physicOffset -> {
    SelectMappedBufferResult result = commitLog.getData(physicOffset, false);
    return isTopicAndKeyMatch(result, topic, key);
});
```

if so, we will do more once IO and protocol analysis

dongeforever:
> There are some commented old code implementing topic and key check. Anyone know Why are these code commented？ https://github.com/apache/rocketmq/pull/3616/files#diff-005813abb0e69e737cbda0fa3dcb04dddd4ab165336fae9619af60ea577fe883L958-L966
> 
> ```java
> //                    String[] keyArray = msg.getKeys().split(MessageConst.KEY_SEPARATOR);
> //                    if (topic.equals(msg.getTopic())) {
> //                        for (String k : keyArray) {
> //                            if (k.equals(key)) {
> //                                match = true;
> //                                break;
> //                            }
> //                        }
> //                    }
> ```

The logic is moved to client  in MQAdminImpl.queryMessage.
Try to reduce the load of brokers.

dongeforever:
@xijiu Don't try to do the check in broker.

The query is usually invoked by the admin client. So it is better to do the check in client side.

If the check is done in server-side, it may affect the brokers.

xijiu:
@dongeforever  I see. I'll correct it again

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -948,24 +952,12 @@ public QueryMessageResult queryMessage(String topic, String key, int maxNum, lon
                 long offset = queryOffsetResult.getPhyOffsets().get(m);
 
                 try {
-
-                    boolean match = true;
```
duhenglucky, line 952: Actually match isn't used in broker, so how about move the check logic to client？

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -993,6 +985,29 @@ public QueryMessageResult queryMessage(String topic, String key, int maxNum, lon
...
+     * @param topic topic
+     * @param key   message key
+     * @return  is match
+     */
+    private boolean isMessageMatch(MessageExt msg, String topic, String key) {
+        String keys = msg.getKeys();
+        if (Objects.equals(topic, msg.getTopic()) && StringUtils.isNotEmpty(keys)) {
+            String[] keyArray = keys.split(MessageConst.KEY_SEPARATOR);
+            if (keyArray.length > 0) {
+                for (String k : keyArray) {
```
duhenglucky, line 1001: Same as last comment

store/src/test/java/org/apache/rocketmq/store/index/IndexFileTest.java
```diff
@@ -21,13 +21,30 @@
...
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
+import org.apache.rocketmq.common.BrokerConfig;
+import org.apache.rocketmq.common.Pair;
 import org.apache.rocketmq.common.UtilAll;
+import org.apache.rocketmq.common.message.MessageConst;
+import org.apache.rocketmq.common.message.MessageDecoder;
+import org.apache.rocketmq.store.*;
```
duhenglucky, line 38: Please don't aggregate all imports with *

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
xijiu: OK, I'll change it soon

xijiu:
@dongeforever @duhenglucky     

I've moved the check logic to the client

duhenglucky (approved):

panzhi33 (approved):

vongosling (approved):

odbozhou (approved):

PR #3658, opened 2021-12-16
add stats all test

zhaohai1299002788:
add stats all test


vongosling:
Would you like to refer to other test cases? BTW, you also follow one issue one pr rule.

Issue #3585, opened 2021-12-06
Improve performance for 4.9.2


areyouok:
We have some performance improvements based on 4.9.2

1. [Part A] eliminate reverse DNS lookup in MessageExt
2. [Part B] Improve header encode/decode performance
3. [Part B] Improve RocketMQSerializable performance with zero-copy
4. [Part C] cache result for parseChannelRemoteAddr()
5. [Part D] improve performance of createUniqID()
6.  [Part E] eliminate duplicated getNamespace() call when where is no namespace
7. [Part F] eliminate regex match in topic/group name check
8. [Part G] [Work in progress] support send batch message with different topic/queue
9. [Part H] eliminate StringBuilder auto resize in PullRequestHoldService.buildKey() when topic length is greater than 14, this method called twice for each message.
10. [Part I] Avoid unnecessary StringBuffer resizing and String Formatting
11. [Part J] Use mmap buffer instead of FileChannel when writing consume queue and slave commit log, which greatly speed up consume tps.
12. [Part K](https://github.com/apache/rocketmq/pull/3659) move execution of notifyMessageArriving() from ReputMessageService thread to PullRequestHoldService thread. 

These commits almost eliminate bad performance methods in the cpu flame graph in producer side.

areyouok:
Part A to F is complete. Unit test fixed.

kimmking:
About Part J:
Actually, mmap is better than fileChannel in small size message case. In large scale size message, fileChannel is more better.
There could be an option to switch mmap/fileChannel.
See https://www.jianshu.com/p/d0b4ac90dbcb

PR #3659, opened 2021-12-16
[ISSUE #3585] [Part K]  move execution of notifyMessageArriving() from ReputMessageService thread to PullRequestHoldService thread

areyouok:
This commit speed up consume qps greatly, in our test up to 200,000 qps.

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullNotifyQueue.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullNotifyQueue.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.broker.longpolling;
+
+import java.util.LinkedList;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+class PullNotifyQueue<T> {
+    private PullNotifyQueueConfig config;
+
+    private final LinkedBlockingQueue<T> queue;
+
+    private int activeConsumeQueueCount = 1;
+    private long activeConsumeQueueEpochTime = System.nanoTime();
+    // only change in unit test
+    private long activeConsumeQueueRefreshTime = 1_000_000_000;
+
+    private long lastBatchDrainTime;
+
+    public PullNotifyQueue(PullNotifyQueueConfig config) {
+        queue = new LinkedBlockingQueue<>(500_000);
+        this.config = config;
+    }
+
+    public void put(T data) throws InterruptedException {
+        queue.put(data);
+    }
+
+    public LinkedList<T> drain() throws InterruptedException {
+        LinkedList<T> result = new LinkedList<>();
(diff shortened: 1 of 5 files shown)


areyouok:
Flame chart before this commit:

thread

caigy:
I found `System.nanoTime()` was used instead of `System.currentTimeMillis()`, will it bring benefits to performance improvement? It also brings  some conversion between nanosecond variables and config values defined in millisecond.

areyouok:
> I found `System.nanoTime()` was used instead of `System.currentTimeMillis()`, will it bring benefits to performance improvement? It also brings some conversion between nanosecond variables and config values defined in millisecond.

There is no performance benefit. 

nanoTime() are more strictly than currentTimeMillis() for measuring time elapse. For example, NTP service may adjust system clock, the nanoTime() method will not affect by it. See javadoc of the method.

areyouok:
I found the current commit may increase consume lantency in performance test (abount 45ms).

I'm working on it.

areyouok:
finished

dongeforever:
is there any impact on the "end to end" latency?

Maybe a performance test report is needed.

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullNotifyQueue.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public void put(T data) throws InterruptedException {
+        queue.put(data);
+    }
+
+    public LinkedList<T> drain() throws InterruptedException {
+        LinkedList<T> result = new LinkedList<>();
+        long tps = config.getTps();
+        if (tps <= config.getTpsThreshold()) {
+            T data = queue.poll(100, TimeUnit.MILLISECONDS);
```
dongeforever, line 48: Why use the store put Tps to control the long polling?

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullNotifyQueue.java
```diff
@@ -0,0 +1,147 @@
...
+            if (data != null) {
+                result.add(data);
+            }
+        } else {
```
dongeforever, line 52: The expectation here is :
* if the queue has data, poll it as quickly as possible
* if the queue doesn't have data, just return the already polled data.

The code maybe be simplified as:
int pollWaitTimeMs = 0;
while(true) {
    T data = queue.poll(pollWaitTimeMs, TimeUnit.MILLISECONDS);
    if (data == null) {
      pollWaitTimeMs = 100;
   } else { 
     pollWaitTimeMs = 0;
     result.add(data);
   }
   if(result.size() > batchMax) {
      break;
   }
   if(result.size() >0 && pollWaitTimeMs > 0) {
     break;
   }
   if(System.nanoTime() - start > maxWaitTimeNonas) {
    break;
 }
}

The tps control seems unnecessary.

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullNotifyQueue.java
areyouok: to improve throughput and decrease cpu cost when tps is greater than threshold.

dongeforever: The end-to-end latency is more important in most cases for rocketmq. How about adjusting the default tps threshold to Integer.MaxValue? 
if someone wants to promote throughput, it could be adjusted.

BTW,  The cpu cost is a problem. The tag filter and property filter are unnecessary in the notification process,  It will be done twice for the pull process will do that work too.
The notification should be lightweight as far as possible.

areyouok: In most case tps is below 100000, we enable this when tps is greater than 100000 to protect rocketmq.

in our tests, 600 queue, 150000+ tps, end-to-end latency is 18ms; 18 queue, 150000+ tps, end-to-end latency is 2ms.

areyouok: we can modify the threshold to a greater number, such as 150000.

In case tps greater than 150000 and a lot of queue is writing, I think the latency is not so important.

vongosling:
Any update?

areyouok:
> Any update?

There is no update. But we have run it in out product environment for 3 months.

Issue #3786, opened 2022-01-24
The length of properties value should be checked before converting them to short.


XiaoyiPeng:
**BUG REPORT**

1. Please describe the issue you observed:

https://github.com/apache/rocketmq/blob/35334e266984c4a3ee6179ffd44bb37372788696/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java#L162

The length of properties value should be checked before converting it to short, **otherwise this value is negative !**

https://github.com/apache/rocketmq/blob/35334e266984c4a3ee6179ffd44bb37372788696/store/src/main/java/org/apache/rocketmq/store/CommitLog.java#L1609

The length of properties value should be checked before converting them to short when encode `MessageExtBatch`. 
The class `CommitLog` **only check** length of properties value when call encode() method for `MessageExtBrokerInner`.

If not do this (check length of properties value), the error message will be ambiguous, as shown below: 

![image](https://user-images.githubusercontent.com/8653312/150728543-e6291189-7ce9-4e95-a370-fd3394073396.png)

2. Please tell us about your environment:

    OS: Windows 10
    branch: develop

PR #3788, opened 2022-01-24
[ISSUE #3786] fix: The length of properties value should be checked before converting them to short.

XiaoyiPeng:

For issue #3786


dongeforever (approved):

dongeforever:
Great Catch.
The check was done for a single message, but not for the batch mode.

duhenglucky (approved):

XiaoyiPeng:
> Great Catch. The check was done for a single message, but not for the batch mode.

Yes, thanks for your review.

tianliuliu:
Whether to add more information to the RuntimeException message, such as the current prop length, limit length...

XiaoyiPeng:
> Whether to add more information to the RuntimeException message, such as the current prop length, limit length...

Thanks for your advice, I've corrected it.

For commit in class `CommitLog`, in order to keep the **code style consistent**, the properties length limit is only recorded in the log, and the exception message is consistent with the code style below: 

image

panzhi33 (approved):

duhenglucky (approved):

Issue #3785, opened 2022-01-22
about MQClientAPIImpl improvement


Kvicii:
hi there.
In MQClientAPIImpl, I see a lot about switch...case code blocks, Is this a preparation for the follow-up or can it be completely improved?

duhenglucky:
@Kvicii please feel free to start a discussion with your design, any optimizations to the code are welcome

Kvicii:
@duhenglucky hi, I think it is possible to merge these code blocks,and I will provide a PR later

Kvicii:
@duhenglucky Does our PR need to be checked out from the develop branch?

tianliuliu:
> I wi

yes, PR based on the develop branch

Kvicii:
[ISSUE#3785](https://github.com/apache/rocketmq/pull/3793)

PR #3793, opened 2022-01-24
[ISSUE #3785] MQClientAPIImpl improvement

Kvicii:
For this class, I see a lot of switch...case blocks, and it's not necessary.Code clutter caused by excessive nesting.
So I merged all the code,but here is another problem,exceptions thrown in switch...case are not the same, so I think if further abstraction is needed.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -173,8 +173,8 @@ import org.apache.rocketmq.remoting.protocol.RemotingSysResponseCode;
 public class MQClientAPIImpl {
 
     private final static InternalLogger log = ClientLogger.getLog();
-	private static final Pattern COMPILE = Pattern.compile(";");
-	private static boolean sendSmartMsg =
+    private static final Pattern COMPILE = Pattern.compile(";");
+    private static boolean sendSmartMsg =
         Boolean.parseBoolean(System.getProperty("org.apache.rocketmq.client.sendSmartMsg", "true"));
 
     static {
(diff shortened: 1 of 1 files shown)


Kvicii:
@duhenglucky hi, I modify this class, As I marked in the commit, do we need to further abstract a top-level exception class and a general checkXXX method

Kvicii:
Can anyone help me review the code?
thx

dongeforever:
@Kvicii  It is better to keep the same code style in MQClientAPIImpl.
Of course, the switch style has disadvantages such as excessive nesting, but it also has advantages. For the RPC layer, the switch style is more readable to handle different kinds of response code and easy to extend further codes.

dongeforever:
@Kvicii The exceptions are a true problem.
It is better to provide a unified exception definition for all the APIs. 
If you are interested, you could collect all the exception types, and introduce a new definition.
The HTTP error code could be a reference, 4xx for client error, 5xx for broker error.

Kvicii:
@dongeforever 
thx
First of all, I think that too many cases cause the code to be unreadable and confusing. If only one of the states needs to be judged, I think redundant judgments can be avoided.
Secondly, is the exception design only for client-related APIs?

dongeforever:
@Kvicii Currently the community is developing the 5.0. Changing the basic exception mechanism may cause bugs and conflicts. It's better to hold on for the moment.

How about involving in the 5.0 rips(if it is at your convenience)

https://github.com/apache/rocketmq/blob/5.0.0-alpha/docs/cn/statictopic/RocketMQ_Static_Topic_Logic_Queue_%E8%AE%BE%E8%AE%A1.md

Kvicii:
Of course.

Issue #3806, opened 2022-01-28
[Feature Request] Support New Batch Producer API that allow multiple topics in one Batch


Zanglei06:
The current batch producer api requires that all messages in the batch belongs to one topic. However, sometimes user want to send batch message that belongs to multiple topics.  It would be good to provide such multi-batch producer api.

duhenglucky:
@Zanglei06 Would you like to review this PR, these two PRs can be better combined.

guyinyou:
The old Broker side can send multiple topics in batches, because the Broker unpacks each Message and redelivers it after receiving the MessageBatch. The new Broker side enhances the MessageBatch capability, but because it does not unpack specific Messages, it does not support batch sending of multiple topics.

guyinyou:
Wecurrently plan to do a feature of sending multiple topics in one batch, but it may more like aggregation of multiple MessageBath requests in the remoting layer

PR #3807, opened 2022-01-28
[ISSUE #3806] add multi-topic batch api for producer

Zanglei06:

Add a new batch producer api that allow multiple topics in one batch.  the new api support both Master-Slave and Dledger Mode.  This New API is already run on our production environment for several months in Master-Slave Mode.

## Brief changelog

SendMessageRequestHeader add two fields:
bool multiTopic, denote that the request is multi-batch when true.
string queueIds, denote that the queueId for each topic. , e.g. "1,9" for two different topics.
string topic, the original topic field is encode to represent the topics in the batch request. e.g. "topicA%%topicB" for two topics.

If the batch message's topics does not have a shared broker, MQClientException will be thrown.

## Verifying this change

XXXX

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
@@ -17,13 +17,13 @@
 package org.apache.rocketmq.broker.processor;
 
 import java.net.SocketAddress;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ThreadLocalRandom;
 
-import io.netty.channel.ChannelHandlerContext;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.mqtrace.ConsumeMessageContext;
 import org.apache.rocketmq.broker.mqtrace.ConsumeMessageHook;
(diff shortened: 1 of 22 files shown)


client/src/main/java/org/apache/rocketmq/client/producer/MQProducer.java
```diff
@@ -91,6 +91,10 @@ TransactionSendResult sendMessageInTransaction(final Message msg,
     SendResult send(final Collection<Message> msgs) throws MQClientException, RemotingException, MQBrokerException,
         InterruptedException;
 
+    // for batch msgs with multi topics.
+    SendResult sendMultiTopicBatch(final Collection<Message> msgs) throws MQClientException, RemotingException, MQBrokerException,
+            InterruptedException;
+
```
dongeforever, line 97: Who has the responsibility to make sure the messages are to be sent to the same broker? How to make sure it?

This API may cause many difficulties to be compatible with advanced features, such as logic queue and message with sharding key?

Only the inner topic route has enough information to decide which queue to be sent for each message. So the multi-topic batch can only be constructed by the inner MQClientInstance, but not by the user directly.

IMO, it is necessary to raise a RIP and provide a detailed design for this feature.

broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
```diff
@@ -588,6 +603,27 @@ private RemotingCommand handlePutMessageResult(PutMessageResult putMessageResult
             return CompletableFuture.completedFuture(response);
         }
 
+        String[] topics = null;
+        int[] queueIds = null;
+        if (requestHeader.isMultiTopic()) {
+            topics = requestHeader.getTopic().split(MixAll.BATCH_TOPIC_SPLITTER); // decode topics
+            if (requestHeader.getQueueIds() == null) {
```
dongeforever, line 610: For reference only. 
How about providing a common BATCH API style?
Encode/Decode the header and body in batch mode, and reuse the processor/storage layer.
It is better not to introduce the batch difficulty to the storage layer. 
For advanced feature batch consumequeue, it is not able to handle the inner message num, if the message has multi topic-queues.

dongeforever:
@duhengforever This feature may need further discussion.

Issue #3827, opened 2022-02-08
Improve Performance of Transactional Message and Delay Message Scheduling


shuangchengsun:
Recently, we conducted a stress test on transaction messages, but found that the performance of transaction messages was significantly lower than the previous version. After our analysis, we found that this phenomenon was caused by a bug in jdk.
![image](https://user-images.githubusercontent.com/33629004/152945487-304c4ca6-001e-49f7-9a68-78b563ada473.png)
As we can see, This method consumes a lot of time, which is obviously unreasonable. With further investigation, it turns out that this is a bug in JDK-8(https://bugs.openjdk.java.net/browse/JDK-8227019), which has been fixed in subsequent versions. Unfortunately, this bug still alive in jdk 1.8.0_202.

The root cause in code as figure show. An easy way to fix it is to use timeout parameter, which can improve 50%
![image](https://user-images.githubusercontent.com/33629004/152948229-5534df54-e8a0-4da1-8e2b-fcdf1721caea.png)

shuangchengsun:
old profile
![image](https://user-images.githubusercontent.com/33629004/152953564-5af592e0-98f8-41e6-9ea4-fade304e0271.png)





new profile
![image](https://user-images.githubusercontent.com/33629004/152953248-0d003ba7-d38e-4627-8494-607f7778fbf4.png)


benchmark result as below. (command: sh tproducer.sh -n xxx.xxx.xxx.xxx:9876 -s 128 -w 64 -sr 0 -su 0 -cu 0 -cr 0 -i 0)
old
![image](https://user-images.githubusercontent.com/33629004/152965364-1337677d-0453-4367-bcab-6e9b8682fadc.png)

new 
![image](https://user-images.githubusercontent.com/33629004/152967124-d1d3694b-3d11-4304-bd84-18c7eb4ea5c3.png)

PR #3828, opened 2022-02-08
[ISSUE #3827]Improve Performance of transactional message and schedule message

shuangchengsun:
#3827 Improve Performance of Transactional Message

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -723,10 +723,6 @@ public class CommitLog {
             }
             if (replicaStatus != PutMessageStatus.PUT_OK) {
                 putMessageResult.setPutMessageStatus(replicaStatus);
-                if (replicaStatus == PutMessageStatus.FLUSH_SLAVE_TIMEOUT) {
-                    log.error("do sync transfer other node, wait return, but failed, topic: {} tags: {} client address: {}",
-                            msg.getTopic(), msg.getTags(), msg.getBornHostNameString());
-                }
             }
             return putMessageResult;
         });
@@ -836,10 +832,6 @@ public class CommitLog {
             }
             if (replicaStatus != PutMessageStatus.PUT_OK) {
                 putMessageResult.setPutMessageStatus(replicaStatus);
-                if (replicaStatus == PutMessageStatus.FLUSH_SLAVE_TIMEOUT) {
-                    log.error("do sync transfer other node, wait return, but failed, topic: {} client address: {}",
-                            messageExtBatch.getTopic(), messageExtBatch.getBornHostNameString());
-                }
             }
             return putMessageResult;
         });
@@ -877,7 +869,7 @@ public class CommitLog {
             if (messageExt.isWaitStoreMsgOK()) {
                 if (service.isSlaveOK(result.getWroteBytes() + result.getWroteOffset())) {
                     GroupCommitRequest request = new GroupCommitRequest(result.getWroteOffset() + result.getWroteBytes(),
-                            this.defaultMessageStore.getMessageStoreConfig().getSyncFlushTimeout());
+                            this.defaultMessageStore.getMessageStoreConfig().getSlaveTimeout());
                     service.putRequest(request);
                     service.getWaitNotifyObject().wakeupAll();
                     return request.future();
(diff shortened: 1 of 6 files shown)


Kvicii:
@shuangchengsun hi, I think if there is a corresponding performance test report, the comparison will be more perfect.

areyouok (approved):

shuangchengsun:
> @shuangchengsun hi, I think if there is a corresponding performance test report, the comparison will be more perfect.

old profile
image





new profile
image


benchmark result as below. (command: sh tproducer.sh -n xxx.xxx.xxx.xxx:9876 -s 128 -w 64 -sr 0 -su 0 -cu 0 -cr 0 -i 0)
old
image

new 
image

HScarb (approved):

XiaoyiPeng:
Good catch！

XiaoyiPeng (approved):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -510,17 +511,20 @@ private boolean isLmqConsumeQueueNumExceeded() {
     @Override
     public PutMessageResult putMessage(MessageExtBrokerInner msg) {
         try {
-            return asyncPutMessage(msg).get();
-        } catch (InterruptedException | ExecutionException e) {
+            // to make sure put timeout bigger than flush timeout, full gc is also considered
+            int putTimeout  = this.messageStoreConfig.getSyncFlushTimeout() + 5000;
```
RongtongJin, line 515: IMO, it would be better to add a new messageStoreConfig like putMessageTimeout than using syncFlushTimeout + 5000.

panzhi33 (approved):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -509,19 +510,27 @@ private boolean isLmqConsumeQueueNumExceeded() {
...
+        return waitForPutResult(asyncPutMessages(messageExtBatch));
+    }
+
+    private PutMessageResult waitForPutResult(CompletableFuture<PutMessageResult> putMessageResultFuture) {
         try {
-            return asyncPutMessages(messageExtBatch).get();
-        } catch (InterruptedException | ExecutionException e) {
+            int putMessageTimeout =
+                    Math.min(this.messageStoreConfig.getSyncFlushTimeout(), this.messageStoreConfig.getSlaveTimeout());
+            return putMessageResultFuture.get(putMessageTimeout, TimeUnit.MILLISECONDS);
```
XiaoyiPeng, line 525: Can we make this parameter `putMessageTimeout` configurable？

shuangchengsun: emmm. i think it is unnecessary to add new config.  it may confused to add new configuration since  here are already enough parameter. The parameter「SlaveTimeout」is used to distinguish which step is timeout while "flush disk" and "flush slave" use the same value 「SyncFlushTimeout」 in old version. However, I just found here still exist some problems in the code above, which will be fixed right away

store/src/main/java/org/apache/rocketmq/store/ha/HAService.java
```diff
@@ -287,17 +287,12 @@ private void doWaitTransfer() {
             if (!this.requestsRead.isEmpty()) {
                 for (CommitLog.GroupCommitRequest req : this.requestsRead) {
                     boolean transferOK = HAService.this.push2SlaveMaxOffset.get() >= req.getNextOffset();
-                    long waitUntilWhen = HAService.this.defaultMessageStore.getSystemClock().now()
-                            + HAService.this.defaultMessageStore.getMessageStoreConfig().getSyncFlushTimeout();
-                    while (!transferOK && HAService.this.defaultMessageStore.getSystemClock().now() < waitUntilWhen) {
+                    long deadLine = req.getDeadLine();
+                    while (!transferOK && deadLine - System.nanoTime() > 0) {
```
ni-ze, line 291: replace 
```java
HAService.this.defaultMessageStore.getSystemClock().now() + req.getTimeoutMillis()
```
with deadLine is not necessary and recommended.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1143,19 +1135,17 @@ public long getJointime() {
...
         private final long nextOffset;
         private CompletableFuture<PutMessageStatus> flushOKFuture = new CompletableFuture<>();
-        private final long startTimestamp = System.currentTimeMillis();
-        private long timeoutMillis = Long.MAX_VALUE;
+        private final long deadLine;
 
         public GroupCommitRequest(long nextOffset, long timeoutMillis) {
             this.nextOffset = nextOffset;
-            this.timeoutMillis = timeoutMillis;
+            this.deadLine = System.nanoTime() + (timeoutMillis << 20);
```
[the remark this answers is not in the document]
shuangchengsun: yes，This operation will cause a 4% inaccuracy, but I think it is acceptable.

store/src/main/java/org/apache/rocketmq/store/ha/HAService.java
areyouok: nanoTIme() is better than currentTimeMillis() for measure elapsed time. currentTimeMillis() may affect by system clock  adjustment, and cause unexpected result.

RongtongJin (approved):

areyouok (approved):

PR #3848, opened 2022-02-15
[ISSUE #3831] Swapping mapped-byte-buffer to save memory on a system with extra-large disks.

Erik1288:
…ge disks.

Swapping mapped-byte-buffer to save memory on a system with ultra-large disks(more than 10TB).

## Brief changelog

Periodically clean up the mmap of "cold data".

## Verifying this change

pass ut

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -1321,6 +1321,47 @@ public class DefaultMessageStore implements MessageStore {
         this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
             @Override
             public void run() {
+                try {
+                    DefaultMessageStore.this.commitLog.swapMap(getMessageStoreConfig().getCommitLogSwapMapReserveFileNum(),
+                        getMessageStoreConfig().getCommitLogForceSwapMapInterval(), getMessageStoreConfig().getCommitLogSwapMapInterval());
+                    for (ConcurrentMap<Integer, ConsumeQueueInterface> maps : getConsumeQueueTable().values()) {
+                        for (ConsumeQueueInterface logic : maps.values()) {
+                            consumeQueueStore.swapMap(
+                                logic,
+                                getMessageStoreConfig().getLogicQueueSwapMapReserveFileNum(),
+                                getMessageStoreConfig().getLogicQueueForceSwapMapInterval(),
+                                getMessageStoreConfig().getLogicQueueSwapMapInterval()
+                            );
+                        }
+                    }
+                } catch (Exception e) {
+                    log.error("swap map exception", e);
+                }
+            }
+        }, 1, 5, TimeUnit.MINUTES);
+
+        this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    DefaultMessageStore.this.commitLog.cleanSwappedMap(getMessageStoreConfig().getCleanSwapedMapInterval());
+                    for (ConcurrentMap<Integer, ConsumeQueueInterface> maps : getConsumeQueueTable().values()) {
+                        for (ConsumeQueueInterface logic : maps.values()) {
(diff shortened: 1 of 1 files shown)


zhouxinyu:
More suggested comments:
1. Add some comments for interface Swappable, it's difficult to understand the specific meaning of `swapMap` and `cleanSwappedMap` directly. 
2. Add unit tests for this PR, and the implementations of `Swappable`. 
3. Maybe only MappedFile and MappedFileQueue implement `Swappable` is enough, it's a little strange that CommitLog and ConsumeQueueStore also is a subclass of this interface.

And a technical question: 
1. We mapped twice for writing when the old buffer hasn't been cleaned, is there a write conflict problem in this situation? Although the cleaner won't clean the last writing file, the file doesn't know it.

zhouxinyu:
Another technical suggestion:

dongeforever:
@zhouxinyu It may have difficulty doing the lazy mapping safely while the mapped file is being used concurrently, which may need to introduce a lock.

The current code is clear,  avoiding the anxiety of concurrent error, and is used for a long time in the production env.

BTW, if you have a simple way to implement the lazy mapping, it is ok to show it in detail.

dongeforever:
@Erik1288 Why do the commitlog and consumequeuestore need to implement the swappable?

It seems a little strange indeed.

zhouxinyu:
> @zhouxinyu It may have difficulty doing the lazy mapping safely while the mapped file is being used concurrently, which may need to introduce a lock.
> 
> The current code is clear, avoiding the anxiety of concurrent error, and is used for a long time in the production env.
> 
> BTW, if you have a simple way to implement the lazy mapping, it is ok to show it in detail.

Yeap, a lock is necessary for this lazy mapping solution, implementing it is not too complex, we convert direct access `this.mappedByteBuffer` to a getter method `this.getBuffer`:


```
public MappedByteBuffer getBuffer() {
        if (this.mappedByteBuffer != null) {
            return this.mappedByteBuffer;
        }

        synchronized (this) {
            if (this.mappedByteBuffer == null) {
                this.mappedByteBuffer = this.fileChannel.map();
            }
        }
        
        return this.mappedByteBuffer;
    }
```

dongeforever:
It will reserve 3 files, which will only swap the readable file for an append-only system. So it is safe in practice.

As for the technical discussion,  the mapped file is implemented with page cache internally by os. So it is safe to write with two mappedbytebuffer as long as the write pos is not intersected.

zhouxinyu:
> and is used for a long time in the production env.

If this swap mechanism is a mature production solution, take it easy.

zhouxinyu:
> 

Yes, `unmap` should be protected by `RefCount`

dongeforever:
@Erik1288 I found other problems.

* getRecentSwapMapTime is fixed to zero, but in fact, it should return swapMapTime
* cleanSwapedMap, shoud check getRefCount() == 1 and it dose not need to sleep but just return.

dongeforever:
@zhouxinyu It come into another design consideration. The current swap mechnism will make sure the existed message can be read correctly.
Otherwise, it may get null temporately, which may cause message loss.

Erik1288:
Thanks for your comments.

> More suggested comments:
> 
> 1. Add some comments for interface Swappable, it's difficult to understand the specific meaning of `swapMap` and `cleanSwappedMap` directly.

Re: Sure, more meaningful comments will be added.

> 2. Add unit tests for this PR, and the implementations of `Swappable`.

Re: Unit-test and Implementation were pushed but not be called.

Erik1288:
> @Erik1288 I found other problems.
> 
> * getRecentSwapMapTime is fixed to zero, but in fact, it should return swapMapTime
> * cleanSwapedMap, shoud check getRefCount() <= 1 and it dose not need to sleep but just return.

Thanks for pointing it out, I'm working on it.

vongosling:
Any update?

lwclover:
Why didn't you update it

Erik1288:
Sorry for the late. Improvements have been made to the code as suggested in the commit: 26beaa2. Since the code base has been changed, should I reopen a new ISSUE based on the latest code base?

Issue #3883, opened 2022-02-23
inner class DefaultLitePullConsumerImpl.ConsumeRequest may be 'static'


wangdi761:
inner class DefaultLitePullConsumerImpl.ConsumeRequest may be 'static'

lizhiboo:
@wangdi761 ConsumeRequest is only used in DefaultLitePullConsumerImpl. In which case should make it static?

ShannonDing:
> @wangdi761 ConsumeRequest is only used in DefaultLitePullConsumerImpl. In which case should make it static?

best to keep it in the status quo.

PR #3884, opened 2022-02-23
[ISSUE #3883] Inner class DefaultLitePullConsumerImpl.ConsumeRequest may be '…

wangdi761:
#3883


ShannonDing:
Best to keep it in the status quo.

PR #3913, opened 2022-03-02
fix [optimization order message example]

web1992:

Optimization example for order message  and fix topic name typo

## Brief changelog

Optimization example for order message

## Verifying this change

I had run the test for this change

### DIFF
--- a/example/src/main/java/org/apache/rocketmq/example/ordermessage/Consumer.java
+++ b/example/src/main/java/org/apache/rocketmq/example/ordermessage/Consumer.java
@@ -29,11 +29,11 @@ import org.apache.rocketmq.common.message.MessageExt;
 
 /**
  * <p>
- * This example shows RocketMq keep order message (local order not global order)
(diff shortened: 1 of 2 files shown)


example/src/main/java/org/apache/rocketmq/example/ordermessage/Consumer.java
```diff
@@ -26,6 +26,35 @@
 import org.apache.rocketmq.common.consumer.ConsumeFromWhere;
 import org.apache.rocketmq.common.message.MessageExt;
 
+
+/**
+ * <p>
+ * This example shows RocketMq keep order message (local order not global order)
```
ShannonDing, line 32: better to use ”RocketMQ“.

web1992: fix it

ni-ze:
Frankly speaking, I can not find where is partially order from output.

Issue #3424, opened 2021-10-21
Update(Close) UnitSubFlag is useless


lushilin:
update topic UnitSubFlag has bug, if close unitSubFlag is not useful.
![image](https://user-images.githubusercontent.com/17663472/138229219-61c0f676-bc2d-4abf-8dc1-4522a6683d65.png)
when close unitSubFlag, do not have else logic to close it.

lizhanhui:
+1 to clean the dead code.

lushilin:
Yes, i think so. 
How about this @lizhanhui

Issue #2870, opened 2021-05-12
fix some problems in sending msg


sunxi92:
1.Repeatedly call the checkMessage(Message msg, DefaultMQProducer defaultMQProducer) to  check the msg and topic in the sync sending msg
![image](https://user-images.githubusercontent.com/25401190/117928717-866ba080-b32e-11eb-832b-94c5e5800f99.png)
![image](https://user-images.githubusercontent.com/25401190/117928805-a4d19c00-b32e-11eb-8eaa-dca4b51e0abc.png)

2.When select a message queue,  the list of faultItem is shuffled before sorted. I think the shuffle operation is redundant
![image](https://user-images.githubusercontent.com/25401190/117930265-5c1ae280-b330-11eb-99a8-d2ace825a067.png)

3.In the following methods， msg and topic(the method of Validators.checkMessage(...))checks are missing:
![image](https://user-images.githubusercontent.com/25401190/117930597-c469c400-b330-11eb-930b-d37992584692.png)
![image](https://user-images.githubusercontent.com/25401190/117930728-eebb8180-b330-11eb-9fe6-6d5193f9ddc1.png)

Git-Yang:
link https://github.com/apache/rocketmq/issues/2831

yuz10:
#2080

PR #3946, opened 2022-03-08
[#2870] move all reset topic with namespace to sendDefaultImpl()

ferrirW:
image

#3424 delete check may cause an NPE.

A better way is that move checkMessage & msg.setTopic to defaultMQProducerImpl.send() and make check before.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/Validators.java
+++ b/client/src/main/java/org/apache/rocketmq/client/Validators.java
@@ -74,6 +74,9 @@ public class Validators {
             throw new MQClientException(ResponseCode.MESSAGE_ILLEGAL,
                 "the message body size over max value, MAX: " + defaultMQProducer.getMaxMessageSize());
         }
+
+        // for support namespace
+        msg.setTopic(defaultMQProducer.withNamespace(msg.getTopic()));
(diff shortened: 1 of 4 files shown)


ferrirW:
Another simple way is move all msg.setTopic() to Validators.checkMessage(). Does it fit？

Git-Yang:
Validators.checkMessage is duplicated.

ni-ze:
add namespace into topic when check message is not a good idea.

ferrirW:
> Validators.checkMessage is duplicated.

We can see git history in above screenshot that duplicate checks are used to prevent Null point exception, and it seem's necessary.

ferrirW:
> add namespace into topic when check message is not a good idea.

yes, it's not best way, but is the simplest change, and I have coverd all invoked of checkMessage().

ferrirW:
> add namespace into topic when check message is not a good idea.

Whether put msg.setTopic after check message is a better way？

ni-ze:
> > add namespace into topic when check message is not a good idea.
> 
> Whether put msg.setTopic after check message is a better way？ @ni-ze

Do not modify it is better, I think.

PR #3963, opened 2022-03-12
[ISSUE #3964] Introduce jmh test for RemotingCommand

### LINKED ISSUE #3964: [Performance] Introduce jmh test for RemotingCommand

WJL3333:
fix #3964
introduce jmh test for RemotingComand decodeCommandCustomHeader

### DIFF
--- a/jmh/pom.xml
+++ b/jmh/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


RongtongJin (commented):
IMO, making jmh a separate module is not a good idea. Why not put it directly into the remoting module？

WJL3333:
> IMO, making jmh a separate module is not a good idea. Why not put it directly into the remoting module？

may be there will be other jmh for other module. the origin propose is for other people to introduce more performance relate test here. a separate module will just make dependency more clearly. may be some jmh test will depend two or more module . Maybe i can make this module not be released.

WJL3333:
close because pr #3965 contains the whole jmh test.

Issue #3973, opened 2022-03-14
[RIP-37] New and unified APIs


aaron-ai:
This issue aims to track the RIP-37, which try to unify a universal APIs spec for RocketMQ.

What can we benefit from this RIP? 
* The behavior of interfaces is more clearly defined, making it easier for newcomers to get started, and the behavior of various interfaces is more predictable.
* The new APIs unbinds the implementation and interface to realize real interface-oriented programming, and the upgrade cost for users will be lower in the future.
* The implementation of polyglot clients has always been a difficult problem, and the lack of a unified specification is one of the important reasons. New specifications allow them to be standardized so that users do not behave inconsistently by switching languages.

or you can view more details from the docs link: https://shimo.im/docs/m5kv92OeRRU8olqX

dugenkui03:
@aaron-ai Some optimized suggestion about api-module, please take a look.
1. Create masker interface `ErrorClassification`, and replace `ErrorCode` with `ErrorClassification` in `ClientException`. I think developer need custom error type, such as limitation;
2. Replace `CompletableFuture` with `Future` in `Producer#sendAsync`, since `Future` is more extensibie. the default implementation of `Producer#sendAsync` still can use `CompletableFuture`;
3. Replace  `BackoffRetryPolicy` with `RetryPolicy` in `ProducerBuilder#setRetryPolicy `.

#### Describe in Chinese
您好，关于 api-module 有些建议请评估是否采纳：
1. 创建标记接口`ErrorClassification`、并使`ErrorCode`实现该接口，`ClientException`中`ErrorCode`替换为`ErrorClassification`。因为我认为开发者经常需要自定义异常和异常类型，例如限流，而现在的接口设计其实也支持用户自定义异常；
2. 在`Producer#sendAsync`中使用`Future`替换`CompletableFuture`，因为前者已经可以满足接口需要，我们可以再具体实现中在返回`CompletableFuture`；
3. 在`ProducerBuilder#setRetryPolicy`中使用`RetryPolicy`替换 `BackoffRetryPolicy`。

aaron-ai:
> @aaron-ai Some optimized suggestion about api-module, please take a look.
> 
> 1. Create masker interface `ErrorClassification`, and replace `ErrorCode` with `ErrorClassification` in `ClientException`. I think developer need custom error type, such as limitation;
[...]

PR #3987, opened 2022-03-14
[RIP-37] Add new APIs for producer

aaron-ai:
As we mentioned in [RIP-37 New and unified APIs](https://shimo.im/docs/m5kv92OeRRU8olqX), establish a new APIs specifications. we divide it into two independent pull requests.

1. The producer part.
2. The consumer part.

this pull request is about the producer part.

related issue: #3973

### DIFF
--- a/apis/src/main/java/org/apache/rocketmq/apis/exception/AuthorisationException.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/exception/AuthorisationException.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.exception;
+
+import org.apache.rocketmq.apis.message.Message;
+import org.apache.rocketmq.apis.producer.Producer;
+
+/**
+ * The difference between {@link AuthenticationException} and {@link AuthorisationException} is that
+ * {@link AuthorisationException} here means current users don't have permission to do current operation.
+ *
+ * <p>For example, current user is forbidden to send message to this topic, {@link AuthorisationException} will be
+ * thrown in {@link Producer#send(Message)}.
+ */
+public class AuthorisationException extends ClientException {
+    public AuthorisationException(ErrorCode code, String message, String requestId) {
+        super(code, message);
+        putMetadata(REQUEST_ID_KEY, requestId);
+    }
+}
--- a/apis/src/main/java/org/apache/rocketmq/apis/exception/ClientException.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/exception/ClientException.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.exception;
+
+import com.google.common.base.MoreObjects;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Base exception for all exception raised in client, each exception should derive from current class.
+ * It should throw exception which is derived from {@link ClientException} rather than {@link ClientException} itself.
+ */
+public abstract class ClientException extends Exception {
+    /**
+     * For those {@link ClientException} along with a remote procedure call, request id could be used to track the
+     * request.
+     */
+    protected static final String REQUEST_ID_KEY = "request-id";
+
+    private final ErrorCode errorCode;
+    private final Map<String, String> context;
+
+    ClientException(ErrorCode errorCode, String message, Throwable cause) {
+        super(message, cause);
+        this.errorCode = errorCode;
+        this.context = new HashMap<>();
+    }
+
+    ClientException(ErrorCode errorCode, String message) {
+        super(message);
+        this.errorCode = errorCode;
+        this.context = new HashMap<>();
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    protected void putMetadata(String key, String value) {
+        context.put(key, value);
+    }
+
+    public Optional<String> getRequestId() {
+        final String requestId = context.get(REQUEST_ID_KEY);
+        return null == requestId ? Optional.empty() : Optional.of(requestId);
+    }
+
+    public ErrorCode getErrorCode() {
+        return errorCode;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(super.toString())
+            .add("errorCode", errorCode)
+            .add("context", context)
+            .toString();
+    }
+}
--- a/apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.message;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.rocketmq.apis.producer.Producer;
+
+/**
+ * Abstract message only used for {@link Producer}.
+ */
+public interface Message {
+    /**
+     * Get the topic of message, which is the first classifier for message.
+     *
+     * @return topic of message.
+     */
+    String getTopic();
+
+    /**
+     * Get the <strong>deep copy</strong> of message body.
+     *
+     * @return the <strong>deep copy</strong> of message body.
+     */
+    byte[] getBody();
+
+    /**
+     * Get the <strong>deep copy</strong> of message properties.
+     *
+     * @return the <strong>deep copy</strong> of message properties.
+     */
+    Map<String, String> getProperties();
+
+    /**
+     * Get the tag of message, which is the second classifier besides topic.
+     *
+     * @return the tag of message.
+     */
+    Optional<String> getTag();
+
+    /**
+     * Get the key collection of message.
+     *
+     * @return <strong>the key collection</strong> of message.
+     */
+    Collection<String> getKeys();
+
+    /**
+     * Get the message group, which make sense only when topic type is fifo.
+     *
+     * @return message group, which is optional.
+     */
+    Optional<String> getMessageGroup();
+
+    /**
+     * Get the expected delivery timestamp, which make sense only when topic type is delay.
+     *
+     * @return message expected delivery timestamp, which is optional.
+     */
+    Optional<Long> getDeliveryTimestamp();
+}
--- a/apis/src/main/java/org/apache/rocketmq/apis/message/MessageBuilder.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/message/MessageBuilder.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.message;
+
+import java.util.Collection;
+
+/**
+ * Builder to config {@link Message}.
+ */
+public interface MessageBuilder {
+    /**
+     * Set the topic for message.
+     *
+     * @param topic topic for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setTopic(String topic);
+
+    /**
+     * Set the body for message.
+     *
+     * @param body body for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setBody(byte[] body);
+
+    /**
+     * Set the tag for message.
+     *
+     * @param tag tag for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setTag(String tag);
+
+    /**
+     * Set the key for message.
+     *
+     * @param key key for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setKey(String key);
+
+    /**
+     * Set the key collection for message.
+     *
+     * @param keys key collection for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setKeys(Collection<String> keys);
+
+    /**
+     * Set the group for message.
+     *
+     * @param messageGroup group for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setMessageGroup(String messageGroup);
+
+    /**
+     * Add user property for message.
+     *
+     * @param key   single property key.
+     * @param value single property value.
+     * @return the message builder instance.
+     */
+    MessageBuilder addProperty(String key, String value);
+
+    /**
+     * Finalize the build of the {@link Message} instance.
+     *
+     * <p>Unique {@link MessageId} is generated after message building.</p>
+     *
+     * @return the message instance.
+     */
+    Message build();
+}
--- a/apis/src/main/java/org/apache/rocketmq/apis/message/MessageId.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/message/MessageId.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.message;
+
+/**
+ * Abstract message id, the implement must override {@link Object#toString()}, which indicates the message id using
+ * string form.
+ */
+public interface MessageId {
+    /**
+     * Get the version of message id.
+     *
+     * @return the version of message id.
+     */
+    MessageIdVersion getVersion();
+
+    /**
+     * The implementation <strong>must</strong> override this method, which indicates the message id using string form.
+     *
+     * @return string-formed string id.
+     */
+    String toString();
+}
--- a/apis/src/main/java/org/apache/rocketmq/apis/message/MessageIdVersion.java
+++ b/apis/src/main/java/org/apache/rocketmq/apis/message/MessageIdVersion.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.message;
+
+public enum MessageIdVersion {
+    /**
+     * V0 version, whose length is 32.
+     */
+    V0,
+    /**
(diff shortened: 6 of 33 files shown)


aaron-ai:
@zhouxinyu @lizhanhui @duhenglucky @RongtongJin @drpmma @chenzlalvin

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    Collection<String> getKeys();
+
+    /**
+     * Get the message group, which make sense only when topic type is fifo.
+     *
+     * @return message group, which is optional.
+     */
+    Optional<String> getMessageGroup();
+
```
dongeforever, line 70: This interface need a method getMessageQueue() too.
It may need to send message to the specified MessageQueue.
The case cannot be covered by getMessageGroup().

The priority of getMessageQueue()  is bigger than  getMessageGroup().

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,132 @@
...
+     * @throws AuthorisationException            if no permission to send message.
+     * @throws AuthenticationException           if identification could not be recognized by server.
+     * @throws ProducerClosedAlreadyException    if producer is closed already.
+     * @throws MessageTypeDoesNotMatchException  if message type does not match with the topic.
+     * @throws NetworkTimeoutException           if encountered network timeout to communicate with server.
+     * @throws NetworkConnectionException        if there is a network connection problem.
+     * @throws PersistenceException              if encountered persistence failure from server.
+     * @throws TransactionCheckerNotSetException if {@link TransactionChecker} is not set.
+     */
+    MessageId send(Message message, Transaction transaction) throws ClientException;
```
dongeforever, line 80: This is not the two-phase API!

How about like this :
try {
TransactionMark  mark = producer.prepareSend(Message message);
//do process
producer.commit(mark);
} catch(Exception e) {
  producer.rollback(mark);
}

Such style is easy to be integrated with db transaction. And the code will not be split.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,132 @@
...
+     * @throws TopicDoesNotExistException       if the topic of message does not exist.
+     * @throws AuthorisationException           if no permission to send message.
+     * @throws AuthenticationException          if identification could not be recognized by server.
+     * @throws ProducerClosedAlreadyException   if producer is closed already.
+     * @throws MessageTypeDoesNotMatchException if message type does not match with the topic.
+     * @throws NetworkTimeoutException          if encountered network timeout to communicate with server.
+     * @throws NetworkConnectionException       if there is a network connection problem.
+     * @throws PersistenceException             if encountered persistence failure from server.
+     */
+    Map<Message, MessageId> send(Collection<Message> messages) throws ClientException;
```
dongeforever, line 110: What will happen if I use a list with the same object?

The return type does not match the argument.

Maybe both using list is ok.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,132 @@
...
+     * @throws TopicDoesNotExistException       if the topic of message does not exist.
+     * @throws AuthorisationException           if no permission to send message.
+     * @throws AuthenticationException          if identification could not be recognized by server.
+     * @throws ProducerClosedAlreadyException   if producer is closed already.
+     * @throws MessageTypeDoesNotMatchException if message type does not match with the topic.
+     * @throws NetworkTimeoutException          if encountered network timeout to communicate with server.
+     * @throws NetworkConnectionException       if there is a network connection problem.
+     * @throws PersistenceException             if encountered persistence failure from server.
+     */
+    MessageId send(Message message) throws ClientException;
```
dongeforever, line 62: Where is the SendResult?

The client at least needs to know the offset and messagequeue for checking!

dongeforever (commented):
Except for the admin operation, the abilities of the new API should be a superset of the old API.

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
aaron-ai: Message here is for producer, user may build message before sending, so Message#getMessageQeue may be not appropriate, but we may consider to add it into the return value of producer sending method.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageView.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Get the message group, which is optional and only make sense only when topic type is fifo.
+     *
+     * @return message group, which is optional.
+     */
+    Optional<String> getMessageGroup();
+
+    /**
+     * Get the expected delivery timestamp, which make sense only when topic type id delay.
```
RongtongJin, line 104: 'when topic type is delay', not 'when topic type id delay'

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Get the expected delivery timestamp, which make sense only when topic type id delay.
```
RongtongJin, line 72: 'when topic type is delay', not 'when topic type id delay'

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
aaron-ai: Well, we may send more than one message in single transaction.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageIdVersion.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.message;
+
+public enum MessageIdVersion {
```
RongtongJin, line 20: Why are there multiple versions of messageId？

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
aaron-ai: Nice catch!

apis/src/main/java/org/apache/rocketmq/apis/message/MessageIdVersion.java
aaron-ai: Well, the old version of messageId could not meet the unique constraint, that's why we have the different version of messageId.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageView.java
aaron-ai: Fixed

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
aaron-ai: Fixed

apis/src/main/java/org/apache/rocketmq/apis/exception/AuthorisationException.java
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.exception;
+
+public class AuthorisationException extends ClientException {
```
WJL3333, line 20: can we add some doc for this exception? it seems the same as AuthenticationException from the naming

apis/src/main/java/org/apache/rocketmq/apis/exception/ClientException.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.apis.exception;
+
+public class ClientException extends Exception {
```
WJL3333, line 20: it seems this is the root Exception for all the rocketmq client exception. when we will throw this exception.
if maintainer want to add an exception which kind of exception can extend this clientException.

add java doc will be more formal

apis/src/main/java/org/apache/rocketmq/apis/message/MessageBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+ * Builder to config {@link Message}.
+ */
+public interface MessageBuilder {
+    /**
+     * Set the topic for message.
+     *
+     * @param topic topic for the message.
+     * @return the message builder instance.
+     */
+    MessageBuilder setTopic(String topic);
```
WJL3333, line 34: if we remove set prefix the api will be more clean to the developer.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageId.java
```diff
@@ -0,0 +1,38 @@
...
+     * @return the version of message id.
+     */
+    MessageIdVersion getVersion();
+
+    /**
+     * The implementation <strong>must</strong> override this method, which indicates the message id using string form.
+     *
+     * @return string-formed string id.
+     */
+    String toString();
```
WJL3333, line 37: this name is the same as the java origin Object.toString. i think we can change the method name. the origin is anti-parttern naming.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageView.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package org.apache.rocketmq.apis.message;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.rocketmq.apis.MessageQueue;
+
+public interface MessageView {
```
WJL3333, line 25: need java doc here. what is an MessageView ?

apis/src/main/java/org/apache/rocketmq/apis/message/MessageView.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    long getOffset();
+
+    /**
+     * Get the <strong>deep copy</strong> of message body, which makes the modification of return value does not
+     * affect the message itself.
+     *
+     * @return the <strong>deep copy</strong> of message body.
+     */
+    byte[] getBody();
```
WJL3333, line 58: deep copy seems will harm the performance. but there seems no better return type here.

WJL3333: how about ByteBuffer here which can be set as unmodifable

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Begin transaction, which follows the
+     * <a href="https://en.wikipedia.org/wiki/Two-phase_commit_protocol">two-phase commit protocol</a>.
+     *
+     * @return a transaction entity to execute commit/rollback operation.
+     * @throws TransactionCheckerNotSetException if {@link TransactionChecker} is not set.
+     * @throws ProducerClosedAlreadyException    if producer is closed already.
+     */
+    Transaction beginTransaction() throws ClientException;
```
WJL3333, line 120: maybe an interface extand producer add some Transaction relate method will be better.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Close the producer and release all related resources.
+     *
+     * <p>This method does not return until all related resource is released. Once producer is closed, <strong>it could
+     * not be started once again.</strong> we maintained an FSM (finite-state machine) to record the different states
+     * for each producer, which is similar to {@link Service.State}.
+     */
+    @SuppressWarnings("UnstableApiUsage")
+    @Override
+    void close();
```
WJL3333, line 131: will close throw exception here?

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+ * Builder to config and start {@link Producer}.
+ */
+public interface ProducerBuilder {
+    /**
+     * Set the client configuration for producer.
+     *
+     * @param clientConfiguration client's configuration.
+     * @return the producer builder instance.
+     */
+    ProducerBuilder setClientConfiguration(ClientConfiguration clientConfiguration);
```
WJL3333, line 40: same as Messagebuilder. no set prefix will make api more clean in fluent api call.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+     * <p>Especially, if this method is invoked more than once, different producer will be created and started.
+     *
+     * @return the producer instance.
+     * @throws TopicDoesNotExistException if there are nonexistent topic(s).
+     * @throws AuthorisationException     if no permission to communicate with server.
+     * @throws AuthenticationException    if identification could not be recognized by server.
+     * @throws NetworkTimeoutException    if encountered network timeout to communicate with server.
+     * @throws NetworkConnectionException if there is a network connection problem.
+     */
+    Producer start() throws ClientException;
```
WJL3333, line 91: may be move start to the producer.
some case will just build the producer without start.

apis/src/main/java/org/apache/rocketmq/apis/producer/Transaction.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public interface Transaction {
+    /**
+     * Try to commit the transaction, which would expose the message before the transaction is closed.
+     *
+     * <p>We don't ensure this operation is successful even though no exception is thrown after invocation,
+     * <strong>actually we omit the exception on purpose because {@link TransactionChecker} is the unique right way
+     * to solve the suspended transaction rather than commit or roll-back repeatedly.</strong>
+     */
+    void commit();
```
WJL3333, line 40: will commit throw exception need the caller process?

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    ProducerBuilder setAsyncThreadCount(int count);
+
+    /**
+     * Set the retry policy to send message.
+     *
+     * @param retryPolicy policy to re-send message when failure encountered.
+     * @return the producer builder instance.
+     */
+    ProducerBuilder setRetryPolicy(BackoffRetryPolicy retryPolicy);
```
WJL3333, line 66: `RetryPolicy` interface here will be better.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
aaron-ai: Actually producerBuilder#start could avoid to start producer repeatedly. What's more, a producer which is not started does not make sense.

apis/src/main/java/org/apache/rocketmq/apis/exception/AuthorisationException.java
aaron-ai: More javadocs will be added here.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
aaron-ai: Yeah, this should be considered, we think that it is more strict if we use `BackoffRetryPolicy` here before we allow users to implement their `RetryPolicy`.

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
dongeforever: I repeat this:
Except for the admin operation, the abilities of the new API should be a superset of the old API.

The old API could send messages to the specified message queue?

So how to do it in the new API?

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
dongeforever: How to send more than one message using such API?
MessageId send(Message message, Transaction transaction)


The old API  for the transaction is not naturally two-phase based. It's better to polish it in the new API.
But currently, the new API is just like the old

apis/src/main/java/org/apache/rocketmq/apis/message/MessageId.java
aaron-ai: `#toString()` here aims to emphasize that the implementation must override this method, maybe there is a better solution?

apis/src/main/java/org/apache/rocketmq/apis/exception/AuthorisationException.java
aaron-ai: > More javadocs will be added here.

done

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
aaron-ai: The code example has been added to the javadocs.

aaron-ai: > How to send more than one message using such API? MessageId send(Message message, Transaction transaction)
> 
> The old API for the transaction is not naturally two-phase based. It's better to polish it in the new API. But currently, the new API is just like the old

The code example has been added to the javadocs.

apis/src/main/java/org/apache/rocketmq/apis/exception/ClientException.java
aaron-ai: Yes, we add `abstract` qualifier into `ClientException`, which makes it could be constructed directly, and java docs also has been added.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageView.java
aaron-ai: > how about ByteBuffer here which can be set as unmodifable

Nice suggestion.

apis/src/main/java/org/apache/rocketmq/apis/message/MessageBuilder.java
aaron-ai: Well, setter/getter style is more common in java, `MessageBuilder#topic()` or `MessageBuilder#topic(string topicName)` may be more common in cpp.

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
zhouxinyu:
```
the abilities of the new API should be a superset of the old API
```
We don't follow this rule, many useless or low-frequency abilities will be removed in the new APIs, the new APIs don't expand the old APIs.

As for sending messages to a specific queue, we were going to add it in the future version. Consider `MessageQueue` is a major model of RocketMQ, we may need to add it in the first version. But, I am not sure it's the right choice, hope more contributors could express opinions on this topic, especially from user perspectives.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
zhouxinyu: We can find many fluent builder samples in JDK which use the setter pattern, such as https://docs.oracle.com/javase/7/docs/api/java/util/Locale.Builder.html

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.rocketmq.apis.exception.ClientException;
+import org.apache.rocketmq.apis.exception.NetworkConnectionException;
+import org.apache.rocketmq.apis.exception.NetworkTimeoutException;
+import org.apache.rocketmq.apis.exception.TopicDoesNotExistException;
+import org.apache.rocketmq.apis.message.Message;
+
+/**
+ * Builder to config and start {@link Producer}.
+ */
+public interface ProducerBuilder {
```
zhouxinyu, line 33: Do we have a chance to force our developers to set the required fields when building the `Producer` or `Message`?

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
zhouxinyu: Could you please provide some code samples? @WJL3333

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
zhouxinyu: It's strange if we don't have a `build` method in a builder interface.

If we only provide a `start` method in the producer builder, that means all the producers are started by default, so why we should have a `start`?

apis/src/main/java/org/apache/rocketmq/apis/producer/Transaction.java
zhouxinyu: `TransactionChecker` is less real-time, which may result in high latency when the commit failed.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
zhouxinyu: @WJL3333 Lombok doesn't support prefixed builders at the beginning, but this project resolved this issue later, please refer to https://github.com/projectlombok/lombok/issues/1805#issuecomment-465140144, there are many reasons support we use prefixed builders.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
chenzlalvin: Builder should throw null field check exception when call build interface.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,140 @@
...
+     * @throws AuthorisationException            if no permission to send message.
+     * @throws AuthenticationException           if identification could not be recognized by server.
+     * @throws ProducerClosedAlreadyException    if producer is closed already.
+     * @throws MessageTypeDoesNotMatchException  if message type does not match with the topic.
+     * @throws NetworkTimeoutException           if encountered network timeout to communicate with server.
+     * @throws NetworkConnectionException        if there is a network connection problem.
+     * @throws PersistenceException              if encountered persistence failure from server.
+     * @throws TransactionCheckerNotSetException if {@link TransactionChecker} is not set.
+     */
+    MessageView send(Message message, Transaction transaction) throws ClientException;
```
zhouxinyu, line 79: Tons of exceptions are thrown, it's daunting.

I suggest we only declare a ClientException with managed code, request-id, message, etc.

These exceptions are unchecked, which means we don't want our users to use these exceptions to control their logic, so a simple ClientException is enough.

apis/src/main/java/org/apache/rocketmq/apis/producer/ProducerBuilder.java
aaron-ai: > It's strange if we don't have a `build` method in a builder interface.
> 
> If we only provide a `start` method in the producer builder, that means all the producers are started by default, so why we should have a `start`?

I agree with the opinon that lack of `#build()` may be a little bit wireld, let's look at it another way. Is it unnecessary for the user to start this process? If so, then I think it's just a naming issue.

apis/src/main/java/org/apache/rocketmq/apis/message/Message.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return message group, which is optional.
+     */
+    Optional<String> getMessageGroup();
+
+    /**
+     * Get the expected delivery timestamp, which make sense only when topic type is delay.
+     *
+     * @return message expected delivery timestamp, which is optional.
+     */
+    Optional<Long> getDeliveryTimestamp();
```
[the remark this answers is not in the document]
aaron-ai: > Is there a corresponding setter for deliverytime ? Why is this property exposed separately？

I might lose setter for `deliveryTimestamp` in a commit, thinks for reminding me.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
aaron-ai: > These exceptions are unchecked, which means we don't want our users to use these exceptions to control their logic, so a simple ClientException is enough.

First of all, exceptions are inevitable. In fact, about the issue that whether to use checked exception or not, maybe no need to prevent users to do responding operation by the exception thrown deliberately? For example, users just want to count how many times the network timeout exception occurred. .

In addition, about the issue to use error codes or specific exception, actuall they can all solve the problem in theory., here's my concern
 1. It is very easy to abuse error code.
 2. Not all exceptions will have the same parameters, such as requestId, they may only exist in some exceptions.

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Begins a transaction.
+     *
+     * <p>For example:
+     *
+     * <pre>{@code
+     * Transaction transaction = producer.beginTransaction();
+     * MessageView messageView1 = producer.send(message1, transaction);
+     * MessageView messageView2 = producer.send(message2, transaction);
```
zhouxinyu, line 84: Return SendReceipt

zhouxinyu (changes requested):

apis/src/main/java/org/apache/rocketmq/apis/producer/Producer.java
aaron-ai: Done

Issue #3996, opened 2022-03-17
fix the acl content of TopicRequestHeader


chaiyx:
**BUG REPORT**

1. Please describe the issue you observed:

- for TopicRequestHeader, the acl content needs to contain all fields of parent class.

PR #3997, opened 2022-03-17
[ISSUE #3996] fix the acl content of TopicRequestHeader

chaiyx:
#3996

### DIFF
--- a/acl/src/main/java/org/apache/rocketmq/acl/common/AclClientRPCHook.java
+++ b/acl/src/main/java/org/apache/rocketmq/acl/common/AclClientRPCHook.java
@@ -17,6 +17,9 @@
 package org.apache.rocketmq.acl.common;
 
 import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


acl/src/main/java/org/apache/rocketmq/acl/common/AclClientRPCHook.java
```diff
@@ -69,7 +72,11 @@ public void doAfterResponse(String remoteAddr, RemotingCommand request, Remoting
             if (null != header) {
                 Field[] fields = fieldCache.get(header.getClass());
                 if (null == fields) {
-                    fields = header.getClass().getDeclaredFields();
+                    Set<Field> fieldList = new HashSet<Field>();
```
caigy, line 75: `parseRequestContent` is used for signature checks, so the order of fields matters. I think a sorted set should be adopted here.

caigy (changes requested):

acl/src/main/java/org/apache/rocketmq/acl/common/AclClientRPCHook.java
chaiyx: at the end of the function, the value of the header will be put into a SortedMap. so it doesn't matter to use HashSet here.

caigy: But the order of fields put into returning SortedMap still depends on the order of `fieldList`.

RongtongJin:
Hi @chaiyx  Could you resubmit your PR to the 5.0.0-beta branch because 5.0.0-alpha already release?

Issue #3933, opened 2022-03-06
Run request reply in the MAC os12 system, and the consumer will time out when returning the result


luky116:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do (The steps to reproduce)?
  Run request reply in the MAC os12 system, and the consumer will time out when returning the result

- What is expected to see?
The consumer can normally return the result to generate this

- What did you see instead?
The consumer reply failed
![image](https://user-images.githubusercontent.com/38887641/156917706-353ea965-f5ee-4ed2-b66d-d38ac0236a10.png)

3. Please tell us about your environment:
    MAC OS 12

5. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):
I ran the following example：
https://github.com/apache/rocketmq/tree/develop/example/src/main/java/org/apache/rocketmq/example/rpc

jasondeng1997:
我也看看

jasondeng1997:
![image](https://user-images.githubusercontent.com/76831112/156918084-30c71280-31f9-4f68-8bac-bd1cc0c364f5.png)

jasondeng1997:
改成配置的服务器，也会报链接为空失败，是我开了vpn吗

jasondeng1997:
vpn关掉也有这个问题

drgnchan:
did you make a breakpoint at `ResponseConsumer.java`?

guyinyou:
There may be a problem with my computer itself, my error is not the same as yours. 
Only throw timeout and no other information.
<img width="1177" alt="FCB6D747-FB2B-4ED9-A33F-7A32F3B569D0" src="https://user-images.githubusercontent.com/36399867/156984215-b213111a-6038-4d5d-87d6-b69a20facad5.png">

maxid:
broker.conf need add brokerIP setting
brokerIP1=公网IP(all consumer will use brokerIP1)
brokerIP2=内网IP
hope rocketmq team can support rocket-client  use ip1 or ip2 option

ni-ze:
I reproduce this bug, root cause is blocked when broker send message to producer. The blocking code is as fellow:
![image](https://user-images.githubusercontent.com/31175234/156995063-bbc550e3-1c30-4719-8b1a-c2e3125f9f78.png)
There are relationships with bornHost and storeHost of reply message.
```xml
bornHost=/127.0.0.1:59653, storeHost=/127.0.0.1:10911
``` 
will be blocked.

```xml
[...]

PR #4010, opened 2022-03-20
fix mac 12+ slow bug

luky116:

When the MAC 12 system uses the request reply function, the response time of the producer is very long,  so the response time should be optimized

## ISSUE
#3933

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/ReplyMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/ReplyMessageProcessor.java
@@ -45,6 +45,7 @@ import org.apache.rocketmq.store.MessageExtBrokerInner;
 import org.apache.rocketmq.store.PutMessageResult;
 import org.apache.rocketmq.store.stats.BrokerStatsManager;
 
+import java.net.InetSocketAddress;
 import java.util.concurrent.ThreadLocalRandom;
 
 public class ReplyMessageProcessor extends AbstractSendMessageProcessor implements NettyRequestProcessor {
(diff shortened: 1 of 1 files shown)


ni-ze (approved):

broker/src/main/java/org/apache/rocketmq/broker/processor/ReplyMessageProcessor.java
```diff
@@ -157,8 +158,10 @@ private PushReplyResult pushReplyMessage(final ChannelHandlerContext ctx,
         final SendMessageRequestHeader requestHeader,
         final Message msg) {
         ReplyMessageRequestHeader replyMessageRequestHeader = new ReplyMessageRequestHeader();
-        replyMessageRequestHeader.setBornHost(ctx.channel().remoteAddress().toString());
-        replyMessageRequestHeader.setStoreHost(this.getStoreHost().toString());
+        InetSocketAddress bornAddress =(InetSocketAddress)(ctx.channel().remoteAddress());
+        replyMessageRequestHeader.setBornHost(bornAddress.getHostName() + ":" + bornAddress.getPort());
```
caigy, line 162: Can we use `getHostString()` here to skip reverse lookup?

tianliuliu (approved):

broker/src/main/java/org/apache/rocketmq/broker/processor/ReplyMessageProcessor.java
```diff
@@ -157,8 +158,10 @@ private PushReplyResult pushReplyMessage(final ChannelHandlerContext ctx,
         final SendMessageRequestHeader requestHeader,
         final Message msg) {
         ReplyMessageRequestHeader replyMessageRequestHeader = new ReplyMessageRequestHeader();
-        replyMessageRequestHeader.setBornHost(ctx.channel().remoteAddress().toString());
-        replyMessageRequestHeader.setStoreHost(this.getStoreHost().toString());
+        InetSocketAddress bornAddress =(InetSocketAddress)(ctx.channel().remoteAddress());
```
yuz10, line 161: Please fix code style to make sure code check pass

yuz10 (approved):

broker/src/main/java/org/apache/rocketmq/broker/processor/ReplyMessageProcessor.java
luky116: I'm afraid not. If the service is in the form of IP address + port, the gethoststring method will lose port

duhenglucky (approved):

PR #4024, opened 2022-03-23
Replace Charset.forName("UTF-8") with StandardCharsets.UTF_8

6U-U9:
Constant charset defined by String (for example, "UTF-8") can be replaced with the predefined StandardCharsets.UTF_8 code.
The code after the fix may work faster, because the charset lookup becomes unnecessary.

## Brief changelog
Replace Charset.forName("UTF-8") with StandardCharsets.UTF_8


Git-Yang:
Can an ISSUE be created for this PR? 
In addition, can the code formatting of SimpleProducer be adjusted?

6U-U9:
#4068 I have created an issue for this PR.
And accoding to the massage sending process in SimplePullConsumer, there is no need for specify Charset in SimpleProducer, so I removed them.

example/src/main/java/org/apache/rocketmq/example/openmessaging/SimpleProducer.java
```diff
@@ -40,15 +39,15 @@ public static void main(String[] args) {
         System.out.printf("Producer startup OK%n");
 
         {
-            Message message = producer.createBytesMessage("OMS_HELLO_TOPIC", "OMS_HELLO_BODY".getBytes(Charset.forName("UTF-8")));
```
lizhiboo, line 43: Why not use StandardCharsets.UTF_8 ? IMO, some os may use gbk as default decoder.

6U-U9: I don't think it's necessary to specify the encode charset for message body.
As long as one encode and decode the message with the same charset, he shall get the correct message.

yuz10 (approved):

Issue #4114, opened 2022-04-04
Add threadPool in DefaultMQAdminExtImpl and execute queryConsumeTimeSpan concurrently 


MatrixHB:
We use the API `queryConsumeTimeSpan` provided by DefaultMQAdminExt when querying message within a certain time period on a user console. We found that the API `queryConsumeTimeSpan` takes a long time, which cause querying message timeout all the time.
 
The reason is that our broker cluster is relatively large, including over 100 brokers. In the implement of `queryConsumeTimeSpan`, the client need to request all the brokers to fetch consumeTimeSpan information, and the request-reply are serial. 
Original Code:
```
        TopicRouteData topicRouteData = this.examineTopicRouteInfo(topic);
        for (BrokerData bd : topicRouteData.getBrokerDatas()) {
            String addr = bd.selectBrokerAddr();
            if (addr != null) {
                spanSet.addAll(this.mqClientInstance.getMQClientAPIImpl().queryConsumeTimeSpan(addr, topic, group, timeoutMillis));
            }
        }
```

In order to improve the performce, we try to add thread pool with 10 core threads in `DefaultMQAdminExtImpl`, and execute queryConsumeTimeSpan concurrently using CompletableFuture. This thread pool can be shared by other APIs in DefaultMQAdminExt.

PR #4115, opened 2022-04-04
[ISSUE #4114] Add threadPool in DefaultMQAdminExtImpl and execute queryConsumeTimeSpan concurrently

MatrixHB:
To solve the problem of [ISSUE #4114](https://github.com/apache/rocketmq/issues/4114)

execute `queryConsumeTimeSpan` and `examineTopicStats` concurrently, which are used when query messages in certain time span.


tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
```diff
@@ -147,6 +157,9 @@ public void start() throws MQClientException {
 
                 mqClientInstance.start();
 
+                executors = new ThreadPoolExecutor(10, 20, 1000 * 30, TimeUnit.MILLISECONDS,
+                    new ArrayBlockingQueue<>(100), new ThreadFactoryImpl("MQAdminThread_"), new ThreadPoolExecutor.CallerRunsPolicy());
```
dugenkui03, line 161: I wonder why not 'CachedThreadPool', like `new ThreadPoolExecutor(0, Integer.MAX_VALUE, 1000 * 30, TimeUnit.MILLISECONDS...`.

MatrixHB: Good suggestion. Actually this thread pool is idle most of the time...

dugenkui03:
Hi, If this change is really accept by rocketmq-team. I suggest create a common method, which can do work concurrently and **throw the original Exception**. Details see review comment.

MatrixHB:
@dugenkui03
 I‘ve just found that in the `5.0.0-beta` branch provided by rocketmq-team, there is already `queryConsumeTimeSpanConcurrent` method to solve my problem.
But the `queryConsumeTimeSpanConcurrent` added in 5.0.0-beta has some problems:
1、Using countDownLatch, `spanSet.addAll()` executed in each thread is not thread-safe, because spanSet is not copy-on-write
2、IMO, add concurrent method separately makes DefaultMQAdminExtImpl method-expansion, not so elegant.
3、The thread pool parameter setting can be optimized.

tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
```diff
@@ -147,6 +157,8 @@ public void start() throws MQClientException {
 
                 mqClientInstance.start();
 
+                executors = Executors.newCachedThreadPool(new ThreadFactoryImpl("MQAdminThread_"));
```
lizhiboo, line 160: IMO，it's better to specify pool core size, max size, length of waiting list and reject conduct.

lwclover:
Wouldn't it be better to default to a thread pool?

Issue #4109, opened 2022-04-03
Updating broker write permission cannot take effect on broker side 


MatrixHB:
We encountered such a case about broker write permission.
When we execute the conmmand `sh mqadmin updateBrokerConfig -b {brokerIP}:10911  -k brokerPermission -v 4`, the Intps of this broker is expected to reduce to 0, but it didn't. 
The possible reasons for this problem are:
1、re-registerBroker to NameServer failed or timeout when executing the `UpdateBrokerConfigSubCommand`
2、exception occurs when `updateTopicRouteInfoFromNameServer` on client side （Maybe lower version of client code hides bug）
3、the client uses the send-message API selecting specific queue, not affected by the new routeInfo. 

Then we found that there is no validation about topic write permission on broker side. The validation and interception is for order topic. 
code： `org.apache.rocketmq.broker.processor.AbstractSendMessageProcessor#msgCheck`
```
if (!PermName.isWriteable(this.brokerController.getBrokerConfig().getBrokerPermission())
          && this.brokerController.getTopicConfigManager().isOrderTopic(requestHeader.getTopic())) {
          response.setCode(ResponseCode.NO_PERMISSION);
          response.setRemark("the broker[" + this.brokerController.getBrokerConfig().getBrokerIP1()
              + "] sending message is forbidden");
          return response;
}
```
Is it necessary to add validation for ordinary topic ?  The validation on broker side will cause send message error with "NO_PERMISSION" message, which means client needs to be aware of this exception.

MatrixHB:
> Use topicRoute to check the real topic route in nameserver after updateBrokerConfig.
> 
> If the client sends with the specific queues, such as send(msg, messagequeue)， the broker tps will not reduce to 0. Otherwise, the client will refresh the new topic route and not send message to the read-only broker.

It is very strange that the result of `topicRoute` is updated and correct, but the producer can still send to the forbidden broker, it seems like the client didn't refresh the new topic route.  This phenomenon lasts a few days. 
The rocketmq client version is 4_1_7.

lwclover:
删除broker写权限建议使用mqadmin中的wipeWritePerm子命令，这个命令是直接修改namesrv上的权限值，client会立刻感知到，broker重启恢复。
You are advised to use the wipeWritePerm subcommand in mqadmin to delete broker write permissions. This command directly changes the permission value on namesrv, which the client will immediately perceive。

ni-ze:
[...]

PR #4121, opened 2022-04-06
[ISSUE #4109] add permission validation on broker side for ordinary topic

MatrixHB:
to solve the problem described in [ISSUE #4109](https://github.com/apache/rocketmq/issues/4109)

This change will cause the broker to throw a "NO_PERMISSION" exception. If the producer uses SYNC mode, it will retry sending to another broker, seeing [DefaultMQProducerImpl#sendDefaultImpl](https://github.com/apache/rocketmq/blob/develop/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java#L562-L627),  so the SYNC mode will not be affected.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AbstractSendMessageProcessor.java
@@ -165,8 +165,7 @@ public abstract class AbstractSendMessageProcessor extends AsyncNettyRequestProc
 
     protected RemotingCommand msgCheck(final ChannelHandlerContext ctx,
         final SendMessageRequestHeader requestHeader, final RemotingCommand response) {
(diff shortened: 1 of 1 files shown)


dongeforever:
Adding the perm check for ordinary topic, the client will get exceptions and cannot failover during the matadata discovery time, which is up to 30 seconds.

For ordinary topics, the original design expects the client to discover the new topic route and send only to the writable brokers, which introduces no exceptions and makes the wipe-perm process more smoothly.

MatrixHB:
> Adding the perm check for ordinary topic, the client will get exceptions and cannot failover during the matadata discovery time, which is up to 30 seconds.
> 
> For ordinary topics, the original design expects the client to discover the new topic route and send only to the writable brokers, which introduces no exceptions and makes the wipe-perm process more smoothly.

Thanks, this change is not so elegant actually. Maybe updating the client version is enough to solve this problem. Next time we meet this trouble we can discuss this again.

Issue #4175, opened 2022-04-15
UpdateTopic from slave broker is not prevented


HScarb:
1. Please describe the issue you observed:

- What did you do (The steps to reproduce)?

Call mqadmin updateTopic and use -b to assign to a slave broker

- What is expected to see?

error message

- What did you see instead?

the command execute success, but topic not created

2. Please tell us about your environment:

3. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

The slave will periodically pull topic and other metadata from master, so updateTopic on slave make no sense

PR #4176, opened 2022-04-15
[ISSUE #4175] Prevent update topic on slave

HScarb:

#4175 

## Brief changelog

When modify topic from slave, return a error response

## Verifying this change

image
image
image
image

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -685,6 +685,9 @@ public class AdminBrokerProcessor extends AsyncNettyRequestProcessor implements 
     private RemotingCommand updateAndCreateSubscriptionGroup(ChannelHandlerContext ctx, RemotingCommand request)
         throws RemotingCommandException {
         final RemotingCommand response = RemotingCommand.createResponseCommand(null);
+        if (validateSlave(response)) {
+            return response;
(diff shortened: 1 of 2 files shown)


yuz10 (approved):

yuz10:
Create subscription group on slave have the same problem, would you correct that?

RongtongJin (approved):

HScarb:
> Create subscription group on slave have the same problem, would you correct that?

I would like to correct that

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -1636,4 +1649,14 @@ private MessageExtBrokerInner toMessageExtBrokerInner(MessageExt msgExt) {
         inner.setWaitStoreMsgOK(false);
         return inner;
     }
+
+    private boolean validateSlave(RemotingCommand response) {
```
ltamber, line 1653: I think the method name `validateSlave` meaning only check but not modify, maybe we should use a name mean `adjust response if slave`

ni-ze:
@HScarb Travis CI failed.

panzhi33 (approved):

duhenglucky (approved):

absolute8511:
Does it need to valid delete action on slave? It may cause the topic remain undeleted on slave if we do this. And by the way why it is not merge into the newest 5.1?

Issue #4154, opened 2022-04-12
maybe there is a Thread Safe problem when send a message in the  'MQFaultStrategy#selectOneMessageQueue' code


mz0113:
![1649761765(1)](https://user-images.githubusercontent.com/42512469/162947438-82a56beb-a899-4022-a7ee-17f44a65d154.png)

mq.setBrokerName(notBestBroker);
mq.setQueueId(tpInfo.getSendWhichQueue().getAndIncrement() % writeQueueNums);

this method of selectOneMessageQueue is executed by multiThread when send a message but the mq (messageQueue) is a shared object , thread A modify it`s brokerName to brokerB,but have not modify it queueId, still 1 , then thread B send msg and use this queue,but now,this queue `s brokerName is brokerB but it`s queueId is 1, in this case , wo can`t sure that brokerB really have queue which queueId is 1

Isn't that really a problem?

lwclover:
> ![1649761765(1)](https://user-images.githubusercontent.com/42512469/162947438-82a56beb-a899-4022-a7ee-17f44a65d154.png)
> 
> mq.setBrokerName(notBestBroker); mq.setQueueId(tpInfo.getSendWhichQueue().getAndIncrement() % writeQueueNums);
> 
> this method of selectOneMessageQueue is executed by multiThread when send a message but the mq (messageQueue) is a shared object , thread A modify it`s brokerName to brokerB,but have not modify it queueId, still 1 , then thread B send msg and use this queue,but now,this queue `s brokerName is brokerB but it`s queueId is 1, in this case , wo can`t sure that brokerB really have queue which queueId is 1
> 
> Isn't that really a problem?

I think you are right, to be more precise, constructors should be used.

ni-ze:
I do not think so.
```java
tpInfo.getSendWhichQueue().incrementAndGet()
```
the index is a thread local variable, would change with thread.if thread switched, index will change.

lwclover:
> I do not think so.
> 
> ```java
> tpInfo.getSendWhichQueue().incrementAndGet()
> ```
> 
> the index is a thread local variable, would change with thread.if thread switched, index will change.

There's a mod

mz0113:
> I do not think so.
> 
> ```java
> tpInfo.getSendWhichQueue().incrementAndGet()
> ```
> 
> the index is a thread local variable, would change with thread.if thread switched, index will change.

It is precisely because the index has changed that other threads have the opportunity to use this queue

[...]

PR #4179, opened 2022-04-17
[ISSUE #4154] fix the thread-safe problem in `selectOneMessageQueue`

MatrixHB:
[#4154 ](https://github.com/apache/rocketmq/issues/4154)

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java
+++ b/client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java
@@ -69,16 +69,17 @@ public class MQFaultStrategy {
                 }
 
                 final String notBestBroker = latencyFaultTolerance.pickOneAtLeast();
-                int writeQueueNums = tpInfo.getQueueIdByBroker(notBestBroker);
-                if (writeQueueNums > 0) {
-                    final MessageQueue mq = tpInfo.selectOneMessageQueue();
-                    if (notBestBroker != null) {
+                if (notBestBroker != null) {
+                    int writeQueueNums = tpInfo.getQueueIdByBroker(notBestBroker);
+                    if (writeQueueNums > 0) {
+                        MessageQueue mq = new MessageQueue();
+                        mq.setTopic(tpInfo.getMessageQueueList().get(0).getTopic());
                         mq.setBrokerName(notBestBroker);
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java
```diff
@@ -69,16 +69,17 @@ public MessageQueue selectOneMessageQueue(final TopicPublishInfo tpInfo, final S
...
                 final String notBestBroker = latencyFaultTolerance.pickOneAtLeast();
-                int writeQueueNums = tpInfo.getQueueIdByBroker(notBestBroker);
-                if (writeQueueNums > 0) {
-                    final MessageQueue mq = tpInfo.selectOneMessageQueue();
-                    if (notBestBroker != null) {
+                if (notBestBroker != null) {
+                    int writeQueueNums = tpInfo.getQueueIdByBroker(notBestBroker);
+                    if (writeQueueNums > 0) {
+                        MessageQueue mq = new MessageQueue();
+                        mq.setTopic(tpInfo.getMessageQueueList().get(0).getTopic());
```
hzh0425, line 76: Why use Topic0 directly here?

tsunghanjacktsai: @MatrixHB This. Same confusion here. Could you please explain why directly apply topic 0 here?

mz0113: > @MatrixHB Second this. Same confusion here. Could you please explain why directly apply topic 0 here?
the topic of all messageQueue in the tpInfo.getMessageQueueList() is same,so you can choose any one,but if you get(1,2,3,4,....),may cause arrayIndexOutOfboundsException,so we get(0),if the size() ==0,it would not go into this method

mz0113: > Why use Topic0 directly here?

回复错人了我晕，我说这么久还没合进去。这个是因为  tpInfo.getMessageQueueList() 的queue全部都是同一个topic下的，所以你取这个list的哪一个元素都行，直接取0是因为代码在这个位置可以明确这个list至少有一个元素在里面，所以取0不会导致数组越界，写其他的值可能导致越界。

mz0113 (commented):
没问题

PR #4218, opened 2022-04-27
delete useless code

alaric27:
These codes confuse me，I think these code is useless


broker/src/main/java/org/apache/rocketmq/broker/processor/PullMessageProcessor.java
```diff
@@ -246,12 +246,6 @@ private RemotingCommand processRequest(final Channel channel, RemotingCommand re
             responseHeader.setMinOffset(getMessageResult.getMinOffset());
             responseHeader.setMaxOffset(getMessageResult.getMaxOffset());
 
-            if (getMessageResult.isSuggestPullingFromSlave()) {
```
hzh0425, line 249: why it useless?

hzh0425: This is the code used to relieve the pressure of the Master

alaric27: I think the following code covers this logic
`
            if (this.brokerController.getBrokerConfig().isSlaveReadEnable()) {
                // consume too slow ,redirect to another machine
                if (getMessageResult.isSuggestPullingFromSlave()) {
                    responseHeader.setSuggestWhichBrokerId(subscriptionGroupConfig.getWhichBrokerWhenConsumeSlowly());
                }
                // consume ok
                else {
                    responseHeader.setSuggestWhichBrokerId(subscriptionGroupConfig.getBrokerId());
                }
            } else {
                responseHeader.setSuggestWhichBrokerId(MixAll.MASTER_ID);
            }
`

hzh0425: The logic is not the same

ni-ze:
It does look repeated.
If this.brokerController.getBrokerConfig().isSlaveReadEnable() equals false, suggest broker is always master.

```java
          if (getMessageResult.isSuggestPullingFromSlave()) {
                responseHeader.setSuggestWhichBrokerId(subscriptionGroupConfig.getWhichBrokerWhenConsumeSlowly());
            } else {
                responseHeader.setSuggestWhichBrokerId(MixAll.MASTER_ID);
            }
```
Result of the above code is covered.

If slaveReadEnable equals true, the result of latter code part has also cover the above.

tsunghanjacktsai (approved):
Both of them do have a duplicate effect in the end actually. I believe we could remove one of them.

broker/src/main/java/org/apache/rocketmq/broker/processor/PullMessageProcessor.java
tsunghanjacktsai: @hzh0425 I believe it's fine to remove it 'cause they do have the same effect in the end, but welcome to provide your thoughts on this.

Issue #4265, opened 2022-05-09
Query message get wrong result when tls enabled


zhanxuefeng:
The tls is enabled in my rocketmq cluster.
The QueryMsgTraceById throw a MQClientException(Code: 208, DESC: query message by key finished, but no message), in fact, the message I queried has a sub trace and a pub trace.

I notice that the QueryMessageTransfer calcute the transferred bytes depends on the return value of target.write method, but the target object defined in FileRegionEncoder return the wrong value, ByteBuffer's capacity is always greater than it's readable bytes, which may cause some bytes in QueryMessageTransfer not read to the ByteBuf in FileRegionEncoder, so super.decode in NettyDecoder get the wrong bytes.

zhanxuefeng:
I add some logs to describe the bug

**RocketMQ version: 4.9.3**
**The message I QueryMsgTraceById has a sub trace and a pub trace**

### QueryMessageProcessor
![image](https://user-images.githubusercontent.com/19700253/167563462-8696a10b-1f33-40bb-9774-d9962de58593.png)

### QueryMessageTransfer
![image](https://user-images.githubusercontent.com/19700253/167563697-48d6ed58-e1fb-432f-b80f-7521f48b83bc.png)

### FileRegionEncoder
![image](https://user-images.githubusercontent.com/19700253/167563788-75113861-ed18-4c9e-975e-25fafc6667f5.png)

### NettyDecoder
![image](https://user-images.githubusercontent.com/19700253/167563848-e04df9e8-5891-4ab7-889f-e6ce956b9c48.png)

### Console log:
![image](https://user-images.githubusercontent.com/19700253/167564898-f0e40e86-78a1-40c9-b42d-8db9b60ef8e6.png)
The `[QueryMessageProcessor] total bytes: 1021` indicate that 1021 bytes should be transferred, and `[QueryMessageProcessor] message count: 2` indicate that 2 messages was found. `[QueryMessageTransfer] header total bytes: 1017` 1017 is the length of data used by LengthFieldBasedFrameDecoder, in QueryMessageTransfer, the header is 193 bytes but 256 transferred, and the first message is 378 bytes but 1024 transferred, 1280(capacity 256 + capacity 1024) is greater than 1021, so the 2nd message will not be transferred, in actual, 571(header 193 + 1st message 378) bytes is transferred,  but LengthFieldBasedFrameDecoder should get 1021 bytes

### NettyDecoder log
[...]

PR #4266, opened 2022-05-09
[ISSUE #4265] Fix query message get wrong result when tls enabled

zhanxuefeng:

Fix query message get the wrong result when tls enabled

## Brief changelog

Correct the return value of WritableByteChannel's write method in FileRegionEncoder


hzh0425:
Can you create an issue to describe this?

zhanxuefeng:
@hzh0425
The same logs mentioned in [ISSUE #4265](https://github.com/apache/rocketmq/issues/4265)

### FileRegionEncoder
image

I search the same msg mentioned in [ISSUE #4265](https://github.com/apache/rocketmq/issues/4265)
image
Header is 193 bytes and transferred 193 bytes, the 1st msg is 378 bytes and transferred 378 bytes, the 2nd msg is 450 bytes and transferred 450 bytes, total 1021 bytes transferred. 
image
NettyDecoder get 1021 bytes and show the sub trace and pub trace.

MatrixHB:
Good catch for this issue. 

However, this bug involves `QueryMessageProcessor`, which is core function for RocketMQ. I wonder why this bug has never been found. Does it only occur when TLS enabled?

Your code change in `FileRegionEncoder` is very underlying. Could you assess the impact on other upper layer functions？

MatrixHB:
Good catch for this issue. 

> ByteBuffer's capacity is always greater than it's readable bytes

However, this bug involves `QueryMessageProcessor`, which is core function for RocketMQ. I wonder why this bug has never been found. Does it only occur when TLS enabled ?

Your code change in `FileRegionEncoder` is very underlying. Does it affect other upper layer functions？

zhanxuefeng:
@MatrixHB
Yes, the bug only occur when tls enabled, `FIleRegionEncoder` will be add to `NettyRemotingServer`'s ChannelPipline only when tlsModel is enforcing.

### HandshakeHandler
image


`ManyMessageTransfer`, `OneMessageTransfer` and `QueryMessageTransfer` are extend from `FileRegion` in RocketMQ

`OneMessageTransfer` may not occur because it has a header and only one ByteBuffer.

But `ManyMessageTransfer` has the problem too, which involves `PullMessageProcessor`, I'll show the test later.

zhanxuefeng:
Test `ManyMessageTransfer`

RocketMQ version: 4.9.3
tls.server.mode: enforcing
transferMsgByHeap=false   this is very important

The topic I created has 1 writeQueueNums and 1 readQueueNums, and I send 2 msgs to the topic.


I add some logs.
### PullMessageProcessor
image
### ManyMessageProcessor
image

### A simple pull consumer
image

### Console log
image
### Consume log
image

Similar to [ISSUE #4265](https://github.com/apache/rocketmq/issues/4265)
610 bytes and 2 msgs was found, but only transferred 413(216 header + 197 1st msg) bytes, because 768(256 capacity + 512 capacity) > 610, so the 2nd msg will not be transferred.
The consume get 413 but 610(4 length + 606 data) needed, so not msg received.

zhanxuefeng:
Prove `ManyMessageProcessor`

With my code changed in `FileRegionEncoder`
image
### Console log
image
### Consumer log
image

One header and 2 msgs were transferred, and consumer received 610 bytes, decode to 2 msgs.

tsunghanjacktsai (approved):

Issue #4263, opened 2022-05-08
Delete topic route info based on cluster


sunxi92:

hzh0425:
Can you add some description?

HScarb:
> 删除 Topic 时，支持按照集群维度删除 Topic 路由信息。

RongtongJin:
Merged

PR #4268, opened 2022-05-09
[ISSUE#4263]Delete topic route info based on cluster when delete topic.

sunxi92:
#4263

### DIFF
--- a/namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
+++ b/namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
@@ -88,6 +88,36 @@ public class RouteInfoManager {
         }
     }
 
+    public void deleteTopic(final String topic, final String clusterName) {
+        try {
+            try {
+                this.lock.writeLock().lockInterruptibly();
+                Set<String> brokerNames = this.clusterAddrTable.get(clusterName);
+                if (brokerNames != null
+                    && !brokerNames.isEmpty()) {
+                    Map<String, QueueData> queueDataMap = this.topicQueueTable.get(topic);
+                    for (String brokerName : brokerNames) {
+                        final QueueData removedQD = queueDataMap.remove(brokerName);
+                        if (removedQD != null) {
+                            log.info("deleteTopic, remove one broker's topic {} {} {}", brokerName, topic,
+                                removedQD);
+                        }
+
+                        if (queueDataMap.isEmpty()) {
+                            log.info("deleteTopic, remove the topic all queue {} {}", brokerName, topic);
+                            this.topicQueueTable.remove(topic);
+                            break;
+                        }
+                    }
+                }
+            } finally {
+                this.lock.writeLock().unlock();
+            }
+        } catch (Exception e) {
+            log.error("deleteTopic Exception", e);
+        }
+    }
+
     public TopicList getAllTopicList() {
         TopicList topicList = new TopicList();
         try {
--- a/namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
+++ b/namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
@@ -96,19 +96,20 @@ public class RouteInfoManager {
                 if (brokerNames != null
                     && !brokerNames.isEmpty()) {
                     Map<String, QueueData> queueDataMap = this.topicQueueTable.get(topic);
-                    for (String brokerName : brokerNames) {
-                        final QueueData removedQD = queueDataMap.remove(brokerName);
-                        if (removedQD != null) {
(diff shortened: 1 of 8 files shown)


ni-ze:
Is the purpose of this pr to support multiple clusterNames?
otherwise, the code is very similar to the original, I think, the same parts should be reused as much as possible.

HScarb (approved):

hzh0425 (approved):

lwclover (changes requested):
In DeleteTopicSubCommand, The parameter 'clusterName' is mandatory.

sunxi92:
> In DeleteTopicSubCommand, The parameter 'clusterName' is mandatory. So deleting a topic with the parameters 'topic' and 'clusterName' is clear.
DeleteTopicSubCommand  should delete topic route info based on clusterName in nameserver. But the processing logic of  nameserver is  deleting all the topic route info in topicQueueTable before updating the code.  @lwclover

lwclover:
> eleteTopicSubCommand should delete topic route info based on clusterNa

In my opion, You don't have to write compatibility code.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -88,6 +88,36 @@ public void deleteTopic(final String topic) {
...
     }
 
+    public void deleteTopic(final String topic, final String clusterName) {
+        try {
+            try {
+                this.lock.writeLock().lockInterruptibly();
+                Set<String> brokerNames = this.clusterAddrTable.get(clusterName);
+                if (brokerNames != null
+                    && !brokerNames.isEmpty()) {
+                    Map<String, QueueData> queueDataMap = this.topicQueueTable.get(topic);
```
lwclover, line 98: QueueDataMap should have a non-null determination to avoid repeating requests and reporting exceptions a second time.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -88,6 +88,36 @@ public void deleteTopic(final String topic) {
...
+                    for (String brokerName : brokerNames) {
+                        final QueueData removedQD = queueDataMap.remove(brokerName);
+                        if (removedQD != null) {
+                            log.info("deleteTopic, remove one broker's topic {} {} {}", brokerName, topic,
+                                removedQD);
+                        }
+
+                        if (queueDataMap.isEmpty()) {
```
lwclover, line 106: My personal recommendation is to put it outside of the for loop, so it's a little more logical.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -88,6 +88,36 @@ public void deleteTopic(final String topic) {
         }
     }
 
+    public void deleteTopic(final String topic, final String clusterName) {
+        try {
+            try {
+                this.lock.writeLock().lockInterruptibly();
+                Set<String> brokerNames = this.clusterAddrTable.get(clusterName);
+                if (brokerNames != null
```
lwclover, line 96: if(brokerNames == null || brokerNames.isEmpty(){
return;
}
Personally, I suggest that the code looks less nested.
It's fine. It's a personal habit.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
sunxi92: you're right!

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
sunxi92: > My personal recommendation is to put it outside of the for loop, so it's a little more logical.


done.

lwclover (approved):

lwclover:
the method deleteTopic(final String topic) is useless, The new method need testcase.

RongtongJin (changes requested):
Hi, @sunxi92 , IMO, compatibility needs to be considered here. If clusterName exists, delete it based on the clusterName. Otherwise, delete the whole topic. Otherwise, there will be issues with older versions of the admin tool or console calls.

lwclover:
> 

@sunxi92  I'm sorry I told you removed the compatibility code.

sunxi92:
> Hi, @sunxi92 , IMO, compatibility needs to be considered here. If clusterName exists, delete it based on the clusterName. Otherwise, delete the whole topic. Otherwise, there will be issues with older versions of the admin tool or console calls.


done!

RongtongJin (approved):

Issue #4280, opened 2022-05-12
We can make some trivial optimizations to the method 'indexKeyHashMethod' in class 'IndexFile'.


XiaoyiPeng:
**BUG REPORT**

1. Please describe the issue you observed:

    We can make some trivial optimizations to the method 'indexKeyHashMethod' in class 'IndexFile', as shown below: 
 https://github.com/apache/rocketmq/blob/edbea4dabf3e1b3a77cc8351586ffc713561921a/store/src/main/java/org/apache/rocketmq/store/index/IndexFile.java#L148-L154

    1.Change the method modifier from public to private.
    2. Remove code from  `Line#151` to `Line#152`

2. Please tell us about your environment:

   OS: Windows10 
   branch: develop

3. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

PR #4281, opened 2022-05-12
[ISSUE #4280] Polish the method 'indexKeyHashMethod' in class 'IndexFile'.

XiaoyiPeng:

For #4280 

## Brief changelog

  1. Change the method modifier from public to private.
  2. Remove code from `Line#151` to `Line#152`

## Verifying this change

XXXX

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/index/IndexFile.java
+++ b/store/src/main/java/org/apache/rocketmq/store/index/IndexFile.java
@@ -145,12 +145,9 @@ public class IndexFile {
         return false;
     }
 
-    public int indexKeyHashMethod(final String key) {
+    private int indexKeyHashMethod(final String key) {
         int keyHash = key.hashCode();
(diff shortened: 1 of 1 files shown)


store/src/main/java/org/apache/rocketmq/store/index/IndexFile.java
```diff
@@ -145,12 +145,9 @@ public boolean putKey(final String key, final long phyOffset, final long storeTi
         return false;
     }
 
-    public int indexKeyHashMethod(final String key) {
+    private int indexKeyHashMethod(final String key) {
         int keyHash = key.hashCode();
-        int keyHashPositive = Math.abs(keyHash);
```
cserwen, line 150: Will hashcode equal the minimum value of int?

XiaoyiPeng: Oh, you remind me, this case can lead to serious problem.

Issue #2706, opened 2021-03-03
SEND_OK will still be returned when the storage is abnormal  ，which may cause message loss


iamqq23ue:
when the flush is abnormal ，SEND_OK will still be returned 。It will cause possible message loss .

I have submitted  a pull request #2707 to fix it. 

RocketMQ version 4.8.0 ,SYNC_FLUSH.

error code show as below ：

You can see that if there is an error, a try catch will be performed .Then   log.error("Error occurred when force data to disk.", e);
But there is no error code returned to the client, and the client still thinks that the send is successful. 

public int flush(final int flushLeastPages) {
    if (this.isAbleToFlush(flushLeastPages)) {
        if (this.hold()) {
            int value = getReadPosition();
            try {
                //We only append data to fileChannel or mappedByteBuffer, never both.
                if (writeBuffer != null || this.fileChannel.position() != 0) {
                    this.fileChannel.force(false);
                } else {
                    this.mappedByteBuffer.force();
                }
            } catch (Throwable e) {
                log.error("Error occurred when force data to disk.", e);
            }

            this.flushedPosition.set(value);
            this.release();
        } else {
            log.warn("in flush, hold failed, flush offset = " + this.flushedPosition.get());
            this.flushedPosition.set(getReadPosition());
        }
    }
    return this.getFlushedPosition();
}

francisoliverlee:
can you give out a description for how to lost message ?

iamqq23ue:
> can you give out a description for how to lost message ?

For example, when the disk read only and flush fails, the actual message is not written to the disk.
In order to solve read only, the server must be restarted, and then the message will be lost.

iamqq23ue:
#2707

lizhiboo:
> when the flush is abnormal ，SEND_OK will still be returned 。It will cause possible message loss .
> 
> I will submit a PR to fix it
> 
> code show as below ：
> 
> ```java
> public int flush(final int flushLeastPages) {
>     if (this.isAbleToFlush(flushLeastPages)) {
>         if (this.hold()) {
>             int value = getReadPosition();
> 
>             try {
>                 //We only append data to fileChannel or mappedByteBuffer, never both.
>                 if (writeBuffer != null || this.fileChannel.position() != 0) {
>                     this.fileChannel.force(false);
>                 } else {
>                     this.mappedByteBuffer.force();
>                 }
>             } catch (Throwable e) {
>                 log.error("Error occurred when force data to disk.", e);
>             }
> 
[...]

PR #4306, opened 2022-05-13
[ISSUE #2706] FIX SEND_OK will still be returned when the storage is abnormal ，which may cause message loss

iamqq23ue:
#2706  I found that the problem still exists in version 4.9.3
when the flush is abnormal ，SEND_OK will still be returned 。It will cause possible message loss .

error code show as below ：

You can see that if there is an error, a try catch will be performed .Then log.error("Error occurred when force data to disk.", e);
But there is no error code returned to the client, and the client still thinks that the send is successful.

`public int flush(final int flushLeastPages) {
        if (this.isAbleToFlush(flushLeastPages)) {
            if (this.hold()) {
                int value = getReadPosition();

                try {
                    //We only append data to fileChannel or mappedByteBuffer, never both.
                    this.flushError = false;
                    if (writeBuffer != null || this.fileChannel.position() != 0) {
                        this.fileChannel.force(false);
                    } else {
                        this.mappedByteBuffer.force();
                    }
                } **catch (Throwable e) {
                    log.error("Error occurred when force data to disk.", e);
                }**

                this.flushedPosition.set(value);
                this.release();
            } else {
                log.warn("in flush, hold failed, flush offset = " + this.flushedPosition.get());
                this.flushedPosition.set(getReadPosition());
            }
        }
        return this.getFlushedPosition();
    }`

I recommend passing the error to the client

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1208,7 +1208,12 @@ public class CommitLog {
                         flushOK = CommitLog.this.mappedFileQueue.getFlushedWhere() >= req.getNextOffset();
                     }
 
-                    req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
+                    if (CommitLog.this.mappedFileQueue.isFlushError()) {
+                        req.wakeupCustomer(PutMessageStatus.FLUSH_DISK_FAILED);
+                    }
+                    else {
+                        req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
+                    }
                 }
 
                 long storeTimestamp = CommitLog.this.mappedFileQueue.getStoreTimestamp();
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1210,8 +1210,7 @@ public class CommitLog {
 
                     if (CommitLog.this.mappedFileQueue.isFlushError()) {
                         req.wakeupCustomer(PutMessageStatus.FLUSH_DISK_FAILED);
-                    }
-                    else {
+                    } else {
                         req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
                     }
                 }
--- a/store/src/main/java/org/apache/rocketmq/store/MappedFile.java
+++ b/store/src/main/java/org/apache/rocketmq/store/MappedFile.java
@@ -66,7 +66,7 @@ public class MappedFile extends ReferenceResource {
     private MappedByteBuffer mappedByteBuffer;
     private volatile long storeTimestamp = 0;
     private boolean firstCreateInQueue = false;
-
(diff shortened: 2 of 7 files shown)


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1208,7 +1208,12 @@ private void doCommit() {
...
                     }
 
-                    req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
+                    if (CommitLog.this.mappedFileQueue.isFlushError()) {
+                        req.wakeupCustomer(PutMessageStatus.FLUSH_DISK_FAILED);
+                    }
+                    else {
+                        req.wakeupCustomer(flushOK ? PutMessageStatus.PUT_OK : PutMessageStatus.FLUSH_DISK_TIMEOUT);
+                    }
                 }
```
hzh0425, line 1217: May be it's better to format:
if () {
  xxx
} else {
  xxx
}

store/src/main/java/org/apache/rocketmq/store/MappedFile.java
```diff
@@ -295,13 +295,15 @@ public int flush(final int flushLeastPages) {
...
                     //We only append data to fileChannel or mappedByteBuffer, never both.
+                    this.flushError = false;
                     if (writeBuffer != null || this.fileChannel.position() != 0) {
                         this.fileChannel.force(false);
                     } else {
                         this.mappedByteBuffer.force();
                     }
                 } catch (Throwable e) {
                     log.error("Error occurred when force data to disk.", e);
+                    this.flushError = true;
```
hzh0425, line 306: If this time flush fails, the next time flush succeeds, but every time you returns 'flush failed' to the client?

iamqq23ue: @hzh0425   “this.flushError = false; ”   will  reset flushError   to  false  when  flush

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
iamqq23ue: @hzh0425 thanks.changed.

store/src/main/java/org/apache/rocketmq/store/MappedFile.java
dongeforever: Flush error rarely happens. And the client does not need to know the flush error.
How about just making it simple, just do not change the flushed position when encountering some unexpected error.

Do not introduce a new response code, and do not need to change other files.

iamqq23ue: @dongeforever Thank you for your review. What you said is also a good idea to solve the problem of data loss.very concise.
But IMO, if "just do not change the flushed position", it may cause "FLUSH_DISK_TIMEOUT" in every flush. when  FLUSH_DISK_TIMEOUT，even if the client resends, it may still be sent to the failed broker. Can it be optimized?

iamqq23ue: @dongeforever When the storage fails, the delayed message cannot be delivered to the target topic, and the delayoffset.json will be synchronized to the slave normally, which will cause the slave to be unable to recover the delayed message .
In addition, the slave is also asynchronous flush. Relying on multiple nodes to ensure that data is not lost, kafka can also do it. The disadvantage of kafka is that there is no synchronous flush

PR #4313, opened 2022-05-13
[ISSUE #2435] Solve the problem that DefaultMQProducer#request() sends messages and waits for timeout synchronously

li-xiao-shuang:
fix : #2435 

1.  Add the default namesrvAddr to the RequestProducer example
2.  Solve the synchronization wait timeout problem because the semaphore does not decrease after DefaultMQProducer#request() message is sent successfully
3. Solve the problem that the callback method is not called after the DefaultMQProducer#request() message is sent successfully

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -1419,7 +1419,6 @@ public class DefaultMQProducerImpl implements MQProducerInner {
             @Override
             public void onException(Throwable e) {
                 requestResponseFuture.setCause(e);
-                requestResponseFuture.executeRequestCallback();
                 requestFail(correlationId);
             }
         }, timeout - cost);
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -484,14 +484,12 @@ public class DefaultMQProducerImpl implements MQProducerInner {
     }
 
     /**
-     * @deprecated
-     * It will be removed at 4.4.0 cause for exception handling and the wrong Semantics of timeout. A new one will be
-     * provided in next version
-     *
      * @param msg
      * @param sendCallback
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -1416,11 +1417,13 @@ public void request(Message msg, final RequestCallback requestCallback, long tim
...
             public void onSuccess(SendResult sendResult) {
                 requestResponseFuture.setSendRequestOk(true);
+                requestResponseFuture.executeRequestCallback();
             }
 
             @Override
             public void onException(Throwable e) {
                 requestResponseFuture.setCause(e);
+                requestResponseFuture.executeRequestCallback();
                 requestFail(correlationId);
```
hzh0425, line 1427: ' requestResponseFuture.executeRequestCallback' has been called in  'requestFail(correlationId)'
you don't need to call it again.

hzh0425 (commented):
According to the above logic, should 'putResponseMessage' also be called in line 1443? and the following request has the same problem?
image

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
li-xiao-shuang: done

li-xiao-shuang:
> According to the above logic, should 'putResponseMessage' also be called in line 1443? and the following request has the same problem? image

Yes, I will deal with this separately. The other interfaces are special. I am still familiar with MQ

hzh0425:
> > According to the above logic, should 'putResponseMessage' also be called in line 1443? and the following request has the same problem? image
> 
> Yes, I will deal with this separately. The other interfaces are special. I am still familiar with MQ

Okay, nice try

duhenglucky (approved):

martincao12:
这么修改相当于在执行sendCallback的onSuccess的时候就立即执行了requestResponseFuture的putResponseMessage或者executeRequestCallback，这俩不是应该在收到replyMsg之后再执行吗，即ClientRemotingProcessor里processReplyMessage做的事情？看Issue #2435里的报错，是因为broker reply msg的时候，request请求的producer已经unregister了。

Issue #4349, opened 2022-05-20
Potential IndexOutOfBoundsException etc.


XiaoyiPeng:
**BUG REPORT**

1. Please describe the issue you observed:

  Hi, @cserwen , In PR #4281,  you reminded me of the risk of method `Math.abs(int a)`, which might return a negative number.
  
  I found this risk elsewhere as well，as shown below:
  
  https://github.com/apache/rocketmq/blob/82a48be00a6d9def14abb871a40cb68cf8a041e9/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java#L79-L80
  
https://github.com/apache/rocketmq/blob/82a48be00a6d9def14abb871a40cb68cf8a041e9/client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java#L75-L77  

I think it would be best to fix it ! 
2. Please tell us about your environment:

OS：MacOS Catalina.
branch: develop.

3. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

PR #4350, opened 2022-05-20
[ISSUE #4349]: Fix the potential 'IndexOutOfBoundsException' etc

XiaoyiPeng:

For #4349 

## Brief changelog

- Fix the potential `IndexOutOfBoundsException`
- `queueId` may be a negative value, fix it.

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
@@ -76,7 +76,9 @@ public class LatencyFaultToleranceImpl implements LatencyFaultTolerance<String> 
             if (half <= 0) {
                 return tmpList.get(0).getName();
             } else {
-                final int i = this.whichItemWorst.incrementAndGet() % half;
+                int i = this.whichItemWorst.incrementAndGet() % half;
+                if (i < 0)
+                    i = 0;
                 return tmpList.get(i).getName();
             }
         }
--- a/client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java
+++ b/client/src/main/java/org/apache/rocketmq/client/latency/MQFaultStrategy.java
@@ -74,7 +74,10 @@ public class MQFaultStrategy {
                     final MessageQueue mq = tpInfo.selectOneMessageQueue();
                     if (notBestBroker != null) {
                         mq.setBrokerName(notBestBroker);
-                        mq.setQueueId(tpInfo.getSendWhichQueue().incrementAndGet() % writeQueueNums);
+                        int queueId = tpInfo.getSendWhichQueue().incrementAndGet() % writeQueueNums;
+                        if (queueId < 0)
+                            queueId = 0;
+                        mq.setQueueId(queueId);
                     }
                     return mq;
                 } else {


Kvicii (approved):
LGTM

hzh0425 (approved):

client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
```diff
@@ -76,7 +76,9 @@ public String pickOneAtLeast() {
             if (half <= 0) {
                 return tmpList.get(0).getName();
             } else {
-                final int i = this.whichItemWorst.incrementAndGet() % half;
+                int i = this.whichItemWorst.incrementAndGet() % half;
+                if (i < 0)
```
Oliverwqcwrw, line 80: IMO,  `i` can not less than zero

XiaoyiPeng: Well, I didn't make it clear in the context.

 `i` is  return from the method `ThreadLocalIndex#incrementAndGet()`, when variable `index` in this method reach to Integer.MAX_VALUE,  `++index`  will be value of `Integer.MIN_VALUE`,  its absolute value (`Math.abs(Integer.MIN_VALUE)`) is itself, that's a negative value.

Oliverwqcwrw: Agree with, I think it would be better to add the comment above the `if(i < 0)`,WDYT

ni-ze:
I think modify this method is better.
image

This method want to return a positive value, but fail in above situation.

HScarb:
Good catch! IMO its better to modify `incrementAdnGet` method.

XiaoyiPeng:
@ni-ze , @HScarb , 

Thanks for your review. 
But IMO,  For code consistency, I recommend not changing this method `ThreadLocalIndex#incrementAndGet()`, because this method is used in many places.
In most scenarios, it determine whether the result returned is positive or negative after the call, as  shown below: 

image

And it we changing this method(`ThreadLocalIndex#incrementAndGet()`), The variable `index` it returned  may still be a positive number, eg: `Integer.MAX_VALUE`.  when `Math.abs(++index)` is called, we still need to judge if this value is negative, just looks like in the screenshot.

Hi @panzhi33 , @RongtongJin  what's your opinion?

panzhi33:
> @ni-ze , @HScarb ,
> 
> Thanks for your review. But IMO, For code consistency, I recommend not changing this method `ThreadLocalIndex#incrementAndGet()`, because this method is used in many places. In most scenarios, it judge whether the result returned is positive or negative after the call, as shown below:
> 
> image
> 
> And if we changing this method(`ThreadLocalIndex#incrementAndGet()`), The variable `index` it returned may be a positive number(eg: `Integer.MAX_VALUE`). when `Math.abs(++index)` is called, we still need to judge if this value is negative, just looks like in the screenshot.
> 
> Hi @panzhi33 , @RongtongJin what's your opinion?

I think it is better to modify org.apache.rocketmq.client.common.ThreadLocalIndex#incrementAndGet.

caigy (commented):
I think  it would be better to change `ThreadLocalIndex#incrementAndGet()`: if `index` reaches `Integer.MAX_VALUE`, set it to 0 instead of incrementing it.

XiaoyiPeng:
> I think it would be better to change `ThreadLocalIndex#incrementAndGet()`: if `index` reaches `Integer.MAX_VALUE`, set it to 0 instead of incrementing it.

OK, Thanks for your review,sir.
 I will check  whether this solution is compatible with the current usage of `ThreadLocalIndex#incrementAndGet() `.

XiaoyiPeng:
Hi, @caigy :

I noticed that this issue has been solved in another PR #4447.  I will close this PR.

Issue #4375, opened 2022-05-25
Make some variables to be final


coderbruis:
1. Please describe the issue you observed:

Make some variables to be final.

PR #4376, opened 2022-05-25
[ISSUE #4375]Make some variables to be final.

coderbruis:

Make some variables to be final.

## Brief changelog

Make some variables to be final.


store/src/main/java/org/apache/rocketmq/store/index/IndexHeader.java
```diff
@@ -22,12 +22,12 @@
...
-    private static int beginPhyoffsetIndex = 16;
-    private static int endPhyoffsetIndex = 24;
-    private static int hashSlotcountIndex = 32;
-    private static int indexCountIndex = 36;
+    private static final int beginTimestampIndex = 0;
+    private static final int endTimestampIndex = 8;
+    private static final int beginPhyoffsetIndex = 16;
+    private static final int endPhyoffsetIndex = 24;
+    private static final int hashSlotcountIndex = 32;
+    private static final int indexCountIndex = 36;
```
hzh0425, line 30: This will not pass the ci format check, so it's best not to change it

tsunghanjacktsai (changes requested):
The compile-time constant (static final) variables should be named with capital letters, which is the reason why the PR cannot get passed.

Moreover, I guess your branch was not updated. Pls make sure it's up to date and clean up the commit history. That would make it more perfect. Thx and appreciate your contribution btw :-)

PR #4383, opened 2022-05-26
[ISSUE #4382]Namesrv nearby route

lwclover:
### RocketMQ同机房就近生产和消费机制。

RocketMQ注册、发现机制
- Broker
  - 间隔30秒发送broker上的Topic全量信息到所有Nameserv。broker启动的时候也是延迟注册到Namesrv上的。当有topic发生变更时TopicConfigManager.dataVersion发生变更。
  - 当有topic创建和更新时，broker立刻发送Topic增量信息到所有Namesrv。同时触发Namesrv的BrokerLiveInfo.dataVersion发生变更。
- Namesrv
  - 保存集群和broker关系、broker信息（brokerName和地址）、broker真实连接和topic版本号，topic信息。
  - 调度线程每10秒钟检查一次，如果发现一个broker 120秒都没有更新注册信息，则删除和这个broker相关的所有信息。
- Producer&Consumer
  - Producer第一次发送Message时，会同步向namesrv请求topic路由信息
  - Consumer启动过程中会向namesrv请求topic路由信息
  - 每隔30秒向namesrv拉取一次topic路由信息，根据路由信息内容（不是版本号）判断是否变化，发生变化更新
- 总结：
Producer第一次发送消息时会同步向namesrv请求topic路由信息，然后默认轮询往每个Queue发送消息。
Consumer启动时获取topic路由信息，然后请求broker获取consumer实例列表，最后根据consumer的数量和负载均衡算法给consumer分配queue。

### 实现rocketmq同机房就近生产和消费：
1.broker增加一个zoneName标识配置，slave的zoneName保持和master相同，通过heartbeat注册到namesrv上。
2.Producer和Consumer可以通过API或者环境变量、系统变量设置zoneName和zoneMode(这是一个开关)。
3.Producer和Consumer获取topic路由信息时，在namesrv上返回对应标识的broker信息。

image

image

### 功能开启方式
broker启动增加系统参数
-Drocketmq.zone=bj

producer或者consumer启动增加系统参数
-Drocketmq.zone=bj
-Drocketmq.zone.mode=true #功能开关


zyl0622:
good job

haqiaolong:
I finally found this function. What a surprise

dongeforever:
@lwclover I have read all the code. This solution is simple and useful in some cases.

And the same time, there are two potential problems.

Firstly,  in the public cloud or hybrid cloud,  with SLB/PrivateLink, the network may be complicated, and we could identify the consumer by their remote address.  So we may need a more common way to identify the client and broker. For example, we may abstract a config named "site", and the client/broker inject the "site" info to the extFields of RemotingCommand.

Secondly, we may need a way to solve the conflict between consumers of the same group in different networks. For example, some consumers on site A, some consumers on site global.

dongeforever:
@lwclover This solution is enough for an inner network inside an organization. It may not be enough for common use cases.

And this issue is so important that it is worth adding a config in the client/broker.

It will nice if you could move forward with a more general solution and I'm glad to help you.

IMO, a RPCHook will be a nice way to add the extField to the API Headers.

lwclover:
> @lwclover This solution is enough for an inner network inside an organization. It may not be enough for common use cases.
> 
> And this issue is so important that it is worth adding a config in the client/broker.
> 
> It will nice if you could move forward with a more general solution and I'm glad to help you.
> 
> IMO, a RPCHook will be a nice way to add the extField to the API Headers.

The optimized version has been submitted

client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
```diff
@@ -340,6 +341,22 @@ public void setEnableStreamRequestType(boolean enableStreamRequestType) {
         this.enableStreamRequestType = enableStreamRequestType;
     }
 
+    public void setZoneName(String zoneName) {
```
duhenglucky, line 344: It is very difficult to let customers to upgrade the client version. If this zone information can be obtained through ip or other environmental information, it will be more friendly

zhouxinyu:
I think we need more discussions about this feature and implementation. I have two points:
1. Is there any chance to implement this feature through the client hooks, e.g., `QueueSelector` for the producer and `AllocateMessageQueueStrategy` for the consumer.
2. The associated implementation seems to introduce more complexity for both client and server, if we really want to bring in this feature, can we consider only implementing it on the server-side, e.g., control the topic route info from namesrv follow the az info of clients?

lizhiboo:
IMO, we can refer to affinity/anti-affinity in k8s. each broker can be labeled some tag, such as idc, rack and so on. namesrv hold broker's tag info through heartbeat. And then, client request route with tags and tolerations through rpchook.

lwclover:
> IMO, we can refer to affinity/anti-affinity in k8s. each broker can be labeled some tag, such as idc, rack and so on. namesrv hold broker's tag info through heartbeat. And then, client request route with tags and tolerations through rpchook.

Yes，That's how I implemented it, I just didn't change the description of the implementation.

lwclover:
> Yes，That's how I implemented it, I just didn't change the description of the implementation.

lwclover:
> I think we need more discussions about this feature and implementation. I have two points:
> 
> 1. Is there any chance to implement this feature through the client hooks, e.g., `QueueSelector` for the producer and `AllocateMessageQueueStrategy` for the consumer.
> 2. The associated implementation seems to introduce more complexity for both client and server, if we really want to bring in this feature, can we consider only implementing it on the server-side, e.g., control the topic route info from namesrv follow the az info of clients?

My original intention was that the user would not need to pay attention to such intricate details

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -90,6 +90,13 @@ public class MixAll {
     public static final String LMQ_PREFIX = "%LMQ%";
     public static final String MULTI_DISPATCH_QUEUE_SPLITTER = ",";
     public static final String REQ_T = "ReqT";
+    public static final String ROCKETMQ_CLIENT_ZONE_ENV = "ROCKETMQ_CLIENT_ZONE";
+    public static final String ROCKETMQ_CLIENT_ZONE_PROPERTY = "rocketmq.client.zone";
+    public static final String ROCKETMQ_CLIENT_ZONE_MODE_ENV = "ROCKETMQ_CLIENT_ZONE_MODE";
+    public static final String ROCKETMQ_CLIENT_ZONE_MODE_PROPERTY = "rocketmq.client.zone.mode";
+    public static final String ZONE_NAME = "zoneName"; 
+    public static final String ZONE_MODE = "zoneMode";
+
```
dongeforever, line 99: To avoid the conflict between normal header fields and extent fields, how about using the prefix "__" to identify the extend fields?

client/src/main/java/org/apache/rocketmq/client/route/hook/AddZoneRPCHook.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.rocketmq.remoting.protocol.RemotingCommand;
+
+public class AddZoneRPCHook implements RPCHook {
+
+    public static final AddZoneRPCHook INSTANCE = new AddZoneRPCHook();
+    /**
+     * The name of zone to which producer or consumer belong.
+     */
+    private String zoneName = System.getProperty(MixAll.ROCKETMQ_CLIENT_ZONE_PROPERTY, System.getenv(MixAll.ROCKETMQ_CLIENT_ZONE_ENV));
+    /**
```
dongeforever, line 32: How about adding more loading mechanisms for zone name configurations?

In some cases, it is important to load the value dynamically. 

For example, load from /etc/rocketmq/extFields.conf  dynamicaly.

And this RPCHook is named to DynamicalExtFieldRPCHook.

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
```diff
@@ -195,6 +195,8 @@ public class BrokerConfig {
      * Whether to distinguish log paths when multiple brokers are deployed on the same machine
      */
     private boolean isolateLogEnable = false;
+    @ImportantField
+    private String zoneName;
```
dongeforever, line 199: Currently, use the rpchook instead.

We need more time to prove this PR's general value.

dongeforever:
@lwclover Currently, we need to make the feature simple enough and prove it along the time.
In the client and broker, we just add a general rpcHook and do not introduce explicit fields. 
Write all the logic inside the nameserver to control the code population.

BTW, the rpcHook could be a general DynamicalExtFieldRPCHook, loading the extent fields from -D, env, etc files.

lwclover:
> 

OK，I will do that

broker/src/main/java/org/apache/rocketmq/broker/plugin/MessageStoreFactory.java
```diff
@@ -1,45 +1,45 @@
-/*
```
dongeforever, line 1: Why this file is labeled as changed?

Maybe some newline delimiter conflicts.

common/src/main/java/org/apache/rocketmq/common/protocol/RequestCode.java
```diff
@@ -194,4 +194,5 @@ public class RequestCode {
     public static final int GET_ALL_PRODUCER_INFO = 328;
 
     public static final int DELETE_EXPIRED_COMMITLOG = 329;
+    
 }
```
dongeforever, line 198: remove these trivial changes.

common/src/main/java/org/apache/rocketmq/common/protocol/route/BrokerData.java
```diff
@@ -29,17 +29,18 @@ public class BrokerData implements Comparable<BrokerData> {
...
     private HashMap<Long/* brokerId */, String/* broker address */> brokerAddrs;
-
+    private String zoneName;
     private final Random random = new Random();
 
     public BrokerData() {
 
     }
 
-    public BrokerData(String cluster, String brokerName, HashMap<Long, String> brokerAddrs) {
```
dongeforever, line 39: Keep the old constructor, and let it be compatible.

common/src/main/java/org/apache/rocketmq/common/rpchook/DynamicalExtFieldRPCHook.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.rocketmq.remoting.RPCHook;
+import org.apache.rocketmq.remoting.protocol.RemotingCommand;
+
+public class DynamicalExtFieldRPCHook implements RPCHook {
+
+    @Override
+    public void doBeforeRequest(String remoteAddr, RemotingCommand request) {
+        String zoneName = System.getProperty(MixAll.ROCKETMQ_ZONE_PROPERTY, System.getenv(MixAll.ROCKETMQ_ZONE_ENV));
+        if (StringUtils.isNotBlank(zoneName)) {
+            request.addExtField(MixAll.ZONE_NAME, System.getProperty(MixAll.ROCKETMQ_ZONE_PROPERTY, System.getenv(MixAll.ROCKETMQ_ZONE_ENV)));
```
dongeforever, line 30: use zoneName instead of duplicated getting code.

dongeforever:
@lwclover It seems ok. 

BTW, some trivial code polishments could be done to make the code clear.

duhenglucky:
LGTM， but would you like to rebase your codebase first? it seems this PR contained so many other's commits

lwclover:
> LGTM， but would you like to rebase your codebase first? it seems this PR contained so many other's commits

Next time. haha

dongeforever (approved):

duhenglucky (approved):

ShannonDing (approved):

Issue #4384, opened 2022-05-27
Expand RocketMQ Topic/Group attributes


drpmma:
1. Please describe the feature you are requesting.

Topic and Group are important concepts in RocketMQ, but they are too simple to provide detailed information and more management ability. Therefore, there is a proposal to expand RocketMQ Topic/Group attributes.  

2. Provide any additional detail on your proposed use case for this feature.

* Topic Message Type
* Consume Timeout for PushConsumer
* Consume Retry Policy
* Max Retry Times
* Group Fifo Consume Flag

Topic Type is planned to implement based on TopicConfig attribute. While remaining Group related attributes will use SubscriptionGroupConfig's fields.

PR #4397, opened 2022-05-31
[ISSUE #4384] Add RetryPolicy interface

drpmma:

Expand RocketMQ Topic/Group attributes

## Brief changelog

Add RetryPolicy interface

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/subscription/RetryPolicy.java
+++ b/common/src/main/java/org/apache/rocketmq/common/subscription/RetryPolicy.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/apache/rocketmq/common/subscription/RetryPolicy.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.common.subscription;
+
+import java.util.concurrent.TimeUnit;
+
+public interface RetryPolicy {
+    long nextDelayDuration(int reconsumeTimes, TimeUnit timeUnit);
+}
```
hzh0425, line 24: It's better to annotate this api

drpmma: fixed

Issue #4437, opened 2022-06-09
[DISCUSS] Optimize design of RpcHook


lizhanhui:
RocketMQ attempts to implement [Intercepting Filter](https://www.oracle.com/technetwork/java/interceptingfilter-142169.html), providing flexibility of executing the hook pipeline before and after each RPC request. See https://github.com/apache/rocketmq/blob/73b9ac82bcd14b2a40ba31888a96e62d06d42d92/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java#L169

https://github.com/apache/rocketmq/blob/73b9ac82bcd14b2a40ba31888a96e62d06d42d92/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java#L177

The overall pattern is awesome while the implementation is pretty limited.

```java

public interface RPCHook {
    void doBeforeRequest(final String remoteAddr, final RemotingCommand request);

    void doAfterResponse(final String remoteAddr, final RemotingCommand request,
        final RemotingCommand response);
}

```

Comparing similar yet better implementation in Netty: https://github.com/netty/netty/blob/d34212439068091bcec29a8fad4df82f0a82c638/transport/src/main/java/io/netty/channel/ChannelPipeline.java#L32

Current RocketMQ implementation is incapable of stopping processing at premature stages.  This limitation results in ugly implementation of features like Authentication, Authorization. Assume a request is found forbidden in the RpcHook implementation, It can only interrupt the execution through throwing exceptions. [Exceptions are not intended for flow control](https://dzone.com/articles/exceptions-as-controlflow-in-java) and handling of the exception in the NettyRemotingServer is also challenging, additionally, this may also impose risks of flooding our logging files.

duhenglucky:
I agree with this optimization. In addition, in the process of request processing, rocketmq will directly write the processing results to the channel, without afterResponse processing, which also brings great challenges to the implementation of various hooks.

lizhanhui:
@duhenglucky  [This PR](https://github.com/apache/rocketmq/pull/4446) resolved all limitations mentioned in this issue.

lizhanhui:
Let's close this issue for now till other guys involved take more time to reach a consensus.

PR #4446, opened 2022-06-10
Revamp interceptor filter of RemotingCommand

lizhanhui:

Fix [issue 4437](https://github.com/apache/rocketmq/issues/4437)

## Brief changelog

Revamp RPC Hook design and implementation

## Verifying this change


remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -166,22 +178,55 @@ public void processMessageReceived(ChannelHandlerContext ctx, RemotingCommand ms
         }
     }
 
+    protected Decision preHandle(final HandlerContext context, final RemotingCommand request,
+                                 final CompletableFuture<RemotingCommand> responseFuture) {
+        Decision decision = Decision.CONTINUE;
+        for (Handler handler : handlers) {
+            try {
+                decision = handler.preHandle(context, request, responseFuture);
+            } catch (Throwable ignore) {
```
zhouxinyu, line 187: Exception means CONTINUE, is that reasonable?

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -166,22 +178,55 @@ public void processMessageReceived(ChannelHandlerContext ctx, RemotingCommand ms
...
                 rpcHook.doBeforeRequest(addr, request);
             }
         }
     }
 
+    @Deprecated
     protected void doAfterRpcHooks(String addr, RemotingCommand request, RemotingCommand response) {
         if (rpcHooks.size() > 0) {
-            for (RPCHook rpcHook: rpcHooks) {
+            for (RPCHook rpcHook : rpcHooks) {
```
zhouxinyu, line 210: Code style conflict

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
lizhanhui: After a second thought, let's change the decision to stop and generate response command according to the exception raised.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
lizhanhui: The code style inconsistency is fixed.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -200,43 +245,79 @@ public void processRequestCommand(final ChannelHandlerContext ctx, final Remotin
...
+                                log.warn(message);
+                                response = RemotingCommand.createResponseCommand(RemotingSysResponseCode.SYSTEM_ERROR,
+                                        message);
+                            }
+                            response.setOpaque(opaque);
+                            ctx.writeAndFlush(response);
+                            return;
+                        }
+                        // TODO: Remove the following line when RPC Hook reaches end of life.
                         doBeforeRpcHooks(remoteAddr, cmd);
```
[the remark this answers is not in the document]
lizhanhui: Done

dongeforever:
The RPCHook has been widely used for almost 10 years, and it has been exposed to end-users. The DefaultMQProducer and DefaultMQConsumer all have API to register RPCHook.

So we should not modify it unless we have to.

In fact, for flow control or any other demands, there is no need to modify the RPCHook.  The PR https://github.com/apache/rocketmq/pull/4454 shows the code. 

This PR needs a wide discussion before merging.

@lizhanhui @zhouxinyu @duhengforever

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -166,22 +173,50 @@ public void processMessageReceived(ChannelHandlerContext ctx, RemotingCommand ms
...
+                break;
             }
         }
+        return decision;
     }
 
+    public void registerRPCHook(RPCHook rpcHook) {
+        if (null != rpcHook) {
+            handlers.add(new HandlerAdaptor(rpcHook));
+        }
```
dongeforever, line 214: If the RPCHook is not exposed to the end-user, this Adapter is great.

But currently, the widely used DefaultMQProduer or  DefaultMQConsumer, or  DefaultAdminExt have the access to RPCHook. 
It is better to come up with other way to solve it.

dongeforever: The flow-control is very useful, So why not apply the ability to the RPCHook, which could benefit the users without adding a new interface.

lizhanhui:
@dongeforever This pull intends to keep RPCHooks as it is. Third-party implementations will continue to work as expected.
Internally, RPCHooks will be wrapped to Handler, which will have better-defined APIs and deliver neat and clear semantics.

Further, we hope new plugins are developed on top of the new interface, bringing them fewer doubts during their development at error handling, logic control, etc.

In addition, this PR solves other closely related defects: 
on the server-side,  post-hooks will get executed when ctx.writeAndFlush is used, as is also pointed out by @duhenglucky
on the client-side, post-hooks are executed for the async code path;

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
lizhanhui: We have two goals: 1) Ensure original RPCHook implementations would continue to work; 2) Better development experience for further developers who develops plugins for RocketMQ

dongeforever:
@lizhanhui Most part of this PR is great.

The only doubt is that there is no need to introduce a new Handler to do this, adding context to the RPCHook works well.

lizhanhui:
@dongeforever Let's wait  to see more feedback from other community developers

fuyou001:
the new feature is only suppport jdk7+,others looks good to me.

zhouxinyu:
> 

Hi @dongeforever, follow the PR https://github.com/apache/rocketmq/pull/4454, it seems that if we want to use the new hook behavior we must override the `AbstractRPCHook` and register it, right? If so, although we don't expose a new method, a new abstract class is exposed.

duhenglucky:
> adding



> @dongeforever This pull intends to keep RPCHooks as it is. Third-party implementations will continue to work as expected. Internally, RPCHooks will be wrapped to Handler, which will have better-defined APIs and deliver neat and clear semantics.
> 
> Further, we hope new plugins are developed on top of the new interface, bringing them fewer doubts during their development at error handling, logic control, etc.
> 
> In addition, this PR solves other closely related defects: on the server-side, post-hooks will get executed when ctx.writeAndFlush is used, as is also pointed out by @duhenglucky on the client-side, post-hooks are executed for the async code path;

A very good change, not only optimizes RPCHook, but also solves the problem that the post method is not called in asynchronous mode, but compatibility is still the biggest bottom line, many developers will implement their own Hook in the process of adapting to their own business .

zhouxinyu:
Back to this PR, `registerHandler` is not a matched name compared to `RPCHook`. Perhaps, we could introduce a `EnhancedRPCHook` interface to make this feature easier to understand for both new and existing developers?

zhouxinyu:
> 

Both the two PRs don't break the compatibility. The discussion point is that should we introduce a new method like `registerHandler` or not.

dongeforever:
@zhouxinyu 
The AbstractRPCHook is a compromised solution for JDK 1.6.

BTW, for most users,  it does not need the HanderContext,  just use the RPCHook as usual. The flow control is usually used on the Server-Side, which wants to provide advanced features.

dongeforever:
The AbstractRPCHook is like an advanced tool. In fact, the third-party project could provide enhanced implementation without requiring the user to change their application code.

duhenglucky:
> > 
> 
> @duhenglucky Both the two PRs don't break the compatibility. The discussion point is that should we introduce a new method like `registerHandler` or not.



> > 
> 
> @duhenglucky Both the two PRs don't break the compatibility. The discussion point is that should we introduce a new method like `registerHandler` or not.

Sorry for didn't notice that this pr moved the registerRPCHook method to the Abstract class, However, for most of the exposed APIs or Hooks, the cost of migration should be avoided as much as possible. I prefer to continue to evolve on RPCHook. RPCHook and Handler seem to be two concepts， AbstractRPCHook provides a solution, but the current implementation, especially in JDK 1.6 doesn't seem to be friendly enough. In the long run, we will also upgrade JDK 1.6 support to 1.8 or higher, but this depends on the outcome of community discussions

lwclover:
Implementation of pipeline by Handlers, each handler call a RPCHook. 
In RPCHook, there may be deserialization and serialization, which is a time-consuming behavior, I don't think this is an appropriate way to work on it.

lizhanhui:
> BTW, for most users, it does not need the HanderContext, just use the RPCHook as usual. The flow control is usually used on the Server-Side, which wants to provide advanced features.

gRPC does split this feature into two interfaces, [ClientInterceptor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientInterceptor.html) and [ServerInterceptor](https://grpc.github.io/grpc-java/javadoc/index.html?io/grpc/ServerInterceptor.html)

lizhanhui:
@lwclover 
> Implementation of pipeline by Handlers, each handler call a RPCHook.
Yes, this is how this PR works.

> In RPCHook, there may be deserialization and serialization, which is a time-consuming behavior, I don't think this is an appropriate way to work on it.

No, there is no serialization or deserialization involved.

lizhanhui:
@duhenglucky 
> the cost of migration should be avoided as much as possible. 
There is no cost or migration required. The existing code would continue to work as expected.

> I prefer to continue to evolve on RPCHook
There is no graceful way of evolving on RpcHook under current constraints. If the 'Default Method' feature of interface is possible, aka, relaxing JDK to 8+, we could as well add an extra pair preXXX and postXXX and delegate current ones to them.

lizhanhui:
> Back to this PR, `registerHandler` is not a matched name compared to `RPCHook`. Perhaps, we could introduce a `EnhancedRPCHook` interface to make this feature easier to understand for both new and existing developers?

Maybe, we could rename / split Handler to ClientInterceptor / ServerInterceptor similar to gRPC.

dongeforever:
@lizhanhui I notice that you want to improve the developer experience.

But compatiblity is the most important experience for developers.

Envolve the RPCHook could continue the user's habit. The only time to break the old habit is that the old habit could not solve the problem any more.
But currently, it is easy to envolve the RPCHook to get the flow-control ability. So there is no need to add a new one to build a new habit?

lizhanhui:
@dongeforever 

> But compatibility is the most important experience for developers.
This PR suffers no compatibility issue at all, which has been explicitly stated previously.

> it is easy to evolve the RPCHook to get the flow-control ability.

This is simply not true. Your proposing alternative suffers in multiple aspects
1.  The proposing alternative forces developers to inherit AbstractRpcHook. Considering Java follows single inheritance, it might be a blocking issue for a few cases.
2. Even if AbstractRpcHook is used, users have to set Decision to HandlerContext. Compared to the return-value approach, this API is less intuitive and error-prone. 
3.  [Context](https://en.wikipedia.org/wiki/Context_(computing)) in most cases are used to feed info to tasks. It is rarely used to return a value, especially when it is not part of the function signature.

dongeforever:
@lizhanhui 

1. In the beginning, without the InterceptorAdaptor, the compatibility problem is serious. Currently, it is not so serious. But two Interfaces(RPCHook vs Interceptor) do a similar thing, which is not so good. And the RPCHook is marked deprecated, the end-user needs to change their code sooner or later, which will cause anxiety to the old user. This is not a good experience.

2. It is important to explain how often the flow control is used.  In practice, the flow control is only used on the server-side, and only be used by the advanced developers(usually the service provider), Which means it is used by a very small number of developers. For the great majority, they do not need the flow control, especially on the client-side. So we'd better protect the end-user experience, which is the great majority.

PR #4469, opened 2022-06-16
[ISSUE#4468] Optimize broker buffer length initialization

shengminw:
Optimize encoder buffer initialization, expand buffer length to avoid message size exceeding when messageBodySize is on the boundry of maxMessageBodySize.

## Brief changelog

1. add maxMessageSize variable differing from maxMessageBodySize（default maxMessageSize = maxMessageBodySize + 64kb）

2. add checking messagesize

3. add junit test  "testEncodeLongMessage()"


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1480,11 +1480,15 @@ public static class MessageExtEncoder {
...
         private final int maxMessageBodySize;
-
+        // The maximum length of the full message.
+        private final int maxMessageSize;
         MessageExtEncoder(final int maxMessageBodySize) {
             ByteBufAllocator alloc = UnpooledByteBufAllocator.DEFAULT;
-            byteBuf = alloc.directBuffer(maxMessageBodySize);
+            //Reserve 64kb for encoding buffer outside body
+            int maxMessageSize = maxMessageBodySize + 64 * 1024;
+            byteBuf = alloc.directBuffer(maxMessageSize);
```
RongtongJin, line 1489: Hi, I wonder to know why it is 64KB?

tsunghanjacktsai (commented):
Hi,

Thx for your contribution in advance. But we could make a discuss if it is worth making this change.

The ByteBuffer used here was from Netty API, and it already had an implementation of autoscaling under it. Whenever the ByteBuf reaches the initialCapacity set by developers, it would start to autoscale by 4MB until maxCapacity (set by devs) or Integer.MAX_VALUE.

Of course, we do not want autoscaling it every time, but if the chance of exceeding the maxMessageBodySize is not that high. I don't see why we need to add another 64KB (from your changes) for ByteBuf.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
shengminw: Most of parameters in the message encoding are fixed length, only the properties and tpopic name content are variable. The maximum length of properties should be less than or equal to Short.MAX_VALUE, and the topic name content should be less than or equal to 127 bytes. Taking the maximum value, it adds up to about 33kb. Considering to reserve some space for subsequent adjustments, 64kb was chosen.
* org.apache.rocketmq.client.Validators#TOPIC_MAX_LENGTH

```java
public static final int TOPIC_MAX_LENGTH = 127;
```
* org.apache.rocketmq.store.CommitLog.MessageExtEncoder#encode

```java
if (propertiesLength > Short.MAX_VALUE) {
    log.warn("putMessage message properties length too long. length={}", propertiesData.length);
    return new PutMessageResult(PutMessageStatus.PROPERTIES_SIZE_EXCEEDED, null);
}
```


Note: The maximum length of the topic is set on the client side, and there is no open interface for modification. Besides, the broker side does not provide the parameter "TOPIC_MAX_LENGTH", so I directly select the extra space as a fixed value.

shengminw:
> 



> Hi,
> 
> Thx for your contribution in advance. But we could make a discuss if it is worth making this change.
> 
> The ByteBuffer used here was from Netty API, and it already had an implementation of autoscaling under it. Whenever the ByteBuf reaches the initialCapacity set by developers, it would start to autoscale by 4MB until maxCapacity (set by devs) or Integer.MAX_VALUE.
> 
> Of course, we do not want autoscaling it every time, but if the chance of exceeding the maxMessageBodySize is not that high, I don't see why we need to add another 64KB (from your changes) for ByteBuf.

Yes, using bytebuf can achieve automatic expansion, so in the current scenario, no error will be reported. But what I consider is that mq should have its own detection of the encoding length of the message body, instead of relying on the self-expansion of bytebuf.

As for why it is 64kb, I have replied in a previous comment.

tsunghanjacktsai:
@shengminw Okay, that makes sense. Thx for your reply. Maybe @ni-ze would like to check this out.

shengminw:
@tsunghanjacktsai ok, thanks~

Kvicii (approved):

store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
```diff
@@ -657,6 +658,30 @@ public void testCleanUnusedLmqTopic() throws Exception {
...
+
+        Field fieldMessageStoreConfig = clazz.getDeclaredField("messageStoreConfig");
+        fieldMessageStoreConfig.setAccessible(true);
+        MessageStoreConfig messageStoreConfig = (MessageStoreConfig) fieldMessageStoreConfig.get(messageStore);
+        messageExtBrokerInner.setBody(new byte[messageStoreConfig.getMaxMessageSize()]);
+
+        CommitLog.PutMessageThreadLocal putMessageThreadLocal = commitLog.getPutMessageThreadLocal().get();
+        PutMessageResult encodeResult1 = putMessageThreadLocal.getEncoder().encode(messageExtBrokerInner);
+        assertTrue(encodeResult1 == null);
+
```
dongeforever, line 678: @shengminw This test is for DefaultMessageStore.
So just use type conversion (DefaultMessageStore) messagestore, instead of reflection.

The reflection is not friendly for further evolution.

dongeforever: BTW, it maybe need another test for body size exactly equal to max message size.

shengminw: ok, I will revise it.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -1482,11 +1482,15 @@ public static class MessageExtEncoder {
...
         // The maximum length of the message body.
         private final int maxMessageBodySize;
-
+        // The maximum length of the full message.
+        private final int maxMessageSize;
         MessageExtEncoder(final int maxMessageBodySize) {
             ByteBufAllocator alloc = UnpooledByteBufAllocator.DEFAULT;
-            byteBuf = alloc.directBuffer(maxMessageBodySize);
+            //Reserve 64kb for encoding buffer outside body
+            int maxMessageSize = maxMessageBodySize + 64 * 1024;
```
ni-ze, line 1490: IMO, variable parameters, which is properties and topic name, is limited and properties before write into commitLog, if this two item length exceed limit, it won't reach here.

and I think it is a little weird to limit length base on an auto scalable ByteBuf.

shengminw: Yes, get to this step, properties(max 32kb) and topic name(max 127 bytes) are limited. But the full message length still may exceed 4M(maxMessageBodySize).
About whether it is neccessary to do this change, it's worth discussing. I think, in the previous version, maxMessageBodySize is also used to limit buffer length. So from my point of view, I think it is necessary to add.

dongeforever (approved):

Issue #4471, opened 2022-06-16
standard JSON dont compat with go client


yuz10:
https://github.com/apache/rocketmq-client-go/issues/845

**BUG REPORT**

 in https://github.com/apache/rocketmq/pull/4432 , the response of GET_ROUTEINFO_BY_TOPIC is changed, for example, the new format is:
`{"brokerDatas":[{"brokerAddrs":{"0":"127.0.0.1:10911"}]}`
the old is:
`{"brokerDatas":[{"brokerAddrs":{0:"127.0.0.1:10911"}]}`

the following go code in https://github.com/apache/rocketmq-client-go/blob/master/internal/route.go#L561
```
func (routeData *TopicRouteData) decode(data string) error {
	res := gjson.Parse(data)
	err := jsoniter.Unmarshal([]byte(res.Get("queueDatas").String()), &routeData.QueueDataList)

	if err != nil {
		return err
	}

	bds := res.Get("brokerDatas").Array()
	routeData.BrokerDataList = make([]*BrokerData, len(bds))
	for idx, v := range bds {
		bd := &BrokerData{
			BrokerName:      v.Get("brokerName").String(),
			Cluster:         v.Get("cluster").String(),
			BrokerAddresses: make(map[int64]string, 0),
		}
		addrs := v.Get("brokerAddrs").String()
		strs := strings.Split(addrs[1:len(addrs)-1], ",")
		if strs != nil {
			for _, str := range strs {
				i := strings.Index(str, ":")
				if i < 0 {
					continue
				}
				id, _ := strconv.ParseInt(str[0:i], 10, 64)
				bd.BrokerAddresses[id] = strings.Replace(str[i+1:], "\"", "", -1)
			}
		}
		routeData.BrokerDataList[idx] = bd
	}
	return nil
}
```
in `id, _ := strconv.ParseInt(str[0:i], 10, 64)`  , if broker id is 1, then `str[0:i]` is `"\"1\""`, and `strconv.ParseInt("\"1\"")` will always be 0
so the producer will send message to slave.

I have submit a pull request in rocketmq-client-go, https://github.com/apache/rocketmq-client-go/pull/846, but I think we should also be compatable with the old versions of go client.

2. Please tell us about your environment:

3. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

duhenglucky:
@yuz10 good catch， would you like to submit a PR to temporarily change back to the previous way to ensure compatibility with old clients?

[...]

PR #4472, opened 2022-06-17
[USSUE #4471] Revert "Let name server generate valid JSON response when process topic route queries"

yuz10:
#4471

Reverts apache/rocketmq#4432


zhouxinyu:
@lizhanhui Why do we close this PR? Do you have a better solution to handle this compatibility issue? e.g., add a config for this behavior and disable it by default?

lizhanhui:
Non-Java clients must have standard JSON. Let's create a patch to let the name server respond conditionally.

PR #4473, opened 2022-06-17
Generate legal JSON response conditionally

lizhanhui:
Fix https://github.com/apache/rocketmq/issues/4471

### DIFF
--- a/namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
+++ b/namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
@@ -62,6 +62,7 @@ import org.apache.rocketmq.remoting.exception.RemotingCommandException;
 import org.apache.rocketmq.remoting.netty.AsyncNettyRequestProcessor;
 import org.apache.rocketmq.remoting.netty.NettyRequestProcessor;
 import org.apache.rocketmq.remoting.protocol.RemotingCommand;
+import org.apache.rocketmq.remoting.protocol.RemotingSerializable;
 
 public class DefaultRequestProcessor extends AsyncNettyRequestProcessor implements NettyRequestProcessor {
     private static InternalLogger log = InternalLoggerFactory.getLogger(LoggerName.NAMESRV_LOGGER_NAME);
@@ -363,9 +364,16 @@ public class DefaultRequestProcessor extends AsyncNettyRequestProcessor implemen
                 topicRouteData.setOrderTopicConf(orderTopicConf);
             }
 
-            byte[] content = topicRouteData.encode(SerializerFeature.BrowserCompatible,
+            byte[] content;
+            Boolean standardJsonOly = requestHeader.getAcceptStandardJsonOnly();
+            if (request.getVersion() >= Version.V4_9_3.ordinal() || (null != standardJsonOly && standardJsonOly)) {
+                content = topicRouteData.encode(SerializerFeature.BrowserCompatible,
                     SerializerFeature.QuoteFieldNames, SerializerFeature.SkipTransientField,
                     SerializerFeature.MapSortField);
+            } else {
+                content = RemotingSerializable.encode(topicRouteData);
+            }
+
             response.setBody(content);
             response.setCode(ResponseCode.SUCCESS);
             response.setRemark(null);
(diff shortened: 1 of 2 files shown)


namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
```diff
@@ -363,9 +364,16 @@ public RemotingCommand getRouteInfoByTopic(ChannelHandlerContext ctx,
                 topicRouteData.setOrderTopicConf(orderTopicConf);
             }
 
-            byte[] content = topicRouteData.encode(SerializerFeature.BrowserCompatible,
+            byte[] content;
+            Boolean standardJsonOly = requestHeader.getAcceptStandardJsonOnly();
+            if (request.getVersion() >= Version.V4_9_3.ordinal() || (null != standardJsonOly && standardJsonOly)) {
+                content = topicRouteData.encode(SerializerFeature.BrowserCompatible,
```
dongeforever, line 370: The 4.9.3 has been released.

So here should > instead of >= ?

lizhanhui: Ahh... MQVersion#current should have been bumped up in the "develop" branch.

zhouxinyu: Or avoid introducing version comparison?   Let clients decide to use `standardJsonOly` or not.

yuz10 (approved):

lizhanhui:
@dongeforever Almost all JSON libraries accept standard JSON only. All JSON libraries for C++ including RapidJSON, jsoncpp, protobuf-json-util, etc. json-go is also true and this is why they parse the response manually. Most popular rust crate serde is exactly same. Java libraries are kind of tolerant, Gson/jackson may tolerate some illegal json produced by FastJSON, but not all (For cases of objects as key, they are not able to tolerate).

Current C++ SDK suffers core-dump from time to time,  one of the root causes is some undefined behavior when parsing illegal JSON output.   

So there is no problem of "they may need to use two different json library.", their problem, however, is that they cannot use a production ready library and are forced to parse manually, the action itself is error-prone.  See the pasted go snippet in the issue.

dongeforever:
@lizhanhui Return the standard JSON is OK.

The only doubt is that we should respect the history and protect the current users.

So a good life circle design is needed for every API/Protocol change, we cannot just change it in a PR without RIP or detailed discussion.

lizhanhui:
"The only doubt is that we should respect the history and protect the current users.
So a good life circle design is needed for every API/Protocol change, we cannot just change it in a PR without RIP or detailed discussion."
The effort itself is helping current users, users who are suffering crashes. There are a number of issues previously raised,  and enough discussions were held. Search the issues first before accusing.

dongeforever:
@lizhanhui "we need to check all the protocol data and make them behave in a consistent way in one version"

 Do not make some APIs is returned standard JSON, others do not.

namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
dongeforever: @zhouxinyu Agrees with you.
It may be difficult to handle all the cases between different versions and different APIs. Just let the client decide.

lizhanhui: Generating invalid JSON is not something ignorable... It is a bug and a serious one, aliening this project from other communities. Anyone who ever writes non-Java code, communicating with RocketMQ,  would see the cumbersomeness.  It is OK to maintain compatibility with previously released clients and this is why this pull request and its predecessor pull request were created. However, this bug has to be fixed and eventually speak standard protocol. 

Back to the proposal of 'letting the client decide'. It seems reasonable at the first sight. But doing so means the future releases of the client would still require the name server continuously generate illegal JSON content. This still looks acceptable, until one day their infrastructure team would make transparent traffic management like envoy, linkerd or dapper. Boom! This issue turns up again.

Previously, I tried to add support for RocketMQ at envoy: https://github.com/envoyproxy/envoy/blob/main/contrib/rocketmq_proxy/filters/network/source/protocol.cc 
See how awkward the code is. I guess someone who read the code must laugh aloud, saying these guys do not understand JSON...amazing.

dongeforever: @lizhanhui It is also OK to force the new clients to use standard JSON. I do not insist on it.

But we'd better do some extra effort to inform other client developers. For example, do notice on the mailing list and on the release notes.

dongeforever: Also, it may be important to keep all the APIs to return standard JSON, not just some of them. This may be another effort.

Anyway a lot of Job.

PR #4477, opened 2022-06-18
implement transaction message escape

azhsmesos:
docs: https://shimo.im/docs/rp3OVGvVLpFav8Am
具体开发思路在 proposal 下


broker/src/test/java/org/apache/rocketmq/broker/transaction/queue/TransactionEscapeBridgeTest.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.rocketmq.broker.transaction.queue;
+
+import org.junit.Ignore;
+
+/**
+ * @author zhaozhenhang <zhaozhenhang@kuaishou.com>
+ * Created on 2022-06-18
+ */
```
hzh0425, line 8: Don't leave author info

broker/src/test/java/org/apache/rocketmq/broker/transaction/queue/TransactionEscapeBridgeTest.java
```diff
@@ -0,0 +1,13 @@
+package org.apache.rocketmq.broker.transaction.queue;
```
hzh0425, line 1: need apache Licence if add new class

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -354,7 +358,13 @@ private PutMessageResult putBackToHalfQueueReturnResult(MessageExt messageExt) {
...
         try {
             MessageExtBrokerInner msgInner = transactionalMessageBridge.renewHalfMessageInner(messageExt);
-            putMessageResult = transactionalMessageBridge.putMessageReturnResult(msgInner);
+            if (this.transactionalMessageBridge.getBrokerController().isSpecialServiceRunning()
+                    && BrokerRole.SLAVE == this.transactionalMessageBridge.getBrokerController().getMessageStoreConfig()
+                    .getBrokerRole()) {
+                putMessageResult = transactionalMessageBridge.getBrokerController().getEscapeBridge().putMessage(msgInner);
+            } else {
+                putMessageResult = transactionalMessageBridge.putMessageReturnResult(msgInner);
+            }
```
caigy, line 367: If retry times of the half message reached maximum value, the broker where it escaped would not check it any more, and the broker it was born also had skipped it, that message would be lost.

caigy, line 367: If retry times of the half message reached maximum value, the broker where it escaped would not check it any more, and the broker it was born also had skipped it, that message would be lost.

caigy (changes requested):
I've left some comments in proposal, hope that it would be helpful.

hzh0425:
Please add some tests to prove correctness

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
azhsmesos: The transactionmessageserviceimpl#putbacktohalfqueuereturnresult() method needs to rewrite the half message. Here,I think we have to rewrite the topic. I understand that this step should also be carried out by escaping to other masters

azhsmesos:
> Please add some tests to prove correctness

thanks！I will add it when i local test later

PR #4493, opened 2022-06-21
Fix the default value of parameter 'sendMsgTimeout'，the default value…

ycfxhsw:
1、Please describe the issue you observed:

As the picture above shows, the default value of the parameter sendMsgTimeout is 10000, but in the code, the default value is 3000.
https://github.com/apache/rocketmq/blob/bda0ae0305863cc5fa2e0578439dfa4f4d9908b3/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java#L102


tsunghanjacktsai (approved):
Great change. We definitely need more contributors to help us improve our documentation.

hzh0425 (approved):
Good

Kvicii (approved):

caigy (approved):

lizhanhui:
Maybe, we should justify the semantics of **sendMsgTimeout**.
Previously, it attempts to limit the overall amount of time, including transparent and explicit retry.

After examining multiple projects, it turns out it would be better if we could separate retry into explicit and transparent ones.  For the explicit retry, jitter in most cases is desirable.  Further, let's users specify RPC tier timeout is more reasonable.

ni-ze (approved):

PR #4500, opened 2022-06-22
Fix static topic test, await the client metadata to be refreshed

dongeforever:

Fix the unstable test StaticTopicIT

### DIFF
--- a/test/src/test/java/org/apache/rocketmq/test/statictopic/StaticTopicIT.java
+++ b/test/src/test/java/org/apache/rocketmq/test/statictopic/StaticTopicIT.java
@@ -21,6 +21,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import org.apache.log4j.Logger;
 import org.apache.rocketmq.broker.BrokerController;
+import org.apache.rocketmq.client.consumer.DefaultMQPushConsumer;
+import org.apache.rocketmq.client.impl.factory.MQClientInstance;
+import org.apache.rocketmq.client.producer.DefaultMQProducer;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.admin.ConsumeStats;
 import org.apache.rocketmq.common.admin.OffsetWrapper;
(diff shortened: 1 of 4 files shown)


lizhanhui:
This pull request is terrible... Magic amount of sleep duration is the root cause of random test failures. Should get rid of them.

[Stackoverflow has a thread discussing similar issues.](https://stackoverflow.com/questions/36283334/how-to-avoid-thread-sleep-in-unit-tests) Maybe we should turn to [awaitility](https://github.com/awaitility/awaitility) as recommended.

hzh0425 (approved):

Kvicii (approved):

test/src/test/java/org/apache/rocketmq/test/statictopic/StaticTopicIT.java
```diff
@@ -117,7 +116,7 @@ public void testCommandsWithBrokers() throws Exception {
         {
             Set<String> brokers = ImmutableSet.of(broker2Name);
             MQAdminTestUtils.remappingStaticTopicWithCommand(topic, brokers, null, nsAddr);
-            Thread.sleep(500);
```
lizhanhui, line 120: Fix the magic number, please.

lizhanhui (changes requested):

dongeforever:
@lizhanhui  I agree with you, await is better than sleeping.

This test is stable on local mac, but it is not stable on the Travis-ci cloud. And I have not found the root cause yet. Could you help to find it, and show the await condition in detail? I am struggling with it for days.

dongeforever:
The root cause seems to be that the client does not refresh the metadata in time.

test/src/test/java/org/apache/rocketmq/test/statictopic/StaticTopicIT.java
dongeforever: done

test/src/test/java/org/apache/rocketmq/test/statictopic/StaticTopicIT.java
```diff
@@ -340,7 +340,7 @@ public void testDoubleReadCheckConsumerOffset() throws Exception {
             sendMessagesAndCheck(producer, targetBrokers, topic, queueNum, msgEachQueue, (i + 1) * TopicQueueMappingUtils.DEFAULT_BLOCK_SEQ_SIZE);
         }
 
-        TestUtils.waitForSeconds(20);
```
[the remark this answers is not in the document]
dongeforever: done

Issue #4520, opened 2022-06-27
[Optimization] Implenment adjusting maxMessageSize dynamicly


shengminw:
## [Optimization] Implenment adjusting maxMessageSize dynamicly

After broker was started, we can still modify the broker's configuration parameters, including maxMessageSize, through tools like mqadmin. 

However, after the maxMessageSize parameter is modified, since the buffer used for encoding in CommitLog has been initialized, it does not actually take effect.

Currently, encode bytebuff is fixed and can't be modify after initialized.

```java
private final ByteBuf byteBuf;
// The maximum length of the message body.
private final int maxMessageBodySize;
// The maximum length of the full message.
private final int maxMessageSize;
MessageExtEncoder(final int maxMessageBodySize) {
    ByteBufAllocator alloc = UnpooledByteBufAllocator.DEFAULT;
    //Reserve 64kb for encoding buffer outside body
    int maxMessageSize = maxMessageBodySize + 64 * 1024;
    byteBuf = alloc.directBuffer(maxMessageSize);
    this.maxMessageBodySize = maxMessageBodySize;
    this.maxMessageSize = maxMessageSize;
}
```

So, I consider to add interface to modify buffer capacity, and implenment adjusting maxMessageSize dynamicly through checking maxMessageSize before encoding bytebuf.

lizhiboo:
@shengminw can u describe any scenario for adjusting maxMessageSize in production? maxMessageSize in broker is corresponding to maxMessageSize in DefaultMQProducer. IMO, maxMessageSize is known to uses before usage, and users should avoid to sender message larger than maxMessageSize.

shengminw:
@lizhiboo 
Yes, maxMessageSize can been set before usage in the MessageStoreConfig. And the default maxMessageSize is 4 * 1024 * 1024. 
Except for setting paramaters before usage, RMQ also provide some tools to update MessageStoreConfig during broker is running. For exmaple, when we want to send a message is 5 * 1024 * 1024, we can use "mqadmin updateBrokerConifig -b *** -k maxMessageSize -v 5 * 1024 * 1024" to update broker config.
However, just modify the MessageStoreConfig cannot tanke effect, because the buffer used for encoding in CommitLog has been initialized.

PR #4521, opened 2022-06-27
[ISSUE#4520] [Optimization] Implenment adjusting maxMessageSize dynamicly

shengminw:
[ISSUE#4520](https://github.com/apache/rocketmq/issues/4520)

In order to make maxMessageSize modification take effect after broker is started, I implenment the function of adjusting maxMessageSize dynamicly.
## Brief changelog
1. add updateEncoderBufferCapacity(int newMaxMessageBodySize), in MessageExtEncoder.
2. add updateMaxMessageSize(PutMessageThreadLocal putMessageThreadLocal), before encode buffer.
3. add junit test  "testDynamicMaxMessageSize()"

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1488,7 +1488,7 @@ public class CommitLog {
     }
 
     public static class MessageExtEncoder {
-        private final ByteBuf byteBuf;
+        private ByteBuf byteBuf;
         // The maximum length of the message body.
         private int maxMessageBodySize;
         // The maximum length of the full message.
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -1709,6 +1709,7 @@ public class CommitLog {
 
         public void updateEncoderBufferCapacity(int newMaxMessageBodySize){
             this.maxMessageBodySize = newMaxMessageBodySize;
+            //Reserve 64kb for encoding buffer outside body
             this.maxMessageSize = Integer.MAX_VALUE - newMaxMessageBodySize >= 64 * 1024 ?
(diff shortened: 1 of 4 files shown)


dongeforever:
@shengminw LGTM.

And it may also need a test for DLedger Mode.

shengminw:
@dongeforever thanks, I will add it.

shengminw:
@dongeforever 
The message encoding buffer in openmessaging/DLedger is of fixed size. https://github.com/openmessaging/dledger/blob/master/src/main/java/io/openmessaging/storage/dledger/store/file/DLedgerMmapFileStore.java
```java
this.localEntryBuffer = ThreadLocal.withInitial(() -> {
    return ByteBuffer.allocate(4194304);
});
```
So temporarily remove the adaptation of DLeger mode.

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -604,6 +604,14 @@ private String generateKey(StringBuilder keyBuilder, MessageExt messageExt) {
         return keyBuilder.toString();
     }
 
+    public void updateMaxMessageSize(PutMessageThreadLocal putMessageThreadLocal) {
+        // dynamically adjust maxMessageSize, but not support DLedger mode temporarily
+        int newMaxMessageSize = this.defaultMessageStore.getMessageStoreConfig().getMaxMessageSize();
+        if (putMessageThreadLocal.getEncoder().getMaxMessageBodySize() != newMaxMessageSize) {
+            putMessageThreadLocal.getEncoder().updateEncoderBufferCapacity(newMaxMessageSize);
+        }
```
dongeforever, line 612: It is better to do some checks to prevent negeative values.

if (newMaxMessageSize < 10) {
  return
}

shengminw: ok, it's neccessary to add this check.

Issue #4550, opened 2022-07-03
the description of getAclConfig command  is inaccurate


meateggmilk:
1. adjust some description
2. remove blank setting to make same as actual setting
![image](https://user-images.githubusercontent.com/87265072/177049000-6dcd84e9-e688-4e44-ace8-f303d4d6853f.png)

lizhiboo:
@meateggmilk IMO, print blank is friendly if this config is not set.

Oliverwqcwrw:
> @meateggmilk IMO, print blank is friendly if this config is not set.

+1

caigy:
> @meateggmilk IMO, print blank is friendly if this config is not set.

+1

lizhiboo:
@meateggmilk thanks for your contribution, i will close this issue, if have any other questions, you can open a new one.

PR #4551, opened 2022-07-03
【ISSUE #4550】adjust description of GetAclConfg command

meateggmilk:

#4550

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/topic/DeleteTopicSubCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/topic/DeleteTopicSubCommand.java
@@ -38,8 +38,8 @@ public class DeleteTopicSubCommand implements SubCommand {
         final String topic
     ) throws InterruptedException, MQBrokerException, RemotingException, MQClientException {
 
-        Set<String> brokerAddressSet = CommandUtil.fetchMasterAndSlaveAddrByClusterName(adminExt, clusterName);
-        adminExt.deleteTopicInBroker(brokerAddressSet, topic);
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/acl/GetAccessConfigSubCommand.java
```diff
@@ -52,20 +52,19 @@ public class GetAccessConfigSubCommand implements SubCommand {
     @Override public Options buildCommandlineOptions(Options options) {
         OptionGroup optionGroup = new OptionGroup();
 
-        Option opt = new Option("b", "brokerAddr", true, "query acl config version for which broker");
+        Option opt = new Option("b", "brokerAddr", true, "query acl config for which broker");
```
tsunghanjacktsai, line 55: Pls help us change the description to "query acl config for specified broker" to make it more formal and consistent with the clusterName one.

tools/src/main/java/org/apache/rocketmq/tools/command/acl/GetAccessConfigSubCommand.java
```diff
@@ -116,8 +115,6 @@ private void printClusterBaseInfo(
                     field.setAccessible(true);
                     if (field.get(config) != null) {
                         System.out.printf("%-1s %-18s: %s\n", "", field.getName(), field.get(config).toString());
-                    } else {
```
tsunghanjacktsai, line 119: I believe leaving it blank makes sense and would not affect its meaning. Do you think it's really necessary to remove it?

tsunghanjacktsai (changes requested):

Issue #4552, opened 2022-07-04
Add asySendBlockMode to implement BackPressure


shengminw:
There are asynchronous sending methods in the current message asysending mechanism:
```java
ExecutorService executor = this.getAsyncSenderExecutor();
try {
    executor.submit(new Runnable() {
        @Override
        public void run() {
            long costTime = System.currentTimeMillis() - beginStartTime;
            if (timeout > costTime) {
                try {
                    sendDefaultImpl(msg, CommunicationMode.ASYNC, sendCallback, timeout - costTime);
                } catch (Exception e) {
                    sendCallback.onException(e);
                }
            } else {
                sendCallback.onException(
                    new RemotingTooMuchRequestException("DEFAULT ASYNC send call timeout"));
            }
        }
    });
} catch (RejectedExecutionException e) {
    throw new MQClientException("executor rejected ", e);
}
```
When the upstream sending message traffic is too heavy, the thread pool will directly refuse to execute and throw  MQClientException.
In this case, the users need to implement a traffic-limit mechanism by themselves, usually  putting the sending thread to sleep for a period of time before sending again. However, this may not be a good solution.
That is to say, asynchronous sending requires a back pressure mechanism.  Within the timeout allowed by the client, the SDK needs to block the sending operation.
This problem is relatively easy to trigger in data synchronization and log collection scenarios.

15712852007:
Imo, if there are too many queued tasks, even if the rejection policy is not triggered, the timeout probability of backlog tasks on the client is still very high, because the queuing time is also within the calculation

shengminw:
> Imo, if there are too many queued tasks, even if the rejection policy is not triggered, the timeout probability of backlog tasks on the client is still very high, because the queuing time is also within the calculation

Yes, after the backpressure mechanism is provided, there will still be timeouts in the case of a large number of requests, but it can be alleviated. 
Besides, IMO, timeout and rejection directly are different in sematic. Users have set the timeout attribute, so this problem can be foreseen, but the return that is rejected directly is not what the users expect.

PR #4553, opened 2022-07-04
[ISSUE#4552] Add asySendBlockMode to implement BackPressure

shengminw:
[ISSUE#4552](https://github.com/apache/rocketmq/issues/4552)

When the upstream sending message traffic is too heavy, the threadpool executor will directly refuse to execute and throw  MQClientException. 

In the IO operation "Netty Remoting", backpressure is implemented by semaphoreAsync:
```java
public void invokeAsyncImpl(final Channel channel, final RemotingCommand request, final long timeoutMillis,
        final InvokeCallback invokeCallback)
        throws InterruptedException, RemotingTooMuchRequestException, RemotingTimeoutException, RemotingSendRequestException {
        long beginStartTime = System.currentTimeMillis();
        final int opaque = request.getOpaque();
        boolean acquired = this.semaphoreAsync.tryAcquire(timeoutMillis, TimeUnit.MILLISECONDS);
        if (acquired) {
            final SemaphoreReleaseOnlyOnce once = new SemaphoreReleaseOnlyOnce(this.semaphoreAsync);
            long costTime = System.currentTimeMillis() - beginStartTime;
            if (timeoutMillis < costTime) {
                once.release();
                throw new RemotingTimeoutException("invokeAsyncImpl call timeout");
            }

            final ResponseFuture responseFuture = new ResponseFuture(channel, opaque, timeoutMillis - costTime, invokeCallback, once);
            this.responseTable.put(opaque, responseFuture);
            try {
                channel.writeAndFlush(request).addListener(new ChannelFutureListener() {
                    @Override
                    public void operationComplete(ChannelFuture f) throws Exception {
                        if (f.isSuccess()) {
                            responseFuture.setSendRequestOK(true);
                            return;
                        }
                        requestFail(opaque);
                        log.warn("send a request command to channel <{}> failed.", RemotingHelper.parseChannelRemoteAddr(channel));
                    }
                });
            } catch (Exception e) {
                responseFuture.release();
                log.warn("send a request command to channel <" + RemotingHelper.parseChannelRemoteAddr(channel) + "> Exception", e);
                throw new RemotingSendRequestException(RemotingHelper.parseChannelRemoteAddr(channel), e);
            }
        } else {
            if (timeoutMillis <= 0) {
                throw new RemotingTooMuchRequestException("invokeAsyncImpl invoke too fast");
            } else {
                String info =
                    String.format("invokeAsyncImpl tryAcquire semaphore timeout, %dms, waiting thread nums: %d semaphoreAsyncValue: %d",
                        timeoutMillis,
                        this.semaphoreAsync.getQueueLength(),
                        this.semaphoreAsync.availablePermits()
                    );
                log.warn(info);
                throw new RemotingTimeoutException(info);
            }
        }
    }
```

I consider to implement backpressure mechanism  relying on BackPressure in NettyRemoting. In this situation, within the timeout allowed by the client, the SDK will block the sending operation, not reject the task. 

## Brief changelog

1. add back pressure mechanism by just run `runnable` in this thread, relying on backpressure in NettyRemoting.
2. add `asySendBlockMode` parameter in `DefaultMQProducer` to let user decide whether to enable this mode.

---

**Notice**: Through the experimental test, the time cost of asynchronous sending is mainly in the IO operation. So, whether  start `runnable` in `DefaultMQProducerImpl` through ThreadPool, there is not much difference in performace.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
+++ b/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
@@ -42,6 +42,7 @@ public class ClientConfig {
     protected String namespace;
     private boolean namespaceInitialized = false;
     protected AccessChannel accessChannel = AccessChannel.LOCAL;
+    private boolean asySendBlockMode = true;
 
     /**
      * Pulling topic information interval from the named server
@@ -338,6 +339,12 @@ public class ClientConfig {
 
     public void setEnableStreamRequestType(boolean enableStreamRequestType) {
         this.enableStreamRequestType = enableStreamRequestType;
+    }
+
+    public boolean isAsySendBlockMode() { return  asySendBlockMode; }
+
+    public void setAsySendBlockMode(boolean asySendBlockMode) {
+        this.asySendBlockMode = asySendBlockMode;
     }
 
     @Override
--- a/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
+++ b/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
@@ -42,7 +42,6 @@ public class ClientConfig {
     protected String namespace;
     private boolean namespaceInitialized = false;
     protected AccessChannel accessChannel = AccessChannel.LOCAL;
-    private boolean asySendBlockMode = true;
 
     /**
      * Pulling topic information interval from the named server
@@ -339,12 +338,6 @@ public class ClientConfig {
 
     public void setEnableStreamRequestType(boolean enableStreamRequestType) {
         this.enableStreamRequestType = enableStreamRequestType;
-    }
-
-    public boolean isAsySendBlockMode() { return  asySendBlockMode; }
-
-    public void setAsySendBlockMode(boolean asySendBlockMode) {
-        this.asySendBlockMode = asySendBlockMode;
     }
 
     @Override
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -492,26 +492,36 @@ public class DefaultMQProducerImpl implements MQProducerInner {
         throws MQClientException, RemotingException, InterruptedException {
         final long beginStartTime = System.currentTimeMillis();
         ExecutorService executor = this.getAsyncSenderExecutor();
-        try {
-            executor.submit(new Runnable() {
-                @Override
-                public void run() {
-                    long costTime = System.currentTimeMillis() - beginStartTime;
-                    if (timeout > costTime) {
-                        try {
-                            sendDefaultImpl(msg, CommunicationMode.ASYNC, sendCallback, timeout - costTime);
-                        } catch (Exception e) {
-                            sendCallback.onException(e);
-                        }
-                    } else {
-                        sendCallback.onException(
-                            new RemotingTooMuchRequestException("DEFAULT ASYNC send call timeout"));
+        Runnable runnable = new Runnable() {
+            @Override
+            public void run() {
+                long costTime = System.currentTimeMillis() - beginStartTime;
+                if (timeout > costTime) {
+                    try {
+                        sendDefaultImpl(msg, CommunicationMode.ASYNC, sendCallback, timeout - costTime);
+                    } catch (Exception e) {
+                        sendCallback.onException(e);
                     }
+                } else {
+                    sendCallback.onException(
+                            new RemotingTooMuchRequestException("DEFAULT ASYNC send call timeout"));
                 }
+            }
 
-            });
+        };
+        try {
+            executor.submit(runnable);
         } catch (RejectedExecutionException e) {
-            throw new MQClientException("executor rejected ", e);
+            if (this.defaultMQProducer.isAsySendBlockMode() && this.asyncSenderExecutor == null) {
+                long costTime = System.currentTimeMillis() - beginStartTime;
+                if (!asyncSenderThreadPoolQueue.offer(runnable, timeout - costTime, TimeUnit.MILLISECONDS)) {
+                    sendCallback.onException(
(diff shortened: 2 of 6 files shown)


client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
```diff
@@ -42,6 +42,7 @@ public class ClientConfig {
     protected String namespace;
     private boolean namespaceInitialized = false;
     protected AccessChannel accessChannel = AccessChannel.LOCAL;
+    private boolean asySendBlockMode = true;
```
cserwen, line 45: It is better to put this parameter in `DefaultMQProducer`

shengminw: thx, I agree with your suggestion. I will modify it.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -1047,37 +1052,42 @@ public void send(final Message msg, final MessageQueue mq, final SendCallback se
...
+
+        if (this.defaultMQProducer.isEnableBackpressureForAsyncMode()  && this.asyncSenderExecutor == null) {
+            runnable.run();
+        } else {
+            try {
+                executor.submit(runnable);
+            } catch (RejectedExecutionException e) {
+                throw new MQClientException("executor rejected ", e);
+            }
         }
```
lizhiboo, line 1091: if enableBackPressureForAsyncMode, it seems like sync send. IMO, backPressure can be triggered by RejectedExecutionException thrown by asyncSenderExecutor.

ni-ze:
IMO, Modify deprecated code is not a good idea.
Maybe [RIP-37](https://shimo.im/docs/m5kv92OeRRU8olqX) can solve your problem.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
shengminw: @lizhiboo Well, time cost of message sending is mainly in the IO operation. So the IO operation remains in an asynchronous mode the sending process can still be regarded as an asynchronous process.
As for your second question, from the users' view, when they revieve the "RejectedExecutionException", they need to solve this problem by themselves, always just sleep the thread for a while. However, this may not be a good solution. We can implement the backpressure in mq better than user implementing by themselves, just throw the exception when "time callout".

shengminw:
@duhenglucky It's a good a idea. I still consider to build a flow control mechanism based on the size of the message memory in the future.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -492,28 +492,33 @@ public void send(final Message msg, final SendCallback sendCallback, final long
...
+            }
 
-            });
-        } catch (RejectedExecutionException e) {
-            throw new MQClientException("executor rejected ", e);
-        }
+        };
 
+        if (this.defaultMQProducer.isEnableBackpressureForAsyncMode()  && this.asyncSenderExecutor == null) {
+            runnable.run();
```
dongeforever, line 514: Why need to check if the asyncSenderExecutor is null here?


If backpressure is enabled, it should keep semantic.

dongeforever:
> 

The new API uses a new protocol model. it is an accomplishment, not a replacement.

BTW, the new API needs a long time to be mature.

Currently, we could also do some necessary polishments to the traditional API without changing the method signatures.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
shengminw: I think if users set custom ThreadPoolExecutors, they may need the ThreadPoolExecutors, not just run in the thread. 

But keeping semantic consitacy is also important, so I think I can fix it , through setting `asyncSenderExecutor` false when user set custom ThreadPoolExecutor.

cserwen:
> > 
> 
> The new API uses a new protocol model. it is an accomplishment, not a replacement.
> 
> BTW, the new API needs a long time to be mature.
> 
> Currently, we could also do some necessary polishments to the traditional API without changing the method signatures.

Agree it.

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
```diff
@@ -133,6 +133,11 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
      */
     private TraceDispatcher traceDispatcher = null;
 
+    /**
+     * Indicate whether to block message when asynchronous sending traffic is too heavy.
+     */
+    private boolean enableBackpressureForAsyncMode = true;
+
```
dongeforever, line 140: By default, this argument should be false to behave as the old client.

And the same time, we could suggest setting enableBackpressureForAsyncMode to true in the example code.

shengminw: ok, I have fix it.

example/src/main/java/org/apache/rocketmq/example/quickstart/Producer.java
```diff
@@ -94,7 +94,8 @@ public static void main(String[] args) throws MQClientException, InterruptedExce
                 /*
                  * if you want to get the send result in a asynchronize way, you can use this send method
                  * {@code
-                 *
```
dongeforever, line 97: It is better to add the suggestions in AsyncProduer

shengminw: thx, I'will modify it.

dongeforever (approved):

PR #4562, opened 2022-07-05
UpdateTopicRouteInfoFromNameServer method in the concurrent lack of detection

zhaowenshuai011:
#4561 
问题：
updateTopicRouteInfoFromNameServer在某些场景下lockNamesrv代码块的内部缺少校验，可能会导致RemotingTooMuchRequestException("sendDefaultImpl call timeout")异常。

描述：
生产者发送消息时，会通过tryToFindTopicPublishInfo获取路由信息，如果没有缓存或没有messageQueue时会调用updateTopicRouteInfoFromNameServer向nameServer查询该topic的路由信息，在此方法内部有一个nameServer锁，这里极端情况下会出现一些问题，如果在同一时刻有很多个线程排队取lockNamesrv锁，假如前面的线程已通过nameServer拉取到路由信息了，在释放锁之后，后续的线程是没有必要重复与nameServer数据交互的，在网络不好等情况下很可能造成后面排队的线程超时(默认值3s)，然后抛throw new RemotingTooMuchRequestException("sendDefaultImpl call timeout")异常。举个较为极端的例子：100个线程使用同一个producer，大家都第一次发送某个topic的消息。


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -671,8 +676,18 @@ private SendResult sendDefaultImpl(
     private TopicPublishInfo tryToFindTopicPublishInfo(final String topic) {
         TopicPublishInfo topicPublishInfo = this.topicPublishInfoTable.get(topic);
         if (null == topicPublishInfo || !topicPublishInfo.ok()) {
-            this.topicPublishInfoTable.putIfAbsent(topic, new TopicPublishInfo());
-            this.mQClientFactory.updateTopicRouteInfoFromNameServer(topic);
+            try {
+                if(lockUpdatePublishInfo.tryLock(LOCK_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS)){
```
lizhiboo, line 680: IMO, add a new lock before lockNamesrv.tryLock do not make effect. We can check topic route info exists or not again after lockNamesrv.tryLock success to avoid redundant getRouteInfoByTopic.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -671,8 +676,18 @@ private SendResult sendDefaultImpl(
     private TopicPublishInfo tryToFindTopicPublishInfo(final String topic) {
         TopicPublishInfo topicPublishInfo = this.topicPublishInfoTable.get(topic);
         if (null == topicPublishInfo || !topicPublishInfo.ok()) {
-            this.topicPublishInfoTable.putIfAbsent(topic, new TopicPublishInfo());
-            this.mQClientFactory.updateTopicRouteInfoFromNameServer(topic);
+            try {
```
duhenglucky, line 679: Add a double-check for this topicRoute, if there's existed topicRoute, return it directly, and it seems that we don't need to add another lock

15712852007: Thank you very much for your reply, but there is such a scenario, the producer during startup, will be intervals of 30 s updated routing information, if add the calibration topicRouter exists within updateTopicRouteInfoFromNameServer method, It may cause topicRouter not to update in time, so I tried to put the lock on a higher level method

15712852007: > Add a double-check for this topicRoute, if there's existed topicRoute, return it directly, and it seems that we don't need to add another lock

Thank you very much for your reply, but there is such a scenario, the producer during startup, will be intervals of 30 s updated routing information, if add the calibration topicRouter exists within updateTopicRouteInfoFromNameServer method, It may cause topicRouter not to update in time, so I tried to put the lock on a higher level method

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
15712852007: > IMO, add a new lock before lockNamesrv.tryLock do not make effect. We can check topic route info exists or not again after lockNamesrv.tryLock success to avoid redundant getRouteInfoByTopic.

Thank you very much for your reply, but there is such a scenario, the producer during startup, will be intervals of 30 s updated routing information, if add the calibration topicRouter exists within updateTopicRouteInfoFromNameServer method, It may cause topicRouter not to update in time, so I tried to put the lock on a higher level method

Oliverwqcwrw:
Today did encounter this problem, the network is bad and many times to send the same topic messages and then throw the RemotingTooMuchRequestException

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
lizhiboo: We can add a parameter isEmpty2FetchFromNamesrv in mQClientFactory.updateTopicRouteInfoFromNameServer method. if isEmpty2FetchFromNamesrv is true, and double-check topic route info is also empty after tryLock, then update from namesrv. if isEmpty2FetchFromNamesrv is false, and double-check topic route info is not empty after tryLock, then do not update from namesrv. MQClientInstance's scheduled task will update topic route info regually.

guyinyou (approved):
The method of modifying to double-check is correct, what you are modifying now is essentially double-check

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
15712852007: It's a good idea, I'll try to optimize it

15712852007:
> The method of modifying to double-check is correct, what you are modifying now is essentially double-check

15712852007:
According to the job log, it is successful. Why is build Failed displayed?

lizhiboo (approved):

lizhiboo:
@15712852007 Travis CI failed, pls check it.

lizhiboo (changes requested):
Travis CI failed.

15712852007:
@duhenglucky @RongtongJin All checks have passed，please help review and merge

caigy (approved):

Hyperkopite:
Encountering same issue here. Looks like this PR didn't get merged in the end. Any plan to fix this in the future?

Issue #4599, opened 2022-07-13
Optimize enableBackpressureForAsyncMode not relying on NettyRemoting


shengminw:
In  [#pull request 4553](https://github.com/apache/rocketmq/pull/4553), I suggest to add  asySendBlockMode to implement BackPressure in producer. And I have implemented this relying on backpressure in NettyRemoting.
```java
if (this.defaultMQProducer.isEnableBackpressureForAsyncMode()) {
    runnable.run();
} else {
    try {
        executor.submit(runnable);
    } catch (RejectedExecutionException e) {
        throw new MQClientException("executor rejected ", e);
    }
}
```
To make a more accurate flow control mechanism, it is neccassary to expose backpressure monitor parameter to the users. 
So I consider to re-build a backpressure mechanism in async-sending message,  not relying on backpressure in NettyRemoting. And at the same time, add message sizes as a backpressure monitor.

### 优化异步消息反压模式 不再复用NettyRemoting的反压
在[#pull request 4553](https://github.com/apache/rocketmq/pull/4553)中，我提议了在客户端增加异步发送反压模式。同时我复用底层的NettyRemoting实现了反压机制。
```java
if (this.defaultMQProducer.isEnableBackpressureForAsyncMode()) {
    runnable.run();
} else {
    try {
        executor.submit(runnable);
    } catch (RejectedExecutionException e) {
        throw new MQClientException("executor rejected ", e);
    }
}
```
为了实现一个更加精确的流控机制，有必要为用户暴露进行反压监控的参数。
因此我考虑在发送消息端重新构建一个反压机制，不复用底层的NettyRemoting。同时增加消息大小也作为反压监控的参数。

PR #4601, opened 2022-07-13
[ISSUE#4599] Optimize enableBackpressureForAsyncMode not relying on NettyRemoting

shengminw:
[issue4599](https://github.com/apache/rocketmq/issues/4599)
In the current backpressure mechanism [pull request#4553](https://github.com/apache/rocketmq/pull/4553), the runner task is running directly relying on backpressure in NettyRemoting.
```java
if (this.defaultMQProducer.isEnableBackpressureForAsyncMode()) {
    runnable.run();
} else {
    try {
        executor.submit(runnable);
    } catch (RejectedExecutionException e) {
        throw new MQClientException("executor rejected ", e);
    }
}
```
To make a more accurate flow control mechanism, I re-build  a backpressure mechanism based on the numbers of asyncSending  messages threads and  size of the message memory.

## Brief changelog
1. Use Semaphore to monitor pressure of sending message.  semaphoreAsyncNum to limit maximum numbers of on-going sending async messages. semaphoreAsyncSize limit maximum message size of on-going sending async messages.
2. reserve using threadpool in backpressure mechanism.


duhenglucky (approved):

hzh0425 (dismissed):

guyinyou:
The back pressure mechanism is a very useful function, but at present, when the traffic of rocketmq is close to the bottleneck, the traffic will drop sharply due to "buzy". I hope to improve this phenomenon. Can you provide the improvement brought by this PR? such as test results

shengminw:
> The back pressure mechanism is a very useful function, but at present, when the traffic of rocketmq is close to the bottleneck, the traffic will drop sharply due to "buzy". I hope to improve this phenomenon. Can you provide the improvement brought by this PR? such as test results

@guyinyou 
IMO, the current design is indeed close to bottleneck, but the processing of messages is always continuous, so it will tend to a steady state. The number of successfully written messages per second does not change much. Only the message sending rate is reduced. If there is no back pressure, there will still be lots of rejections.
Due to limited testing equipment, I just ran a simple test on my laptop.
I set the producer to 4 threads for continuous asynchronous message sending.
At the same time, I set up a thread to monitor the number of successfully sent messages, the number of timeouts messages, and the number of rejected messages within this 1s.

* When **off enableBackpressureForAsyncMode**:
off-enable

* When **on enableBackpressureForAsyncMode**:
on-enable

Although the local test cannot completely simulate the heavy-traffic environment in the server, it can still be seen, that after turning on the backpressure, **the async-sending rate is reduced, but the rate of successful message writing remains unchanged. Besides, a large number of rejections can be avoided.**

guyinyou (approved):

guyinyou:
When I benchmarked rocketmq before, there was a need for back pressure. My approach was to maintain a semaphore to control the concurrent number of asynchronous sending, which achieved a very ideal effect. You can try it.

```
 UniformRateLimiter rateLimiter = new UniformRateLimiter(1024);
 AtomicLong windowsCnt = new AtomicLong(0);
 while(true) {
    try {
        if(Producer.windowsCnt.incrementAndGet() >= windowsSize){
            while (Producer.windowsCnt.get() >= windowsSize) {
                Thread.yield();
            }
        }
        producer.send(new Message(xxxxxxx), new SendCallback() {
            @Override public void onSuccess(SendResult sendResult) {
                windowsCnt.decrementAndGet();
            }

            @Override public void onException(Throwable e) {
                windowsCnt.decrementAndGet();
                throw new RuntimeException(e);
            }
        });
    } catch (Exception e) {
        e.printStackTrace();
    }
 }
```
Of course, this implementation is very rough. In actual scenarios, this rateLimiter should be a dynamic value, which may be increased or decreased according to the number of failed messages sent recently, the delay in sending, etc. If you are interested, you can communicate and discuss together.

shengminw:
@guyinyou It seems that your approach is similar to my implementation, both use the semaphore to control the concurrent number of asynchronous sendings. 
Your idea about dynamic rateLimiter is a great suggestion. Do you mean a dynamic threshold to control whether to block sending?

guyinyou:
> @guyinyou It seems that your approach is similar to my implementation, both use the semaphore to control the concurrent number of asynchronous sendings. Your idea about dynamic rateLimiter is a great suggestion. Do you mean a dynamic threshold to control whether to block sending?

Yes, and it is enough to only need to send failures and send delays

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -491,27 +492,68 @@ public void send(Message msg,
     public void send(final Message msg, final SendCallback sendCallback, final long timeout)
```
Oliverwqcwrw, line 492: I notice that this method is marked as deprecated. Is there any need to modify this method? It may be removed in the future,WDYT

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
```diff
@@ -133,6 +134,21 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
...
+    /**
+     * Indicate whether to block message when asynchronous sending traffic is too heavy.
+     */
+    private boolean enableBackpressureForAsyncMode = false;
+
+    /**
+     * on BackpressureForAsyncMode, limit maximum number of on-going sending async messages
+     */
+    private Semaphore semaphoreAsyncNum = new Semaphore(60000, true);
+
```
dongeforever, line 146: expose the integer value instead of Semaphore directly.

It is better to initialize the semaphore in the constructor of DefaultMQProducerImpl

dongeforever: default to 10000 is enough.

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
```diff
@@ -133,6 +134,21 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
...
+    /**
+     * on BackpressureForAsyncMode, limit maximum message size of on-going sending async messages
+     */
+    private Semaphore semaphoreAsyncSize = new Semaphore(512 * 1024 * 1024, true);
+
```
dongeforever, line 151: default to 100M is enough.

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
shengminw: It's a good suggestion. I have fixed it.

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -142,6 +147,8 @@ public Thread newThread(Runnable r) {
                     return new Thread(r, "AsyncSenderExecutor_" + this.threadIndex.incrementAndGet());
                 }
             });
+        semaphoreAsyncSendNum = new Semaphore(defaultMQProducer.getBackPressureForAsyncSendNum(), true);
+        semaphoreAsyncSendSize = new Semaphore(defaultMQProducer.getBackPressureForAsyncSendSize(), true);
     }
```
dongeforever, line 152: Maybe we need to do some checks for illegal arguments.

for example:
asyncSendNum cannot be smaller than 10;
asyncSize cannot be smaller than 1M;

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -491,27 +506,66 @@ public void send(Message msg,
...
+                            new RemotingTooMuchRequestException("send message tryAcquire semaphoreAsyncNum timeout"));
+                    return;
+                }
+                costTime = System.currentTimeMillis() - beginStartTime;
+                isSemaphoreAsyncSizeAquired = semaphoreAsyncSendSize.tryAcquire(msgLen, timeout - costTime, TimeUnit.MILLISECONDS);
+                if (!isSemaphoreAsyncSizeAquired) {
+                    sendCallback.onException(
+                            new RemotingTooMuchRequestException("send message tryAcquire semaphoreAsyncSize timeout"));
+                    return;
                 }
```
dongeforever, line 552: what will happen if timeout - costTime  is negative?

shengminw: In tryAquire, if timeout is negative, it will return false. I also added a conditional judgment to ensure that `timeout-costTime` is positive

tianliuliu (dismissed):

dongeforever (approved):

PR #4643, opened 2022-07-20
add message property PROPERTY_BORN_TIMESTAMP

ingdex:
add bornTimeStamp property to message when calling MessagingProcessor::sendMessage


tsunghanjacktsai (commented):
Hi @ingdex ,

Is there any reason why you wanna make this change? Could you please explain what scenarios you'd like to apply it in?

PR #4653, opened 2022-07-21
add message property PROPERTY_BORN_TIMESTAMP and set bornTimeStamp in…

ingdex:
add message property PROPERTY_BORN_TIMESTAMP and set bornTimeStamp in sendMessageRequestHeader as PROPERTY_BORN_TIMESTAMP

Currently, org.apache.rocketmq.common.protocol.header.SendMessageRequestHeader.bornTimestamp is set as the timestamp proxy sending message to broker, change it to message's bornTimestamp. 

In RocketMQ 4.0, org.apache.rocketmq.common.protocol.header.SendMessageRequestHeader.bornTimestamp represents the timestamp when the message was born. In RocketMQ 5.0, the bornTimestamp of the message is a message attribute, that is, the timestamp when message was constructed, and there is no bornTimestamp in the SendMessageRequest of RocketMQ 5.0.
Note: When using grpc sendMessage to send multiple messages, the bornTimestamp in the requestHeader is set to the bornTimestamp of the first message, which may not be accurate. When a bornTimestamp is required, the bornTimestamp of the message property should be used.


lollipopjin (approved):
LGTM, already done with decoding milliseconds to TimeStamp on receiving messages.

xdkxlk (commented):
LGTM

RongtongJin (approved):

PR #4752, opened 2022-08-01
[ISSUE #4749] make topic length , consumer group length configable

francisoliverlee:

fix https://github.com/apache/rocketmq/issues/4749

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/Validators.java
+++ b/client/src/main/java/org/apache/rocketmq/client/Validators.java
@@ -33,8 +33,12 @@ import org.apache.rocketmq.common.topic.TopicValidator;
  * Common Validator
  */
 public class Validators {
-    public static final int CHARACTER_MAX_LENGTH = 255;
-    public static final int TOPIC_MAX_LENGTH = 127;
+    public static final int DEFAULT_TOPIC_MAX_LENGTH = 192;
+    public static final int DEFAULT_CONSUMER_GROUP_MAX_LENGTH = 255;
+    // not thread safe
+    public static int CONSUMER_GROUP_MAX_LENGTH = DEFAULT_CONSUMER_GROUP_MAX_LENGTH;
+    // not thread safe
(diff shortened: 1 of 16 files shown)


ni-ze:
```java
 void setMaxTopicLength(int maxLength);
    int getMaxTopicLength();
```
IMO,  Put those two method into **ClientConfig** will be better. MQConsumer,LitePullConsumer and MQConsumerInner contains code about send/consume message only.  Set/get config should use ClientConfig.

guyinyou:
I also think that it is a very good choice to put configuration-related variables in ClientConfig. At the same time, it is necessary to consider how to deal with the inconsistency of the MAX_TOPIC_LENGTH between the producer and the consumer.

client/src/main/java/org/apache/rocketmq/client/Validators.java
```diff
@@ -33,8 +33,12 @@
  * Common Validator
  */
 public class Validators {
-    public static final int CHARACTER_MAX_LENGTH = 255;
-    public static final int TOPIC_MAX_LENGTH = 127;
+    public static final int DEFAULT_TOPIC_MAX_LENGTH = 192;
+    public static final int DEFAULT_CONSUMER_GROUP_MAX_LENGTH = 255;
+    // not thread safe
+    public static int consumerGroupMaxLength = DEFAULT_CONSUMER_GROUP_MAX_LENGTH;
```
YangJodie, line 39: If I actually set this length to 255, the retry topic may be limited by the system topic length. we need reduce the length to safe water.

lizhiboo: It seems that  Validators.checkTopic() method only used in MQProducer. it does not harm sendmessageback logic in consume case.

lizhiboo:
@francisoliverlee following test would not throw exception
```java
AbstractMessingActivityTest.testValidateTopic()
        assertThrows(GrpcProxyException.class, () -> messingActivity.validateTopic(Resource.newBuilder().setName(createString(128)).build()));
```

Issue #3870, opened 2022-02-20
Optimize topic route data notification


xijiu:
# 1.Background

Topic routing data is very important information in RocketMQ, which meta data is stored on the NameServer. And the client (including producer and consumer) obtains the routing info through the rotation training of the NameServer every 30 seconds. Such design can ensure that the latest configuration information is periodically obtained, but there are also some bad cases:

1. If the client frequently requests a topic that does not exist, it will cause each request to send a pull request to the nameServer, increasing the network overhead
2. For some applications, the client may access many topics, and some topics may no longer be accessed after one visit, or the access may be very infrequent. But the client will still pull routing information from the NameServer every time during round-robin training. While increasing network overhead, zombie topics also occupy more memory
3. When the topic routing information changes, it takes up 30 seconds to get the latest data

# 2.Transformation Ideas
Under the premise that the **30-second rotation is unchanged**, the notification and storage of Topic routing data are optimized to a certain extent, which are divided into the following three parts:

* a. **Active notification**   NameServer actively pushes data to the client; but also pay attention to 2 problems

	*  Hotspot data; For example, when the broker restarts, all related topics send notifications and causing data hotspots. The solution is to not trigger the active notification mechanism when the channel closed or created.
	*  When sending an active notification request, check the load and memory of the machine. If the resource is found to be tight, the active notification will be temporarily closed.

* b. **Topic expired** The client sets an expiration mechanism for the topic. If a topic has not been accessed for a period of time, it will be automatically removed from the client.
* c. **Topic not existed** Cache the "non-existent" status of a topic on the Client to ensure that when the client frequently accesses a topic that does not exist, it can hit the local cache and reduce network overhead.

# 3.Detailed Design

## 3.1 Active Notification
NameServer finds that topic routing has changed, and actively notifies Client to pull the latest routing data

* NameServer send Topic name to Client
* After the client receives the message, it sends a request to the NameServer to pull the latest route data of the topic

![rmq中NameServer轮训Topic路由](https://user-images.githubusercontent.com/19780771/154846756-ce064f35-ee5c-48fe-89ab-42a159a0aa21.png)

### 3.1.1 NameServer
All Topic routing data operations are in class `org.apache.rocketmq.namesrv.routeinfo.RouteInfoManager`, The notification is triggered **async**. All topics are rotated every 1 second. If a topic route is found to change, it will actively notifies the client.

#### 3.1.1.1、Topic and Channel
[...]

PR #4772, opened 2022-08-04
[ISSUE #3870] Optimize topic route data notification

xijiu:

Optimize topic route data notification
[Github issue 3870](https://github.com/apache/rocketmq/issues/3870)

## Brief changelog

* NameServer
* Broker
* Client


namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/TopicRouteNotifier.java
```diff
@@ -0,0 +1,92 @@
...
+        for (Channel channel : channelSet) {
+            RemotingCommand remotingCommand = transToCommand(topic);
+            try {
+                remotingServer.invokeOneway(channel, remotingCommand, 50);
+            } catch (Exception e) {
+                log.error("invoke client exception. topic={}, channel={}, error={}", topic, channel, e.toString());
+            }
+        }
+    }
+
```
dongeforever, line 75: Here may need to support multi-topic mode,  that is to notify the channel multi topics at one time.

Maybe the getTopicRoute needs to support multi-mode too.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/TopicRouteNotifier.java
```diff
@@ -0,0 +1,92 @@
...
+        this.routeInfoManager = routeInfoManager;
+        this.remotingServer = remotingServer;
+    }
+
+    /**
+     * if topic route info has changed in the period, then notify client
+     */
+    public void notifyClients() {
+        Map<String, Set<Channel>> topicAndChannelMap = routeInfoManager.getAndResetChangedTopicMap();
+        if (MapUtils.isEmpty(topicAndChannelMap)) {
```
dongeforever, line 53: here may need a dynamic config to control whether to notify or not.

namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -455,6 +470,10 @@ private void createAndUpdateQueueData(final String brokerName, final TopicConfig
                 log.info("topic changed, {} OLD: {} NEW: {}", topicConfig.getTopicName(), existedQD,
                     queueData);
                 queueDataMap.put(brokerName, queueData);
+                // if broker restart, many topic will register, for avoid hot, then ignore
+                if (singleTopicRouteChanged) {
```
dongeforever, line 474: In case of "broker restart/failure", it also needs to be notified.

How about using the multi-node to improve performance and reduce CPU cost?

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -738,19 +768,22 @@ private SendResult sendDefaultImpl(
 
     private TopicPublishInfo tryToFindTopicPublishInfo(final String topic) {
         TopicPublishInfo topicPublishInfo = this.topicPublishInfoTable.get(topic);
+
+        // topic route info not in local cache, then make a request to nameServer
         if (null == topicPublishInfo || !topicPublishInfo.ok()) {
             this.topicPublishInfoTable.putIfAbsent(topic, new TopicPublishInfo());
```
dongeforever, line 774: If the topicPublishInfo is not ok(i.e. the topic does not exist), it will generate RPC for each message, which will cause too many requests to the nameserver, and make the nameserver to be hot.

For the failed topic,  we may need to add a placeholder for it.

Issue #4778, opened 2022-08-04
Design of optimize log transfer protocol


lizhimins:
# Design of log replication transfer protocol

In a distributed system, broker or namesrv nodes may go down, so we introduce a dynamic multiple replica architecture to achieve high availability at the application service level. In the traditional active-standby mode, the two replicas cannot achieve majority election and do not have the ability to switch by itself. However, the cost of three replicas required by Raft is relatively high, the storage layer does not reuse the persistence model of RocketMQ's native implementation, and there are some problems such as inflexible message confirmation. In the design of “RIP 44”, it has been proposed to implement the DLedger Controller consistency agreement on NameServer https://shimo.im/docs/N2A1Mz9QZltQZoAD.

The proposal has the following core elements:

1. Multiple DLedger Controllers elect the master through consensus protocol.

   1. The assignment of the identity of a single broker replica group is completed by the master controller (BrokerId = 0 means the master)
   2. For the non-switching version, you only need to control the controller not to elect a new master. It is controlled by the parameter allowElectMaster, which realizes the dynamic switching between the selected master and the non-selected master version.

2. The Broker obtains the replica group information of the BrokerMemberGroup from the Controller by means of event notification, polling and redirection, and modifies its own identity.

3. In the traditional architecture, the lease mechanism is generally used to avoid multi-master situation.

   Considering the network partition, the Controller is unavailable but not actively downgraded. In this design, without lease mechanism is not adopted, and dual masters cannot be avoided in the case of network partitions.

   For example, if the old master is isolated by the network, and allAckInSyncStateSet / minInSyncReplica is not configured, the message will still be written successfully, and the master and slave are not synchronized.

4. Use Epoch-StartOffset and MaxOffset to determine lastConsistentPoint (Offset) and maintain CommitLog and EpochMap

## Transfer frame format

The newly implemented log service protocol communicates based on frames in the format of

frame length (int) + message type (int) + write timestamp (long) + epoch (long) + payload (byte[], length = total-24) 

![image-20220804143508950](https://user-images.githubusercontent.com/22487634/182796635-e1c1280a-2d5f-456f-baa7-b829417d63c5.png)

### 1. Five states of the state machine

They are Ready - Handshake - Transfer - Suspend - Shutdown

**Ready phase**: After recover the commitLog, consumeQueue, TimerLog and other states, broker will enter the ready state.

[...]

PR #4780, opened 2022-08-04
[ISSUE #4778] Design of optimize log transfer protocol

lizhimins:
[ISSUE #4778] Design of optimize log transfer protocol


common/src/main/java/org/apache/rocketmq/common/EpochEntry.java
```diff
@@ -64,10 +65,10 @@ public void setEndOffset(long endOffset) {
     @Override
     public String toString() {
         return "EpochEntry{" +
-            "epoch=" + epoch +
-            ", startOffset=" + startOffset +
-            ", endOffset=" + endOffset +
-            '}';
+                "epoch=" + epoch +
```
lizhanhui, line 68: Can we use JSON library to generate string representation instead of manual concatenation?

store/src/main/java/org/apache/rocketmq/store/ha/netty/NettyTransferDecoder.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    protected TransferMessage decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+        try {
+            ByteBuf frame = (ByteBuf) super.decode(ctx, in);
+            if (null == frame) {
+                return null;
+            }
+            int bodyLength = frame.readInt();
```
zhouxinyu, line 51: Do you handle the half `TransferMessage`?

store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAClient.java
```diff
@@ -17,152 +17,132 @@
...
-    public void setLocalAddress(String localAddress) {
-        this.localAddress = localAddress;
+    public void updateSlaveId(Long newId) {
+        this.slaveId.set(newId);
     }
 
-    public void updateSlaveId(Long newId) {
-        Long currentId = this.slaveId.get();
-        if (this.slaveId.compareAndSet(currentId, newId)) {
-            LOGGER.info("Update slave Id, OLD: {}, New: {}", currentId, newId);
```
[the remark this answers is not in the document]
lizhimins: Why do we need CAS here？

hzh0425:
Cool, thanks for your work. @lizhimins 
Can you provide a performance comparison and stability test of the old and new versions?
If we switch to the netty version, I'm more curious about how much performance can be improved and how stable it is

PR #4797, opened 2022-08-08
[ISSUE #4771]In HA mode, if you do not truncate consumerOffsetMap, it may cause a bug

### LINKED ISSUE #4771: In HA mode, if you do not truncate consumerOffsetMap, it may cause a bug
**BUG REPORT**
image
1.nodes in syncStatusSet  all not available.
2.enableElectUnCleanMaster = true, slave3 was chosen to be master.
3.slave4 tried to truncate commitLog.
4.slave4  was chosen to be master and continued to append new message to commitLog.
5.consumer started to consume message from consumerOffset.

fujian-zfj:
BUG REPORT
close #4771 
image


store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -81,7 +81,7 @@ public class TimerMessageStore {
     public static final int DAY_SECS = 24 * 3600;
     // The total days in the timer wheel when precision is 1000ms.
     // If the broker shutdown last more than the configured days, will cause message loss
-    public static final int TIMER_WHELL_TTL_DAY = 7;
+    public static final int TIMER_WHEEL_TTL_DAY = 7;
```
RongtongJin, line 84: Modifications not related to this PR can be split

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -168,6 +168,8 @@ public class DefaultMessageStore implements MessageStore {
     private final ConcurrentMap<Integer /* level */, Long/* delay timeMillis */> delayLevelTable =
         new ConcurrentHashMap<Integer, Long>(32);
 
+    private ConcurrentMap<String/* topic@group */, ConcurrentMap<Integer, Long>> offsetTable;
```
RongtongJin, line 171: There is no need to add a consumer offset map here. Perhaps we can modify the consummeQueueStore of the broker module through the register hook.

store/src/main/java/org/apache/rocketmq/store/queue/ConsumeQueueStore.java
```diff
@@ -341,6 +343,50 @@ private void putConsumeQueue(final String topic, final int queueId, final Consum
...
+    public void truncateTimerConsumerOffset(TimerMessageStore timerMessageStore) {
+        if (this.consumeQueueTable.get(TimerMessageStore.TIMER_TOPIC) == null
+            || this.consumeQueueTable.get(TimerMessageStore.TIMER_TOPIC).get(0) == null) {
+            return;
+        }
+        ConsumeQueueInterface consumerQueue = this.consumeQueueTable.get(TimerMessageStore.TIMER_TOPIC).get(0);
+        if (timerMessageStore.getQueueOffset() > consumerQueue.getMaxOffsetInQueue()) {
+            timerMessageStore.setQueueOffset(consumerQueue.getMaxOffsetInQueue());
+        }
+    }
```
RongtongJin, line 355: It would be better to handle timer consumer offset separately in consume queue store.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -646,6 +648,11 @@ public void truncateDirtyFiles(long offsetToTruncate) {
 
         recoverTopicQueueTable();
 
+        // truncate consumer offset
+        truncateConsumerOffsetTable(this.offsetTable);
```
ShannonDing, line 652: truncateConsumerOffsetTable() is OK, the parameter is not required.

hzh0425:
Good point!

dongeforever:
@RongtongJin The truncation may also need to be done when the broker role changes.

RongtongJin:
> @RongtongJin The truncation may also need to be done when the broker role changes.

Yes, when the broker role changes, commit log and consume queue have been truncated, but consumer offset has not been corrected. This PR will correct consumer offset.

store/src/main/java/org/apache/rocketmq/store/TruncateFilesHook.java
```diff
@@ -0,0 +1,14 @@
+package org.apache.rocketmq.store;
```
RongtongJin, line 1: Missing ASF header

lizhanhui:
As I pointed out [here](https://github.com/apache/rocketmq/issues/4771#issuecomment-1211515886), this fix is kind of walkaround. Should we invest more efforts on the design flaw or correct the design issue itself?

lizhanhui (commented):
Need to discuss the solution first.

PR #4881, opened 2022-08-24
send async,time calculation rules Multiple method patches

15712852007:
For issues that were closed yesterday, there are still some missing methods. Please try merging again


15712852007:
@lizhanhui  Please help to merge again, thank you.

lizhanhui:
@15712852007 1, The synchronous path shall also be updated; 2, For async method, we should have a timer, helping to respect timeout specified by the method.  
For the async send, its workflow should be
1. client call async method;
2. create a timeout timer;
3. invoke async rpc; perform transparent retries if timeout is not reached;
4. if RPC is successful within time frame, cancel the timer; if timeout timer fires, call callback with timeout;

15712852007:
> @15712852007 1, The synchronous path shall also be updated; 2, For async method, we should have a timer, helping to respect timeout specified by the method. For the async send, its workflow should be
> 
> 1. client call async method;
> 2. create a timeout timer;
> 3. invoke async rpc; perform transparent retries if timeout is not reached;
> 4. if RPC is successful within time frame, cancel the timer; if timeout timer fires, call callback with timeout;

In my opinion, the introduction of timers may cause unnecessary waste of resources and may cause performance problems, we can make smaller and more reasonable designs based on the current version

lizhanhui:
@15712852007 It would be nice if you describe *smaller and more reasonable designs based on the current version* in finer detail.

15712852007:
> @15712852007 It would be nice if you describe _smaller and more reasonable designs based on the current version_ in finer detail.

@lizhanhui  In my opinion, there are two main areas where the client takes the most time to send messages:
1. The thread is waiting to be scheduled.
2. Network request correlation,
As a user, I care more about the timeout of the program during running, rather than the timeout of the program before it starts running. In my opinion, it is the best choice to eliminate the time consuming before the program runs, because complex solutions often cause performance loss, so it is not as practical as simple solutions

aaron-ai:
Hi, think you for contributing!

You may need to rebase the develop branch to pass the residual status check because we re-defined the mergeability recently.

15712852007:
> Hi, think you for contributing!
> 
> You may need to rebase the develop branch to pass the residual status check because we re-defined the mergeability recently.

The check have been passed, please help merge the branches, thank you

lizhanhui (approved):
OK, let's simplify this part

PR #4903, opened 2022-08-26
[ISSUE 4902]fix-warmMappedFile：Place a safepoint before the loop, the processing effect takes precedence over sleep

isysc1:
Place a safepoint before the loop, the processing effect takes precedence over sleep

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/logfile/DefaultMappedFile.java
+++ b/store/src/main/java/org/apache/rocketmq/store/logfile/DefaultMappedFile.java
@@ -535,10 +535,10 @@ public class DefaultMappedFile extends AbstractMappedFile {
 
         long beginTime = System.currentTimeMillis();
         ByteBuffer byteBuffer = this.mappedByteBuffer.slice();
-        int flush = 0;
-        long time = System.currentTimeMillis();
-        for (int i = 0, j = 0; i < this.fileSize; i += DefaultMappedFile.OS_PAGE_SIZE, j++) {
-            byteBuffer.put(i, (byte) 0);
+        long flush = 0;
+        // long time = System.currentTimeMillis();
+        for (long i = 0, j = 0; i < this.fileSize; i += DefaultMappedFile.OS_PAGE_SIZE, j++) {
+            byteBuffer.put((int) i, (byte) 0);
             // force flush when flush disk type is sync
             if (type == FlushDiskType.SYNC_FLUSH) {
                 if ((i / OS_PAGE_SIZE) - (flush / OS_PAGE_SIZE) >= pages) {
@@ -548,15 +548,15 @@ public class DefaultMappedFile extends AbstractMappedFile {
             }
 
             // prevent gc
-            if (j % 1000 == 0) {
-                log.info("j={}, costTime={}", j, System.currentTimeMillis() - time);
-                time = System.currentTimeMillis();
-                try {
-                    Thread.sleep(0);
-                } catch (InterruptedException e) {
-                    log.error("Interrupted", e);
-                }
-            }
+            // if (j % 1000 == 0) {
+            //     log.info("j={}, costTime={}", j, System.currentTimeMillis() - time);
+            //     time = System.currentTimeMillis();
+            //     try {
+            //         Thread.sleep(0);
+            //     } catch (InterruptedException e) {
+            //         log.error("Interrupted", e);
+            //     }
+            // }
         }
 
         // force flush when prepare load finished


Hccake (commented):
注释掉 sleep 逻辑后，j 变量就没有用了。
另外 i 一定小于 fileSize，而 fileSize 又是 int 类型，现在把 i 改为 long 类型完全没有意义

shy-share:
> 而 fileSize 又是 int 类型，现在把 i 改为 long 类型完全没有意义

@Hccake  filesize是不是int类型和jvm能不能在循环前放置安全点并没有什么关系，它看的就是循环的这个i是不是long类型的
参考这篇[文章](https://juejin.cn/post/6844903878765314061#comment)

Hccake:
> > 而 fileSize 又是 int 类型，现在把 i 改为 long 类型完全没有意义
> 
> @Hccake filesize是不是int类型和jvm能不能在循环前放置安全点并没有什么关系，它看的就是循环的这个i是不是long类型的 参考这篇[文章](https://juejin.cn/post/6844903878765314061#comment)

受教，学习了

lizhanhui:
The existing solution is to write one byte per page as warming. According to https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/reference_guide/using_mlock_to_avoid_page_io, it looks sufficient to mlock;

Another thing, actual empty write is actually preferred for some reason, it would be better to iterate from file tail to head, as the head of the file region is more likely to be written with data in the near future.

shy-share (changes requested):
good.Why not just delete the unwanted code After all, there are git records

shy-share:
> The existing solution is to write one byte per page as warming. According to https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/reference_guide/using_mlock_to_avoid_page_io, it looks sufficient to mlock;

I checked some information, and some places said that mlock alone cannot guarantee that memory can be locked, because there may be copy-on-write problems. I'm not sure if this is the problem.
[refer to this article](https://www.cnblogs.com/lixiaofei1987/p/3205002.html)

lizhanhui:
@shy-share The article you posted may talk about anonymous memory regions.  For file-backed memory area, mlock should be the behavior discussed in the redhat doc.

Knowden:
why the comment of sleep call is 'prevent gc'? In fact it's helping gc, right?

shy-share:
> why the comment of sleep call is 'prevent gc'? In fact it's helping gc, right?

@Knowden  In fact, it can't prevent gc, it main purpose is to increase gc frequency and reduce single gc time.
refer to [why-thread-sleep0-can-prevent-gc-in-rocketmq](https://stackoverflow.com/questions/53284031/why-thread-sleep0-can-prevent-gc-in-rocketmq)

Oliverwqcwrw:
Would it be better to place safepoint if we had GC problems in the current situation ?

lizhanhui:
@Oliverwqcwrw Why not perform an A-B  test between the current solution and the proposed one, let data talk.

Oliverwqcwrw:
> @Oliverwqcwrw Why not perform an A-B test between the current solution and the proposed one, let data talk.

agree with, The data makes it easier for us to choose specific solutions

Oliverwqcwrw:
> @Oliverwqcwrw Why not perform an A-B test between the current solution and the proposed one, let data talk.

agree with, The data makes it easier for us to choose specific solutions

shy-share:
@lizhanhui @Oliverwqcwrw This [article](https://mp.weixin.qq.com/s/AhoTKmXfRW0RDC3rS5SZcw) has done a simple test, it seems that there is no big difference, and it is only limited to versions below jdk8, this problem has been optimized in jdk10

lizhanhui:
@shy-share Pretty good analysis. Well done.

lizhanhui:
@shy-share the article seems deprecated https://stackoverflow.com/questions/72753599/counted-uncounted-loops-and-safepoints-is-while-i-someint-considered-u 
loop with a long variable becomes COUNTED after JDK16.

shy-share:
> @shy-share the article seems deprecated https://stackoverflow.com/questions/72753599/counted-uncounted-loops-and-safepoints-is-while-i-someint-considered-u loop with a long variable becomes COUNTED after JDK16.

@lizhanhui ok，so I think it is still necessary to modify this code. For jdk8, there does not seem to be a big performance gap after modification, but in order to be compatible with subsequent versions of jdk, it seems better to do so

shy-share:
@lizhanhui @Oliverwqcwrw 

i write a benchmark code，refer to [this](https://juejin.cn/post/7141541647447228429)

Oliverwqcwrw (approved):
I think it's a little easier to understand than Sleep's way

lizhanhui:
@Oliverwqcwrw Either Uncounted Loop or Sleep(0) is fine, IMO. Be sure to add proper documentation, explaining the purpose of  `sleep(0)` or `long-uncounted loop`.

Oliverwqcwrw:
> @Oliverwqcwrw Either Uncounted Loop or Sleep(0) is fine, IMO. Be sure to add proper documentation, explaining the purpose of `sleep(0)` or `long-uncounted loop`.

agree with :)

lizhanhui (approved):

kimmking:
实际上，这里的改动无效：for循环里有方法调用，就不存在这个问题。方法调用栈退出的地方，是可以加check safepoint的。

kimmking:
for循环退不出的场景，也可以通过-XX:+UseCountedLoopSafepoints解决，不需要改代码。

shy-share:
> 实际上，这里的改动无效：for循环里有方法调用，就不存在这个问题。方法调用栈退出的地方，是可以加check safepoint的。

方法可能会被JIT编译器优化，从而导致无法正确识别safepoint

shy-share:
> for循环退不出的场景，也可以通过-XX:+UseCountedLoopSafepoints解决，不需要改代码。

一般很少修改jdk参数，如果线上用

yinyinnie:
循环体内不是已经有方法调用了嘛： `byteBuffer.put((int) i, (byte) 0);`  难道这个方法调用不能当做“安全点”？
我猜测不能当安全点的原因是：JIT在编译byteBuffer.put((int) i, (byte) 0); 这段代码的时候，会进行内联操作？

yinyinnie:
还有一个疑问，我理解无论是 1. Thread.sleep 2. 还是把int 换成long, 都是为了手动在for循环体内放置一个安全点。 为什么手动放置一个安全点就达到优化性能的效果了呢？

原因难道是这样：

有安全点的话，虚拟机通过“短暂的停顿”，就能枚举需要被回收的对象，所有“短暂的停顿”所累积的时间记为t1
如果不放置安全点，可能会导致full GC, 将GC的时间记为t2
t1 远小于 t2?
如果原因真是这样，想请问 t1 和 t2之间的量级差多少？

TonyMistark:
https://www.zhihu.com/question/287421003/answer/1944526317195343386 从这个老哥文章来瞻仰源码来的涨知识了

Issue #3799, opened 2022-01-26
[RIP-30] Support Compaction topic


ltamber:
The issue tracker is **ONLY** used for bug report(feature request need to follow [RIP process](https://github.com/apache/rocketmq/wiki/RocketMQ-Improvement-Proposal)). Keep in mind, please check whether there is an existing same report before your raise a new one.

Alternately (especially if your communication is not a bug report), you can send mail to our [mailing lists](http://rocketmq.apache.org/about/contact/). We welcome any friendly suggestions, bug fixes, collaboration and other improvements.

Please ensure that your bug report is clear and that it is complete. Otherwise, we may be unable to understand it or to reproduce it, either of which would prevent us from fixing the bug. We strongly recommend the report(bug report or feature request) could include some hints as the following:

**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently RocketMQ does not support the compaction topic, which results in that if the application(such as connector/streams) uses messages to store state and only needs the latest state data at a certain time , it needs to pull all the messages from the broker, this process will consume a lot of time; At the same time, a lot of data is not required by the client. Therefore, we need a special data cleaning mechanism, which is to compress the same type of data (same key), and only keep the latest data, In this way, only a small amount of data is loaded when the application starts.

2. Provide any additional detail on your proposed use case for this feature.

refer to [dev-maillist](https://lists.apache.org/thread/c85oj18whdjnxc331n1fxqhb8z50bpnm)

3. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

4. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:

- [sub-task1-issue-number](example_sub_issue1_link_here): sub-task1 description here, 
- [sub-task2-issue-number](example_sub_issue2_link_here): sub-task2 description here,
- ...

duhenglucky:
https://github.com/apache/rocketmq/wiki/RIP-30-Support-Compaction-topic

ferrirW:
I am very interested in this. How is the current progress? Is there something I can get involved in

ltamber:
> I am very interested in this. How is the current progress? Is there something can get involved in

almost done, the code will be pushed soon.

PR #4998, opened 2022-09-04
Support replicate between master and slave

ltamber:
optimize #3799
when rebuild slave, then slave will replicate message from master.


duhenglucky (approved):
Temporary branch, merge first

Issue #4949, opened 2022-08-31
Fix 'Address already in use' in AutoSwitchRoleIntegrationTest#testCheckSyncStateSet


aaron-ai:
https://github.com/apache/rocketmq/runs/8110128184?check_suite_focus=true

![image](https://user-images.githubusercontent.com/19537356/187649959-e261306e-fb82-4f23-a964-42fed1470e94.png)

PR #5000, opened 2022-09-05
[ISSUE#4949] fix Address already in use

complone:

Related to [issue-4949](https://github.com/apache/rocketmq/issues/4949)

Fixed the port occupancy problem caused by the different order of shutdown calls
## Verifying this change

WeChat34a292217717a19ca08b5fab2b751e6c

### DIFF
--- a/test/src/test/java/org/apache/rocketmq/test/autoswitchrole/AutoSwitchRoleBase.java
+++ b/test/src/test/java/org/apache/rocketmq/test/autoswitchrole/AutoSwitchRoleBase.java
@@ -18,8 +18,10 @@
 package org.apache.rocketmq.test.autoswitchrole;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.net.ServerSocket;
 import java.net.SocketAddress;
 import java.util.ArrayList;
 import java.util.List;
@@ -59,7 +61,9 @@ public class AutoSwitchRoleBase {
     protected List<BrokerController> brokerList;
     private SocketAddress BornHost;
     private SocketAddress StoreHost;
-
+    private static Integer No= 0;
+    
+    
     protected void initialize() {
         this.brokerList = new ArrayList<>();
         try {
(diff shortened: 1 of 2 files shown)


lizhanhui:
Hi @complone , nextPort() can only fix the issue when unit tests are running in sequential order. It would be better if we ask the OS to pick up an available port. 
Aka, socket.listen(0).

complone:
> 细节

Hi, do you mean to apply for a free available port as follows:
```
            ServerSocket serverSocket = new ServerSocket(0);
            int port = serverSocket.getLocalPort();
```

aaron-ai:
> > 细节
> 
> Hi, do you mean to apply for a free available port as follows:
> 
> ```
>             ServerSocket serverSocket = new ServerSocket(0);
>             int port = serverSocket.getLocalPort();
> ```

Thank you for contributing!

It seems that the only way to find an available port is try to occupy a port using 0 and release it, but this approach is not thread-safe. You could get more information from the [deprecation of SocketUtils#findAvailableTcpPort](https://github.com/spring-projects/spring-framework/issues/28052)

lizhanhui:
>  only way to find an available port is try to occupy a port using 0 and release it, but this approach is not thread-safe.

No, we should not use this approach. Modify the main source code to accept a configured listening port `0` and reflect/update the actual listening port in server-config objects. Use the config in test cases as we do in `RemotingServerTest`

aaron-ai:
> > only way to find an available port is try to occupy a port using 0 and release it, but this approach is not thread-safe.
> 
> No, we should not use this approach. Modify the main source code to accept a configured listening port `0` and reflect/update the actual listening port in server-config objects. Use the config in test cases as we do in `RemotingServerTest`

I agree with that.

complone:
> > only way to find an available port is try to occupy a port using 0 and release it, but this approach is not thread-safe.
> 
> No, we should not use this approach. Modify the main source code to accept a configured listening port `0` and reflect/update the actual listening port in server-config objects. Use the config in test cases as we do in `RemotingServerTest`
I think such [changes](https://github.com/apache/rocketmq/pull/5000/files#diff-ee85830d5ccf3051cea017b2ea12299dd9c282bade28bc50061490f963c5c1ffR77) are in line with expectations. In fact, the default value of nettyServerConfig.getListenPort is 0. We do not need to implement the method for netty port compatibility in NettyRemotingServer.

complone:
As shown here, if the port of nettyserver is temporarily occupied by port 0, we need to add additional logic to release port 0 in NettyRemoteServer

https://github.com/complone/rocketmq/tree/fix-issue4949-port

lizhanhui:
>  we need to add additional logic to release port 0 in NettyRemoteServer

Read https://www.baeldung.com/cs/binding-available-ports#different-ways-to-bind-a-port

> But I am worried that this method is too invasive because we need to interact based on the Remoting protocol, which will cause the thread timing to be out of order.

Actually, this is the standard way to run tests in parallelism. I've spent some time investigating the potential changes required, finding we indeed need a few changes to the current code and dledger dependency. I am OK with the current status of this PR now that there is no achievable solution with changes less than 10 lines

lizhanhui (approved):

PR #5014, opened 2022-09-07
[ISSUE #5013] allow broker to bind specified network interface

### LINKED ISSUE #5013: Bind broker to specific network interface
When we set brokerIP1 in config file, broker will not only listen specified IP but also other network interface. This may cause security problem. It is necessary to add a new configuration item to make broker only bind to a specific IP.

ShadowySpirits:

close #5013 

## Brief changelog

Add a new configuration item "bindIP" to make rocketmq only bind to a specific IP.

## Verifying this change

Does this pr affect the original behavior?
No. The default value of the new config is 0.0.0.0, which means binding all network interfaces.

What are the limitations of the new configuration items?
It must be "0.0.0.0" or any IP assigned to a network interface. And if it's not 0.0.0.0, it needs to be the same as brokerIP1 to avoid potential network problems.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
@@ -138,12 +138,6 @@ public class BrokerStartup {
                 System.exit(-2);
             }
 
-            if (!nettyServerConfig.getBindIP().equals("0.0.0.0") &&
-                !nettyServerConfig.getBindIP().equals(brokerConfig.getBrokerIP1())) {
-                System.out.printf("Broker bind ip: %s should be 0.0.0.0 or equal to broker ip1: %s", nettyServerConfig.getBindIP(), brokerConfig.getBrokerIP1());
-                System.exit(-2);
-            }
-
             String namesrvAddr = brokerConfig.getNamesrvAddr();
             if (null != namesrvAddr) {
                 try {
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
@@ -18,6 +18,11 @@ package org.apache.rocketmq.broker;
 
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.joran.JoranConfigurator;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.util.Properties;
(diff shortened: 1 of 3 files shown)


broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
```diff
@@ -139,6 +138,12 @@ public static BrokerController createBrokerController(String[] args) {
                 System.exit(-2);
             }
 
+            if (!nettyServerConfig.getBindIP().equals("0.0.0.0") &&
+                !nettyServerConfig.getBindIP().equals(brokerConfig.getBrokerIP1())) {
```
lizhanhui, line 142: The host to bind may be different with wildcard address and broker IP1. Broker IP1 is for clients to connect to brokers. For example, in the cloud environment, brokerIP1 is configured to its public IP. 
The ECS itself may have multiple network interfaces wiring virtually to different networks.

ShadowySpirits: In the cloud environment, if public IP is assigned to one virtual network interface, just set brokerIP1 and bindIP to public IP. More generally case is cloud host use nat, which means you can not bind to public IP directly because of OS limition.

image

In this case, you can set brokerIP1 to public IP and set bindIP to 0.0.0.0. The same is true for binding to multiple network interfaces.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
```diff
@@ -240,15 +239,18 @@ public void initChannel(SocketChannel ch) {
...
+                log.info("Server is listening {}:{}", this.nettyServerConfig.getBindIP(), this.nettyServerConfig.getListenPort());
             }
             this.remotingServerTable.put(this.nettyServerConfig.getListenPort(), this);
-        } catch (InterruptedException e1) {
-            throw new RuntimeException("this.serverBootstrap.bind().sync() InterruptedException", e1);
+        } catch (Exception e) {
+            if (e instanceof BindException) {
+                throw new RuntimeException("bind to " + this.nettyServerConfig.getBindIP() + ":" + this.nettyServerConfig.getListenPort() + " failed", e);
+            }
+            throw new RuntimeException("this.serverBootstrap.bind().sync() InterruptedException", e);
```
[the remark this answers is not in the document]
lizhanhui: Let's simply propagate/wrap the internal IllegalStateException, with our intention attached.

lizhanhui (dismissed):

lizhanhui (approved):

PR #5029, opened 2022-09-08
[ISSUE #5028] Add echo server for broker and name server to check service status

### LINKED ISSUE #5028: [feature] Add echo server to check service status
At present, rocketmq lacks means to quickly detect network connectivity and service status. SRE usually suggest user ping broker or name server ip, or telnet their port. But sometimes this way doesn't work out. eg. icmp is allowed but tcp not, or tcp connection can be established but following packets cannot be processed.

Therefore, broker and name server needs to provide a way to detect whether the service is currently available. And it should be as simple as possible. OS build-in command is enough, don't bother users to install additional tools.

The example is as follows:

Snipaste_2022-09-08_11-38-53

ShadowySpirits:
Tested on windows, linux(debian, centos) and macos.
close #5028

### DIFF
--- a/namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
+++ b/namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
@@ -18,8 +18,8 @@ package org.apache.rocketmq.namesrv.processor;
 
 import io.netty.channel.ChannelHandlerContext;
 import java.io.UnsupportedEncodingException;
-import java.util.Properties;
 import java.util.List;
+import java.util.Properties;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.rocketmq.common.DataVersion;
 import org.apache.rocketmq.common.MQVersion;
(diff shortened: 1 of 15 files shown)


remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyDecoder.java
```diff
@@ -45,6 +46,14 @@ public Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
             }
             return RemotingCommand.decode(frame);
         } catch (Exception e) {
+            in.resetReaderIndex();
+            // user input may be HiRMQ, HiRMQ\n or HiRMQ\r\n
+            if (in.readableBytes() >= 5 || in.readableBytes() <= 7) {
+                String magicCode = in.readCharSequence(5, StandardCharsets.UTF_8).toString();
+                if (magicCode.equalsIgnoreCase("HiRMQ")) {
```
Oliverwqcwrw, line 53: I think that it was important to let the user know that the response was sent by `hirmq`, so It needs to be marked on the document

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyDecoder.java
```diff
@@ -45,6 +46,14 @@ public Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
             }
             return RemotingCommand.decode(frame);
         } catch (Exception e) {
+            in.resetReaderIndex();
```
lizhanhui, line 49: Features should not rely on malformed frames.  This is the hot data path, and is likely to impact performance too much.

namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/DefaultRequestProcessor.java
```diff
@@ -674,4 +676,10 @@ private RemotingCommand getClientConfigs(ChannelHandlerContext ctx, RemotingComm
         return response;
     }
 
+    private RemotingCommand healthCheck(ChannelHandlerContext ctx, RemotingCommand request) {
```
lizhanhui, line 679: Can we just build a command line script, querying existing broker status info? The server just respond with memory info to the cli,

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyDecoder.java
ShadowySpirits: Good idea, I have added this feature to document.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyDecoder.java
ShadowySpirits: Under normal circumstances, there will not be many malformed frames. So this feature has almost no impact on performance. If there are a lot of malformed frames, it may be a packet flood attack. We should ban attacker ip.

@zhouxinyu do you have any better suggestions?

mxsm: > Under normal circumstances, there will not be many malformed frames. So this feature has almost no impact on performance. If there are a lot of malformed frames, it may be a packet flood attack. We should ban attacker ip.
> 
> @zhouxinyu do you have any better suggestions?

How about use RMQ admin CL?  @ShadowySpirits

ShadowySpirits: > > Under normal circumstances, there will not be many malformed frames. So this feature has almost no impact on performance. If there are a lot of malformed frames, it may be a packet flood attack. We should ban attacker ip.
> > @zhouxinyu do you have any better suggestions?
> 
> How about use RMQ admin CL? @ShadowySpirits

This tool is designed for application developers who use rocketmq sdk. They may not have experience using mqadmin, even if their runtime doesn't have jdk. So I hope to be able to use linux built-in software to achieve this function. @mxsm

Issue #5091, opened 2022-09-15
RocketMQ Store Module Improvement


lizhimins:
RocketMQ Store Module Improvement. If you want more information, please visit https://github.com/apache/rocketmq/issues/5087

- [x] [Store-A] Speed up Broker initialization by concurrently loading ConsumeQueue
- [ ] [Store-B] Support automatic deletion of ConsumeQueue files without valid data to save disk space
- [ ] [Store-C] WIP

Store-A: [#5093] When the number of ConsumeQueues on the Broker is large, the startup speed is very slow and the recovery time is long. After testing, it takes about 150 seconds to "cold recover" 200,000 ConsumeQueue files in a virtual machine (16C 32G 1T SSD) environment. After using this optimization, it only takes 48 seconds, saving 2/3 of the startup time.

Store-B: There are a large number of topics (include RETRY and DLQ) that are no longer used on the Broker occupies a lot of disk space. We use scheduled tasks to delete these index files, but after deleting the ConsumeQueue, the offset will start from 0 again. There are some side effects for subscription groups that consume using pull mode.

azhsmesos:
I'd like to try Store B. Can I try it？

nowinkeyy:
I'm interested in Store-A. Can I have a try? 😃

shy-share:
@lizhimins @azhsmesos Has store-B been resolved?

azhsmesos:
No one assigned it to me, so I think it had been claimed by someone else and I didn't do it. sorry~

azhsmesos:
> @lizhimins @azhsmesos Has store-B been resolved?

No one assigned it to me, so I think it had been claimed by someone else and I didn't do it

shy-share:
> > @lizhimins @azhsmesos Has store-B been resolved?
> 
> No one assigned it to me, so I think it had been claimed by someone else and I didn't do it. sorry~

hhhhhhhhhhhhhhhhhh，in fact, you don’t need to wait for others to assign you, you can directly submit a PR.

PR #5093, opened 2022-09-16
[ISSUE #5091] [Store-A] Speed up Broker initialization by concurrently loading ConsumeQueue

lizhimins:


ShannonDing (approved):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -1629,22 +1629,33 @@ private boolean isTempFileExist() {
         return file.exists();
     }
 
-    private void recover(final boolean lastExitOK) {
-        long recoverCqStart = System.currentTimeMillis();
-        long maxPhyOffsetOfConsumeQueue = this.recoverConsumeQueue();
-        long recoverCqEnd = System.currentTimeMillis();
+    private boolean recover(final boolean lastExitOK, final boolean recoverConcurrently) {
```
RongtongJin, line 1632: How about keeping the original function signature？ We can also get the brokerConfig  in the recover function.

RongtongJin, line 1632: How about keeping the original function signature？ We can also get the brokerConfig in the recover function.

lizhimins: ok, I will fix it

RongtongJin (approved):

caigy (approved):

Oliverwqcwrw (approved):

Issue #5162, opened 2022-09-22
There may be a bug in DefaultMessageStore#getMessage


Knowden:
`maxFilterMessageCount` is calculated as `maxMsgNums * ConsumeQueue.CQ_STORE_UNIT_SIZE` and it's seems like meaning the maxFilterMessageSize.

```java
final int maxFilterMessageCount = Math.max(16000, maxMsgNums * ConsumeQueue.CQ_STORE_UNIT_SIZE);
```

and it's used to judge wether this pull has scanned enough message as follow
```java
if (cqUnit.getQueueOffset() - offset > maxFilterMessageCount) {
    break;
}
```
the cqUnit.queueOffset is logic offset of consumer queue which is calculated as follow
```java
long queueOffset = (sbr.getStartOffset() + sbr.getByteBuffer().position() -  relativePos) / CQ_STORE_UNIT_SIZE;
CqUnit cqUnit = new CqUnit(queueOffset,
        sbr.getByteBuffer().getLong(),
        sbr.getByteBuffer().getInt(),
        sbr.getByteBuffer().getLong());
```
Is there something going wrong? or if I made some mistake when reading source code and submitting issue, please figure it out thanks

HScarb:
IMO you're right.
`maxFilterMessageCount` is physical offset of `ConsumeQueue` file, while `cqUnit.getQueueOffset() - offset` is logical offset

Knowden:
> For some pulled messages for maxMsgNums could be filter-out. so `Message Count ` is not a good way to limit. Change Count-Limit to Bytes-Limit, you can check code on 4.9.x.
> 
> https://github.com/apache/rocketmq/blob/c88686f13e6f4acef1f00c6d3e18589683ec3368/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java#L623
> 
> but it seems need some updates on
> 
> ```java
> if ((cqUnit.getQueueOffset() - offset) *  CQ_STORE_UNIT_SIZE  > maxFilterMessageCount) {
>     break;
> }
> ```
ok. I will submit a PR to fix this later on.

Knowden:
https://github.com/apache/rocketmq/pull/5171

ni-ze:
LGTM~

PR #5171, opened 2022-09-23
[ISSUE #5162] Fix bug about DefaultMessageStore maxFilterMessageCount calculating

Knowden:

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -729,7 +729,7 @@ public class DefaultMessageStore implements MessageStore {
                 status = GetMessageStatus.OFFSET_OVERFLOW_BADLY;
                 nextBeginOffset = nextOffsetCorrection(offset, maxOffset);
             } else {
-                final int maxFilterMessageCount = Math.max(16000, maxMsgNums * ConsumeQueue.CQ_STORE_UNIT_SIZE);
+                final int maxFilterMessageSize = Math.max(16000, maxMsgNums * consumeQueue.getUnitSize());
                 final boolean diskFallRecorded = this.messageStoreConfig.isDiskFallRecorded();
 
                 long maxPullSize = Math.max(maxTotalMsgSize, 100);
@@ -764,7 +764,7 @@ public class DefaultMessageStore implements MessageStore {
 
                             boolean isInDisk = checkInDiskByCommitOffset(offsetPy, maxOffsetPy);
 
-                            if ((cqUnit.getQueueOffset() - offset) * ConsumeQueue.CQ_STORE_UNIT_SIZE > maxFilterMessageCount) {
+                            if ((cqUnit.getQueueOffset() - offset) * consumeQueue.getUnitSize() > maxFilterMessageSize) {
                                 break;
                             }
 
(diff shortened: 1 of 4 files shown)


store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -764,7 +764,7 @@ public GetMessageResult getMessage(final String group, final String topic, final
 
                             boolean isInDisk = checkInDiskByCommitOffset(offsetPy, maxOffsetPy);
 
-                            if (cqUnit.getQueueOffset() - offset > maxFilterMessageCount) {
+                            if ((cqUnit.getQueueOffset() - offset) * ConsumeQueue.CQ_STORE_UNIT_SIZE > maxFilterMessageCount) {
```
HScarb, line 767: IMO, it's better to modify `maxFilterMessageCount` like this
```java
final int maxFilterMessageCount = Math.max(800 /* 16000 / 20 */, maxMsgNums);
```

There are 2 kinds of ConsumeQueue in RocketMQ 5.0
`ConsumeQueue` which unit size is 20, and `BatchConsumeQueue` which unit size is 46.
`(cqUnit.getQueueOffset() - offset) * ConsumeQueue.CQ_STORE_UNIT_SIZE` may be confused

Knowden: > For some pulled messages for maxMsgNums could be filter-out. so `Message Count ` is not a good way to limit. Change Count-Limit to Bytes-Limit, you can check code on 4.9.x.
> 
> https://github.com/apache/rocketmq/blob/c88686f13e6f4acef1f00c6d3e18589683ec3368/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java#L623
> 
> but it seems need some updates on
> 
> ```java
> if ((cqUnit.getQueueOffset() - offset) *  CQ_STORE_UNIT_SIZE  > maxFilterMessageCount) {
>     break;
> }
> ```
I'm agree with HScarb cause UNIT_SIZE will be different between BatchConsumeQueue and ConsumeQueue. @francisoliverlee so I wander why we have to use Byte-Limit rather than Count-Limit

lizhanhui: count-based comparison is more understandable. If there are implementation-specific subtle differences, there should be a concept to abstract away from them.

lizhanhui (changes requested):
count-based comparison is more understandable. If there are implementation-specific subtle differences, there should be a concept to abstract away from them.

lizhanhui (approved):

Knowden:
@lizhanhui some checks of workflow were not successful. I have runned these failed test local and successed. I want to re-run this workflow job but it seems that I don't have permission to do this. Could you pls help me re-run this job

ltamber (approved):

caigy (approved):

Knowden:
> @Knowden Pls modify title as instructions in pull request template.

done

Oliverwqcwrw (approved):
LGTM

ShannonDing (approved):

tsunghanjacktsai (approved):

RongtongJin (approved):

Issue #5095, opened 2022-09-16
RocketMQ Remoting Module Improvement


lizhimins:
RocketMQ Remoting Module Improvement. If you want more information, please visit [#5087](https://github.com/apache/rocketmq/issues/5087)

- [x] [Remoting-A] Support logging rpc distribution in remoting protocol
- [x] [Remoting-B] Support socks proxy in remoting client
- [x] [Remoting-C] Support output thread exception information to the error log
- [x] [Remoting-D] Not handle more request when channel write ability changed to unwritable

Remoting-A: [#5114] Logging rpc distribution can help developer quickly locate related requests with problems
Remoting-B: [#5123] In some special network scenarios, the client needs to connect to the server by proxy
Remoting-C: [#5129] When thread abnormal exit, output thread exception information to the error log
Remoting-D: [#5176] To avoid the system from going OOM we have to override the method channelWritabilityChanged of the ChannelInboundHandler and do back pressure handling. The outbound data will be queued in an internal buffer maintained by netty. [Referral link](https://netty.io/4.1/api/io/netty/channel/Channel.html#isWritable--)

PR #5176, opened 2022-09-23
[ISSUE #5095] [Remoting-D] Not handle more request when channel write ability changed to unwritable

lizhimins:


lizhanhui (approved):

chenjinkai:
more detail about this problem [https://stackoverflow.com/questions/53441150/netty-back-propogate-pressure-in-channel-handler-pipeline-to-slow-down-sender/53442091#53442091](detail)

Issue #5254, opened 2022-10-09
ServiceProvider Stream Closure Optimization


RapperCL:
# serviceProvider
1 inputStream close 
The load and loadClass methods do not actively close the inputStream
<img width="1049" alt="image" src="https://user-images.githubusercontent.com/44110731/194739366-39d9ee61-f3b5-4a60-9e28-47e9ea57f2d2.png">
2 BufferedReader close
When the BufferReader file stream reads an exception, the close() method will not be executed
<img width="1143" alt="image" src="https://user-images.githubusercontent.com/44110731/194739416-c7d66212-2230-478d-ab7b-f5299e24a57b.png">

We should actively close the byte stream inputStream and character stream BufferedReader, and consider exceptions.

PR #5255, opened 2022-10-09
[ISSUE #5254] Stream close optimization

RapperCL:
[issue](https://github.com/apache/rocketmq/issues/5254)


lizhanhui (changes requested):
Why not use try-with-resource?

RapperCL:
> Why not use try-with-resource?

Hmm, I'll change it

RapperCL:
> > Why not use try-with-resource?
> 
> Hmm, I'll change it

I've optimized using try-with-resource and removed the optional if null judgment from the method

lizhanhui (approved):

lizhanhui (approved):

Issue #5415, opened 2022-10-28
TransactionalMessageBridge.writeOp Duplicate recording of metrics data 


RapperCL:
In the putMessageReturnResult method, the relevant data has been recorded, and the result is repeated in the writeOp method.
![image](https://user-images.githubusercontent.com/44110731/198508045-28745cb7-ce06-407d-8044-1c4d9cd692ba.png)
![image](https://user-images.githubusercontent.com/44110731/198508132-519c420e-84b8-4e50-a2fa-1cf98d8584ab.png)

RongtongJin:
@RapperCL Good catch!

PR #5416, opened 2022-10-28
[ISSUE #5415]fix duplicate records

RapperCL:
[issue](https://github.com/apache/rocketmq/issues/5415)

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageBridge.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageBridge.java
@@ -303,9 +303,7 @@ public class TransactionalMessageBridge {
 
     public boolean writeOp(Integer queueId,Message message) {
         MessageQueue opQueue;
-        if (opQueueMap.containsKey(queueId)) {
-            opQueue = opQueueMap.get(queueId);
-        } else {
+        if ((opQueue = opQueueMap.get(queueId)) == null) {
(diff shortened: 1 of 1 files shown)


shy-share (changes requested):
`        if ((opQueue = opQueueMap.get(queueId)) == null) {`
You write this way although it looks cleaner, but it is not easy to read the code

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageBridge.java
```diff
@@ -303,9 +303,7 @@ private TopicConfig selectTopicConfig(String topic) {
 
     public boolean writeOp(Integer queueId,Message message) {
         MessageQueue opQueue;
-        if (opQueueMap.containsKey(queueId)) {
-            opQueue = opQueueMap.get(queueId);
-        } else {
+        if ((opQueue = opQueueMap.get(queueId)) == null) {
```
caigy, line 306: Assigning `opQueue` at first may be better than initializing it in `if` condition.

RapperCL: Ok, I think this can reduce one query, although the time complexity is only O(1) and I don't think it's a good practice to judge first and then get it. If the opQueueMap has a removal operation, there will be problems.

RongtongJin (approved):

PR #5466, opened 2022-11-04
Fix order consumer's lock issue when first start

drpmma:

Fix order consumer's lock issue when first start.

## Brief changelog

Add lock when the order consumer first start.


jasonjavaYY:
Can I ask how to pass all checks? The green pass mark makes me feel so comfortable~

lizhanhui (approved):

Issue #5472, opened 2022-11-07
The main method of class “MonitorService” could be removed


jasonjavaYY:
**BUG REPORT**
There is no call to Class “org.apache.rocketmq.tools.monitor.MonitorService” startup method in the scripts under the “bin directory”, which means that the main method of class “MonitorService” will not be executed, as shown in the following figure: 
![11](https://user-images.githubusercontent.com/52964650/200215699-a2e52793-ba13-4811-95e2-2dd112835b18.png)

So I think the main method of class “MonitorService” should be removed.
1、The steps to reproduce
run the “  grep -rn "MonitorService"  ” command in the bin directory.
2、expected to see:
Startup of class "MonitorService" exists in some scripts.
3、the real output instead:
No script calls Class "MonitorService" startup, but Class "MonitorService" defines the main method.

PR #5476, opened 2022-11-07
[ISSUE #5472] Add the annotation @Deprecated to the methods that are not used in Class "MonitorService"

jasonjavaYY:

There is no call to Class “org.apache.rocketmq.tools.monitor.MonitorService” startup method in the scripts under the “bin directory”, which means that the main method of class “MonitorService” will not be executed, so I add the annotation @Deprecated to the methods that are not used in Class "MonitorService".

## Brief changelog

Add the annotation @Deprecated to the methods that are not used in Class "MonitorService".

## Verifying this change

Trivial changes: This pr only adds the annotation @Deprecated and does not involve logic code modification.
I have run mvn clean install to make sure unit-test pass.


tsunghanjacktsai (commented):
@jasonjavaYY Please keep your local branch updated with the upstream one and only include the necessary commit.

PR #5483, opened 2022-11-08
[ISSUE #5481] Decrease the repeated consumption probability of expired message

### LINKED ISSUE #5481: Decrease the repeated consumption probability of expired message
RocketMQ client will send back the expired message in push consumer, which should not be put into message listener anymore.

Before message is consumed,  we should ignore the message which is removed from process queue.

aaron-ai:
Fixes #5481


client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
```diff
@@ -284,6 +284,13 @@ public void processConsumeResult(
                 List<MessageExt> msgBackFailed = new ArrayList<>(consumeRequest.getMsgs().size());
                 for (int i = ackIndex + 1; i < consumeRequest.getMsgs().size(); i++) {
                     MessageExt msg = consumeRequest.getMsgs().get(i);
+                    // Maybe message is expired and cleaned, just ignore it.
+                    if (!consumeRequest.getProcessQueue().hasMessage(msg)) {
+                        log.info("Message does not exist in process queue, skip to send it back, topic={}, "
```
lizhanhui, line 289: The logging content looks ambiguous...

`Message does not exist in process queue, skip to send it back, topic={},` -->
`Message is not found in its process queue; skip send-back-procedure.`

lizhanhui (approved):

PR #5503, opened 2022-11-10
invert dependency between remoting and common module

### LINKED ISSUE #5502: Invert dependency between remoting and common module
It is usually understood that the `common` module is the fundamental dependency for other modules and should not depend on other modules in the project. But in RocketMQ `common` module depends on `remoting` module, which looks weird and is an obstacle to the expansion of remoting. So I want to invert the dependency between `remoting` and `common` module to make `common` module really common.

ShadowySpirits:
close #5502 

## Detailed changes

- move org.apache.rocketmq.common.protocol to org.apache.rocketmq.remoting.protocol
- move all classes that extend RemotingSerializable to org.apache.rocketmq.remoting.protocol
- move org.apache.rocketmq.remoting.common.RemotingUtil to org.apache.rocketmq.common.utils.NetworkUtil
- move org.apache.rocketmq.remoting.common.RemotingHelper#exceptionSimpleDesc to org.apache.rocketmq.common.UtilAll#exceptionSimpleDesc
- remove some functions in TopicValidator: no longer handle response in TopicValidator
- fix tests


lizhanhui (dismissed):

drpmma:
The class `ServiceThread` is repeated.

ShadowySpirits:
> The class `ServiceThread` is repeated.

we can not remove org.apache.rocketmq.remoting.common.ServiceThread for now because dledger depends on it. I removed all uses of it in RocketMQ, and this duplicate class will be removed in the next pr after fixing dledger.

lizhanhui:
> > The class `ServiceThread` is repeated.
> 
> We can not remove org.apache.rocketmq.remoting.common.ServiceThread for now because dledger depends on it. I removed all uses of it in RocketMQ, and this duplicate class will be removed in the next pr after fixing dledger.

OK, it makes sense

lizhanhui (approved):

Issue #5536, opened 2022-11-17
[Static Topic] Is it necessary to consider the compatible while upgrade namesrv and broker? 


humkum:
We know that static topic introduced the logic queue and physic queue's mapping configs. And broker register itself would carry these configs. If we update the cluster by rolling_up, there would be 2 situations:
1. Old version broker didn't carry topic queue mapping configs but the new version name server would try to decode these info, which would failed.
2. New version broker would carry topic queue mapping configs but the old version name server couldn't decode the heart beat data.
![image](https://user-images.githubusercontent.com/50660789/202347722-74719249-5564-4ac6-94fa-672ab129dafc.png)
Should we consider these cases?

lizhanhui:
We should define an upgrade blueprint, allowing clusters of previous versions to update without disruption.

A viable way is to upgrade name servers first, which provides capabilities of serving both broker versions;
After all name server nodes are updated, then start to upgrade brokers.

Name servers, thus, have to be able to process the transitional states: 1) all brokers miss placement configs; 2) partial brokers miss placement configs; 3) all brokers have placement configs.

humkum:
> We should define an upgrade blueprint, allowing clusters of previous versions to update without disruption.
> 
> A viable way is to upgrade name servers first, which provides capabilities of serving both broker versions; After all name server nodes are updated, then start to upgrade brokers.
> 
> Name servers, thus, have to be able to process the transitional states: 1) all brokers miss placement configs; 2) partial brokers miss placement configs; 3) all brokers have placement configs.

That's right！I didn't find the compatible operation in name server in current version, is there some logic I missed, or these are just missing features. If name server did not have these compatible ability, should we support this?

PR #5539, opened 2022-11-18
[ISSUE #5536] fix:new version name server should be compatible with old version broker

humkum:

fixes https://github.com/apache/rocketmq/issues/5536
Make the new name srv compatible with the old broker.
While broker configure compressedRegister=true，if we update the cluster by rolling, upgrading broker after upgrading name server or vice versa, would cause exception as follows:
image

So, we should consider the broker version to decode the register broker body


cserwen (approved):

hzh0425 (approved):
LGTM

PR #5565, opened 2022-11-22
delete 'j' variate

Tanky-Zhang:

When I read the source code,It seems that there's a ‘j’ variable here that doesn't work，so I delete it

## Brief changelog

image

Hi @RongtongJin  could you please  see if this pr works.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/logfile/DefaultMappedFile.java
+++ b/store/src/main/java/org/apache/rocketmq/store/logfile/DefaultMappedFile.java
@@ -560,7 +560,7 @@ public class DefaultMappedFile extends AbstractMappedFile {
         ByteBuffer byteBuffer = this.mappedByteBuffer.slice();
         long flush = 0;
         // long time = System.currentTimeMillis();
-        for (long i = 0, j = 0; i < this.fileSize; i += DefaultMappedFile.OS_PAGE_SIZE, j++) {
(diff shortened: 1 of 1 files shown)


store/src/main/java/org/apache/rocketmq/store/logfile/DefaultMappedFile.java
```diff
@@ -560,7 +560,7 @@ public void warmMappedFile(FlushDiskType type, int pages) {
         ByteBuffer byteBuffer = this.mappedByteBuffer.slice();
         long flush = 0;
         // long time = System.currentTimeMillis();
-        for (long i = 0, j = 0; i < this.fileSize; i += DefaultMappedFile.OS_PAGE_SIZE, j++) {
+        for (long i = 0; i < this.fileSize; i += DefaultMappedFile.OS_PAGE_SIZE) {
```
RongtongJin, line 563: j is used in the code commented below

Tanky-Zhang: now the code commented below doesn't work

Tanky-Zhang: > j is used in the code commented below

so，It's not necessary to delete j

PR #5587, opened 2022-11-24
[ISSUE #5531] Apply MDC in broker container mode

### LINKED ISSUE #5531: Logging isolation of broker container is hacky
Current solution of logging isolation is hacky: https://github.com/apache/rocketmq/blob/develop/docs/cn/BrokerContainer.md#%E6%97%A5%E5%BF%97%E5%8F%98%E5%8C%96

Maybe we should try [logback MDC](https://logback.qos.ch/manual/mdc.html)

aaron-ai:
Fixes #5531


lizhanhui (approved):
Good solution.

PR #5606, opened 2022-11-28
[ISSUE #5605]Introduce tag estimation for lag calculation

### LINKED ISSUE #5605: [OPTIMIZATION] Introduce tag estimation for lag calculation
Currently, if the consumer subscribes to a topic with a tag filter, the lag is not very accurate, especially since messages with this tag are infrequent. So I want to introduce tag estimation for lag calculation: sampling the ratio of messages with this tag and multiplying the total lag by this ratio.

You can read [this blog (Chinese Ver.)](https://blog.lv5.moe/p/explanation-and-optimization-of-apache-rocketmq-lag) for more information about more improvement in the lag calculation.

ShadowySpirits:
Close #5605

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -2666,4 +2666,33 @@ public class DefaultMessageStore implements MessageStore {
     public boolean isShutdown() {
         return shutdown;
     }
+
+    @Override
+    public long estimateMessageCount(String topic, int queueId, long from, long to, MessageFilter filter) {
+        if (from < 0) {
+            from = 0;
+        }
+
+        if (from >= to) {
+            return 0;
+        }
+
+        if (null == filter) {
+            return to - from;
+        }
+
+        ConsumeQueueInterface consumeQueue = findConsumeQueue(topic, queueId);
+        if (null == consumeQueue) {
+            return 0;
+        }
+
+        long minOffset = consumeQueue.getMinOffsetInQueue();
+        from = Math.max(from, minOffset);
+        if (to < minOffset) {
+            to = Math.min(minOffset + messageStoreConfig.getMaxConsumeQueueScan(), consumeQueue.getMaxOffsetInQueue());
+        }
+
+        long msgCount = consumeQueue.estimateMessageCount(from, to, filter);
+        return msgCount == -1 ? to - from : msgCount;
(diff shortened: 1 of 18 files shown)


store/src/main/java/org/apache/rocketmq/store/queue/BatchConsumeQueue.java
```diff
@@ -1005,4 +1005,83 @@ public void cleanSwappedMap(long forceCleanSwapIntervalMs) {
...
+            }
+
+            SelectMappedBufferResult slice = mappedFile.selectMappedBuffer(start, len);
+            if (null != slice) {
+                try {
+                    ByteBuffer buffer = slice.getByteBuffer();
+                    int current = 0;
+                    while (current < len) {
+                        // Skip physicalOffset and message length fields.
+                        buffer.position(current + 8 + 4);
```
RongtongJin, line 1048: It would be better to use constants instead of 4+8, like BatchConsumeQueue.MSG_TAG_OFFSET_INDEX

store/src/main/java/org/apache/rocketmq/store/queue/BatchConsumeQueue.java
```diff
@@ -1005,4 +1005,83 @@ public void cleanSwappedMap(long forceCleanSwapIntervalMs) {
...
+                        long tagCode = buffer.getLong();
```
RongtongJin, line 1049: Since there may be multiple messages in one BCQ index, IMO, It would be better to consider using the batch size field to estimate a more accurate value.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2666,4 +2666,33 @@ public SendMessageBackHook getSendMessageBackHook() {
...
+        }
+
+        ConsumeQueueInterface consumeQueue = findConsumeQueue(topic, queueId);
+        if (null == consumeQueue) {
+            return 0;
+        }
+
+        long minOffset = consumeQueue.getMinOffsetInQueue();
+        from = Math.max(from, minOffset);
+        if (to < minOffset) {
```
fuyou001, line 2691: minOffset  maybe using from arguments semantics is more readable

RongtongJin (approved):

fuyou001 (approved):

hzh0425 (approved):

PR #5611, opened 2022-11-29
[ISSUE #5568] Support long length group and topic for pop mode consumption

lizhimins:
see #5568

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
@@ -603,6 +603,8 @@ public class PopMessageProcessor implements NettyRequestProcessor {
                             String ckInfo = ExtraInfoUtil.buildExtraInfo(offset, popTime, requestHeader.getInvisibleTime(),
                                 reviveQid, messageExt.getTopic(), brokerName, messageExt.getQueueId(), messageExt.getQueueOffset());
                             messageExt.getProperties().putIfAbsent(MessageConst.PROPERTY_POP_CK, ckInfo);
+
+                            // Set retry message topic to origin topic and clear message store size to recode
                             messageExt.setTopic(requestHeader.getTopic());
                             messageExt.setStoreSize(0);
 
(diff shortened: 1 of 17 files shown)


common/src/main/java/org/apache/rocketmq/common/message/MessageExtBrokerInner.java
```diff
@@ -27,6 +27,8 @@ public class MessageExtBrokerInner extends MessageExt {
 
     private ByteBuffer encodedBuff;
 
+    private MessageVersion version = MessageVersion.MESSAGE_VERSION_V1;
+
```
fuyou001, line 31: version default null  may be safe.

common/src/main/java/org/apache/rocketmq/common/message/MessageExtBrokerInner.java
```diff
@@ -27,6 +27,8 @@ public class MessageExtBrokerInner extends MessageExt {
 
     private ByteBuffer encodedBuff;
 
+    private MessageVersion version = MessageVersion.MESSAGE_VERSION_V1;
```
fuyou001, line 30: version default null may be safe.

broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -586,9 +587,37 @@ private long popMsgFromQueue(boolean isRetry, GetMessageResult getMessageResult,
...
+                        try {
+                            String ckInfo = ExtraInfoUtil.buildExtraInfo(offset, popTime, requestHeader.getInvisibleTime(),
+                                reviveQid, messageExt.getTopic(), brokerName, messageExt.getQueueId(), messageExt.getQueueOffset());
+                            messageExt.getProperties().putIfAbsent(MessageConst.PROPERTY_POP_CK, ckInfo);
+
+                            // Set retry message topic to origin topic and clear message store size to recode
+                            messageExt.setTopic(requestHeader.getTopic());
+                            messageExt.setStoreSize(0);
+
+                            byte[] encode = MessageDecoder.encode(messageExt, true);
```
xdkxlk, line 611: message is not been decompressed at MessageDecoder.decodesBatch(mapedBuffer.getByteBuffer(),true, false, true)
and will be compressed here
which means, messages will be compressed twice

xdkxlk: In my opinion, this should be `MessageDecoder.encode(messageExt, false)`

xdkxlk (approved):

xdkxlk (approved):

Issue #5695, opened 2022-12-14
Optimize broker startup


lizhimins:
The broker should not be initialized immediately when call createBrokerController.
For compatibility reasons, I want to divide the startup process into several parts but won't modify the function name createBrokerController.

PR #5696, opened 2022-12-14
[ISSUE #5695] Optimize broker startup

lizhimins:


broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
```diff
@@ -79,163 +79,165 @@ public static void shutdown(final BrokerController controller) {
...
+            MixAll.properties2Object(properties, messageStoreConfig);
+        }
 
-            MixAll.properties2Object(ServerUtil.commandLine2Properties(commandLine), brokerConfig);
+        MixAll.properties2Object(ServerUtil.commandLine2Properties(commandLine), brokerConfig);
+        if (null == brokerConfig.getRocketmqHome()) {
+            System.out.printf("Please set the %s variable in your environment " +
+                "to match the location of the RocketMQ installation", MixAll.ROCKETMQ_HOME_ENV);
+            System.exit(-2);
+        }
```
aaron-ai, line 121: Could we add a console appender to replace system print here?

aaron-ai (approved):

broker/src/main/java/org/apache/rocketmq/broker/BrokerStartup.java
lizhimins: User may not have correctly loaded the log related configuration file(such as logback.xml), resulting in the error message not being displayed. It would be better to keep current situation.

Issue #5823, opened 2023-01-04
Clean unused topic semantic error in default message store


lizhimins:
Clean unused topic semantic error in default message store. It should be understood as deleting the topic which not be required.

PR #5824, opened 2023-01-04
[ISSUE #5823] Add delete topic in message store interface

lizhimins:


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -1960,11 +1960,20 @@ public RemotingCommand deleteExpiredCommitLog() {
...
-        LOGGER.warn("invoke cleanUnusedTopic start.");
+        LOGGER.info("AdminBrokerProcessor#cleanUnusedTopic: start.");
         final RemotingCommand response = RemotingCommand.createResponseCommand(null);
-        brokerController.getMessageStore().cleanUnusedTopic(brokerController.getTopicConfigManager().getTopicConfigTable().keySet());
-        LOGGER.warn("invoke cleanUnusedTopic end.");
-        response.setCode(ResponseCode.SUCCESS);
+        Set<String> topicManagerTopics = Sets.newHashSet(
+            brokerController.getTopicConfigManager().getTopicConfigTable().keySet());
+        int responseCode = ResponseCode.SUCCESS;
+        if (this.brokerController.getMessageStore() instanceof DefaultMessageStore) {
```
ShadowySpirits, line 1968: Why should we add this restriction？ IMO, there is no need to care about the underlying implementation outside the store module.

RongtongJin (commented):
We have deleteTopics and cleanUnusedTopic API, I don't think there is anything wrong with the semantics of cleanUnusedTopic. Or we can use deleteTopic instead of cleanUnusedTopic in message store.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
RongtongJin: I guess only the implementation class DefaultMessageStore can obtain consumeQueueTable

lizhimins:
image

"delete topic" is a remoting API，"clean unused topic" is a store API

RongtongJin:
> image
> 
> "delete topic" is a remoting API，"clean unused topic" is a store API

“clean unused topic” is also a remoting API, but it has different meanings from the store API "clean unused topic", so why not rename store API "clean unused topic" to "delete topics" ?

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
ShadowySpirits: Then we maybe consider exposing the interface to get all cq or sinking the implementation to the store module.

lizhimins: I already change it.

ShadowySpirits (approved):

aaron-ai (approved):

PR #5826, opened 2023-01-05
[ISSUE #5825] Enhance Message body readability

### LINKED ISSUE #5825: The message body readability is not very friendly
**BUG REPORT**

> 
    public String toString() {
        return "Message{" +
            "topic='" + topic + '\'' +
            ", flag=" + flag +
            ", properties=" + properties +
            ", body=" + Arrays.toString(body) +
            ", transactionId='" + transactionId + '\'' +
            '}';
    }

body=[72, 101, 108, 108, 111, 32, 82, 111, 99, 107, 101, 116, 77, 81, 32, 57, 57, 56]

- What is expected to see?

enhance readability

- What did you see instead?

develop

4. Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

Oliverwqcwrw:

Close #5825 

## Brief changelog

Enhance readability

## Verifying this change

XXXX

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/Message.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/Message.java
@@ -17,7 +17,7 @@
 package org.apache.rocketmq.common.message;
 
 import java.io.Serializable;
-import java.util.Arrays;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
@@ -210,7 +210,7 @@ public class Message implements Serializable {
             "topic='" + topic + '\'' +
             ", flag=" + flag +
             ", properties=" + properties +
-            ", body=" + Arrays.toString(body) +
+            ", body=" + getBodyStr(body) +
             ", transactionId='" + transactionId + '\'' +
             '}';
     }
@@ -227,5 +227,15 @@ public class Message implements Serializable {
 
     public long getDeliverTimeMs() {
         return Long.parseLong(this.getUserProperty(MessageConst.PROPERTY_TIMER_DELIVER_MS));
+    }
+
+    private String getBodyStr(byte[] body) {
+        if (body == null) {
+            return "body is null";
+        }
+        if (body.length > 3000) {
+            return "body size is " + body.length;
+        }
+        return new String(body, StandardCharsets.UTF_8);
     }
 }


common/src/main/java/org/apache/rocketmq/common/message/Message.java
```diff
@@ -210,7 +210,7 @@ public String toString() {
             "topic='" + topic + '\'' +
             ", flag=" + flag +
             ", properties=" + properties +
-            ", body=" + Arrays.toString(body) +
+            ", body=" + (body != null ? new String(body, StandardCharsets.UTF_8) : "") +
```
ShadowySpirits, line 213: The message body may be very large and compressed or serialized. It's better not to print the message body in `toString` function.

shy-share: > The message body may be very large and compressed or serialized. It's better not to print the message body in `toString` function.

Wasn't it also printed before this modification? Do you mean to delete this directly

Oliverwqcwrw: IMO, user can choose if to print message body,

so we can provide a readable format when users need to print message body

caigy: > IMO, user can choose if to print message body,
> 
> so we can provide a readable format when users need to print message body

+1. Maybe we can check the size of message body first and only print the message body if the size is small.

ShadowySpirits: > > The message body may be very large and compressed or serialized. It's better not to print the message body in `toString` function.
> 
> Wasn't it also printed before this modification? Do you mean to delete this directly

Maybe print body size instead.

shy-share: Recommended print body size when body size more than 3000,otherwise, it is impossible to distinguish whether the body is empty just by looking at the log

shy-share (changes requested):
Recommended print body size when body size more than 3000,otherwise, it is impossible to distinguish whether the body is empty just by looking at the log

Oliverwqcwrw:
> Recommended print body size when body size more than 3000,otherwise, it is impossible to distinguish whether the body is empty just by looking at the log

done

common/src/main/java/org/apache/rocketmq/common/message/Message.java
ni-ze: Do not print body is better, because the serialize of body is unknown.

shy-share (approved):

common/src/main/java/org/apache/rocketmq/common/message/Message.java
```diff
@@ -228,4 +228,14 @@ public void setDeliverTimeMs(long timeMs) {
     public long getDeliverTimeMs() {
         return Long.parseLong(this.getUserProperty(MessageConst.PROPERTY_TIMER_DELIVER_MS));
     }
+
+    private String getBodyStr(byte[] body) {
+        if (body == null) {
+            return "body is null";
+        }
+        if (body.length > 3000) {
```
leizhiyuan, line 236: maybe a better way is use the first 3000 characters？

PR #5840, opened 2023-01-09
[ISSUE #5837] Deprecate MessageStore#checkInDiskByConsumeOffset

### LINKED ISSUE #5837: Deprecate MessageStore#checkInDiskByConsumeOffset
link to #5754 

We work on introducing tiered storage for RocketMQ so that the semantics of MessageStore#checkInDiskByConsumeOffset is no longer appropriate.

https://github.com/apache/rocketmq/blob/00da3e7249d2339717d96f7ea2ab8e4a2fd4242c/store/src/main/java/org/apache/rocketmq/store/MessageStore.java#L379-L387

This issue will deprecate MessageStore#checkInDiskByConsumeOffset and add two new interfaces in MessageStore:

```java
    /**
     * Check if the given message is in the page cache.
     *
     * @param topic         topic.
     * @param queueId       queue ID.
     * @param consumeOffset consume queue offset.
     * @return true if the message is in page cache; false otherwise.
     */
    boolean checkInMemByConsumeOffset(final String topic, final int queueId, long consumeOffset, int batchSize);

    /**
     * Check if the given message is in store.
     *
     * @param topic         topic.
     * @param queueId       queue ID.
     * @param consumeOffset consume queue offset.
     * @return true if the message is in store; false otherwise.
     */
    boolean checkInStoreByConsumeOffset(final String topic, final int queueId, long consumeOffset);
```

- `checkInMemByConsumeOffset` function use [mincore](https://man7.org/linux/man-pages/man2/mincore.2.html) syscall to determine whether message data are resident in memory accurately.
- `checkInStoreByConsumeOffset` function check if the message data is in the local disk for DefaultMessageStore or in the blob storage for TieredMessageStore.

ShadowySpirits:
close #5837

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManager.java
@@ -116,7 +116,7 @@ public class BroadcastOffsetManager extends ServiceThread {
                 brokerController.getConsumerOffsetManager().queryOffset(broadcastGroupId(groupId), topic, queueId);
         }
         if (storeOffset < 0) {
-            if (!this.brokerController.getMessageStore().checkInDiskByConsumeOffset(topic, queueId, 0)) {
+            if (this.brokerController.getMessageStore().checkInMemByConsumeOffset(topic, queueId, 0, 1)) {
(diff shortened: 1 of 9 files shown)


lollipopjin (approved):
LGTM

PR #5856, opened 2023-01-10
[ISSUE #5740] Fix util.getIP sometimes choose localhost over LAN ip.

### LINKED ISSUE #5740: [code discussion] Should the method UtilAll.getIP() return a localhost ?
https://github.com/apache/rocketmq/blob/72786535067310b79138e8ae4509851d85d35b95/common/src/main/java/org/apache/rocketmq/common/UtilAll.java#L604-L619

This method basically traversal network interfaces and find a non-private ip.And if failed to find one, fall back to the whatever ip it found in the first netInterface, which in my setting is localhost(from lo0).
It is used to create unique id, so this is no good behavior right ? 
How about let the lan ip have a higher priority than localhost ?

RagingSpud:

fix #5740

## Brief changelog
According to RFC 1122, { 127, any } are reserved for internal host loopback address.So `isInternalIP` function return true when ip is between 127.0.0.0~127.255.255.255.
And now it prefers LAN ip to localhost(internal host loopback address).
## Verifying this change

XXXX

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
@@ -517,8 +517,10 @@ public class UtilAll {
         //10.0.0.0~10.255.255.255
         //172.16.0.0~172.31.255.255
         //192.168.0.0~192.168.255.255
+        //127.0.0.0~127.255.255.255
         if (ip[0] == (byte) 10) {
-
+            return true;
+        } else if (ip[0] == (byte) 127) {
             return true;
         } else if (ip[0] == (byte) 172) {
             if (ip[1] >= (byte) 16 && ip[1] <= (byte) 31) {
@@ -604,7 +606,7 @@ public class UtilAll {
                             if (ipCheck(ipByte)) {
                                 if (!isInternalIP(ipByte)) {
                                     return ipByte;
-                                } else if (internalIP == null) {
+                                } else if (internalIP == null || internalIP[0] == (byte) 127) {
                                     internalIP = ipByte;
                                 }
                             }


common/src/main/java/org/apache/rocketmq/common/UtilAll.java
```diff
@@ -604,7 +606,7 @@ public static byte[] getIP() {
                             if (ipCheck(ipByte)) {
                                 if (!isInternalIP(ipByte)) {
                                     return ipByte;
-                                } else if (internalIP == null) {
+                                } else if (internalIP == null || internalIP[0] == (byte) 127) {
```
ShadowySpirits, line 609: This judgment changes the behavior of the getIP function. If we only have lo interface, getIP will throw RuntimeException.

ShadowySpirits (changes requested):

common/src/main/java/org/apache/rocketmq/common/UtilAll.java
RagingSpud: If we only have lo interface,`internalIP == null || internalIP[0] == (byte) 127)` will be true，so it will set `internalIP` to localhost and return.
If we have lo and LAN ip，when it find the LAN ip，`internalIP == null || internalIP[0] == (byte) 127)` also will be true, so it will override the value of `internalIP`.
Am I missing something？

ShadowySpirits: You are right. Please ignore this review.

ShadowySpirits (approved):

hzh0425 (approved):
LGTM

hzh0425:
Welcome @RagingSpud

PR #5887, opened 2023-01-15
[ISSUE #5884] Concurrent check CommitLog messages

### LINKED ISSUE #5884: [RIP-52] [Task 1] Concurrent check CommitLog messages
_No description provided._

nowinkeyy:

close #5884 

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
+++ b/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
@@ -399,10 +399,6 @@ public class BrokerConfig extends BrokerIdentity {
      */
     private boolean estimateAccumulation = true;
 
-    /**
-     * Build ConsumeQueue concurrently with multi-thread
-     */
-    private boolean enableBuildConsumeQueueConcurrently = false;
 
     public long getMaxPopPollingSize() {
         return maxPopPollingSize;
@@ -1650,13 +1646,5 @@ public class BrokerConfig extends BrokerIdentity {
 
     public void setBatchDispatchRequestThreadPoolNums(int batchDispatchRequestThreadPoolNums) {
         this.batchDispatchRequestThreadPoolNums = batchDispatchRequestThreadPoolNums;
-    }
-
-    public boolean isEnableBuildConsumeQueueConcurrently() {
-        return enableBuildConsumeQueueConcurrently;
-    }
-
-    public void setEnableBuildConsumeQueueConcurrently(boolean enableBuildConsumeQueueConcurrently) {
-        this.enableBuildConsumeQueueConcurrently = enableBuildConsumeQueueConcurrently;
     }
 }
--- a/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
+++ b/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
@@ -392,8 +392,6 @@ public class BrokerConfig extends BrokerIdentity {
     private long channelExpiredTimeout = 1000 * 120;
     private long subscriptionExpiredTimeout = 1000 * 60 * 10;
 
-    private int batchDispatchRequestThreadPoolNums = 16;
-
     /**
      * Estimate accumulation or not when subscription filter type is tag and is not SUB_ALL.
      */
@@ -1638,13 +1636,5 @@ public class BrokerConfig extends BrokerIdentity {
 
     public void setEstimateAccumulation(boolean estimateAccumulation) {
         this.estimateAccumulation = estimateAccumulation;
-    }
-
-    public int getBatchDispatchRequestThreadPoolNums() {
-        return batchDispatchRequestThreadPoolNums;
-    }
-
-    public void setBatchDispatchRequestThreadPoolNums(int batchDispatchRequestThreadPoolNums) {
-        this.batchDispatchRequestThreadPoolNums = batchDispatchRequestThreadPoolNums;
     }
 }
--- a/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
+++ b/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
@@ -392,6 +392,8 @@ public class BrokerConfig extends BrokerIdentity {
     private long channelExpiredTimeout = 1000 * 120;
     private long subscriptionExpiredTimeout = 1000 * 60 * 10;
 
+    private int batchDispatchRequestThreadPoolNums = 16;
+
     /**
      * Estimate accumulation or not when subscription filter type is tag and is not SUB_ALL.
      */
@@ -1635,5 +1637,13 @@ public class BrokerConfig extends BrokerIdentity {
 
     public void setEstimateAccumulation(boolean estimateAccumulation) {
         this.estimateAccumulation = estimateAccumulation;
+    }
+
+    public int getBatchDispatchRequestThreadPoolNums() {
+        return batchDispatchRequestThreadPoolNums;
+    }
+
+    public void setBatchDispatchRequestThreadPoolNums(int batchDispatchRequestThreadPoolNums) {
+        this.batchDispatchRequestThreadPoolNums = batchDispatchRequestThreadPoolNums;
     }
 }
--- a/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
+++ b/common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
@@ -399,6 +399,11 @@ public class BrokerConfig extends BrokerIdentity {
      */
     private boolean estimateAccumulation = true;
 
+    /**
+     * Build ConsumeQueue concurrently with multi-thread
+     */
+    private boolean enableBuildConsumeQueueConcurrently = false;
+
     public long getMaxPopPollingSize() {
         return maxPopPollingSize;
     }
@@ -1645,5 +1650,13 @@ public class BrokerConfig extends BrokerIdentity {
 
     public void setBatchDispatchRequestThreadPoolNums(int batchDispatchRequestThreadPoolNums) {
         this.batchDispatchRequestThreadPoolNums = batchDispatchRequestThreadPoolNums;
+    }
+
+    public boolean isEnableBuildConsumeQueueConcurrently() {
+        return enableBuildConsumeQueueConcurrently;
+    }
+
+    public void setEnableBuildConsumeQueueConcurrently(boolean enableBuildConsumeQueueConcurrently) {
+        this.enableBuildConsumeQueueConcurrently = enableBuildConsumeQueueConcurrently;
     }
 }
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -54,7 +54,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import org.apache.commons.lang3.StringUtils;
@@ -3086,7 +3085,7 @@ public class DefaultMessageStore implements MessageStore {
                         this.reputFromOffset);
             }
 
-            this.shutdown();
+            super.shutdown();
         }
 
         @Override
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -2889,7 +2889,7 @@ public class DefaultMessageStore implements MessageStore {
                     batchDispatchRequestQueue.poll();
                 }
             } catch (Exception e) {
-                LOGGER.warn(e.getMessage());
+                DefaultMessageStore.LOGGER.warn(this.getServiceName() + " service has exception. ", e);
             }
         }
 
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -42,6 +42,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.ExecutionException;
@@ -134,6 +135,10 @@ public class DefaultMessageStore implements MessageStore {
 
     private ReputMessageService reputMessageService;
 
+    private MainBatchDispatchRequestService mainBatchDispatchRequestService;
+
+    private DispatchService dispatchService;
+
     private HAService haService;
 
     // CompactionLog
@@ -189,6 +194,14 @@ public class DefaultMessageStore implements MessageStore {
 
     private int maxDelayLevel;
 
+    private final AtomicInteger mappedPageHoldCount = new AtomicInteger(0);
+
+    private final ConcurrentLinkedQueue<BatchDispatchRequest> batchDispatchRequestQueue = new ConcurrentLinkedQueue<>();
+
+    private int dispatchRequestOrderlyQueueSize = 16;
+
+    private final DispatchRequestOrderlyQueue dispatchRequestOrderlyQueue = new DispatchRequestOrderlyQueue(dispatchRequestOrderlyQueueSize);
+
     public DefaultMessageStore(final MessageStoreConfig messageStoreConfig, final BrokerStatsManager brokerStatsManager,
         final MessageArrivingListener messageArrivingListener, final BrokerConfig brokerConfig) throws IOException {
         this.messageArrivingListener = messageArrivingListener;
@@ -225,7 +238,13 @@ public class DefaultMessageStore implements MessageStore {
             }
         }
 
-        this.reputMessageService = new ReputMessageService();
+        if (!messageStoreConfig.isEnableBuildConsumeQueueConcurrently()) {
+            this.reputMessageService = new ReputMessageService();
+        } else {
+            this.reputMessageService = new ConcurrentReputMessageService();
+            this.mainBatchDispatchRequestService = new MainBatchDispatchRequestService();
+            this.dispatchService = new DispatchService();
+        }
 
         this.transientStorePool = new TransientStorePool(this);
 
@@ -361,6 +380,11 @@ public class DefaultMessageStore implements MessageStore {
         this.reputMessageService.setReputFromOffset(this.commitLog.getConfirmOffset());
         this.reputMessageService.start();
 
+        if (messageStoreConfig.isEnableBuildConsumeQueueConcurrently()) {
+            this.mainBatchDispatchRequestService.start();
+            this.dispatchService.start();
+        }
+
         // Checking is not necessary, as long as the dLedger's implementation exactly follows the definition of Recover,
         // which is eliminating the dispatch inconsistency between the commitLog and consumeQueue at the end of recovery.
         this.doRecheckReputOffsetFromCq();
@@ -458,6 +482,12 @@ public class DefaultMessageStore implements MessageStore {
             }
             this.commitLog.shutdown();
             this.reputMessageService.shutdown();
+            if (mainBatchDispatchRequestService != null) {
+                mainBatchDispatchRequestService.shutdown();
+            }
+            if (dispatchService != null) {
+                dispatchService.shutdown();
+            }
             this.flushConsumeQueueService.shutdown();
             this.allocateMappedFileService.shutdown();
             this.storeCheckpoint.flush();
@@ -675,7 +705,12 @@ public class DefaultMessageStore implements MessageStore {
 
         this.recoverTopicQueueTable();
 
-        this.reputMessageService = new ReputMessageService();
+        if (!messageStoreConfig.isEnableBuildConsumeQueueConcurrently()) {
+            this.reputMessageService = new ReputMessageService();
+        } else {
+            this.reputMessageService = new ConcurrentReputMessageService();
+        }
+
         this.reputMessageService.setReputFromOffset(Math.min(oldReputFromOffset, offsetToTruncate));
         this.reputMessageService.start();
     }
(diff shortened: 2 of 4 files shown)


guyinyou (dismissed):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,10 +2576,181 @@ public long getJoinTime() {
...
+                    this.batchDispatchRequestThreadPoolNums,
+                    this.batchDispatchRequestThreadPoolNums,
+                    1000 * 60,
+                    TimeUnit.MICROSECONDS,
+                    new LinkedBlockingDeque<>(),
+                    new ThreadFactoryImpl("BatchDispatchRequestServiceThread_"));
+        }
+
+        private void startBatchDispatchRequestService() {
+            new Thread(() -> {
```
RongtongJin, line 2685: It would be better to use org.apache.rocketmq.common.ServiceThread instead of java native thread.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,10 +2576,181 @@ public long getJoinTime() {
...
+            return null;
+        }
+
+        public synchronized boolean isEmpty() {
+            return maxPtr.get() == ptr;
+        }
+
+    }
+
     class ReputMessageService extends ServiceThread {
```
RongtongJin, line 2647: How about creating a new class that extends ReputMessageService?  We can initialize different implementations according to isEnableBuildConsumeQueueConcurrently config. On the other hand, the ReputMessageService code will not change and the code readability will be better.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,10 +2576,181 @@ public long getJoinTime() {
...
+                        try {
+                            Thread.sleep(1);
+                        } catch (InterruptedException e) {
+                            throw new RuntimeException(e);
+                        }
+                    }
+                }
+            }, "MainBatchDispatchRequestServiceThread").start();
+
+            new Thread(() -> {
```
RongtongJin, line 2715: It would be better to use org.apache.rocketmq.common.ServiceThread instead of java native thread.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2689,6 +2865,97 @@ private void doReput() {
...
+                                this.reputFromOffset = DefaultMessageStore.this.commitLog.rollNextFile(this.reputFromOffset);
+                                readSize = result.getSize();
+                            }
+                            this.createBatchDispatchRequest(byteBuffer, batchDispatchRequestStart, batchDispatchRequestSize);
+                            batchDispatchRequestStart = -1;
+                            batchDispatchRequestSize = -1;
+                        }
+                    }
+                } catch (Throwable e) {
+                    throw e;
```
RongtongJin, line 2942: This line seems redundant.

RongtongJin:
Hi @nowinkeyy, It would be better to provide the performance test report.

mxsm (commented):
@nowinkeyy  It would be best to provide specific data or similar comparison charts to test whether this implementation can improve efficiency. The number of concurrent threads should also be taken into account, and corresponding test data is hoped to be provided.

fuyou001:
please add unit test

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2689,6 +2865,97 @@ private void doReput() {
...
+                                totalSize = 0;
+                                break;
+                            default:
+                                totalSize = -1;
+                                doNext = false;
+                        }
+
+                        if (totalSize > 0) {
+                            if (batchDispatchRequestStart == -1) {
+                                batchDispatchRequestStart = byteBuffer.position() - 8;
```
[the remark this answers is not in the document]
fuyou001: using byteBuffer.mark(),byteBuffer.reset() may be more code readability

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
nowinkeyy: > This line seems redundant.

Yes, it has been deleted.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2689,6 +2865,97 @@ private void doReput() {
...
+                        }
+                    }
+                } catch (Throwable e) {
+                    throw e;
+                } finally {
+                    this.createBatchDispatchRequest(result.getByteBuffer(), batchDispatchRequestStart, batchDispatchRequestSize);
+                    boolean over = this.mappedPageHoldCount.get() == 0;
+                    while (!over) {
+                        try {
+                            Thread.sleep(1);
```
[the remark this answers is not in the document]
nowinkeyy: > TimeUnit.MILLISECONDS.sleep(1) replace Thread.sleep

okk.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,10 +2576,181 @@ public long getJoinTime() {
         }
     }
 
+    class BatchDispatchRequest {
+
```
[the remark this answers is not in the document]
nowinkeyy: > using ServiceThread notify and wait logic

Does it mean countdownlatch2? Specifically, MainBatchDispatchRequestServiceThread notifications and waits between ReputMessageService?

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2689,6 +2865,97 @@ private void doReput() {
...
+                            }
+                            byteBuffer.position(byteBuffer.position() + totalSize - 8);
+                            this.reputFromOffset += totalSize;
+                            readSize += totalSize;
+                        } else {
+                            if (totalSize == 0) {
+                                this.reputFromOffset = DefaultMessageStore.this.commitLog.rollNextFile(this.reputFromOffset);
+                                readSize = result.getSize();
+                            }
+                            this.createBatchDispatchRequest(byteBuffer, batchDispatchRequestStart, batchDispatchRequestSize);
```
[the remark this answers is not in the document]
nowinkeyy: > duplicate code，please refactor into a method

Did you say 2933 and 2934?

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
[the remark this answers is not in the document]
nowinkeyy: > using byteBuffer.mark(),byteBuffer.reset() may be more code readability

mark() and reset() have been used.(byteBuffer.mark(); byteBuffer.getInt(); byteBuffer.reset();) May I ask which part of the code comments are needed

nowinkeyy:
Test data was provided by @guyinyou

image
image

guyinyou (approved):
After testing, it meets expectations

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2850,258 @@ public String getServiceName() {
...
+
+        private final ExecutorService batchDispatchRequestExecutor;
+
+        public MainBatchDispatchRequestService() {
+            batchDispatchRequestExecutor = new ThreadPoolExecutor(
+                    DefaultMessageStore.this.getBrokerConfig().getBatchDispatchRequestThreadPoolNums(),
+                    DefaultMessageStore.this.getBrokerConfig().getBatchDispatchRequestThreadPoolNums(),
+                    1000 * 60,
+                    TimeUnit.MICROSECONDS,
+                    new LinkedBlockingQueue<>(1024),
```
fuyou001, line 2863: when runnable tasks greater than  1024, batchDispatchRequestExecutor miss RejectedExecutionHandler  logic，the default RejectedExecutionHandler is right?

nowinkeyy: > when runnable tasks greater than 1024, batchDispatchRequestExecutor miss RejectedExecutionHandler logic，the default RejectedExecutionHandler is right?

The reason I set the value to 1024 is that I think this size is sufficient. Because a task size is 4MB. A commitlog file size of 1G can generate 256tasks. So the queue can hold 1024 tasks generated by four commitlog files. So I set the default reject policy.

nowinkeyy: Does it need a policy to temporarily save more than 1024 tasks? I can adjust it if necessary.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,9 +2610,77 @@ public long getJoinTime() {
...
+            this.size = size;
+            this.id = id;
+        }
+    }
+
+    class DispatchRequestOrderlyQueue {
+
+        DispatchRequest[][] buffer;
+
+        int ptr = 0;
```
fuyou001, line 2635: prt maybe long type is better.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,9 +2610,77 @@ public long getJoinTime() {
...
+
+        AtomicInteger maxPtr = new AtomicInteger();
+
```
fuyou001, line 2638: maxPtr maybe AtomicLoing type is better.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
fuyou001: > > when runnable tasks greater than 1024, batchDispatchRequestExecutor miss RejectedExecutionHandler logic，the default RejectedExecutionHandler is right?
> 
> The reason I set the value to 1024 is that I think this size is sufficient. Because a task size is 4MB. A commitlog file size of 1G can generate 256tasks. So the queue can hold 1024 tasks generated by four commitlog files. So I set the default reject policy.

I got it .

adjust 4096 ?

nowinkeyy: > 

okk.

fuyou001: @nowinkeyy , sorry.

I thought for a moment，it must be se right RejectedExecutionHandler,because if it occur,may can lost message. the 
reject policy can block task,but cant't throw Exception cause message lost

nowinkeyy: > @nowinkeyy , sorry.
> 
> I thought for a moment，it must be se right RejectedExecutionHandler,because if it occur,may can lost message. the reject policy can block task,but cant't throw Exception cause message lost

You are right. And i have an idea that I don't know if it is feasible. Before putting a task on a thread pool queue, determine if the queue size is less than 1024? As shown in the figure:
image

fuyou001: i can't confirm that batchDispatchRequestQueue queue getSize method is concurrent safe

common usage  set right  reject policy of Executor.

nowinkeyy: > i can't confirm that batchDispatchRequestQueue queue getSize method is concurrent safe
> 
> common usage set right reject policy of Executor.

You are right, it is more appropriate to set a right reject policy of Executor. 

( The getsize() method is concurrency-safe because it is LinkedBlockingQueue. )

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2852,269 @@ public String getServiceName() {
...
+            }
+            return DispatchService.class.getSimpleName();
+        }
+    }
+
+    class ConcurrentReputMessageService extends ReputMessageService {
+
+        private static final int BATCH_SIZE = 1024 * 1024 * 4;
+
+        private int batchId = 0;
```
fuyou001, line 2990: batchId type change long

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2852,269 @@ public String getServiceName() {
...
+                            }
+                        }
+                    });
+        }
+
+        private void pollBatchDispatchRequest() {
+            if (!batchDispatchRequestQueue.isEmpty()) {
+                BatchDispatchRequest task = batchDispatchRequestQueue.poll();
+                batchDispatchRequestExecutor.execute(() -> {
+                    ByteBuffer tmpByteBuffer = task.byteBuffer.duplicate();
```
fuyou001, line 2884: batchDispatchRequestExecutor  execute  runnable  must be have try catch final protect

nowinkeyy: Is that ok?

image

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2852,269 @@ public String getServiceName() {
...
+                    DefaultMessageStore.this.getBrokerConfig().getBatchDispatchRequestThreadPoolNums(),
+                    1000 * 60,
+                    TimeUnit.MICROSECONDS,
+                    new LinkedBlockingQueue<>(4096),
+                    new ThreadFactoryImpl("BatchDispatchRequestServiceThread_"),
+                    new RejectedExecutionHandler() {
+                        @Override
+                        public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+                            try {
+                                LOGGER.warn("Task {} is blocking put into the workQueue", r);
```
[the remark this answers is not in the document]
nowinkeyy: > you can use JDK CallerRunsPolicy

I don't think using CallerRunsPolicy works well. Because when run() is executed, the calling thread blocks waiting for a long time. Even after the workQueue is idle, the calling thread continues to block for some time. In the current way, the calling thread simply blocks and waits until the workQueue is idle.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
fuyou001: Give you the demo, you can think about different

```
package org.apache.rocketmq.test;

import java.util.concurrent.ExecutorService;
import java.util.concurrent.LinkedBlockingQueue;
import java.util.concurrent.ThreadPoolExecutor;
import java.util.concurrent.TimeUnit;

import org.apache.rocketmq.common.ThreadFactoryImpl;

public class Test {

    
        good(false);
        good(true);
        bad(false);
        bad(true);

    }

    public static void good(boolean throwException) {
        ExecutorService executor = new ThreadPoolExecutor(1, 1,
                1000 * 60,
                TimeUnit.MICROSECONDS,
                new LinkedBlockingQueue<>(2),
                new ThreadFactoryImpl("GoodBatchDispatchRequestServiceThread_"),
                new ThreadPoolExecutor.CallerRunsPolicy());
        for (int i = 0; i < 15; i++) {
            executor.execute(() -> {
                try {
                    TimeUnit.SECONDS.sleep(1);
                    System.out.println("ThreadName:" + Thread.currentThread().getName());
                    if (throwException)
                        throw new RuntimeException("test");
                } catch (Exception e) {

                }
            });
        }

        for (int i = 0; i < 5; i++) {
            executor.execute(() -> {
                System.out.println("testbb");
            });
        }
        System.out.println("test-cccccccc");

    }

    public static void bad(boolean throwException) {
        ExecutorService executor = new ThreadPoolExecutor(1, 1,
                1000 * 60,
                TimeUnit.MICROSECONDS,
                new LinkedBlockingQueue<>(2),
                new ThreadFactoryImpl("BadBatchDispatchRequestServiceThread_"));
        for (int i = 0; i < 15; i++) {
            executor.execute(() -> {
                try {
                    TimeUnit.SECONDS.sleep(1);

                }
                System.out.println("ThreadName:" + Thread.currentThread().getName());
                if (throwException)
                    throw new RuntimeException("test");
            });
        }

        for (int i = 0; i < 5; i++) {
            executor.execute(() -> {
                System.out.println("testbb");
            });
        }
        System.out.println("test-cccccccc");

    }
}
```

nowinkeyy: The differences I think are: 
1) good() has a deny reject CallerRunsPolicy, and the calling thread main is responsible for performing the overflowed task. Bad() does not have a reject policy, and overflowed tasks are discarded. 
2) Exceptions thrown within the scope of try will be catch. Exceptions thrown outside the scope of the try are thrown to a higher level and interrupt the current program, and seem to cause Executor to create a new thread (the previous thread may have been destroyed)

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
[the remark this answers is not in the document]
nowinkeyy: > I don't think using CallerRunsPolicy works well. Because when run() is executed, the calling thread blocks waiting for a long time. Even after the workQueue is idle, the calling thread continues to block for some time. In the current way, the calling thread simply blocks and waits until the workQueue is idle.

The run () execution process blocks because only requests within a window (default size 16) can be put into the dispatchRequestOrderlyQueue, and requests beyond the window range block waiting for placement. 
If you use the CallerRunsPolicy strategy, when the work queue is full, the calling thread must be executing run () requests that exceed the window scope, causing the calling thread to wait too long to finish executing run ().

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
fuyou001: you can refactor related code as soon as possible. @nowinkeyy

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2851,264 @@ public String getServiceName() {
...
+                                LOGGER.error("[BUG]read total count not equals msg total size.");
+                            }
+                        }
+                        dispatchRequestOrderlyQueue.put(task.id, dispatchRequestList.toArray(new DispatchRequest[dispatchRequestList.size()]));
+                        mappedPageHoldCount.getAndDecrement();
+                    });
+                    batchDispatchRequestQueue.poll();
+                }
+            } catch (Exception e) {
+                LOGGER.warn(e.getMessage());
```
fuyou001, line 2892: like DefaultMessageStore.LOGGER.warn(this.getServiceName() + " service has exception. ", e);

guyinyou:
image

After testing, it still meets expectations

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2851,264 @@ public String getServiceName() {
...
+                    new ThreadFactoryImpl("BatchDispatchRequestServiceThread_"),
+                    new ThreadPoolExecutor.AbortPolicy());
+        }
+
+        private void pollBatchDispatchRequest() {
+            try {
+                if (!batchDispatchRequestQueue.isEmpty()) {
+                    BatchDispatchRequest task = batchDispatchRequestQueue.peek();
+                    batchDispatchRequestExecutor.execute(() -> {
+                        ByteBuffer tmpByteBuffer = task.byteBuffer.duplicate();
```
fuyou001, line 2874: try..catch

fuyou001 (approved):

guyinyou (dismissed):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,282 @@ public String getServiceName() {
...
+                    TimeUnit.MICROSECONDS,
+                    new LinkedBlockingQueue<>(4096),
+                    new ThreadFactoryImpl("BatchDispatchRequestServiceThread_"),
+                    new ThreadPoolExecutor.AbortPolicy());
+        }
+
+        private void pollBatchDispatchRequest() {
+            try {
+                if (!batchDispatchRequestQueue.isEmpty()) {
+                    BatchDispatchRequest task = batchDispatchRequestQueue.peek();
```
fuyou001, line 2856: batchDispatchRequestQueue.polll

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,282 @@ public String getServiceName() {
...
+                                    LOGGER.error("[BUG]read total count not equals msg total size.");
+                                }
+                            }
+                            dispatchRequestOrderlyQueue.put(task.id, dispatchRequestList.toArray(new DispatchRequest[dispatchRequestList.size()]));
+                            mappedPageHoldCount.getAndDecrement();
+                        } catch (Exception e) {
+                            LOGGER.error("There is an exception in task execution.", e);
+                        }
+                    });
+                    batchDispatchRequestQueue.poll();
```
fuyou001, line 2877: delete batchDispatchRequestQueue.poll()

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,282 @@ public String getServiceName() {
...
+
+        @Override
+        public void start() {
+            super.start();
+            this.mainBatchDispatchRequestService.start();
+            this.dispatchService.start();
+        }
+
+        @Override
+        public void doReput() {
```
drpmma, line 3001: The doReput method in ConcurrentReputMessageService is similar with the method in class ReputMessageService. Is it possible to abstract some code fragments to reuse the code?

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,282 @@ public String getServiceName() {
...
+                    break;
+                }
+
+                int batchDispatchRequestStart = -1;
+                int batchDispatchRequestSize = -1;
+                try {
+                    this.reputFromOffset = result.getStartOffset();
+
+                    for (int readSize = 0; readSize < result.getSize() && reputFromOffset < DefaultMessageStore.this.getConfirmOffset() && doNext; ) {
+                        ByteBuffer byteBuffer = result.getByteBuffer();
```
drpmma, line 3021: It's better to have a method that is responsible for building DispatchRequest including byteBuffer decode like checkMessageAndReturnSize.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,282 @@ public String getServiceName() {
...
+
+                        int totalSize = byteBuffer.getInt();
+                        if (reputFromOffset + totalSize > DefaultMessageStore.this.getConfirmOffset()) {
+                            doNext = false;
+                            break;
+                        }
+
+                        int magicCode = byteBuffer.getInt();
+                        switch (magicCode) {
+                            case CommitLog.MESSAGE_MAGIC_CODE:
```
drpmma, line 3033: MESSAGE_MAGIC_CODE_V2 should be put into consideration.
See https://github.com/apache/rocketmq/blob/aa8ca48d478e58309f9a6b026e6fea7f4772c6e9/store/src/main/java/org/apache/rocketmq/store/CommitLog.java#L386-L391

drpmma (changes requested):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
nowinkeyy: > The doReput method in ConcurrentReputMessageService is similar with the method in class ReputMessageService. Is it possible to abstract some code fragments to reuse the code?

If I abstracted a method, it would probably require some changes to ReputMessageService, which I don't want to do. And I don't think the reusability requirements for this part of the code are high.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
nowinkeyy: I think a synthesis method is a bit difficult, because this part of the code to follow below stronger affinity, less like checkMessageAndReturnSize method can very good independence. The following parts can then be combined into a method:

image

drpmma: I agree about the part that the process of decoding byteBuffer should be combined into a method.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
drpmma: I would accept it.

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
drpmma: Do you want to submit another commit to combine it into a method?

nowinkeyy: > Do you want to submit another commit to combine it into a method?

yeah, i will.

drpmma (dismissed):
LGTM

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,291 @@ public String getServiceName() {
...
+    class DispatchService extends ServiceThread {
+
+        private final List<DispatchRequest[]> dispatchRequestsList = new ArrayList<>();
+
+        private void dispatch() {
+            dispatchRequestsList.clear();
+            dispatchRequestOrderlyQueue.get(dispatchRequestsList);
+            if (!dispatchRequestsList.isEmpty()) {
+                for (DispatchRequest[] dispatchRequests : dispatchRequestsList) {
+                    for (DispatchRequest dispatchRequest : dispatchRequests) {
```
ShadowySpirits, line 2919: Can you give these variables a more clear name or add some comments?

ShadowySpirits (changes requested):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2571,9 +2595,77 @@ public long getJoinTime() {
...
+                }
+            }
+            return null;
+        }
+
+        public synchronized boolean isEmpty() {
+            return maxPtr.get() == ptr;
+        }
+
+    }
```
ShadowySpirits, line 2664: Ditto. We should avoid abbreviated variable names like idx, obj or rets.

ShadowySpirits: And DispatchRequestOrderlyQueue#get always returns null, which looks weird. Changed to the following form to be more object-oriented.

```
public List<DispatchRequest[]> get()
```

ShadowySpirits (changes requested):

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
nowinkeyy: > And DispatchRequestOrderlyQueue#get always returns null, which looks weird. Changed to the following form to be more object-oriented.

How about I change it to void? Because it doesn't need a return value.

ShadowySpirits: > > And DispatchRequestOrderlyQueue#get always returns null, which looks weird. Changed to the following form to be more object-oriented.
> 
> How about I change it to void? Because it doesn't need a return value.

`public void get(List<DispatchRequest[]> rets)` seems like a function in C code style to avoid returning local variables causing dangling pointer problems. In Java, we are more accustomed to using the return value of the `get` method.

nowinkeyy: I agree with you. The intention here is to reduce the new List<DispatchRequest[]>, so the return value is not set.

RongtongJin (approved):

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
```diff
@@ -392,11 +392,14 @@ public boolean isEnable() {
     private long channelExpiredTimeout = 1000 * 120;
     private long subscriptionExpiredTimeout = 1000 * 60 * 10;
 
+    private int batchDispatchRequestThreadPoolNums = 16;
```
RongtongJin, line 395: Why not put this config to MessageStoreConfig?

RongtongJin (changes requested):

common/src/main/java/org/apache/rocketmq/common/BrokerConfig.java
guyinyou: > Why not put this config to MessageStoreConfig?

a small detail, niubility

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2743,6 +2835,291 @@ public String getServiceName() {
...
+
+                        int totalSize = preCheckMessageAndReturnSize(byteBuffer);
+
+                        if (totalSize > 0) {
+                            if (batchDispatchRequestStart == -1) {
+                                batchDispatchRequestStart = byteBuffer.position();
+                                batchDispatchRequestSize = 0;
+                            }
+                            batchDispatchRequestSize += totalSize;
+                            if (batchDispatchRequestSize > BATCH_SIZE) {
```
drpmma, line 3031: Maybe BATCH_SIZE should also be put into MessageStoreConfig.

nowinkeyy: I agree with you.

nowinkeyy: Sorry, on second thought, I don't provide a way to modify BATCH_SIZE yet, because BATCH_SIZE may affect the number of tasks in a thread pool, which could result in a large number of reject exceptions.

drpmma: BATCH_SIZE can be assigned from MessageStoreConfig here which will only take effect when broker startup.

https://github.com/apache/rocketmq/blob/02577ed7230a3587cb04ba65abdd6870230c4d7e/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java#L2970

nowinkeyy: What you said is right. But we don't want to expose BATCH_SIZE right now. It doesn't make much sense to debug this parameter right now because it's not the bottleneck. If this parameter is set too small, problems may occur. (If you have any more reasons, please let me know and I will make changes.)

drpmma: LGTM.

RongtongJin (approved):

Issue #5974, opened 2023-02-03
Nameserver address is verified when the producer client is started


socutes:
Currently, Nameserver addresses are not verified in producer.start(). Instead, the check is done at the end of the send() function after a valid TopicPublishInfo cannot be found when the message is sent.

Reasonable logic: check when start() is executed. If the validation fails, producer.start() throws an exception.

<img width="1268" alt="image" src="https://user-images.githubusercontent.com/7479320/216535151-e502f509-6db6-4574-a4d4-3dbb3cd257f5.png">

socutes:
@aaron-ai @zhouxinyu PTAL

PR #5978, opened 2023-02-03
[ISSUE #5974] Nameserver address is verified when the producer client is started

socutes:
Currently, Nameserver addresses are not verified in producer.start(). Instead, the check is done at the end of the send() function after a valid TopicPublishInfo cannot be found when the message is sent.

Reasonable logic: check when start() is executed. If the validation fails, producer.start() throws an exception.

fix: #5974


socutes:
@aaron-ai @zhouxinyu PTAL

mxsm (commented):
Hi @socutes  CI not pass PLS check it

client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -611,7 +618,7 @@ private SendResult sendDefaultImpl(
         final long invokeID = random.nextLong();
         long beginTimestampFirst = System.currentTimeMillis();
         long beginTimestampPrev = beginTimestampFirst;
-        long endTimestamp = beginTimestampFirst;
+        long endTimestamp;
```
Oliverwqcwrw, line 621: it is better to keep the origin

PR #6004, opened 2023-02-07
[RIP-41] Add E2E test to CI

deepsola:

- Add new CI config to run e2e test(apache/rocketmq-e2e) when PUSH or PR.

## Brief changelog

1. New CI provides an independent RocketMQ environment for E2E testing by using the Alibaba Cloud ASK capability for each PUSH and PR, and deploys using the helm chart (apache/rocketmq-docker), and the mirror image is the submitted code.
2. The test report is visualized, and the report of each UT and E2E test is displayed visually on the current execution page of CI, no need to check the log, which is convenient for troubleshooting.
3. Compatibility testing is supported, and testing of multiple OS and JDK versions can be performed concurrently.
4. Due to security considerations, the PR and PUSH action CIs are separated into two pipelines.

## Example
- PUSH: https://github.com/deepsola/rocketmq/actions/runs/4103445477
- PR: 
          [UT] https://github.com/deepsola/rocketmq/actions/runs/4109646599
          [E2E TEST] https://github.com/deepsola/rocketmq/actions/runs/4109709225

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -1,0 +1,75 @@
+name: PR-CI
+
+on:
+  pull_request:
+
+jobs:
+  unit-test:
+    if: always()
+    name: Unit test
+    runs-on: ${{ matrix.os }}-latest
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        os: [ubuntu, macos, windows]
+        java-version: [8]
+        include:
+          - os: ubuntu
+            java-version: 11
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: Cache maven repository
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-maven-
+      - uses: actions/setup-java@v3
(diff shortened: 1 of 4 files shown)


RongtongJin (commented):
@deepsola It seems that only one workflow for building distributed tar was added when submitting PR and no e2e test was run.

deepsola:
> @deepsola It seems that only one workflow for building distributed tar was added when submitting PR and no e2e test was run.

According to the security issues raised in this article, https://securitylab.github.com/research/github-actions-preventing-pwn-requests/, secrets cannot be accessed in the CI execution generated by pr, while e2e The test needs to use secrets, so I use [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) to split the CI of the pull request into two parts, pr.yml and pr-e2e-test.yml, where pr-e2e-test.yml will be executed after the end of pr.yml, but only pr-e2e-test.yml will be triggered only if it exists in the default branch. Since the pr-e2e-test.yml file is not merged now, the e2e test will not be triggered in this pr, but it will be triggered in other pr after this (but still will not appear in the CI associated with pr), refer to demo: https://github.com/deepsola/start-world/pull/6, 
 and https://github.com/deepsola/start-world/actions, the "Second" CI is triggered by the pr.

drpmma (approved):
LGTM

PR #6175, opened 2023-02-25
[ISSUE #6174] Small optimization of Proxy code and logic 

### LINKED ISSUE #6174: Small optimization of Proxy code and logic
Optimize two points:
1.The init() method in GrpcChannelManager is not called, so the cleanup thread scanExpireResultFuture does not take effect.
2.scanExpireResultFuture scans every 10 seconds. Unable to accurately handle timeout requests.GrpcProxyRelayRequestTimeoutInSeconds units for seconds, should be adjusted to scan once per second.

image

socutes:
close #6174


socutes:
@zhouxinyu PTAL

xdkxlk (approved):
LGTM

PR #6216, opened 2023-03-01
[ISSUE #6215]Make benchmark cover compress msg situation

### LINKED ISSUE #6215: Make benchmark cover the situation of compressing message using different compress algorithm.
There are ZLIB、ZSTD、LZ4 algorithm to compress message body at present. 
We can make it configurable in benchmark.

humkum:
closes #6215 

We could use benchmark like this to compress message body which is larger than 2 KB with "ZSTD" compress algorithm:
```sh producer.sh -n namesrv -t topic -c true -ct ZSTD -cl 5 -ch 2048```

### DIFF
--- a/example/src/main/java/org/apache/rocketmq/example/benchmark/BatchProducer.java
+++ b/example/src/main/java/org/apache/rocketmq/example/benchmark/BatchProducer.java
@@ -40,6 +40,7 @@ import org.apache.rocketmq.client.producer.SendResult;
 import org.apache.rocketmq.client.producer.SendStatus;
 import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.UtilAll;
+import org.apache.rocketmq.common.compression.CompressionType;
 import org.apache.rocketmq.common.message.Message;
 import org.apache.rocketmq.remoting.RPCHook;
 import org.apache.rocketmq.remoting.exception.RemotingException;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by lizhimins — APPROVED** [review]

File: example/src/main/java/org/apache/rocketmq/example/benchmark/BatchProducer.java [hunk 1]
```diff
@@ -93,6 +96,19 @@ public static void main(String[] args) throws MQClientException {
         }
 
         final DefaultMQProducer producer = initInstance(namesrv, msgTraceEnable, rpcHook);
+
+        if (enableCompress) {
+            String compressType = commandLine.hasOption("ct") ? commandLine.getOptionValue("ct").trim() : "ZLIB";
+            int compressLevel = commandLine.hasOption("cl") ? Integer.parseInt(commandLine.getOptionValue("cl")) : 5;
+            int compressOverHowMuch = commandLine.hasOption("ch") ? Integer.parseInt(commandLine.getOptionValue("ch")) : 4;
```
Oliverwqcwrw, line 103: How about changing the default value of compressOverHowMuch to 4096?

humkum: You are right. It's my carelessness.

Oliverwqcwrw (approved):

PR #6255, opened 2023-03-06
[ISSUE #6026] Remove uncompatible sed in runserver.sh

aaron-ai:
`sed -r` is not compatible with MacOS, replace it by `awk`.

related issue: #6026


drpmma (approved):
LGTM

Issue #6269, opened 2023-03-07
correct delay offset larger than max offset


yiduwangkai:
issue tracker is used for bug reporting purposes ONLY whereas feature request needs to follow the [RIP process](https://github.com/apache/rocketmq/wiki/RocketMQ-Improvement-Proposal). To avoid unnecessary duplication, please check whether there is a previous issue before filing a new one.

It is recommended to start a discussion thread in the [mailing lists](http://rocketmq.apache.org/about/contact/) or [github discussions](https://github.com/apache/rocketmq/discussions) in cases of discussing your deployment plan, API clarification, and other non-bug-reporting issues.
We welcome any friendly suggestions, bug fixes, collaboration, and other improvements.

Please ensure that your bug report is clear and self-contained. Otherwise, it would take additional rounds of communication, thus more time, to understand the problem itself.

Generally, fixing an issue goes through the following steps:

Understand the issue reported;
Reproduce the unexpected behavior locally;
Perform root cause analysis to identify the underlying problem;
Create test cases to cover the identified problem;
Work out a solution to rectify the behavior and make the newly created test cases pass;
Make a pull request and go through peer review;
As a result, it would be very helpful yet challenging if you could provide an isolated project reproducing your reported issue. Anyway, please ensure your issue report is informative enough for the community to pick up. At a minimum, include the following hints:

BUG REPORT

Please describe the issue you observed:
What did you do (The steps to reproduce)?
1、run rocketmq 5.0 on test env

2、The current offset is greater than maxoffset when the exact delay has not been used for some time 
![image](https://user-images.githubusercontent.com/16093572/223415741-6891aa9b-e478-4ede-ad9b-211e0c2c0418.png)
i add some log
` log.info("Timer EnqueueGetService enqueue bufferCQ {}, currQueueOffset {} offset result {}", bufferCQ, currQueueOffset, (offset * CQ_STORE_UNIT_SIZE) > cq.getMinLogicOffset());`

Other information (e.g. detailed explanation, logs, related issues, suggestions on how to fix, etc):

echooymxq:
> 2、The current offset is greater than maxoffset when the exact delay has not been used for some time

I don't understand this point. What situations would cause this, can you describe it in more detail?

PR #6270, opened 2023-03-07
[ISSUE #6269] correct delay offset when larger than max offset

yiduwangkai:
Make sure set the target branch to develop

What is the purpose of the change
When i run rocketmq 5.0 in test, The current offset is greater than maxoffset when the exact delay has not been used for some time
This submission is to add max offset assign to fix this bug

Brief changelog
change TimerMessageStore

Verifying this change
i run delay message for many days, when i try to sole it like i summitted code

Verifying this change
You can compare the changed versions before and after. Build a k8s cluster by yourself, deploy a rocketmq cluster in it, use the nodePort mode of k8s, keep the port of nodePort mode different from the proxy grpcServerPort, and try to send and receive messages to verify

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -68,6 +68,8 @@ import org.apache.rocketmq.store.metrics.DefaultStoreMetricsManager;
 import org.apache.rocketmq.store.stats.BrokerStatsManager;
 import org.apache.rocketmq.store.util.PerfCounter;
 
+import static org.apache.rocketmq.store.ConsumeQueue.CQ_STORE_UNIT_SIZE;
+
 public class TimerMessageStore {
     public static final String TIMER_TOPIC = TopicValidator.SYSTEM_TOPIC_PREFIX + "wheel_timer";
     public static final String TIMER_OUT_MS = MessageConst.PROPERTY_TIMER_OUT_MS;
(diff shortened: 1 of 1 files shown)


store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -632,22 +637,23 @@ public boolean enqueue(int queueId) {
                     int sizePy = bufferCQ.getByteBuffer().getInt();
                     bufferCQ.getByteBuffer().getLong(); //tags code
                     MessageExt msgExt = getMessageByCommitOffset(offsetPy, sizePy);
+                    currQueueOffset = offset + (i / CQ_STORE_UNIT_SIZE);
```
drpmma, line 640: It's recommend to use CQ_STORE_UNIT_SIZE or ConsumeQueue.CQ_STORE_UNIT_SIZE in all occurrence.

drpmma: Why relocate the line from the end of the loop to this position?

yiduwangkai: The isRunningEnqueue method checkpoint the offset when the resource role changes. The offset in this case should be nextOffset

yiduwangkai: @drpmma @zhouxinyu help review it

yiduwangkai: @RongtongJin help review it

Issue #6398, opened 2023-03-19
Remove duplicate code in TopicPublishInfo


zyx333:
The code block below in the selectOneMessageQueue method is identical to another overloaded method (2nd code block).

https://github.com/apache/rocketmq/blob/7d6748445617dea43bc8c2c0410292e2eb423dc1/client/src/main/java/org/apache/rocketmq/client/impl/producer/TopicPublishInfo.java#L74-L76

https://github.com/apache/rocketmq/blob/7d6748445617dea43bc8c2c0410292e2eb423dc1/client/src/main/java/org/apache/rocketmq/client/impl/producer/TopicPublishInfo.java#L85-L90

IMO, it can be rewritten like this.
```
for (int i = 0; i < this.messageQueueList.size(); i++) {
         MessageQueue mq = selectOneMessageQueue()
         if (!mq.getBrokerName().equals(lastBrokerName)) {
               return mq;
         }
 }
```

If it's acceptable, I will submit a pull request.

Oliverwqcwrw:
IMO, It's a good idea, welcome your PR :-)

AdityaP7649:
Hi there! I'm interested in working on this issue. It aligns with my skills. Could you please assign this issue to me, or guide me on how to get started? Thanks!

benedikt-roth:
@Oliverwqcwrw I would like to work on this in case it is not picked up yet.

farukhfrk:
I would like to work on this, Could you please assign it to me?

ThestralWarrior:
i see this hasn't been assigned yet, can I work on this?

bitWizrd:
@Oliverwqcwrw   Kindly to ask if there is somebody working for this issue? lf not, I'm glad to undertake it.

Oliverwqcwrw:
Thank you very much for your attention to the community, I noticed that the latest code has changed, please take a look at the latest code to check whether the problem still exists :)

ling0900:
Is it possible for this issue to be closed in the near future? If not, I'll try to follow up on it.

PR #6499, opened 2023-03-29
ISSUE #6398

Swapnapratim:
To clarify, while there is indeed some duplicated code between the two selectOneMessageQueue methods, there are some differences as well, particularly in the handling of the lastBrokerName argument. Therefore, simply merging the two methods into one may not be the best solution in this case.

However, there are still some opportunities to refactor the code to reduce duplication. One option could be to extract the common code into a private helper method, which both selectOneMessageQueue methods can call as needed.
In this refactored version, the selectOneMessageQueue methods now simply call a private helper method, passing in null or the lastBrokerName argument as appropriate. The helper method contains the common code for selecting a message queue, including the loop, incrementing the queue index, and checking whether the broker name matches the given parameter.

Note that in the refactored version, I also added a fallback option.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/TopicPublishInfo.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/TopicPublishInfo.java
@@ -66,28 +66,54 @@ public class TopicPublishInfo {
         this.haveTopicRouterInfo = haveTopicRouterInfo;
     }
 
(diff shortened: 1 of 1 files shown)


mxsm (commented):
Hi @Swapnapratim  ISSUE #6398  has submit pr #6412 to fix it.

drpmma:
Close for duplication.

PR #6506, opened 2023-03-29
[ISSUE #6505] Fix the incorrect incremental epoch when autoSwitch ha.

### LINKED ISSUE #6505: The  incorrect incremental epoch when autoswitch ha.
image

echooymxq:

close #6505 

fix #<xxx> 

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/controller/src/main/java/org/apache/rocketmq/controller/impl/manager/ReplicasInfoManager.java
+++ b/controller/src/main/java/org/apache/rocketmq/controller/impl/manager/ReplicasInfoManager.java
@@ -533,7 +533,7 @@ public class ReplicasInfoManager {
             } else {
                 // If new master was not elected, which means old master was shutdown and the newSyncStateSet list had no more replicas
                 // So we should delete old master, but retain newSyncStateSet list.
-                syncStateInfo.updateMasterInfo(null);
+                syncStateInfo.cleanMasterInfo();
             }
             return;
         }
--- a/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
@@ -662,14 +662,16 @@ public class ReplicasManager {
                                 // If this broker is now the slave, and master has been changed
                                 changeToSlave(newMasterAddress, newMasterEpoch, masterBrokerId);
                             }
-                        } else {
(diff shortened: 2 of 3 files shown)


controller/src/main/java/org/apache/rocketmq/controller/impl/manager/ReplicasInfoManager.java
```diff
@@ -533,7 +533,7 @@ private void handleElectMaster(final ElectMasterEvent event) {
             } else {
                 // If new master was not elected, which means old master was shutdown and the newSyncStateSet list had no more replicas
                 // So we should delete old master, but retain newSyncStateSet list.
-                syncStateInfo.updateMasterInfo(null);
+                syncStateInfo.cleanMasterInfo();
```
RongtongJin, line 536: If the master's epoch is not increased and is set to null, it appears that the broker will not actively trigger a re election in the current implementation.If the master's epoch is not increased and is set to null, it appears that the broker will not actively trigger a re election in the current implementation.

echooymxq: Ok, let me understand what you mean.
> it appears that the broker will not actively trigger a re election in the current implementation.

Because in current implementation, broker will send elect request proactive when the epoch increased and master is null. code as follow:
```java
if (newMasterEpoch > this.masterEpoch) {
      if (StringUtils.isNoneEmpty(newMasterAddress) && masterBrokerId != null) {
          ....
      } else {
          // In this case, the master in controller is null, try elect in controller, this will trigger the electMasterEvent in controller.
          brokerElect();
      }
  }
```
If the epoch not increased when the master shutdown, this step maybe never occur. But discontinuous incremental epochs can be very confusing.

tsunghanjacktsai:
Hey @echooymxq , I understand your confusion, but maybe we could just add a log message after the master epoch increased. This might simply solve the unnecessary misunderstanding of the discontinuous epoch.

echooymxq:
> Hey @echooymxq , I understand your confusion, but maybe we could just add a log message after the master epoch increased. This might simply solve the unnecessary misunderstanding of the discontinuous epoch.

@tsunghanjacktsai  Really thank your reply. But in my opinion, the epoch should only increased when a new master elected. Maybe we should listen to more people.

Issue #5768, opened 2022-12-23
back-incompatible the RequestCode(`REGISTER_TOPIC_IN_NAMESRV`) with `217` value


TheNorthMemory:
1. Please describe the issue you observed:

There was described the `217` code as `GET_KV_CONFIG_BY_VALUE` label, see [here](https://github.com/apache/rocketmq-client-cpp/blob/master/src/protocol/MQProtos.h#L98-L100)

```c++
  GET_KV_CONFIG_BY_VALUE = 217,
```

But since `5.0.0-beta`,  the RequestCode(`217`) may be reusable as `REGISTER_TOPIC_IN_NAMESRV`

https://github.com/apache/rocketmq/blob/ec5d3236f65786be7dde712dda5f5ee152d0dce6/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RequestCode.java#L133

Those two kind of `RemotingCommand` operations looks like incompatible, the one `GET_KV_CONFIG_BY_VALUE` mostly same as fetching and another `REGISTER_TOPIC_IN_NAMESRV` sames to pushing.

It's suggest that toggle the `REGISTER_TOPIC_IN_NAMESRV` code to another one for the back compatible.

leizhiyuan:
It seems only be defined in cpp api

there is no use in cpp now。https://github.com/search?q=repo%3Aapache%2Frocketmq-client-cpp+GET_KV_CONFIG_BY_VALUE&type=code

TheNorthMemory:
> It seems only be defined in cpp api
> 
> there is no use in cpp now。https://github.com/search?q=repo%3Aapache%2Frocketmq-client-cpp+GET_KV_CONFIG_BY_VALUE&type=code

please take a look at this one [[rocketmq-client-nodejs](https://github.com/apache/rocketmq-client-nodejs)](https://github.com/apache/rocketmq-client-nodejs)
> This official Node.js client is a lightweight wrapper around [rocketmq-client-cpp](https://github.com/apache/rocketmq-client-cpp), a finely tuned CPP client.

leizhiyuan:
I see

Abhijeetmishr:
@TheNorthMemory @leizhiyuan @Oliverwqcwrw can I pick this up ?

TheNorthMemory:
> @TheNorthMemory @leizhiyuan @Oliverwqcwrw can I pick this up ?

Sure, I'm not good at JAVA and glad to see it is solved.

[...]

PR #6514, opened 2023-03-30
changed code of REGISTER_TOPIC_IN_NAMESRV to 1217

Abhijeetmishr:
## What is the purpose of the change - #5768 

> There was described the 217 code as GET_KV_CONFIG_BY_VALUE label, see [here](https://github.com/apache/rocketmq-client-cpp/blob/master/src/protocol/MQProtos.h#L98-L100)

> But since 5.0.0-beta, the RequestCode(217) may be reusable as REGISTER_TOPIC_IN_NAMESRV

> Those two kind of RemotingCommand operations looks like incompatible, the one GET_KV_CONFIG_BY_VALUE mostly same as fetching and another REGISTER_TOPIC_IN_NAMESRV sames to pushing.


lizhanhui:
I guess this would cause an incompatibility issue.

PR #6554, opened 2023-04-09
[#6535] Unify proxy loadingcache to caffeine LoadingCache

### LINKED ISSUE #6535: Replace guava LoadingCache with caffeine LoadingCache in proxy service

In proxy ClusterMetadataService, we use guava LoadingCache to cache topicConfig and subscriptionGroupConfig
image

In proxy TopicRouteService, we use caffeine LoadingCache to cache topic MessageQueueView
image

Factually, we may replace guava cache with caffeine, which is compatible and performs better.

redlsz:

fix #6535

## Brief changelog

Unify proxy loadingcache to caffeine

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/AbstractCacheLoader.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/AbstractCacheLoader.java
@@ -16,31 +16,25 @@
  */
 package org.apache.rocketmq.proxy.common;
 
-import com.google.common.cache.CacheLoader;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListenableFutureTask;
-import java.util.concurrent.ThreadPoolExecutor;
+import com.github.benmanes.caffeine.cache.CacheLoader;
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/common/AbstractCacheLoader.java
```diff
@@ -16,31 +16,25 @@
...
-    @Override
-    public ListenableFuture<V> reload(@Nonnull K key, @Nonnull V oldValue) throws Exception {
-        ListenableFutureTask<V> task = ListenableFutureTask.create(() -> {
+    @Override @Nonnull
+    public CompletableFuture<V> asyncReload(@Nonnull K key, @Nonnull V oldValue, @Nonnull Executor executor) {
+        CompletableFuture<V> future = new CompletableFuture<>();
+        executor.execute(() -> {
             try {
-                return getDirectly(key);
+                future.complete(getDirectly(key));
```
drpmma, line 31: Now that the cacheloader uses an async interface. Why not implement AsyncCacheLoader directly？
And would it be better that `getDirectly` also returns a future to adapt to the potential async interface?

PR #6562, opened 2023-04-10
Upgrade the image to the latest version

### LINKED ISSUE #6560: Upgrade github runner image to the latest version
See: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

In general, Java programs are not affected by the version of the operating system, so we should always use the latest version. By the way, it seems that we are missing a GitHub action runner for Ubuntu 18.04(see [here](https://github.com/apache/rocketmq/actions/runs/4634974063/jobs/8201637590)), which has been deprecated.

aaron-ai:
Fixes #6560

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -42,6 +42,9 @@ github:
         contexts:
           - misspell-check
           - check-license
+          - maven-compile (ubuntu-latest, JDK-8)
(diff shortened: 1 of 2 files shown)


xdkxlk (approved):
LGTM

PR #6623, opened 2023-04-20
[ISSUE #6598] Fix pop cause out memory error when message store not writeable

### LINKED ISSUE #6598: Pop consume cause out of memory  when not writeable.
https://github.com/apache/rocketmq/blob/3fe81bfbac11ee9a3bf9c21155d2c9c69570bb72/broker/src/main/java/org/apache/rocketmq/broker/processor/PopBufferMergeService.java#L262-L269
If the messageStore is not writeable, the pop buffer will cause OutOfMemoryError when the check point put to store  fail.

echooymxq:

close #6598 

Add return value for appendCheckPoint to determine whether to release message byte buffer.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopBufferMergeService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopBufferMergeService.java
@@ -402,8 +402,13 @@ public class PopBufferMergeService extends ServiceThread {
      * @param nextBeginOffset
      * @return
      */
-    public void addCkJustOffset(PopCheckPoint point, int reviveQueueId, long reviveQueueOffset, long nextBeginOffset) {
+    public boolean addCkJustOffset(PopCheckPoint point, int reviveQueueId, long reviveQueueOffset, long nextBeginOffset) {
         PopCheckPointWrapper pointWrapper = new PopCheckPointWrapper(reviveQueueId, reviveQueueOffset, point, nextBeginOffset, true);
+
+        if (this.counter.get() > brokerController.getBrokerConfig().getPopCkMaxBufferSize()) {
+            POP_LOGGER.warn("[PopBuffer]add ck, max size, {}, {}", point, this.counter.get());
+            return false;
+        }
 
         this.putCkToStore(pointWrapper, !checkQueueOk(pointWrapper));
 
@@ -413,6 +418,8 @@ public class PopBufferMergeService extends ServiceThread {
         if (brokerController.getBrokerConfig().isEnablePopLog()) {
             POP_LOGGER.info("[PopBuffer]add ck just offset, {}", pointWrapper);
         }
+
+        return true;
     }
 
     public void addCkMock(String group, String topic, int queueId, long startOffset, long invisibleTime,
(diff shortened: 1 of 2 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -602,6 +602,28 @@ private CompletableFuture<Long> popMsgFromQueue(boolean isRetry, GetMessageResul
...
+                            requestHeader.getConsumerGroup(),
+                            queueId, popTime, requestHeader.getInvisibleTime(), result.getMessageQueueOffset(),
+                            orderCountInfo);
+                        this.brokerController.getConsumerOffsetManager().commitOffset(channel.remoteAddress().toString(),
+                            requestHeader.getConsumerGroup(), topic, queueId, finalOffset);
+                    } else {
+                        boolean success = appendCheckPoint(requestHeader, topic, reviveQid, queueId, finalOffset, result, popTime, this.brokerController.getBrokerConfig().getBrokerName());
+                        if (!success) {
+                            // If append checkPoint error, will not return these messages.
+                            for (SelectMappedBufferResult mapedBuffer : result.getMessageMapedList()) {
```
drpmma, line 616: mappedBuffer

drpmma, line 616: mappedBuffer

drpmma (changes requested):

broker/src/main/java/org/apache/rocketmq/broker/processor/PopBufferMergeService.java
```diff
@@ -402,9 +402,14 @@ private boolean checkQueueOk(PopCheckPointWrapper pointWrapper) {
      * @param nextBeginOffset
      * @return
      */
-    public void addCkJustOffset(PopCheckPoint point, int reviveQueueId, long reviveQueueOffset, long nextBeginOffset) {
+    public boolean addCkJustOffset(PopCheckPoint point, int reviveQueueId, long reviveQueueOffset, long nextBeginOffset) {
         PopCheckPointWrapper pointWrapper = new PopCheckPointWrapper(reviveQueueId, reviveQueueOffset, point, nextBeginOffset, true);
 
+        if (this.counter.get() > brokerController.getBrokerConfig().getPopCkMaxBufferSize()) {
```
drpmma, line 408: The 

The `addCkJustOffset` will only be invoked when add ck to buffer failed.
So this condition will always be true and return false.

See addCk
https://github.com/apache/rocketmq/blob/8410785ef9d0bd2d2387bb5a706d84a4971356be/broker/src/main/java/org/apache/rocketmq/broker/processor/PopBufferMergeService.java#L447-L467

echooymxq: Yeah, if the buffer full, because the limit condition, the add ck to buffer will be failed, but the `addCkJustOffset ` will add buffer success.

lizhimins (approved):
Good catch

broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -602,6 +602,28 @@ private CompletableFuture<Long> popMsgFromQueue(boolean isRetry, GetMessageResul
...
+                                mapedBuffer.release();
+                            }
+                            atomicRestNum.set(result.getMaxOffset() - atomicOffset.get() + atomicRestNum.get());
```
lizhimins, line 619: Suggest adding comments here

PR #6663, opened 2023-04-27
[ISSUE #6662] Optimize the process of HA's confirmOffset calculation

### LINKED ISSUE #6662: The HA's confirmOffset calculation may be incorrect when the network is partitioned

### Describe the Bug

In controller mode, dual masters may occur when the network is partitioned. When the abnormal master (not recognized by the controller) is isolated by the network with a slave, its confirmOffset may change, causing the consumption point to move forward.
This may result in message loss.

在controller模式下，当网络分区时可能会导致双主。当异常的主（未被controller认可）有slave一同被网络隔离时，它的confirmOffset可能变化，导致消费位点前移。
这可能导致消息丢失。

In the case of two replicas, since the two masters have no slave Ack after the partition, the calculation of confirmOffset will not be triggered.
So this problem only occurs when master/slave are isolated at the same time.

在两副本的情况下，由于分区后两个master均没有slave Ack，因此不会触发confirmOffset的计算。
所以该问题仅出现于master/slave同时被隔离时。

### Steps to Reproduce

3 controllers, 3 brokers.

Using openchaos' random-partition may reproduce the bug. 

image

no lost message.

### What Did You See Instead?

image

### What Version Are You Using?

v5.1.0

### Environment

Compiler: openjdk version "1.8.0_362"
OS:  CentOS

### Additional Context

Maybe bug is in these code:
image

When the isolated master (A) is not linked to the current abnormal master (B) for a long time, the HAConnection (A) of the abnormal master (B) may be cleaned up.
当隔离后的master(A)长时间未链接至当前异常master(B)，异常master(B)的HAConnection(A)可能被清理。
image

When slave(C) synchronizes messages with master(B), it will request confirmOffset. At this time, calculating confirmOffset will exclude A, so confirmOffset moves forward.
当slave(C) 向master(B)同步消息时会请求confirmOffset, 此时计算confirmOffset会将A排除在外，因而confirmOffset向前移动。
image

When the confirmOffset is moved, the reportMessageService will build this part of the message into the consumeQueue.
当confirmOffset移动后，reputMessageService会将该部分消息构建进入consumeQueue。
image

After the part of the message is consumed, the part of the message will be truncated when the network partition ends, and the new message will not be consumed.
当该部分消息被消费后，网络分区结束时该部分消息将被截断，新的消息将无法被消费。

Since the abnormal master is isolated from the controller, the syncStateSet is not updated successfully. We can check whether all the brokers in the syncStateSet have established connections.
由于异常master与controller隔离，因此syncStateSet未更新成功，可以检查syncStateSet中的broker是否都建立有connection。

GenerousMan:

Fixes #6662

### Brief Description

Determine whether all brokers in syncStateSet have established connections. If not, the confirmOffset cannot be updated.
判断syncStateSet中的broker是否都建立有连接。如果没有，则无法更新confirmOffset.

using openchaos. 
So far no messages have been lost.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
@@ -114,7 +114,6 @@ public class ReplicasManager {
         this.scanExecutor = new ThreadPoolExecutor(4, 10, 60, TimeUnit.SECONDS,
             new ArrayBlockingQueue<>(32), new ThreadFactoryImpl("ReplicasManager_scan_thread_", brokerController.getBrokerIdentity()));
         this.haService = (AutoSwitchHAService) brokerController.getMessageStore().getHaService();
-        this.haService.setBrokerControllerId(this.brokerControllerId);
         this.brokerConfig = brokerController.getBrokerConfig();
         this.availableControllerAddresses = new ConcurrentHashMap<>();
         this.syncStateSet = new HashSet<>();
@@ -601,6 +600,7 @@ public class ReplicasManager {
         if (this.brokerMetadata.isLoaded()) {
             this.registerState = RegisterState.CREATE_METADATA_FILE_DONE;
             this.brokerControllerId = brokerMetadata.getBrokerId();
+            this.haService.setBrokerControllerId(this.brokerControllerId);
             return;
         }
         // 2. check if temp metadata exist
--- a/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
@@ -114,6 +114,7 @@ public class ReplicasManager {
         this.scanExecutor = new ThreadPoolExecutor(4, 10, 60, TimeUnit.SECONDS,
             new ArrayBlockingQueue<>(32), new ThreadFactoryImpl("ReplicasManager_scan_thread_", brokerController.getBrokerIdentity()));
         this.haService = (AutoSwitchHAService) brokerController.getMessageStore().getHaService();
+        this.haService.setBrokerControllerId(this.brokerControllerId);
         this.brokerConfig = brokerController.getBrokerConfig();
         this.availableControllerAddresses = new ConcurrentHashMap<>();
         this.syncStateSet = new HashSet<>();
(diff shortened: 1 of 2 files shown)


broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
```diff
@@ -268,6 +268,7 @@ public void changeToMaster(final int newMasterEpoch, final int syncStateSetEpoch
 
                 // Notify ha service, change to master
                 this.haService.changeToMaster(newMasterEpoch);
+                this.haService.setBrokerControllerId(this.brokerControllerId);
```
RongtongJin, line 271: After registration is completed, the brokerControllerId is determined, and there is no need to set it every time the master is switched.

GenerousMan: OK, I'll modify this logic.

store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAService.java
```diff
@@ -432,7 +434,7 @@ private long computeConfirmOffset() {
         long confirmOffset = this.defaultMessageStore.getMaxPhyOffset();
```
fuyou001, line 434: refactor confirmOffset to newConfirmOffset

GenerousMan: done, it has been renamed to newConfirmOffset.

broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
```diff
@@ -600,6 +600,7 @@ private void confirmNowRegisteringState() {
         if (this.brokerMetadata.isLoaded()) {
             this.registerState = RegisterState.CREATE_METADATA_FILE_DONE;
             this.brokerControllerId = brokerMetadata.getBrokerId();
+            this.haService.setBrokerControllerId(this.brokerControllerId);
```
RongtongJin, line 603: If it is the first time to start (without brokerIdentity file), the brokerControllerId will be forgotten to be set to HAService.

GenerousMan: Yes, you are right.
I have modify the code to set brokerControllerId when it starts without brokerIdentity file.

RongtongJin (approved):

PR #6669, opened 2023-04-29
[ISSUE #6668] Fix error unlocking issue during interruption

### LINKED ISSUE #6668: If an interrupt occurs, the lock is not acquired at this time, but it will be unlocked incorrectly
If an interrupt occurs, the lock is not acquired at this time, but it will be unlocked incorrectly

image

luozongle:
fix #6668

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
@@ -404,30 +404,31 @@ public class ProcessQueue {
     public void fillProcessQueueInfo(final ProcessQueueInfo info) {
         try {
             this.treeMapLock.readLock().lockInterruptibly();
+            try {
+                if (!this.msgTreeMap.isEmpty()) {
+                    info.setCachedMsgMinOffset(this.msgTreeMap.firstKey());
+                    info.setCachedMsgMaxOffset(this.msgTreeMap.lastKey());
+                    info.setCachedMsgCount(this.msgTreeMap.size());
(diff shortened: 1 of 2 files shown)


namesrv/src/main/java/org/apache/rocketmq/namesrv/routeinfo/RouteInfoManager.java
```diff
@@ -719,8 +719,8 @@ public TopicList getHasUnitSubUnUnitTopicList() {
     private TopicList topicQueueTableIter(Predicate<QueueData> pickCondition) {
         TopicList topicList = new TopicList();
         try {
+            this.lock.readLock().lockInterruptibly();
```
ni-ze, line 722: what the purpose of move code position?

client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
```diff
@@ -404,30 +404,31 @@ public void incTryUnlockTimes() {
...
-            info.setLastConsumeTimestamp(this.lastConsumeTimestamp);
+                info.setDroped(this.dropped);
+                info.setLastPullTimestamp(this.lastPullTimestamp);
+                info.setLastConsumeTimestamp(this.lastConsumeTimestamp);
+            } finally {
+                this.treeMapLock.readLock().unlock();
+            }
         } catch (Exception e) {
-        } finally {
-            this.treeMapLock.readLock().unlock();
```
ni-ze, line 430: the catch-ignored block can be removed.

caigy (commented):
IMO interrupted status should be restored when `InterruptedException` is thrown.

luozongle:
> 

What does IMO mean？
Should I restore the interrupt state when an interrupt occurs？

PR #6672, opened 2023-04-30
[ISSUE #6671]Enhance getting PID

### LINKED ISSUE #6671: Enhance getting PID

### Summary

use reflection to improve compatibility with Android Platform which does not support JMX extensions to get PID

### Motivation

image

use reflection to improve compatibility with Android Platform which does not support JMX extensions to get PID

mxsm:

Fixes #6671

### Brief Description

Enhance getting PID

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/UtilAll.java
@@ -82,7 +82,7 @@ public class UtilAll {
                 try {
                     // try a Linux-specific way
                     return Integer.parseInt(new File("/proc/self").getCanonicalFile().getName());
-                } catch (final IOException ignoredUseDefault) {
+                } catch (final Exception ignoredUseDefault) {
                     // Ignore exception.
                 }
             }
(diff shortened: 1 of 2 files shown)


fuyou001 (dismissed):

common/src/main/java/org/apache/rocketmq/common/UtilAll.java
```diff
@@ -69,12 +68,25 @@ public class UtilAll {
...
+
+                final Object runtimeMXBean = getRuntimeMXBean.invoke(null);
+                final String name = (String) getName.invoke(runtimeMXBean);
+
+                return Integer.parseInt(name.split("@")[0]);
+            } catch (final Exception ex) {
+                try {
+                    // try a Linux-specific way
+                    return Integer.parseInt(new File("/proc/self").getCanonicalFile().getName());
+                } catch (final IOException ignoredUseDefault) {
```
lizhiboo, line 85: Are ther any other types of exceptions thrown here?

mxsm: Catching Exception might be better, I will modify it and resubmit again

tsunghanjacktsai (approved):

PR #6680, opened 2023-05-02
[ISSUE #6679]SubscriptionData may not be updated when the client's system time is rolled back

### LINKED ISSUE #6679: SubscriptionData may not be updated when the client's system time is rolled back

### Describe the Bug

When the system time is set back to a previous time, it may cause the notification of client subscription information to Broker, but the message in Broker will not be modified.
image
image
image
If the client’s time rolls back to before, then here it may delete the subscription topic that is still within the expiration time normally.

### Steps to Reproduce

 system time is set back to a previous time

Update the changed subscription information

### What Did You See Instead?

Maybe not update it

### What Version Are You Using?

develop

mxsm:

Fixes #6679 

### Brief Description

- Fix SubscriptionData may not be updated when the client's system time is rolled back

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
@@ -57,7 +57,8 @@ public class RebalancePushImpl extends RebalanceImpl {
          * Fix: inconsistency subscription may lead to consumer miss messages.
          */
         SubscriptionData subscriptionData = this.subscriptionInner.get(topic);
-        long newVersion = System.currentTimeMillis();
+        //Ensure that the version number is greater than the previous one.
+        long newVersion = subscriptionData.getSubVersion() + 1;
         log.info("{} Rebalance changed, also update version: {}, {}", topic, subscriptionData.getSubVersion(), newVersion);
         subscriptionData.setSubVersion(newVersion);
 
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/heartbeat/SubscriptionData.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/heartbeat/SubscriptionData.java
@@ -38,6 +38,9 @@ public class SubscriptionData implements Comparable<SubscriptionData> {
     @JSONField(serialize = false)
     private String filterClassSource;
 
+    @JSONField(serialize = false)
+    private transient long lastUpdateTime = System.currentTimeMillis();
+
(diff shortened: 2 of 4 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
```diff
@@ -57,7 +57,8 @@ public void messageQueueChanged(String topic, Set<MessageQueue> mqAll, Set<Messa
          * Fix: inconsistency subscription may lead to consumer miss messages.
          */
         SubscriptionData subscriptionData = this.subscriptionInner.get(topic);
-        long newVersion = System.currentTimeMillis();
```
ferrirW, line 60: Maybe it could be nanoseconds

mxsm: This is also a good way, but I think using the increment method is simpler. Both have their own advantages.

drpmma:
Is there a possibility that introducing changes to Subversion will disrupt other functions?

mxsm:
> Is there a possibility that introducing changes to Subversion will disrupt other functions?

1. The generation of subversion only made changes to the client side, and no changes were made to the broker side. 
2. The expiration and removal of SubscriptionData is handled through the newly added lastUpdateTime.

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/heartbeat/SubscriptionData.java
```diff
@@ -38,6 +38,9 @@ public class SubscriptionData implements Comparable<SubscriptionData> {
     @JSONField(serialize = false)
     private String filterClassSource;
 
+    @JSONField(serialize = false)
+    private transient long lastUpdateTime = System.currentTimeMillis();
```
caigy, line 42: Should we use nanoTime?

mxsm: I don’t think it’s necessary. The lastUpdateTime is generated on the server where the Broker is located. Even if the client generates it, it will not be serialized and sent to the Broker, so there is no problem with changes

ni-ze (approved):
well done.

ni-ze:
image
Heatbeat does not update the timestamp of subscriptionData. It updated at pullRequest. so the consumer group will not be removed in removeExpireConsumerGroupInfo. The above modification is correct, IMO.

PR #6729, opened 2023-05-10
[ISSUE #6728] Compute the confirmOffset without considering new connections

### LINKED ISSUE #6728: [Bug] The confirmOffset may be set to -1 due to the new connection

OS: centOS

### RocketMQ version

5.1.0

### JDK Version

_No response_

### Describe the Bug

The confirmOffset may be set to -1 due to the new connection.

image

If a connection is inited, but has yet been TRANSFER state, then the slaveAckOffset will be -1.
It will make the newConfirmOffset be a wrong value, which is dangerous.

### Steps to Reproduce

add a log, then you will see the confirmOffset may be -1 value.

Compute the confirmOffset without those new connections.

### What Did You See Instead?

The confirmOffset was set to -1.

### Additional Context

_No response_

GenerousMan:
[ISSUE #6728] Compute the confirmOffset without considering new connections.

Fixes #6728 

### Brief Description

1. When calculate the newConfirmOffset, skip those connections with -1 ackOffset value.
2. Add a new method, getConfirmOffsetDirectly(). This method will only return the confirmOffset without checking and re-computing. Because in some cases (If set confirmOffset wrongly) endless calling will occur between the getConfirmOffset() and computeConfirmOffset().

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAService.java
@@ -421,13 +421,20 @@ public class AutoSwitchHAService extends DefaultHAService {
         for (Long syncId : currentSyncStateSet) {
             if (!idList.contains(syncId) && this.brokerControllerId != null && !Objects.equals(syncId, this.brokerControllerId)) {
                 LOGGER.warn("Slave {} is still in syncStateSet, but has lost its connection. So new offset can't be compute.", syncId);
-                return this.defaultMessageStore.getConfirmOffset();
+                // Without check and re-compute, return the confirmOffset's value directly.
+                return this.defaultMessageStore.getConfirmOffsetDirectly();
             }
         }
 
         for (HAConnection connection : this.connectionList) {
             final Long slaveId = ((AutoSwitchHAConnection) connection).getSlaveId();
             if (currentSyncStateSet.contains(slaveId)) {
+                long slaveAckOffset = connection.getSlaveAckOffset();
+                if (slaveAckOffset <= 0) {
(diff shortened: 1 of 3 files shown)


store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAService.java
```diff
@@ -421,14 +421,21 @@ public long computeConfirmOffset() {
...
         for (HAConnection connection : this.connectionList) {
             final Long slaveId = ((AutoSwitchHAConnection) connection).getSlaveId();
             if (currentSyncStateSet.contains(slaveId)) {
-                newConfirmOffset = Math.min(newConfirmOffset, connection.getSlaveAckOffset());
+                long slaveAckOffset = connection.getSlaveAckOffset();
+                if (slaveAckOffset <= 0) {
+                    // Slave's connection is just inited, the ack hasn't been calculated.
+                    // So skip this ackOffset.
+                    continue;
+                }
```
echooymxq, line 437: if (currentSyncStateSet.contains(slaveId) && connection.getSlaveAckOffset() > 0) {
 ...
}

echooymxq (commented):
currentSyncStateSet.contains(brokerAddr) && connection.getSlaveAckOffset() > 0

store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAService.java
GenerousMan: Yes, I have optimized this logic as you said.

fuyou001 (approved):
LGTM

mxsm (approved):
LGTM~

RongtongJin (approved):

Issue #6726, opened 2023-05-09
[Bug] org.apache.rocketmq.client.producer.DefaultMQProducer#request invoke timeout


haiyanghan:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

os: windows, macos

### RocketMQ version

develop

### JDK Version

jdk8

### Describe the Bug

1. Request timeout 
2. The message did not wait for consumers to consume before sending a callback

### Steps to Reproduce

Just change the send method in org. apache. locketmq. example. quickstart. Producer to request method

### What Did You Expect to See?

Synchronous return or asynchronous callback after consumer consumption

### What Did You See Instead?

After the message arrives at the broker, the broker sends a reply message to the producer,Actually, it should be when consumers consume before sending a reply message to the producer

PR #6730, opened 2023-05-10
[ISSUE apache#6726] Fix bug DefaultMQProducer#request method invoke t…

haiyanghan:
[ISSUE #6726]


tsunghanjacktsai (commented):
Hey @haiyanghan , thanks for your effort at first. I probably see what you're trying to do here according to the issue you created, but the meaning of message sending in MQ just simply means that the Broker would return a corresponding response once the message arrived without waiting it to be consumed. Of course, you could detail your thought if I've misunderstood your intention for doing this.

haiyanghan:
> Hey @haiyanghan , thanks for your effort at first. I probably see what you're trying to do here according to the issue you created, but the meaning of message sending in MQ just simply means that the Broker would return a corresponding response once the message arrived without waiting it to be consumed. Of course, you could detail your thought if I've misunderstood your intention for doing this.

Yes, when the message arrives at the broker, it returns, which is the logic of the original request method.
But the request method should be based on the request reply mode, waiting for the message to be consumed

haiyanghan:
> Hey @haiyanghan , thanks for your effort at first. I probably see what you're trying to do here according to the issue you created, but the meaning of message sending in MQ just simply means that the Broker would return a corresponding response once the message arrived without waiting it to be consumed. Of course, you could detail your thought if I've misunderstood your intention for doing this.



> Hey @haiyanghan , thanks for your effort at first. I probably see what you're trying to do here according to the issue you created, but the meaning of message sending in MQ just simply means that the Broker would return a corresponding response once the message arrived without waiting it to be consumed. Of course, you could detail your thought if I've misunderstood your intention for doing this.

Sorry, my English proficiency is not very good and the problem description is not very clear

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -1025,9 +1025,7 @@ public void registerProcessor() {
         replyMessageProcessor.registerSendMessageHook(sendMessageHookList);
 
         this.remotingServer.registerProcessor(RequestCode.SEND_REPLY_MESSAGE, replyMessageProcessor, replyMessageExecutor);
-        this.remotingServer.registerProcessor(RequestCode.SEND_REPLY_MESSAGE_V2, replyMessageProcessor, replyMessageExecutor);
```
aaron-ai, line 1028: Why is the `SEND_REPLY_MESSAGE_V2` being removed? Will this cause any compatibility issues?

haiyanghan: there may compatibility issues, A better way is to add a new one SEND_REPLY_MESSAGE_V3 message code

Guiqu1aixi (approved):

haiyanghan:
@Guiqu1aixi  hi, look here

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
haiyanghan: @aaron-ai hi, I have optimized the compatibility issue, Preserved `SEND_ REPLY_ MESSAGE` and `SEND_ REPLY_ MESSAGE_ V2` request code And the corresponding `processor`

PR #6735, opened 2023-05-11
[MINOR][Enhancement]Used string.format() instead of '+' operator concatenation.

nikam14:
### Brief Description

It is more suitable to use String.format() instead of '+' operator.


caigy:
What's the benefit of this modification?

nikam14:
Usage of String.format() method is slowing down the code as compare to the + operator concatenation.
I think closing the pr is good.

acl/src/main/java/org/apache/rocketmq/acl/common/AuthorizationHeader.java
```diff
@@ -51,14 +51,14 @@ public AuthorizationHeader(String header) throws DecoderException {
             String[] kv = keyValue.trim().split("=", 2);
             int kvLength = kv.length;
             if (kv.length != AUTH_HEADER_KV_LENGTH) {
-                throw new DecoderException("authorization keyValues length is incorrect, actual length=" + kvLength);
```
ferrirW, line 54: In my tests, + is much faster, so why change this here?

PR #6740, opened 2023-05-12
Remove redundant variables

haiyanghan:
This AtomicInteger variable is meaningless, there is no thread safety issue here, and the variable does not return


haiyanghan:
> Hi @haiyanghan Pleas link to issue.

ok

haiyanghan:
@mxsm PTAL

ferrirW (approved):

PR #6746, opened 2023-05-12
feat:add INNER_MULTI_DISPATCH check, can not contains '/' path separator

humkum:
Fixes #issue_id

### Brief Description

Check from client side and broker side:
Client side: INNER_MULTI_DISPATCH contains "/" throws MQClientException to avoid illegal LMQ path
Broker side: If receive old version clients message, generate CQ file for valid LMQ and skip generate CQ file for invalid LMQ

Client exception:
image
Broker log:
image

To avoid too many logs with every message, the code had delete the error log when skip generate the invalid LMQ.


lizhimins (approved):

DongyuanPan:
LMQ is currently used in the MQTT scenario. The Topic in the MQTT protocol contains multiple levels, so it will contain / characters. Therefore, such transmission cannot be prevented at the client level.

DongyuanPan (approved):

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -96,6 +96,7 @@ public class MixAll {
     public static final String REPLY_MESSAGE_FLAG = "reply";
     public static final String LMQ_PREFIX = "%LMQ%";
     public static final String MULTI_DISPATCH_QUEUE_SPLITTER = ",";
+    public static final String PATH_SEPARATOR = "/";
```
RongtongJin, line 99: Should we consider other operating systems, such as Windows, whose path delimiter is \?

ferrirW: +1, could use File.separator or System.getProperty("file.separator") @humkum

ferrirW (approved):

ferrirW (approved):

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -96,6 +96,7 @@ public class MixAll {
     public static final String REPLY_MESSAGE_FLAG = "reply";
     public static final String LMQ_PREFIX = "%LMQ%";
     public static final String MULTI_DISPATCH_QUEUE_SPLITTER = ",";
+    public static final String PATH_SEPARATOR = File.separator;
```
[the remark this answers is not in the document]
humkum: Plz help review again~

RongtongJin (approved):

PR #6786, opened 2023-05-22
[ISSUE #6785] Isolate the remoteChannel by group

### LINKED ISSUE #6785: [Bug] The subscription data of Remoting may overlap in proxy

Ubuntu

### RocketMQ version

branch: develop

### JDK Version

OpenJDK 1.8

### Describe the Bug

For the order version client of remoting, the channel is shared between different pushConsumers. As a result, subscription data from different groups may overlap when receiving channel data from other proxies.

### Steps to Reproduce

Start two remoting pushConsumers with different groups with proxy

The subscription data is consistent with the subscription set in the client

### What Did You See Instead?

The subscription data is not consistent with the subscription set in the client

### Additional Context

_No response_

xdkxlk:

Fixes #6785

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/service/sysmessage/HeartbeatSyncer.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/service/sysmessage/HeartbeatSyncer.java
@@ -49,7 +49,7 @@ public class HeartbeatSyncer extends AbstractSystemMessageSyncer {
 
     protected ThreadPoolExecutor threadPoolExecutor;
     protected ConsumerManager consumerManager;
-    protected final Map<String /* channelId as longText */, RemoteChannel> remoteChannelMap = new ConcurrentHashMap<>();
+    protected final Map<String /* group @ channelId as longText */, RemoteChannel> remoteChannelMap = new ConcurrentHashMap<>();
     protected String localProxyId;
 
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/service/sysmessage/HeartbeatSyncer.java
```diff
@@ -49,7 +49,7 @@ public class HeartbeatSyncer extends AbstractSystemMessageSyncer {
 
     protected ThreadPoolExecutor threadPoolExecutor;
     protected ConsumerManager consumerManager;
-    protected final Map<String /* channelId as longText */, RemoteChannel> remoteChannelMap = new ConcurrentHashMap<>();
+    protected final Map<String /* group @ channelId as longText */, RemoteChannel> remoteChannelMap = new ConcurrentHashMap<>();
```
chenzlalvin, line 52: Will the "@" symbol conflict with the Group character?

xdkxlk: This is not a problem, as long as clients from different groups can be distinguished here.

lollipopjin (approved):
LGTM

PR #6796, opened 2023-05-23
[ISSUE #6795] feat: support to send transaction msg with lmqQueue property

### LINKED ISSUE #6795: [Feature] Support to send transaction msg with lmqQueue property

When I send a msg with lmq property, two consumer indexes will be built for lmq topic.

Just build once.

Don't build lmq for msgs of `RMQ_SYS_TRANS_HALF_TOPIC` and `RMQ_SYS_TRANS_CHECK_MAX_TIME_TOPIC`

### Additional Context

_No response_

cserwen:

Fixes #6795


HScarb (commented):
Do we need to also add a check in `checkMultiDispatchQueue` method?

cserwen:
> checkMultiDispatchQueue

I don't think this is necessary. We have checked the message and assigned the queueOffset when writing it. If the index is not built because the check fails in the `checkMultiDispatchQueue`, the index offset will be discontinuous. However, in `checkMultiDispatchQueue` format validation is necessary.

cserwen:
@HScarb  Plz help me to review this.

PR #6801, opened 2023-05-23
[ISSUE #6800] fix: change the config variable isEnableBatchPush to enableBatchPush 

### LINKED ISSUE #6800: [Bug] isEnableBatchPush can not be changed since the setter name is wrong

Linux

### RocketMQ version

4.9.x 

### JDK Version

_No response_

### Describe the Bug

isEnableBatchPush can not be changed since the setter name is wrong

### Steps to Reproduce

1. add `isEnableBatchPush=true` in the broker.properties config
2. use `enableDLegerCommitLog=true`
3. saw batch push is not enabled in dledger mode and the var isEnableBatchPush in MessageStoreConfig  still false

using test to reproducer
```Java
@Test
    public void testBrokerControllerConfigInit() throws Exception {
        Properties properties = new Properties();
        properties.setProperty("isEnableBatchPush", "true");
        MessageStoreConfig messageStoreConfig = new MessageStoreConfig();
        MixAll.properties2Object(properties, messageStoreConfig);
        assertThat(messageStoreConfig.isEnableBatchPush()).isTrue();
    }

```

isEnableBatchPush should be changed after set  `isEnableBatchPush=true` in the broker.properties config

### What Did You See Instead?

isEnableBatchPush did not change in the MessageStoreConfig

### Additional Context

_No response_

absolute8511:
Change-Id: I1ebf4748c26c42efb058c11a8b5084b69213892c

Fixes #6800

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java
+++ b/store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java
@@ -235,7 +235,7 @@ public class MessageStoreConfig {
     private String dLegerPeers;
     private String dLegerSelfId;
     private String preferredLeaderId;
-    private boolean isEnableBatchPush = false;
+    private boolean enableBatchPush = false;
 
     private boolean enableScheduleMessageStats = true;
 
@@ -1126,11 +1126,11 @@ public class MessageStoreConfig {
     }
 
     public boolean isEnableBatchPush() {
-        return isEnableBatchPush;
+        return enableBatchPush;
     }
 
-    public void setIsEnableBatchPush(boolean enableBatchPush) {
-        isEnableBatchPush = enableBatchPush;
+    public void setEnableBatchPush(boolean enableBatchPush) {
+        this.enableBatchPush = enableBatchPush;
     }
 
     public boolean isEnableScheduleMessageStats() {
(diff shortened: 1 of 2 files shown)


store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java
```diff
@@ -1129,7 +1129,7 @@ public boolean isEnableBatchPush() {
         return isEnableBatchPush;
     }
 
-    public void setEnableBatchPush(boolean enableBatchPush) {
+    public void setIsEnableBatchPush(boolean enableBatchPush) {
         isEnableBatchPush = enableBatchPush;
```
ferrirW, line 1133: IMO，change the isEnableBatchPush to enableBatchPush is better

absolute8511: If change the var name I think it will break the compatibility.

cserwen: Now we can set it in broker.cfg by `enableBatchPush`, so use `enableBatchPush` directly is better. @absolute8511

absolute8511: @cserwen since old isEnableBatchPush is ignored, and I think we can change the variable name `isEnableBatchPush` to `enableBatchPush`  without break compatibility.

cserwen (approved):

DongyuanPan (approved):
LGTM

RongtongJin (approved):

PR #6812, opened 2023-05-25
fix: litepull consumer send trace data failed

### LINKED ISSUE #6809: [Bug] DefaultLitePullConsumer send trace data failed

centos7

### RocketMQ version

4.9.0+

### JDK Version

jdk8

### Describe the Bug

Fail to send trace data when use DefaultLitePullConsumer, with trace and acl enabled. 
image

The message consumption is ok.

### Steps to Reproduce

Enable broker acl.
Send some messages to a topic.
Start a DefaultLitePullConsumer to consume, with trace enabled.
Check the client WARN logs.

Succeed to send consumption trace data.

### What Did You See Instead?

Fail to send consumption trace data.

### Additional Context

The cause is:
image

We should use the custom rpchook, not hardcoding null.

redlsz:
Fixes #6809

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
@@ -180,6 +180,11 @@ public class DefaultLitePullConsumer extends ClientConfig implements LitePullCon
     private String customizedTraceTopic;
 
     /**
+     * The rpc hook
+     */
+    private RPCHook rpcHook;
+
+    /**
      * Default constructor.
      */
     public DefaultLitePullConsumer() {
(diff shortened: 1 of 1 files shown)


echooymxq (changes requested):
Maybe we should refactor it to like ` DefaultMQPushConsumer`, initialize the `TraceDispatcher` on the constructor.

redlsz:
> Maybe we should refactor it to like ` DefaultMQPushConsumer`, initialize the `TraceDispatcher` on the constructor.

It was before the [commit](https://github.com/apache/rocketmq/commit/2dc75d32099cf12e3caff0b32c7fa5fd1ce66a4c).

And bug in this issue is also introduced by the commit above.

leizhiyuan:
any update?

yuz10:
fixed in #7988

Issue #6813, opened 2023-05-25
[Enhancement] In controller mode, the master's BrokerRole will be set to SYNC_ MASTER


fengcharly:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

![image](https://github.com/apache/rocketmq/assets/35131874/8ade10c0-f051-4dd5-9cf4-88f7474b05e5)

org.apache.rocketmq.broker.controller.ReplicasManager#changeToMaster：

In controller mode, the master's BrokerRole will be set to SYNC_ MASTER
But this will cause my BrokerRole configuration to fail, such as setting it to ASYNC_ MASTER, how can I forcibly change my BrokerRole to SYNC_ MASTER, I don't think it's reasonable

### Motivation

In controller mode, the master's BrokerRole will be set to SYNC_ MASTER
But this will cause my BrokerRole configuration to fail, such as setting it to ASYNC_ MASTER, how can I forcibly change my BrokerRole to SYNC_ MASTER, I don't think it's reasonable

### Describe the Solution You'd Like

Delete the following line of code for changeToMaster:
this.brokerController.getMessageStoreConfig().setBrokerRole(BrokerRole.SYNC_MASTER);

### Describe Alternatives You've Considered

not have

fengcharly:
In election mode, the parameters brokerRole and brokerId become invalid, and the controller ultimately determines the master-slave relationship.

PR #6814, opened 2023-05-25
[ISSUE ##6813] Resolve inconsistency between brokerRole and configura…

fengcharly:
Resolve inconsistency between brokerRole and configuration in Controller mode
https://github.com/apache/rocketmq/issues/6813


RongtongJin (commented):
In the controller mode, setting role configuration is useless as it is determined and set solely by the controller.

fengcharly:
> In the controller mode, setting role configuration is useless as it is determined and set solely by the controller.

I think it is unacceptable to stand on the customer's side, for example, if the customer sets BrokerRole=ASYNC_ MASTER, as a result, it was changed to SYNC on the dashboard after deployment_ MASTER, don't mention the customer, even the testing phase won't let you pass

fengcharly:
> In the controller mode, setting role configuration is useless as it is determined and set solely by the controller.

Ok, I understand

Issue #6873, opened 2023-06-08
[Bug] If dns resolve controller address exception will update controllerAddresses to null


weihubeats:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

Ubuntu 20.04

### RocketMQ version

5.1.1

### JDK Version

1.8

### Describe the Bug

If dns resolve controller address exception will update controllerAddresses to null

### Steps to Reproduce

Let controller resolve dns exceptions

### What Did You Expect to See?

dns resolution exception does not update controller address

### What Did You See Instead?

dns resolution exception update controller address to null

PR #6874, opened 2023-06-08
[ISSUE #6873] Not update controller address list when the resolved result is empty

weihubeats:
https://github.com/apache/rocketmq/issues/6873


TheR1sing3un (approved):
LGTM~

xdkxlk:
I think we should differentiate between whether it throws an exception or actually returns null. And `dnsLookupAddressByDomain` is also used by `updateNameServerAddressListByDnsLookup` in `BrokerOuterAPI` .

weihubeats:
I think even if the actual is null, it should not be updated to null, because the actual is null is also abnormal.
Exceptions or actual null can be distinguished by logging

RongtongJin (commented):
If the parsing fails, it should be printed in the log. What are the benefits of setting it to null here?

ferrirW:
If parsing fails, the api will return the empty list, not null

you refered case can be provide more information for repetition ？

Issue #6881, opened 2023-06-09
[Bug] the messages in the delay queue are replayed when I upgraded broker from 4.9.1 to 5.1.1


gaoyf:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

OS: CentOS 6.9

### RocketMQ version

branch: (develop|tag 5.1.1) version: 5.1.1

### JDK Version

JDK: 1.8.0_202

### Describe the Bug

When I upgrade the broker from 4.9.1 to 5.1.1, when the broker starts to load data for more than 10 seconds, then the [persist timing task](https://github.com/apache/rocketmq/blob/develop/broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java#L98) in ScheduleMessageService will be executed, but the load method has not been executed yet, so the existing deleyOffset.json overwrites, eventually causing the schedule message to be replayed.

### Steps to Reproduce

As long as you find a broker with a scheduled message, and the commitlog reaches hundreds of G, it can be reproduced after starting with 5.x for more than 10 seconds. You can refer to the [case I wrote](https://blog.csdn.net/a417930422/article/details/131126969) for details.

### What Did You Expect to See?

the messages in the delay queue are not replayed when I upgraded broker.

### What Did You See Instead?

the messages in the delay queue are replayed when I upgraded broker.

gaoyf:
PR is https://github.com/apache/rocketmq/pull/6882

PR #6882, opened 2023-06-09
[ISSUE #6881] Fix scheduled messages are replayed bug

gaoyf:

https://github.com/apache/rocketmq/issues/6881

Fixes #issue_id

### Brief Description

scheduled messages are replayed bug

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
@@ -82,7 +82,7 @@ public class ScheduleMessageService extends ConfigManager {
     private DataVersion dataVersion = new DataVersion();
     private boolean enableAsyncDeliver = false;
     private ScheduledExecutorService handleExecutorService;
-    private final ScheduledExecutorService scheduledPersistService;
+    private ScheduledExecutorService scheduledPersistService;
     private final Map<Integer /* level */, LinkedBlockingQueue<PutResultProcess>> deliverPendingTable =
         new ConcurrentHashMap<>(32);
     private final BrokerController brokerController;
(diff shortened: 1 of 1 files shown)


broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
```diff
@@ -161,15 +152,15 @@ public void start() {
                 }
             }
 
-            this.deliverExecutorService.scheduleAtFixedRate(() -> {
+            scheduledPersistService = new ScheduledThreadPoolExecutor(1,
```
xdkxlk, line 155: Why do we need to change the code here? In my opinion, removing the scheduleAtFixedRate in the constructor is enough.

broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
```diff
@@ -82,7 +82,7 @@ public class ScheduleMessageService extends ConfigManager {
     private DataVersion dataVersion = new DataVersion();
     private boolean enableAsyncDeliver = false;
     private ScheduledExecutorService handleExecutorService;
-    private final ScheduledExecutorService scheduledPersistService;
+    private ScheduledExecutorService scheduledPersistService;
```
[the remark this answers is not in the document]
gaoyf: OK, I have changed the code.

broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
gaoyf: I have change the code. I think persist task should execute by scheduledPersistService but not deliverExecutorService.

RongtongJin (approved):

PR #6883, opened 2023-06-09
[ISSUE #6879] ConcurrentHashMapUtils fails to solve the loop bug in JDK8

### LINKED ISSUE #6879: [Bug] ConcurrentHashMapUtils fails to solve the loop bug in JDK8.

Mac  OS 13.0

### RocketMQ version

branch: (develop|tag 5.1.1) version: 5.1.1

### JDK Version

jdk8

### Describe the Bug

[ConcurrentHashMapUtils](https://github.com/apache/rocketmq/blob/develop/common/src/main/java/org/apache/rocketmq/common/utils/ConcurrentHashMapUtils.java) 中提供的computeIfAbsent方法没有解决jdk8的循环问题，复现代码如下

```
@Test
    public void computeIfAbsent() {
        ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();
//        map.computeIfAbsent("AaAa", key->map.computeIfAbsent("BBBB",key2->"42"));
        ConcurrentHashMapUtils.computeIfAbsent(map, "AaAa", key -> map.computeIfAbsent("BBBB", key2 -> "42"));
    }
```

leeyiyu:

Fixes #6879

Test Case
```
@Test
    public void computeIfAbsent() {

        ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();
//        map.computeIfAbsent("AaAa", key->map.computeIfAbsent("BBBB",key2->"42"));
        ConcurrentHashMapUtils.computeIfAbsent(map, "AaAa", key -> map.computeIfAbsent("BBBB", key2 -> "42"));
    }

```

Solution
```
public static <K, V> V computeIfAbsent(ConcurrentMap<K, V> map, K key, Function<? super K, ? extends V> func) {
        Objects.requireNonNull(func);
        if (isJdk8) {
            V v = map.get(key);
            if (null == v) {
                // v = map.computeIfAbsent(key, func);

                // this bug fix methods maybe cause `func.apply` multiple calls.
                v = func.apply(key);
                if (null == v) {
                    return null;
                }
                final V res = map.putIfAbsent(key, v);
                if (null != res) {
                    // if pre value present, means other thread put value already, and putIfAbsent not effect
                    // return exist value
                    return res;
                }
            }
            return v;
        } else {
            return map.computeIfAbsent(key, func);
        }
    }
```

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/ConcurrentHashMapUtils.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/ConcurrentHashMapUtils.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.common.utils;
 
+import java.util.Objects;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
 
@@ -40,10 +41,23 @@ public abstract class ConcurrentHashMapUtils {
      * @see <a href="https://bugs.openjdk.java.net/browse/JDK-8161372">https://bugs.openjdk.java.net/browse/JDK-8161372</a>
      */
     public static <K, V> V computeIfAbsent(ConcurrentMap<K, V> map, K key, Function<? super K, ? extends V> func) {
+        Objects.requireNonNull(func);
         if (isJdk8) {
             V v = map.get(key);
             if (null == v) {
-                v = map.computeIfAbsent(key, func);
+//                v = map.computeIfAbsent(key, func);
+
+                // this bug fix methods maybe cause `func.apply` multiple calls.
+                v = func.apply(key);
+                if (null == v) {
+                    return null;
+                }
+                final V res = map.putIfAbsent(key, v);
+                if (null != res) {
+                    // if pre value present, means other thread put value already, and putIfAbsent not effect
+                    // return exist value
+                    return res;
+                }
             }
             return v;
         } else {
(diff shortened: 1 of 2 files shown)


xiaoyifang:
check this ,https://juejin.cn/post/7094561581631012878

leeyiyu:
> 



> 



> check this ,https://juejin.cn/post/7094561581631012878
> 
> IMO, The original code has no problem.

You can run this Case,CPU utilization will be 100%

```
@Test
    public void computeIfAbsent() {

        ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();
        ConcurrentHashMapUtils.computeIfAbsent(map, "AaAa", key -> map.computeIfAbsent("BBBB", key2 -> "42"));
    }
```

Codeprh:
Reference dubbo link：https://github.com/apache/dubbo/pull/11987

xiaoyifang:
> ConcurrentHashMapUtils.computeIfAbsent(map, "AaAa", key -> map.computeIfAbsent("BBBB", key2 -> "42"));

This has caused a deadlock.  I think This should left for the user to avoid .

the above code is similar like 
```
synchronized(obj){
    synchronized(obj){
           //logic
    }

}
```

ferrirW:
> check this ,https://juejin.cn/post/7094561581631012878
> 
> IMO, The original code has no problem.

It doesn't seem to be same bug. The blog mentions is https://bugs.openjdk.org/browse/JDK-8161372

I think @leeyiyu fixed is: https://bugs.openjdk.org/browse/JDK-8062841

caigy (approved):

mxsm (approved):

PR #6903, opened 2023-06-14
fix fail to start broker if mapped file size is 0

yuz10:

Fixes #issue_id


store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
```diff
@@ -252,6 +252,12 @@ public boolean doLoad(List<File> files) {
                 continue;
             }
 
+            if (file.length() == 0) {
+                boolean ok = file.delete();
```
ferrirW, line 256: IMO, It is safer to add a judgment which ensure that it is the last file before deleting it

ferrirW (approved):

store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java
```diff
@@ -247,11 +247,18 @@ public boolean doLoad(List<File> files) {
...
         files.sort(Comparator.comparing(File::getName));
 
-        for (File file : files) {
+        for (int i = 0; i < files.size(); i++) {
+            File file = files.get(i);
             if (file.isDirectory()) {
                 continue;
             }
 
+            if (file.length() == 0 && i == files.size() - 1) {
```
lizhimins, line 256: We should investigate why there are files with a length of 0. Is it caused by pre-allocation or file switching?
If any file has a length of zero, it can be considered as damaged, Not only the last one.

yuz10: The file is empty because broker exits after file created and before mmap, new RandomAccessFile will create an empty file.

image

RongtongJin (approved):

PR #6918, opened 2023-06-18
[ISSUE #6917] fix cannot query all messages of the same msgId

### LINKED ISSUE #6917: [Bug] cannot query all messages of the same msgId

Ubuntu 20.04

### RocketMQ version

5.1.1

### JDK Version

1.8.0_362-8u362-ga-0ubuntu1~22.04-b09

### Describe the Bug

cannot query all messages of the same msgId, using mqadmin queryMsgByUniqueKey

### Steps to Reproduce

1. create some retry messages using benchmark: `./consumer.sh -n localhost:9876 -g group1 -r 1`
2. query messages: `./mqadmin queryMsgByUniqueKey -n localhost:9876 -i <msgId> -t %RETRY%group1 -a true`
  "-a true" means show all messages.

show multiple messages

### What Did You See Instead?

only one message

### Additional Context

_No response_

yuz10:

Fixes #6917

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
@@ -410,19 +410,7 @@ public class MQAdminImpl {
                     for (MessageExt msgExt : qr.getMessageList()) {
                         if (isUniqKey) {
                             if (msgExt.getMsgId().equals(key)) {
-
-                                if (messageList.size() > 0) {
-
-                                    if (messageList.get(0).getStoreTimestamp() > msgExt.getStoreTimestamp()) {
-
-                                        messageList.clear();
-                                        messageList.add(msgExt);
-                                    }
-
-                                } else {
-
-                                    messageList.add(msgExt);
-                                }
+                                messageList.add(msgExt);
                             } else {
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
```diff
@@ -410,19 +410,7 @@ public void operationComplete(ResponseFuture responseFuture) {
...
                         if (isUniqKey) {
                             if (msgExt.getMsgId().equals(key)) {
-
-                                if (messageList.size() > 0) {
-
-                                    if (messageList.get(0).getStoreTimestamp() > msgExt.getStoreTimestamp()) {
-
-                                        messageList.clear();
-                                        messageList.add(msgExt);
-                                    }
```
RongtongJin, line 420: If only one is to be kept, do we need to retain the original semantics and ensure that the storeTimestamp is the greatest for this one to be kept?

yuz10: Actually, A consumer that achieves idempotence only takes the first message, and ignores others. So I think it's not appropriate to keep the last message.

RongtongJin (approved):

PR #6925, opened 2023-06-21
Upgrading sonar-maven-plugin.version 

asia-zengtao:
Upgrading sonar-maven-plugin.version from 3.0.2 to 3.6.1.1688. The lower version of Sonar is causing an error.


Oliverwqcwrw:
Please create an issue to associate with this PR

PR #6988, opened 2023-07-03
[ISSUE #6987]feat:Add LMQ quantity check

### LINKED ISSUE #6987: [Enhancement] Add LMQ quantity check

### Summary

In the early stages of LMQ design，there used to be the limit of LMQ quantity, which was 20,000 every broker defaulted. But the logic of LMQ quantity limit had been removed later.

### Motivation

The introduction of LMQ will lead to the introduction of a large number of small index files into the RocketMQ system. Considering the limitation of the operating system file system inode, and the maintenance and dataStore of metadata is a very big challenge, so this limitation should be added.

Add the LMQ quantity check logic back

none

### Additional Context

_No response_

humkum:

Fixes #6987

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -529,6 +529,13 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
             }
         }
 
+        if (MixAll.isLmq(topic)) {
+            this.brokerController.getMessageStore().cleanUnusedLmqTopic(topic);
+            response.setCode(ResponseCode.SUCCESS);
+            response.setRemark(null);
+            return response;
+        }
+
         this.brokerController.getTopicConfigManager().deleteTopicConfig(requestHeader.getTopic());
         this.brokerController.getTopicQueueMappingManager().delete(requestHeader.getTopic());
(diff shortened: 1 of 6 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -529,6 +529,13 @@ private synchronized RemotingCommand deleteTopic(ChannelHandlerContext ctx,
             }
         }
 
+        if (MixAll.isLmq(topic)) {
```
ferrirW, line 532: asynchronously delete is better

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -545,6 +545,20 @@ public void destroyLogics() {
...
+    private PutMessageStatus checkLmqMessage(MessageExtBrokerInner msg) {
+        if (msg.getProperties() != null
+            && StringUtils.isNotBlank(msg.getProperty(MessageConst.PROPERTY_INNER_MULTI_DISPATCH))
+            && this.isLmqConsumeQueueNumExceeded()) {
+            return PutMessageStatus.LMQ_CONSUME_QUEUE_NUM_EXCEEDED;
+        }
+        return PutMessageStatus.PUT_OK;
+    }
+
+    private boolean isLmqConsumeQueueNumExceeded() {
```
ferrirW, line 557: IMO, you should split two patches:
1. add delete cq logic
2. limit queue number

PR #7012, opened 2023-07-11
fix receive message activity attempt id not correct

lizhimins:
fix receive message activity attempt id not correct


xdkxlk (approved):
LGTM

PR #7014, opened 2023-07-11
[ISSUE #7013]polish ColdDataCheckService's logic

### LINKED ISSUE #7013: [Enhancement] The logic of coldDataCheckService() needs to be optimized

Mac, M1

### RocketMQ version

latest

### JDK Version

_No response_

### Describe the Bug

Regardless of whether the cold read control switch is turned on, initPageSize() will be executed. However in some cases it may trigger an error.

### Steps to Reproduce

turn off the related switch: 
coldDataScanEnable                                =  false
coldDataFlowControlEnable                    =  false

then start the broker, there will still be such log in store.log: 
2023-07-03 13:54:01 INFO main - initPageSize pageSize: 4096

However, this discussion shows that in some cases, it will lead to an error:
https://github.com/apache/rocketmq/discussions/7007

The initialization should only be triggered after the switch is turned on.

### What Did You See Instead?

then start the broker, there is such log in store.log: 
2023-07-03 13:54:01 INFO main - initPageSize pageSize: 4096

### Additional Context

_No response_

GenerousMan:

Fixes #7013

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -2091,12 +2091,7 @@ public class CommitLog implements Swappable {
                     }
 
                     if (pageSize < 0) {
-                        try {
-                            initPageSize();
-                        } catch (Exception e) {
-                            log.error("Before scanning files in pagecache, initing has failed.", e);
-                            return;
-                        }
+                        initPageSize();
                     }
 
                     long beginClockTimestamp = this.systemClock.now();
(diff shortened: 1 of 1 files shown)


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -2089,6 +2089,16 @@ public void run() {
...
                         this.waitForRunning(defaultMessageStore.getMessageStoreConfig().getTimerColdDataCheckIntervalMs());
                     }
+
+                    if (pageSize < 0) {
+                        try {
+                            initPageSize();
+                        } catch (Exception e) {
+                            log.error("Before scanning files in pagecache, initing has failed.", e);
+                            return;
+                        }
```
RongtongJin, line 2099: 这里无法抓住java.lang.UnsatisfiedLinkError

RongtongJin (approved):

PR #7023, opened 2023-07-14
Fix completes TransactionSendResult

tianchaohongyu:
Fix completes TransactionSendResult


RongtongJin:
@tianchaohongyu Could you create a related issue to explain this PR?

tianchaohongyu:
> @tianchaohongyu Could you create a related issue to explain this PR?

The current modification involves copying the attribute values of SendResult to TransactionSendResult after sending a transactional message. 

The problem I am currently facing is that in Spring Cloud development, a transaction sends multiple transactional messages. Therefore, each time the "executeLocalTransaction()" method is automatically called after sending a transactional message, it can only return the "UNKNOWN" status. As a result, I need to additionally call "endTransaction()" immediately after the local transaction ends to notify the Broker of the transaction status. However, the TransactionSendResult object returned by the "sendMessageInTransaction()" method does not have complete attributes, which prevents me from calling "endTransaction()".

Furthermore, from a software design perspective, since TransactionSendResult inherits from SendResult, it should adhere to the Liskov Substitution Principle and have all the attribute values of SendResult.

PR #7030, opened 2023-07-17
[ISSUE #7029] Add a config to determine whether pop response should return the actual retry topic or tamper with the original topic

### LINKED ISSUE #7029: [Enhancement] Add a config to determine whether pop response should return the actual retry topic or tamper with the original topic

### Summary

Add a config to determine whether pop response should return the actual retry topic or tamper with the original topic

### Motivation

In #5568, the pop retry topic was directly changed to the origin topic on the broker (which was originally done on the client side), resulting in incompatibility for old versions using the remoting protocol to pop retry consumption.

I propose to add a config that can return the actual retry topic, thus making it compatible with older client versions.

Add a config to determine whether pop response should return the actual retry topic or tamper with the original topic

No

### Additional Context

_No response_

RongtongJin:

Fixes #7029 

### Brief Description

Add a config to determine whether pop response should return the actual retry topic or tamper with the original topic

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
@@ -589,8 +589,8 @@ public class PopMessageProcessor implements NettyRequestProcessor {
                 atomicRestNum.set(result.getMaxOffset() - result.getNextBeginOffset() + atomicRestNum.get());
                 String brokerName = brokerController.getBrokerConfig().getBrokerName();
                 for (SelectMappedBufferResult mapedBuffer : result.getMessageMapedList()) {
-                    // We should not recode buffer for normal topic message
-                    if (!isRetry) {
+                    // We should not recode buffer when popResponseReturnActualRetryTopic is true or topic is not retry topic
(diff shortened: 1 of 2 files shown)


caigy (approved):

broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -589,8 +589,8 @@ private CompletableFuture<Long> popMsgFromQueue(String attemptId, boolean isRetr
                 atomicRestNum.set(result.getMaxOffset() - result.getNextBeginOffset() + atomicRestNum.get());
                 String brokerName = brokerController.getBrokerConfig().getBrokerName();
                 for (SelectMappedBufferResult mapedBuffer : result.getMessageMapedList()) {
-                    // We should not recode buffer for normal topic message
-                    if (!isRetry) {
+                    // We should not recode buffer when popResponseReturnActualRetryTopic is true or topic is not retry topic
+                    if (brokerController.getBrokerConfig().isPopResponseReturnActualRetryTopic() || !isRetry) {
```
lizhimins, line 593: 不对吧，默认是 false，全走下面了

lizhimins (changes requested):

lizhimins (approved):

Issue #5068, opened 2022-09-14
java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;


yushan-yaojin:
java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;

1、开发环境、测试环境、生产环境 都是jdk11
2、在开发环境、测试环境正常 ，在生产环境报以上错误
3、测试、生产 环境服务器均是 centos

ShadowySpirits:
Please give more detailed information like exception stack.

RongtongJin:
> java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
> 
> 1、开发环境、测试环境、生产环境 都是jdk11 2、在开发环境、测试环境正常 ，在生产环境报以上错误 3、测试、生产 环境服务器均是 centos

使用的rocketmq版本是什么

aaron-ai:
This issue may be cause by cross-compilation, which is not supported with RocketMQ yet.

caigy:
It occurs when building with JDK version >=9 and running on JDK8.
Ref: https://www.morling.dev/blog/bytebuffer-and-the-dreaded-nosuchmethoderror/

aaron-ai:
We could consider to upgrade the JDK to 9+, which could use `--release` flag to support cross-compilation.

related link: https://openjdk.org/jeps/247

yushan-yaojin:
> > java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
> > 1、开发环境、测试环境、生产环境 都是jdk11 2、在开发环境、测试环境正常 ，在生产环境报以上错误 3、测试、生产 环境服务器均是 centos
> 
> 使用的rocketmq版本是什么
   mq服务端版本是：4.9.4 
   springboot应用端的mq是：rocketmq-spring-boot-starter  2.2.1
   jdk版本都是：openjdk version "11.0.10"   

  错误在：org.apache.rocketmq.common.message.MessageExt.socketAddress2ByteBuffer ,第80行

yushan-yaojin:
> Please give more detailed information like exception stack.
[...]

PR #7033, opened 2023-07-17
[ISSUE #5068] Fix return types  in ByteBuffer to Support JAVA9+

qianye1001:

Fixes #issue_id
https://github.com/apache/rocketmq/issues/5068

### Brief Description

Java 9 introduces overridden methods with covariant return types for the following methods in java.nio.ByteBuffer:

- position
- limit
- flip
- clear

After compiling in the Java11 environment, it can run in the Java8 environment and send and receive messages normally

### DIFF
--- a/acl/src/test/java/org/apache/rocketmq/acl/plain/PlainAccessControlFlowTest.java
+++ b/acl/src/test/java/org/apache/rocketmq/acl/plain/PlainAccessControlFlowTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.rocketmq.acl.plain;
 
+import java.nio.Buffer;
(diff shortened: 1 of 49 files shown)


aaron-ai:
Nice catch!

HattoriHenzo:
[late answer] To solve this issue I had to recompile the jar with the `JAVA_HOME `set to `Java 8 or 1.8`  because the maven settings for the version of Java are set like this:

```
<!-- Compiler settings properties -->
<maven.compiler.source>1.8</maven.compiler.source>
<maven.compiler.target>1.8</maven.compiler.target>
```
If you are using docker, use the **eclipse-temurin:8-jdk-alpine** image.

PR #7038, opened 2023-07-18
[#ISSUE 7035] Fix correct min offset behavior in tiered storage

### LINKED ISSUE #7035: [Bug] Fix correct min offset behavior in tiered storage

MacOS, JDK 11

### RocketMQ version

branch: develop latest

### JDK Version

JDK11

### Describe the Bug

When resetting the consumer's offset to an earlier time to start consuming data from tiered storage, we use "clean expired" to remove older data because commitlog and consume queue file have their own lifecycles. However, it is possible for cq file to still exist while the commitlog no longer exists. In this scenario, the data pointed to by cq unit is empty and there is no reasonable "next begin offset" to provide, preventing the client from correcting the position and self-healing the consumption process.

### Steps to Reproduce

No.

Fix offset success.

### What Did You See Instead?

No.

### Additional Context

_No response_

lizhimins:
fix #7035


drpmma (approved):
LGTM

PR #7045, opened 2023-07-19
passing the renew event type to create the correct context

### LINKED ISSUE #7044: [Enhancement] Passing the renew event type to create the correct context

### Summary

Passing the renew event type to create the correct context.

### Motivation

All renew event will use the same action name to changeInvisibleTime, causes cannot distinguish origin of calls.

Add event type in RenewEvent.

None.

### Additional Context

_No response_

xdkxlk:

Fixes #7044


drpmma (approved):
LGTM

lizhimins (commented):
ok

PR #7062, opened 2023-07-21
[ISSUE #7061] Support forward HAProxyMessage for Multi Protocol server.

### LINKED ISSUE #7061: [Enhancement] Support forward HAProxyMessage for Multi Protocol server.

### Summary

When we want to support both Remoting protocol and gRPC protocol on the same port, and enable Proxy Protocol, the information for that protocol will be lost during data forwarding. Therefore, we need to re-add the information for that protocol during forwarding.

### Motivation

In a multi-protocol environment, it is necessary to forward Proxy Protocol data simultaneously when forwarding other protocols.

1. In a multi-protocol service, when determining that a request is using gRPC protocol, data forwarding is required.
2. Before forwarding, it is necessary to determine whether there is Proxy Protocol data. 
3. If there is, these data should be repackaged as Proxy Protocol and forwarded first, and then the business-related data can be forwarded.
4. During this process, compatibility with SSL protocol encryption needs to be considered.

No.

### Additional Context

_No response_

dingshuangxi888:

Fixes #7061

### Brief Description

When we want to support both Remoting protocol and gRPC protocol on the same port, and enable Proxy Protocol, the information for that protocol will be lost during data forwarding. Therefore, we need to re-add the information for that protocol during forwarding.

1. Start proxy server with multi protocol.
2. Open the proxy protocol server.
3. Send message with grpc protocol.
4. Judge the proxy protocol is pass to the grpc server.


xdkxlk (approved):
LGTM

lollipopjin (approved):
LGMT

PR #7065, opened 2023-07-23
[ISSUE #7064] [RIP-66] Support KV(RocksDB) Storage

### LINKED ISSUE #7064: [RIP-66] Support KV(Rocksdb) Storage

### Summary

We will introduce kv(rocksdb) storage to solve the possible Full GC caused by large object memory allocation when metadata such as topic is serialized, and the performance degradation caused by random writing of a large number of small ConsumeQueue files in a million-topic scenario

### Motivation

we have encountered and been plagued by some probelms in the actual production environment. In some actual production scenarios, millions of topics and subsriptions will be created, and both of them may be deleted and created frequently. In current architecture, both topics and subcriptions are persisted in real time, which means each request of topics and subsciptions updating will trigger the persist interface, and the persistence of such metadata is written in full rather than append-only.  

In the scenario of millions of topics, frequent persistence generates the large memory object jsonString of the topicConfigTable. When the memory is tight, the large memory object jsonString will be directly allocated to the old generation, resulting in frequent Full GC.

In addition, millions of topics will inevitably bring millions of indexed ConsumeQueue small files, a large number of indexed small files will destroy the advantages of rocketmq sequential writing, and the random writing of ConsumeQueue will make the performance drop sharply.

This proposal mainly optimizes and solves the existing perfermance problems in the million-topic scenario from two levels:
1. Metadata like topic, subscription, consumerOffset implement kv storage
2. ConsumeQueue index file implements kv storage

no

### Additional Context

No response

fujian-zfj:

Fixes #7064

### Brief Description

This proposal mainly optimizes and solves the existing perfermance problems in the million-topic scenario from two levels:
1. Metadata like topic, subscription, consumerOffset implement kv storage
2. ConsumeQueue index file implements kv storage


common/src/main/java/org/apache/rocketmq/common/TopicConfig.java
```diff
@@ -27,7 +29,7 @@
 
 import static org.apache.rocketmq.common.TopicAttributes.TOPIC_MESSAGE_TYPE_ATTRIBUTE;
 
-public class TopicConfig {
+public class TopicConfig implements Serializable {
```
lizhimins, line 32: 不建议这么改

common/pom.xml
```diff
@@ -104,5 +104,10 @@
             <groupId>io.github.aliyunmq</groupId>
             <artifactId>rocketmq-logback-classic</artifactId>
         </dependency>
+        <!-- rocksdb -->
+        <dependency>
+            <groupId>io.github.aliyunmq</groupId>
```
lizhimins, line 109: Why rely on a commercial version? How to ensure the stability of this version?

lizhimins (changes requested):

common/pom.xml
fujian-zfj: It's still open source, but we need a new compactionFilter for ConsumeQueue. If you're interested, see how flink uses rocksdb.

pom.xml
```diff
@@ -973,6 +974,12 @@
                 <artifactId>jul-to-slf4j</artifactId>
                 <version>${jul-to-slf4j.version}</version>
             </dependency>
+            <!-- rocksdb -->
+            <dependency>
+                <groupId>io.github.aliyunmq</groupId>
+                <artifactId>rocketmq-rocksdb</artifactId>
+                <version>${rocksdb.version}</version>
+            </dependency>
```
aaron-ai, line 982: Is this feature compatible with other platforms (Windows and macOS)? If not, should we set it as optional?

fujian-zfj: support Windows, MacOS and Linux

common/src/main/java/org/apache/rocketmq/common/TopicConfig.java
fujian-zfj: 已删除

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
```diff
@@ -0,0 +1,1296 @@
...
+        super(messageStore);
+
+        this.storePath = StorePathConfigHelper.getStorePathConsumeQueue(messageStoreConfig.getStorePathRootDir());
+        this.rocksDBStorage = new ConsumeQueueRocksDBStorage(messageStore, storePath, 4);
+        this.flushOptions = new FlushOptions();
+        this.flushOptions.setWaitForFlush(true);
+        this.flushOptions.setAllowWriteStall(false);
+
+        this.writeBatch = new WriteBatch();
+        this.bufferDRList = new ArrayList(BATCH_SIZE);
```
HScarb, line 165: Raw use of parameterized class

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
```diff
@@ -0,0 +1,1296 @@
...
+    }
+
+    private boolean shutdownInner() {
+        this.consumeQueueCache.shutdown();
+        return this.rocksDBStorage.shutdown();
+    }
+
+    @Override
+    public void putMessagePositionInfoWrapper(DispatchRequest request) throws Exception {
+        if (request == null || this.bufferDRList.size() >= BATCH_SIZE) {
```
HScarb, line 231: Bulk put is used here, if the number of requests never reaches BATCH_SIZE, does it never put?

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
fujian-zfj: OK

lizhanhui:
It's a huge pull request, containing massive changes to core parts of this project with more than **6000** lines of change. Thus, it's challenging to have a quality review. Normally, developers would break it into a series of small and verifiable ones to keep community participation.

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -746,7 +760,12 @@ public boolean initializeMetadata() {
     public boolean initializeMessageStore() {
         boolean result = true;
         try {
-            DefaultMessageStore defaultMessageStore = new DefaultMessageStore(this.messageStoreConfig, this.brokerStatsManager, this.messageArrivingListener, this.brokerConfig, topicConfigManager.getTopicConfigTable());
+            DefaultMessageStore defaultMessageStore;
+            if (isEnableRocksDBStore()) {
```
lizhanhui, line 764: Will SPI be a better choice than `if...else`?

broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
```diff
@@ -225,7 +225,7 @@ public void shutdown() {
 
     public synchronized void changeBrokerRole(final Long newMasterBrokerId, final String newMasterAddress,
         final Integer newMasterEpoch,
-        final Integer syncStateSetEpoch, final Set<Long> syncStateSet) {
+        final Integer syncStateSetEpoch, final Set<Long> syncStateSet) throws Exception {
```
lizhanhui, line 228: `Exception` is too general...Create one within the project exception hierarchy

broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
```diff
@@ -235,7 +235,7 @@ public synchronized void changeBrokerRole(final Long newMasterBrokerId, final St
         }
     }
 
-    public void changeToMaster(final int newMasterEpoch, final int syncStateSetEpoch, final Set<Long> syncStateSet) {
+    public void changeToMaster(final int newMasterEpoch, final int syncStateSetEpoch, final Set<Long> syncStateSet) throws Exception {
```
lizhanhui, line 238: Same here

common/pom.xml
lizhanhui: @fujian-zfj  Your purpose of building custom rocksdb is to define a compaction filter. What a custom compaction filter expects is native function. Will this way be better alternative:
1. Use official rocksdb-java jar;
2. Build a shared library(dynamically linked with rocksdb shared library) for expected custom compaction filters;
3. When configuring RocksDB, pass the native function to it, because it does not matter which shared library the compaction filter stays in;
As a matter of fact, this is how the native programming languages use rocksdb in case of dynamic link.

broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
fujian-zfj: OK

lizhanhui (changes requested):
Please break this huge PR into smaller ones

fujian-zfj:
> Please

fujian-zfj:
> Please break this huge PR into smaller ones

OK, I will split it into two PRS, metadata and consumeQueue

Issue #7075, opened 2023-07-25
[Bug] NPE in proxy.processor.ConsumerProcessor#lockBatchMQ and #unlockBatchMQ


wuyoudexiao:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

mac os

### RocketMQ version

5.1.3

### JDK Version

java11

### Describe the Bug

> Set<AddressableMessageQueue> addressableMessageQueueSet = buildAddressableSet(ctx, mqSet);

addressableMessageQueueSet can be null when Function buildAddressableSet catch exception

> Map<String, List<AddressableMessageQueue>> messageQueueSetMap = buildAddressableMapByBrokerName(addressableMessageQueueSet);

in function buildAddressableMapByBrokerName, null will cause NTE

### Steps to Reproduce

Specially handle the returned null and print the log

### What Did You Expect to See?

no exception

### What Did You See Instead?

null pointer exception

PR #7078, opened 2023-07-25
fix: npe in lockBatchMQ and unlockBatchMQ

wuyoudexiao:

Fixes #issue_id
https://github.com/apache/rocketmq/issues/7075
### Brief Description
Set addressableMessageQueueSet = buildAddressableSet(ctx, mqSet);

addressableMessageQueueSet can be null when Function buildAddressableSet catch exception

Map<String, List> messageQueueSetMap = buildAddressableMapByBrokerName(addressableMessageQueueSet);

in function buildAddressableMapByBrokerName, null will cause NTE

print log and do not add null to addressableMessageQueueSet


xdkxlk (approved):
LGTM

Issue #7077, opened 2023-07-25
[Bug]  schedule CQ offset invalid. offset=77, cqMinOffset=0, cqMaxOffset=74, queueId=1


weihubeats:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

mac os(linux)

### RocketMQ version

5.1.0

### JDK Version

jdk8

### Describe the Bug

2023-07-25 10:33:49 ERROR BrokerControllerScheduledThread1 - schedule CQ offset invalid. offset=77, cqMinOffset=0, cqMaxOffset=74, queueId=1
2023-07-25 10:33:50 ERROR BrokerControllerScheduledThread1 - correct delay offset [ delayLevel 2 ] from 77 to 74

in reality
![image](https://github.com/apache/rocketmq/assets/42484192/dfe19282-6e84-4f54-867d-6f13d1b8d8e7)

Offset corrections were made here
![image](https://github.com/apache/rocketmq/assets/42484192/0fb97ac3-4793-4e75-a68d-d94ed0fadedf)
But it doesn't refresh to delayOffset.json

Then re-added in the delayOffset.json file for a corrective override
![image](https://github.com/apache/rocketmq/assets/42484192/6964457e-2ef6-4d65-8f20-12779755043a)

### Steps to Reproduce

1. star broker slave
2. delete store file

### What Did You Expect to See?

no error

### What Did You See Instead?

error
2023-07-25 10:33:49 ERROR BrokerControllerScheduledThread1 - schedule CQ offset invalid. offset=77, cqMinOffset=0, cqMaxOffset=74, queueId=1
2023-07-25 10:33:50 ERROR BrokerControllerScheduledThread1 - correct delay offset [ delayLevel 2 ] from 77 to 74

RongtongJin:
why delayOffset.json will reload? It seems that the two images are duplicated

weihubeats:
This is because SlaveSynchronize calls syncDelayOffset to synchronise delayed messages.
[...]

PR #7084, opened 2023-07-26
[Bug] schedule CQ offset invalid. offset=77, cqMinOffset=0, cqMaxOffset=74, queueId=1

weihubeats:
https://github.com/apache/rocketmq/issues/7077


weihubeats:
@RongtongJin I understand that it was repaired in this way, take a look

broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
```diff
@@ -250,7 +250,6 @@ public boolean correctDelayOffset() {
                 }
                 if (correctDelayOffset != currentDelayOffset) {
                     log.error("correct delay offset [ delayLevel {} ] from {} to {}", delayLevel, currentDelayOffset, correctDelayOffset);
-                    offsetTable.put(delayLevel, correctDelayOffset);
```
RongtongJin, line 253: If it is not put in the table, the correction will have no effect.

weihubeats: So there should be no reloading of the `delayOffset.json` file here, right?

broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
```diff
@@ -218,8 +218,7 @@ public String encode() {
 
     @Override
     public boolean load() {
-        boolean result = super.load();
-        result = result && this.parseDelayLevel();
```
RongtongJin, line 222: If super.load() is not called, the JSON file will not be parsed, even during startup. IMO, we can add a new method called loadWhenSyncDelayOffset, which will not correct the delay offset when syncDelayOffset is called.

weihubeats: I added a `loadWhenSyncDelayOffset` method, called at `this.brokerController.getScheduleMessageService().loadWhenSyncDelayOffset();`. `loadWhenSyncDelayOffset `doesn't read `delayOffset.json` repeatedly.

RongtongJin: IMO, we should read delayOffset.json and should not correct the delay offset when slave broker sync delay offset.

weihubeats: I don't know if I understand correctly, but we correct the offset when we initialize the startup, not when we synchronize it

weihubeats: This doesn't feel right either.

weihubeats:
Do we not need to call `correctDelayOffset` in syncDelayOffset, because in `syncDelayOffset` we have already called correctDelayOffset on`delayOffset.json` via `MixAll.string2File(delayOffset, fileName);`

RongtongJin (approved):

PR #7090, opened 2023-07-28
[ISSUE #7087] fix:consumer offset compatible problem caused by LmqConsumerOffsetManager

### LINKED ISSUE #7087: [Bug] Compatible problem caused by LMQ(Light Message Queue)

CentOS 7.3

### RocketMQ version

4.8.x upgrade to 4.9.3+

### JDK Version

OpenJDK 1.8.0_202

### Describe the Bug

During the process of broker upgrade, the consumer will get the consume offset from the broker when client rebalance occurs. When the RocketMQ broker version is upgraded from 4.8.x to 4.9.x and the LMQ function is enabled. The broker will only initialize the LmqConsumerOffsetManager as follows:
```this.consumerOffsetManager = messageStoreConfig.isEnableLmq() ? new LmqConsumerOffsetManager(this) : new ConsumerOffsetManager(this);```
0da1dc62-ebb6-4ef6-a69d-bd98a747c403
And the persistent file name of LmqConsumerOffsetManager is **lmqConsumerOffset.json**, which is different from the original consume offset file name **consumerOffset.json**. For the above reasons, the new version LmqConsumerOffsetManager cannot load the previous consume offset when the broker restarts. The broker side will return consume offset 0 to the client. As follows:
image
Based on the above logic, when the consume offset is 0 at the consumer side, which is less than the minimum consume offset, it will start to pull from the minimum offset. Lead to consumption accumulation and consumption duplication.
image
Consumer clients would consume from the oldest offset when get consume offset 0.

### Steps to Reproduce

1. Broker version less than 4.9.3.
2. There's topics who's min offset is bigger than 0.
3. Broker version upgrade to version 4.9.3+ 

Consumer offset should not be reset to minimum.

### What Did You See Instead?

Consumer offset was reset to minimum.

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by humkum**
…ager

Fixes #7087

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/LmqConsumerOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/LmqConsumerOffsetManager.java
@@ -20,12 +20,18 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.BrokerPathConfigHelper;
 import org.apache.rocketmq.common.MixAll;
+import org.apache.rocketmq.common.constant.LoggerName;
+import org.apache.rocketmq.logging.org.slf4j.Logger;
+import org.apache.rocketmq.logging.org.slf4j.LoggerFactory;
 import org.apache.rocketmq.remoting.protocol.RemotingSerializable;
 
 public class LmqConsumerOffsetManager extends ConsumerOffsetManager {
+    protected static final Logger LOG = LoggerFactory.getLogger(LoggerName.BROKER_LOGGER_NAME);
+
     private ConcurrentHashMap<String, Long> lmqOffsetTable = new ConcurrentHashMap<>(512);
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ferrirW** [comment]
@humkum It seems that changes involving file names are generic, so consider placing the compatibility logic in the parent class ConfigManager

Please help review whether such modify is appropriate. @RongtongJin @tianliuliu

File: broker/src/main/java/org/apache/rocketmq/broker/offset/LmqConsumerOffsetManager.java [hunk 1]
```diff
@@ -36,6 +40,92 @@ public LmqConsumerOffsetManager(BrokerController brokerController) {
         super(brokerController);
     }
 
+    @Override
+    public boolean load() {
```
ferrirW, line 44: IMO, compatibility in the ConfigManager would be better

humkum: Compatibility logic can not remove to parent class, because it involves many different implementation methods, such as configFilePath, which can not obtain in parent class.

humkum:
I made some changes to fix the compatibility issues for upgrades and rollbacks, please help to review, thanks. @RongtongJin @tianliuliu

PR #7150, opened 2023-08-10
Update mqbroker to use runbroker.sh instead of runserver.sh when enabling --enable-proxy

iamgd67:

no issue

### Brief Description

Update mqbroker to use runbroker.sh instead of runserver.sh when enabling `--enable-proxy` this allow JVM `heap` and `gc` configuration using broker's settings instead of other common serverices'(proxy,namenode, etc). our main purpose, like the filename `mqbroker` suggest, is to start broker (which embeds a proxy), so use broker's config is reasonable

chinese version
mqbroker的--enable-proxy选项是启动内嵌了proxy的broker，而不是内嵌broker的proxy，而且broker的工作量和重要程度大于proxy，所以使用broker的gc和heap配置更合适

after the change, started broker using configs in  runbroker.sh and work well

### DIFF
--- a/distribution/bin/mqbroker
+++ b/distribution/bin/mqbroker
@@ -68,7 +68,7 @@ if [ "$enable_proxy" = true ]; then
   if [ "$broker_config" != "" ]; then
       args_for_proxy=${args_for_proxy}" -bc "${broker_config}
   fi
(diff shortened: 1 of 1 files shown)


iamgd67:
@iamgd67 
Looks like some checks have been hung for 17 hourse. Can you retrigger them?

xdkxlk:
> @xdkxlk cleaned commit history, can you trigger the workflows?

OK

iamgd67:
@xdkxlk are we ready to merge this or waiting for more reviews

PR #7217, opened 2023-08-21
[ISSUE #7171] add misspell check in configuration during broker startup. 

### LINKED ISSUE #7171: [Enhancement] add misspell check in configuration during broker startup.

### Summary

Make a clear warning log when user set a wrong property in Configuration, It would be more friendly and got greater experience to users, especially new users.

### Motivation

In previous usage of RocketMQ, I incorrectly set a configuration item `brokerIp1 = xxx.xxx` (which correct property name is `brokerIP1` instead of `brokerIp1`). but RocketMQ cannot give a clear log or any other thing to warning me that you got a wrong configuration item.

Then I got many problems after start RocketMQ.（Because I didn't know that I make an incorrect configuration item）

If RocketMQ can check the configuration file when loading it and make an explicit warning log to user. It would be more friendly and make a greater experience for user, especially new users.

Check the property name and make a clear warning log when got a wrong property in Loading Configuration Process.

or ignore it , then keep going.

### Additional Context

_No response_

yao-wenbin:

Fixes #7171 

### Brief Description
add misspell check during load configuration file to config instance.
when user's configuration file has incorrect or misspell properties, they will get clearly warning log in console.
such as below：
image

image

add ``testBuildBrokerController``  test case in ``BrokerStartupTest`` to assert should have warning log print when configuration file havs invalid properties (misspell occur).

add ``testInvalidProperties`` test case in ``MixAllTest`` to assert ``invalidProperties`` method work correctly

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/MixAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/MixAll.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.common;
 
+import com.google.common.collect.Sets;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
@@ -36,6 +37,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -518,4 +520,25 @@ public static boolean isSysConsumerGroupForNoColdReadLimit(String consumerGroup)
...
+            for (Method method : methods) {
+                String mn = method.getName();
+                if (mn.startsWith("get")) {
+                    // get getter's field name and to camel name.
+                    fieldNames.add(mn.substring(3, 4).toLowerCase() +  mn.substring(4));
+                }
+            }
+        }
+
+        return Sets.difference(propertyNames, fieldNames);
```
joeCarf, line 542: I think it may miss logging when propertyNames not contains all fieldNames

yao-wenbin: I'm a little confused. The difference method  only return a set of property in property name set (which is user configure) but not in field names set (which is RocketMQ Configuration Item). So I think it will correctly record the invalid property name set in user's configuration file.
I have some test case in ``MixAllTest.java``, It would be  better explain that when logging will be missed if you could give me some clearly test case.~

joeCarf: Gotcha！there are tests failures, plz fix it :)
image

Issue #7055, opened 2023-07-20
[Feature] Micrometer Observation Support


marcingrzejszczak:
### Is Your Feature Request Related to a Problem?

I'm a co-maintainer of [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth/) and [Micrometer](https://micrometer.io/) projects (together with @shakuzen and @jonatan-ivanov).

[Micrometer Observation](https://micrometer.io/docs/observation) is part of the Micrometer 1.10 release and [Micrometer Tracing](https://micrometer.io/docs/tracing) is a new project. The idea of Micrometer Observation is that you instrument code once but you get multiple benefits out of it - e.g. you can get tracing, metrics, logging or whatever you see fit).

Since this project supports OpenTracing that is deprecated I was curious if there's interest in adding Micrometer Observation support so that except for metrics, spans could be created and tracing context propagation could happen too. Via Micrometer Tracing one can use OpenTelemetry or OpenZipkin Brave Tracer, but with the handler mechanism the possibilities are endless :)

If there's such interest we could provide a PR to add support for that.

### Describe the Solution You'd Like

I would add a PR that uses Micrometer Observation to instrument the project

### Describe Alternatives You've Considered

In order to have both metrics and tracing one would need to use the deprecated opentracing instrumentation + metrics instrumentation. With Micrometer Observation you instrument once and you can get metrics, tracing, logging etc. just by configuring it

marcingrzejszczak:
I've added a draft PR here https://github.com/apache/rocketmq/pull/7248

jonatan-ivanov:
Can we keep this alive or, which is better, review and merge the PR?

jonatan-ivanov:
Could somebody take a look at the connected PR?
https://github.com/apache/rocketmq/pull/7248

PR #7248, opened 2023-08-24
Micrometer observation

marcingrzejszczak:
This is a draft PR since there are a couple of questions / issues that need to be addressed

- I've created a separate `micrometer` package so that the code is isolated, but I can move it next to OpenTracing packages
- I tried to follow the current (24.08.2023) Semantic conventions for messaging https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/messaging/ but need help with all the remaining `TODO:` comments as far as tags are concerned
- Where should we put the documentation to this integration?
- How do we deal with scoping (span scopes) when consuming batches? How often do we process a single message? If we left an open scope for a span how would users then close it? Can it be done automatically?

https://github.com/apache/rocketmq/issues/7055

### Brief Description

Adds Micrometer Observation support via the `ConsumeMessageHook` and `SendMessageHook`

I've added 2 tests that reuse OpenTracing's testing approach. They test for the presence of spans and metrics.

Got the following results in Zipkin:

.Receive Span
image

.Produce Span
image


marcingrzejszczak:
hey @vintagewang @YangJodie , WDYT about this feature? Are you willing to accept it in any way?

jonatan-ivanov:
@vintagewang @YangJodie Could you please add no stale label?

jonatan-ivanov:
Could somebody please add no stale label?

jonatan-ivanov:
Could somebody please add no stale label?

Issue #7267, opened 2023-08-26
[Bug] Failed to obtain semaphore before sending request and threw an error exception


CLFutureX:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

window 10

### RocketMQ version

develop(5.1.3)

### JDK Version

JDK 1.8.0_372

### Describe the Bug

Condition judgment error
code path: org.apache.rocketmq.remoting.netty.NettyRemotingAbstract#invokeAsyncImpl
When obtaining semaphore resource timeout, if the timeout <= 0, then a timeout exception should be thrown , otherwise a invoke that is too fast exception will be thrown.
But now the situation is this ： When obtaining semaphore resource timeout, if the timeout <= 0, then a invoke that is too fast exception be thrown , otherwise a timeout exception be thrown.
![image](https://github.com/apache/rocketmq/assets/44110731/64853ac8-9541-4ffc-a0b9-af8e396d20b6)

### Steps to Reproduce

Code level defects

### What Did You Expect to See?

When obtaining semaphore resource timeout, if the timeout <= 0, then a timeout exception should be thrown , otherwise a invoke that is too fast exception will be thrown.

### What Did You See Instead?

When obtaining semaphore resource timeout, if the timeout <= 0, then a a invoke that is too fast exception  be thrown , otherwise a timeout exception be thrown.

PR #7268, opened 2023-08-26
[ISSUE #7267]fix : Failed to obtain semaphore before sending request and threw an error exception

CLFutureX:
[issue](https://github.com/apache/rocketmq/issues/7267)

Fixes #issue_id

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -534,7 +534,7 @@ public abstract class NettyRemotingAbstract {
                 throw new RemotingSendRequestException(RemotingHelper.parseChannelRemoteAddr(channel), e);
             }
         } else {
-            if (timeoutMillis <= 0) {
+            if (timeoutMillis > 0) {
                 throw new RemotingTooMuchRequestException("invokeAsyncImpl invoke too fast");
             } else {
                 String info =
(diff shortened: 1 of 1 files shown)


mxsm (commented):
@jimjag Please link this pr to issue

RapperCL:
> @jimjag Please link this pr to issue

Okay

joeCarf (commented):
IMO, the `timeoutMillis` stands for the time-limit, there should be timeoutException only when time-limit is above 0; if the time limit is 0, it is not a timeout. So i think the original way has no problem

RapperCL:
> IMO, the `timeoutMillis` stands for the time-limit, there should be timeoutException only when time-limit is above 0; if the time limit is 0, it is not a timeout. So i think the original way has no problem

However, in the current context semantics, timeout represents the remaining time of the current operation. Moreover, the underlying timeout time is usually mandatory and should not be less than 0, unless in higher application layers, we can set it as follows: when the timeout is less than 0, we set a default timeout time.

joeCarf:
`timeout represents the remaining time of the current operation` only when `acquired` is false, which means too many request comes in and over the limit

RapperCL:
> `timeout represents the remaining time of the current operation` only when `acquired` is false, which means too many request comes in and over the limit

When timeoutMillis is greater than 0, if tryAcquire fails, does it not mean that requests are more frequent at this time？

RapperCL:
> `timeout represents the remaining time of the current operation` only when `acquired` is false, which means too many request comes in and over the limit

I want to throw a timeout exception when timeout<=0, considering that the previous operation consumed too much time, resulting in a direct interruption after obtaining the semaphore failed

PR #7272, opened 2023-08-27
[ISSUE #7271]Optimize the configuration for setting the quantity of TimerDequeuePutMessageService

### LINKED ISSUE #7271: [Enhancement] Optimize the configuration for setting the quantity of TimerDequeuePutMessageService

### Summary

image
The quantity of TimerDequeuePutMessageService and TimerDequeueGetMessageService services is governed by a common configuration.

### Motivation

The number of TimerDequeuePutMessageService and TimerDequeueGetMessageService services varies according to different configurations.

Set the number of service threads for TimerDequeuePutMessageService using the timerPutMessageThreadNum configuration."

Set the number of service threads for TimerDequeuePutMessageService using the timerPutMessageThreadNum configuration.

### Additional Context

_No response_

mxsm:

Fixes #7271 

### Brief Description

- Optimize the configuration for setting the quantity of TimerDequeuePutMessageService

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -222,7 +222,7 @@ public class TimerMessageStore {
             dequeueGetMessageServices[i] = new TimerDequeueGetMessageService();
         }
 
-        int putThreadNum = Math.max(storeConfig.getTimerGetMessageThreadNum(), 1);
+        int putThreadNum = Math.max(storeConfig.getTimerPutMessageThreadNum(), 1);
         dequeuePutMessageServices = new TimerDequeuePutMessageService[putThreadNum];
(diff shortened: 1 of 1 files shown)


joeCarf (commented):
Is this a bug? or will it cause some problem ?

mxsm:
> 

@joeCarf I personally don't consider this a bug, but the quantities of the TimerDequeuePutMessageService and TimerDequeueGetMessageService can be configured separately. It's like having separate controls for the inlet and outlet sizes of a pipeline – they can be adjusted individually rather than being the same for both. It's also analogous to a consumer queue, where TimerDequeuePutMessageService puts data into the queue, and TimerDequeuePutMessageService retrieves data from it. The producer and consumer don't necessarily need to match; adjustments can be made according to the actual situation. This approach offers more flexibility. Of course, there's no issue with the current approach of having both quantities be the same. Personally, I believe that separate configuration would be a better choice.

RongtongJin (approved):

PR #7287, opened 2023-08-31
[Enhancement 7285] Broker don't check name server host is correct or not

### LINKED ISSUE #7285: [Enhancement] Broker don't check name server host is correct or not

### Summary

RemotingUtil.string2SocketAddress(addr); 

don't check addr is correct.

### Motivation

like this 

The broker[broker-abc, 1x.xxx.xx.xxx:10911] boot success. serializeType=JSON and name server is xxx.yyy.cxxxx.dddd:9876,aaaa.yyy.cxxxx.dddd:9876

the name server address contains ',' which should use ';', but the console says  "The broker[broker-abc, ip:10911] boot success.“

check name server addr. if the address is not correcnt , throw a exception 

don't say The broker[broker-abc, ip:10911] boot success.

print  name serve address is not correct, please fix it 

### Additional Context

_No response_

onejimmyboy:

Fixes #7285

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
@@ -28,6 +28,8 @@ import java.nio.channels.Selector;
 import java.nio.channels.spi.SelectorProvider;
 import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.commons.validator.routines.InetAddressValidator;
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.logging.org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
```diff
@@ -153,6 +154,25 @@ public static SocketAddress string2SocketAddress(final String addr) {
...
+
+    public static boolean isValidIPAddress(String ipAddress) {
+        InetAddressValidator validator = InetAddressValidator.getInstance();
+        if (validator.isValidInet4Address(ipAddress)) {
+            return true;
+        }
+        if (validator.isValidInet6Address(ipAddress)) {
+            return true;
+        }
+        return ipAddress.equals("localhost");
```
joeCarf, line 173: what if users give a hostname like `broker01`, will it work as expected?

Issue #7311, opened 2023-09-06
[Enhancement]  The grpcChannelManager object  should be added to the appendStartAndShutdown method  


CLFutureX:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

window 10

### RocketMQ version

develop(5.1.3)

### JDK Version

JDK 1.8.0_372

### Describe the Bug
**path:**  org.apache.rocketmq.proxy.grpc.v2.DefaultGrpcMessingActivity#init
T The grpcChannelManager object  should be added to the appendStartAndShutdown method ， Because the grpcChannelManager has a thread pool object, it needs to be shutdown.
![1694010786329](https://github.com/apache/rocketmq/assets/44110731/b78ddbab-528a-466c-9f1c-ca61acfe4023)

### Steps to Reproduce

Code level defects

### What Did You Expect to See?

 The grpcChannelManager object  should be added to the appendStartAndShutdown method 

### What Did You See Instead?

T The grpcChannelManager object  was not  added to the appendStartAndShutdown method

PR #7312, opened 2023-09-06
[ISSUE #7311]  fix The object added by this appendStartAndShutdown method is incorrect

CLFutureX:
[issue](https://github.com/apache/rocketmq/issues/7311)

Fixes #issue_id


proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/DefaultGrpcMessingActivity.java
```diff
@@ -87,7 +87,7 @@ protected void init(MessagingProcessor messagingProcessor) {
         this.routeActivity = new RouteActivity(messagingProcessor, grpcClientSettingsManager, grpcChannelManager);
         this.clientActivity = new ClientActivity(messagingProcessor, grpcClientSettingsManager, grpcChannelManager);
 
-        this.appendStartAndShutdown(this.grpcClientSettingsManager);
+        this.appendStartAndShutdown(this.grpcChannelManager);
```
joeCarf, line 90: IMO, there is thread in grpcClientSettingsManager, it also need to be shut down

proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/DefaultGrpcMessingActivity.java
```diff
@@ -87,6 +87,7 @@ protected void init(MessagingProcessor messagingProcessor) {
         this.routeActivity = new RouteActivity(messagingProcessor, grpcClientSettingsManager, grpcChannelManager);
         this.clientActivity = new ClientActivity(messagingProcessor, grpcClientSettingsManager, grpcChannelManager);
 
+        this.appendShutdown(this.grpcChannelManager);
```
xdkxlk, line 90: Although the start method of grpcChannelManager is currently an empty implementation, it would be better to use appendStartAndShutdown.

joeCarf (approved):

PR #7320, opened 2023-09-07
[ISSUE #7319] Optimize fault-tolerant mechanism for sending messages and hot update switch

### LINKED ISSUE #7319: [Enhancement] Optimize fault-tolerant mechanism for sending messages and hot update switch

### Summary

Optimize fault-tolerant mechanism for sending messages and hot update switch

### Motivation

The fault-tolerant mechanism's switch in the proxy needs to be hot updated. So that this function can be turned on in time in case of emergency.

On every update of the faultItemTable, a switch check is performed.

no more.

### Additional Context

_No response_

GenerousMan:

Fixes #7319

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
@@ -38,7 +38,7 @@ public class LatencyFaultToleranceImpl implements LatencyFaultTolerance<String> 
     private int detectInterval = 2000;
     private final ThreadLocalIndex whichItemWorst = new ThreadLocalIndex();
 
-    private boolean startDetectorEnable = false;
+    private volatile boolean startDetectorEnable = false;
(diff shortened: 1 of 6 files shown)


GenerousMan:
@xdkxlk please help review~

drpmma (dismissed):
LGTM

client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
```diff
@@ -37,6 +37,8 @@ public class LatencyFaultToleranceImpl implements LatencyFaultTolerance<String>
     private int detectTimeout = 200;
     private int detectInterval = 2000;
     private final ThreadLocalIndex whichItemWorst = new ThreadLocalIndex();
+
+    private boolean startDetectorEnable = false;
```
xdkxlk, line 41: It is better to use volatile here

xdkxlk (approved):

PR #7327, opened 2023-09-07
[ISSUE #7336] Validators.checkGroup method verified that the group name is null,so …

liushangqing9803:
Fixes https://github.com/apache/rocketmq/issues/7336
Validators.checkGroup method verified that the group name is null,so there is no need for repeated verification.


joeCarf (approved):
LGTM

liushangqing9803:
> LGTM

Hello, does this mean it has already been merged? Do I need to manually close this pull request and issue?

PR #7354, opened 2023-09-13
[ISSUE #7353] Topic creation and deletion commands act on slave nodes

### LINKED ISSUE #7353: [Enhancement] Topic creation and deletion commands act on slave nodes

### Summary

Topic creation and deletion commands do not act on slave nodes

### Motivation

Topic creation and deletion commands act on slave nodes

Topic creation and deletion requests are sent to the slave node at the same time.

none

### Additional Context

_No response_

guyinyou:

Fixes #7353

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -41,6 +41,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.rocketmq.common.ServiceThread;
 import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.TopicFilterType;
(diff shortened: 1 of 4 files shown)


tools/src/main/java/org/apache/rocketmq/tools/command/topic/DeleteTopicSubCommand.java
```diff
@@ -38,7 +38,7 @@ public static void deleteTopic(final DefaultMQAdminExt adminExt,
         final String topic
     ) throws InterruptedException, MQBrokerException, RemotingException, MQClientException {
 
-        Set<String> masterBrokerAddressSet = CommandUtil.fetchMasterAddrByClusterName(adminExt, clusterName);
+        Set<String> masterBrokerAddressSet = CommandUtil.fetchMasterAndSlaveAddrByClusterName(adminExt, clusterName);
```
RongtongJin, line 41: It would be better to rename `masterBrokerAddressSet` to `brokerAddressSet`.

tools/src/main/java/org/apache/rocketmq/tools/command/topic/UpdateTopicSubCommand.java
```diff
@@ -168,7 +168,7 @@ public void execute(final CommandLine commandLine, final Options options,
                 defaultMQAdminExt.start();
 
                 Set<String> masterSet =
-                    CommandUtil.fetchMasterAddrByClusterName(defaultMQAdminExt, clusterName);
+                    CommandUtil.fetchMasterAndSlaveAddrByClusterName(defaultMQAdminExt, clusterName);
```
RongtongJin, line 171: Same as above

guyinyou:
done

RongtongJin (commented):
Need to consider the following: 1. If the topic is deleted from the master first and then from the slave, is it possible that after the synchronization, the topic no longer exists in the slave? 2. If the topic is deleted from the slave first and then from the master, is it possible that due to synchronization, data of the topic remains in the slave?

PR #7359, opened 2023-09-14
[ISSUE #7358] Optimize pull request hold service

### LINKED ISSUE #7358: [Enhancement] Low and discontinuous production traffic may cause delays in message consumption

### Summary

https://github.com/apache/rocketmq/discussions/7021

### Motivation

image
image

As the arthas watch results show,  message arriving notify is ~0.1ms ahead of suspending pull request, result in about 5 seconds consumption delay of this message.

We can do a check after suspend pull request.

Not yet.

### Additional Context

_No response_

redlsz:
Fixes #7358

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
@@ -16,11 +16,14 @@
  */
 package org.apache.rocketmq.broker.longpolling;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.common.ServiceThread;
 import org.apache.rocketmq.common.SystemClock;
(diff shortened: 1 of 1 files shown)


joeCarf (approved):
LGTM

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
```diff
@@ -54,6 +56,11 @@ public void suspendPullRequest(final String topic, final int queueId, final Pull
 
         pullRequest.getRequestCommand().setSuspended(true);
         mpr.addPullRequest(pullRequest);
+        // check the last notify maxOffset
```
joeCarf, line 59: if the pull request been put in the `pullRequestTable`, will it be processed repeated?

redlsz: Won't. The implementation of notifyMessageArriving guarantees that each request will only be processed once even in concurrent scenarios.

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
```diff
@@ -36,6 +36,8 @@ public class PullRequestHoldService extends ServiceThread {
     private final SystemClock systemClock = new SystemClock();
     protected ConcurrentMap<String/* topic@queueId */, ManyPullRequest> pullRequestTable =
         new ConcurrentHashMap<>(1024);
+    protected ConcurrentMap<String/* topic@queueId */, Long/* offset */> lastNotifyMaxOffset =
```
xdkxlk, line 39: When deleting a topic or reducing the number of queues, the memory of lastNotifyMaxOffset may be leaked.

redlsz: > When deleting a topic or reducing the number of queues, the memory of lastNotifyMaxOffset may be leaked.

Indeed. But the leak is very slight? The pullRequestTable has the same issue.

redlsz: > When deleting a topic or reducing the number of queues, the memory of lastNotifyMaxOffset may be leaked.

Resolved. It is more suitable to use expired cache here. @xdkxlk PTAL.

broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
```diff
@@ -122,6 +134,12 @@ public void notifyMessageArriving(final String topic, final int queueId, final l
     public void notifyMessageArriving(final String topic, final int queueId, final long maxOffset, final Long tagsCode,
         long msgStoreTime, byte[] filterBitMap, Map<String, String> properties) {
         String key = this.buildKey(topic, queueId);
+
+        // record the maxOffset if notify from message store
+        if (msgStoreTime > 0) {
+            this.lastNotifyMaxOffsetCache.put(key, maxOffset);
```
drpmma, line 140: Should we examine whether the incoming maxOffset exceeds the value of the cache?

redlsz: > Should we examine whether the incoming maxOffset exceeds the value of the cache?

You mean the cached offset may be larger than the incoming one? The ReputMessageService executes in the order of commitlog offset, so the notifies would arrive in order of queue logic offset.

PR #7382, opened 2023-09-22
[ISSUE #7381] Fix the problem of inaccurate timer message metric

### LINKED ISSUE #7381: [Bug] The metrics of timerMessage are inaccurate and may even be negative.

macos

### RocketMQ version

5.1.3

### JDK Version

_No response_

### Describe the Bug

After the timer message is enqueueed, if the entry time wheel is slow, it will be directly thrown into the dequePutQueue, and the increment operation of metrics will not be performed.
This will cause the metrics of timerMessage to be inaccurate and may even be negative.

### Steps to Reproduce

just start a local broker, and send timerMessage into it.
you need to set the deliver time now + 3~5ms.

metrics should be the num of timer messages in the timing wheel.

### What Did You See Instead?

image

### Additional Context

_No response_

GenerousMan:

Fixes #7381

### Brief Description

When entering the time wheel, if it is judged that it can be dequeued, its enqueue time is set to infinity, which means that it has never entered the time wheel.

Skip messages with an infinite enqueue time when counting metrics.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -41,6 +41,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.rocketmq.common.ServiceThread;
 import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.TopicFilterType;
@@ -600,7 +601,7 @@ public class TimerMessageStore {
                 return;
             }
             if (msg.getProperty(TIMER_ENQUEUE_MS) != null
-                    && Long.parseLong(msg.getProperty(TIMER_ENQUEUE_MS)) == Long.MAX_VALUE) {
+                    && NumberUtils.toLong(msg.getProperty(TIMER_ENQUEUE_MS)) == Long.MAX_VALUE) {
                 return;
             }
             // pass msg into addAndGet, for further more judgement extension.
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -599,6 +599,10 @@ public class TimerMessageStore {
             if (null == msg || null == msg.getProperty(MessageConst.PROPERTY_REAL_TOPIC)) {
                 return;
             }
+            if (msg.getProperty(TIMER_ENQUEUE_MS) != null
+                    && Long.parseLong(msg.getProperty(TIMER_ENQUEUE_MS)) == Long.MAX_VALUE) {
+                return;
+            }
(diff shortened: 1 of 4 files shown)


store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -599,7 +599,12 @@ public void addMetric(MessageExt msg, int value) {
             if (null == msg || null == msg.getProperty(MessageConst.PROPERTY_REAL_TOPIC)) {
                 return;
             }
-            timerMetrics.addAndGet(msg.getProperty(MessageConst.PROPERTY_REAL_TOPIC), value);
+            if (msg.getProperty(TIMER_ENQUEUE_MS) != null
```
joeCarf, line 602: NumberUtils.toLong() might be more clear

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -1452,9 +1458,14 @@ public void run() {
...
                             try {
                                 perfCounterTicks.startTick(DEQUEUE_PUT);
-                                DefaultStoreMetricsManager.incTimerDequeueCount(getRealTopic(tr.getMsg()));
-                                addMetric(tr.getMsg(), -1);
-                                MessageExtBrokerInner msg = convert(tr.getMsg(), tr.getEnqueueTime(), needRoll(tr.getMagic()));
+                                MessageExt msgExt = tr.getMsg();
+                                DefaultStoreMetricsManager.incTimerDequeueCount(getRealTopic(msgExt));
+                                if (tr.getEnqueueTime() == Long.MAX_VALUE) {
+                                    // never enqueue, mark it.
+                                    MessageAccessor.putProperty(msgExt, TIMER_ENQUEUE_MS, Long.MAX_VALUE+"");
```
joeCarf, line 1465: will `String.valueOF(Long.MAX_VALUE)` be more clear?

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
GenerousMan: done

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
GenerousMan: done

RongtongJin (approved):

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
joeCarf: no need of `msg.getProperty(TIMER_ENQUEUE_MS) != null` if using `NumberUtils.toLong`, you can check this out

mxsm (approved):
LGTM

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
GenerousMan: Under normal situation, this indicator is still null when exiting the queue, so this judgment is necessary.
When leaving the queue, the 'enququ_time' attribute in the message has not been retrieved from the timerRequest yet. When leaving the queue, this attribute will not be null unless it is determined that it has never entered the time wheel.

PR #7384, opened 2023-09-23
[ISSUE #7306] Create system topic subscription when autoCreateSubscriptionGroup is false.

### LINKED ISSUE #7306: [Enhancement]Create sys topic subscription when autoCreateSubscriptionGroup false

### Summary

At proxy mode, we use the broadcast topic to manage consumer connection, it should create the sys topic subscription  when the `autoCreateSubscriptionGroup` is false.

### Motivation

In daily usage, we often disable the automatic creation of subscription groups. Consequently, if proxy starts and a subscription group hasn't been created manually,  the consumers connected to different proxy. the broadcast message can't be consumed by proxy, the proxy can't detect the other proxy consumers. it can adversely affect consumer rebalancing. Moreover, this issue is not easy for users to identify, leading to incorrect usage.

After the check and create sys Topic, it also create the appropriate subscription group config.

N/A

### Additional Context

_No response_

echooymxq:

fix #7306 

Fixes #issue_id


joeCarf (commented):
I think unit test of these new functions might necessary :)

echooymxq:
@drpmma Help review it. 🙏

PR #7405, opened 2023-09-27
 Enhancement MQClientInstance #7402

### LINKED ISSUE #7402: [Enhancement] Enhancement MQClientInstance

### Summary

MQClientInstance#start

image

DefaultMQProducer 这个内部的生产者客户端只要在消费者客户端创建的时候才需要这个； **CLIENT_INNER_PRODUCER**
它的作用是在消费失败需要重试的时候，可以利用这个Producer将消息发送回去。

但是在创建生产者客户端的时候它并不需要这么一个内部的生产者客户端; 它没没有做任何事情，而且还占用了资源

---
同理PullMessageService也是一样，在创建生产者客户端的时候它不需要使用到这个服务。

DefaultMQProducer This internal producer client only needs this when the consumer client is created; **CLIENT_INNER_PRODUCER**
Its function is to use this Producer to send the message back when consumption fails and needs to be retried.

But when you create a producer client, it doesn't need such an internal producer client; it doesn't do anything, and it also takes up resources.

---
The same goes for PullMessageService. It does not need to use this service when creating a producer client.

### Motivation

1. 建议根据条件判断是否启用这些服务
2. 现在是不管什么客户端创建之后都会启动这些服务，连DashBoard这种毫无相关性的MQAdminExt客户端被创建时都启动了这些服务
3. MQClientInstance 秩序作为一个抽象的客户端实例就行了，只启动远程通信客户端能力即可。
4. 这些莫名其妙创建的客户端会带来一些困扰，比如`producer`命令查询所有生产者客户端的时候 列举了非常多的客户端

下图展示的是我只启动了 Dashboard而已，居然启动了3个生产者客户端,并且每个ClientId都不一样(要知道不同的ClientId都需要创建属于自己的网络通道)，无故浪费了资源
image

---

1. It is recommended to determine whether to enable these services based on conditions
2. Now no matter what client is created, these services will be started. Even the unrelated MQAdminExt client like DashBoard starts these services when it is created.
3. MQClientInstance can be used as an abstract client instance, and only the remote communication client capability can be enabled.
4. These inexplicably created clients will cause some trouble. For example, when the `producer` command queries all producer clients, a lot of clients are listed.

同时. 可以大幅度减少没有必要的心跳请求，例如 MQAdminExt客户端就不需要发起心跳(目前情况因为内置的Producer的创建导致会发送心跳请求)
At the same time, unnecessary heartbeat requests can be greatly reduced. For example, the MQAdminExt client does not need to initiate heartbeats (currently, heartbeat requests will be sent due to the creation of the built-in Producer).

image

.

### Additional Context

_No response_

shirenchuang:

Fixes #7402 

通过判断consumerTable的数量判断是否有消费者存在

1. 在初次启动的时候，如果是注册的消费者，则直接将那几个服务启动，否则不启动
2. 如果初次启动不是消费者，但是后续有加入新的消费者，则判断那几个服务是否已经启动，如果没有则启动

Determine whether there are consumers by judging the number of consumerTables

1. At the initial startup, if it is a registered consumer, directly start those services, otherwise it will not start.
2. If the initial startup is not a consumer, but new consumers are added later, determine whether those services have been started, and if not, start them

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -257,6 +257,7 @@ public class MQClientInstance {
         return mqList;
     }
 
+    @SuppressWarnings("checkstyle:UnnecessaryParentheses")
     public void start() throws MQClientException {
 
         synchronized (this) {
@@ -276,7 +277,7 @@ public class MQClientInstance {
                     if (!consumerTable.isEmpty()
                             || clientConfig instanceof LitePullConsumer
                             || clientConfig instanceof MQPushConsumer) {
-                        startConsumerService();
+                        tryStartConsumerService();
                     }
 
                     log.info("the client factory [{}] start OK", this.clientId);
@@ -284,8 +285,8 @@ public class MQClientInstance {
                     break;
                 case RUNNING:
                     // If a Consumer is added later, make sure to start the Consumer-related services
-                    if (!consumerTable.isEmpty() && !pullMessageService.isStopped()) {
-                        startConsumerService();
+                    if (!consumerTable.isEmpty()) {
+                        tryStartConsumerService();
                     }
                     break;
                 case START_FAILED:
@@ -296,7 +297,7 @@ public class MQClientInstance {
         }
     }
 
-    private void startConsumerService() throws MQClientException {
+    private void tryStartConsumerService() throws MQClientException {
         // Start pull service
         this.pullMessageService.start();
         // Start rebalance service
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -286,6 +296,17 @@ public void start() throws MQClientException {
         }
     }
 
+    private void start4ConsumerService() throws MQClientException {
```
joeCarf, line 299: I think naming `startConsumerService` will be enough

joeCarf (approved):
LGTM

PR #7408, opened 2023-09-27
Optimize ConsumeMessageOrderlyService enumerated types of comparison

fqgsoftwareengineer:
Optimized performance

Optimized performance：
The equals of enumeration types can be replaced by ==, which improves performance and is safer at compile and run time.
see:
1. https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.9
2. https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals
Just compare the pressure measurements with the previous ones

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
@@ -86,7 +86,7 @@ public class ConsumeMessageOrderlyService implements ConsumeMessageService {
     }
 
     public void start() {
-        if (MessageModel.CLUSTERING==ConsumeMessageOrderlyService.this.defaultMQPushConsumerImpl.messageModel()) {
+        if (MessageModel.CLUSTERING == ConsumeMessageOrderlyService.this.defaultMQPushConsumerImpl.messageModel()) {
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
```diff
@@ -86,7 +86,7 @@ public ConsumeMessageOrderlyService(DefaultMQPushConsumerImpl defaultMQPushConsu
     }
 
     public void start() {
-        if (MessageModel.CLUSTERING.equals(ConsumeMessageOrderlyService.this.defaultMQPushConsumerImpl.messageModel())) {
+        if (MessageModel.CLUSTERING==ConsumeMessageOrderlyService.this.defaultMQPushConsumerImpl.messageModel()) {
```
joeCarf, line 89: need blank space around the equal operator

joeCarf: for good code style

fqgsoftwareengineer: Space formatting has been added

Issue #7100, opened 2023-07-31
[Enhancement] Need to show the confirmation message when the user creates topic of the same name


AayushSaini101:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Currently, We can able to create the same topic by having a similar name again and again without any kind of confirmation message. This will cause problems and lost of the configuration of the older topic of the RocketMQ. 
![image](https://github.com/apache/rocketmq/assets/60972989/655f78ce-685f-4a62-a0e8-ae4dfd050613)
loss

### Motivation

It will cause a lot of issues to suppose by mistake new topic is created of the same name, the older topic configuration will be lost.

### Describe the Solution You'd Like

We can provide a confirmation message whenever the user tries to create a topic of the same name.

### Describe Alternatives You've Considered

Don't allow the user to create the topic of the same name

AayushSaini101:
@xiaoyifang @jimjag Can you please check this issue? if this is valid to work I can start working on it

francisoliverlee:
api should have a certain response, you are welcome to work it out.

AayushSaini101:
Thank you @francisoliverlee

AayushSaini101:
Hi, @francisoliverlee How can I debug the changes? Can you please tell

PR #7416, opened 2023-09-30
Added condition to check topic is present or not

AayushSaini101:

Resolve [Issue](https://github.com/apache/rocketmq/issues/7100)

Fixes #issue_id

### Brief Description

Added a condition to check topic is present or not in the cluster

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -381,6 +381,11 @@ public class MQClientAPIImpl implements NameServerUpdateCallback {
         throws RemotingException, MQBrokerException, InterruptedException, MQClientException {
         Validators.checkTopicConfig(topicConfig);
 
+        if(getTopicListFromNameServer(timeoutMillis).getTopicList().contains(topicConfig.getTopicName())){
+
+             throw new MQClientException(405,"Topic is already present");
+        }
+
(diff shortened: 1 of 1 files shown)


joeCarf (commented):
you may start an issue and describe this change

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -381,6 +381,11 @@ public void createTopic(final String addr, final String defaultTopic, final Topi
         throws RemotingException, MQBrokerException, InterruptedException, MQClientException {
         Validators.checkTopicConfig(topicConfig);
 
+        if (getTopicListFromNameServer(timeoutMillis).getTopicList().contains(topicConfig.getTopicName())) {
+
```
joeCarf, line 385: redundant line

joeCarf (commented):
you may change the title like `[ISSUE #xxxx]xxxxx`

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
AayushSaini101: Thank you @joeCarf I will change.  Actually, i want to fetch all the topics present in the cluster but unable to find out. I checked this approach but this is not working, and causing a test case to fail. When i debug the test case. I found the getTopicList() method returning null response.

PR #7439, opened 2023-10-10
add getter for class Message ,fix json serialize bug

LetLetMe:
These getter and setter methods indeed cause the JSON serialization of the Message subclasses to fail.


common/src/main/java/org/apache/rocketmq/common/message/Message.java
```diff
@@ -218,14 +218,36 @@ public String toString() {
...
         this.putProperty(MessageConst.PROPERTY_TIMER_DELAY_MS, String.valueOf(timeMs));
     }
+
+    public long getDelayTimeMs() {
+        String t = this.getProperty(MessageConst.PROPERTY_TIMER_DELAY_MS);
+        if (t != null) {
+            return Long.parseLong(t);
+        }
+        return 0;
+    }
```
drpmma, line 241: format

drpmma (approved):
LGTM

LetLetMe:
#7440

PR #7449, opened 2023-10-11
Optimize StatsItem

kang-hl:

Fixes #issue_id


common/src/main/java/org/apache/rocketmq/common/stats/StatsItem.java
```diff
@@ -248,11 +224,23 @@ public long getTimestamp() {
         return timestamp;
     }
 
+    public void setTimestamp(long timestamp) {
```
joeCarf, line 227: why need set function?

kang-hl: 复用CallSnapshot对象，需要修改其属性。

PR #7452, opened 2023-10-12
[ISSUE #7451] Override toString for TopicConfigAndQueueMapping

### LINKED ISSUE #7451: [Enhancement] Override toString for TopicConfigAndQueueMapping

### Summary

TopicConfigAndQueueMapping extends from the TopicConfig class but does not override the toString method. This results in situations where they appear equal when printed, but in reality, the equals method returns false. T his can be highly confusing for anyone observing this behavior.

### Motivation

Avoid unnecessary misunderstandings

as mentioned above

as mentioned above

### Additional Context

_No response_

LetLetMe:
TopicConfigAndQueueMapping extends from the TopicConfig class but does not override the toString method. This results in situations where they appear equal when printed, but in reality, the equals method returns false. T his can be highly confusing for anyone observing this behavior.

close  #7451


drpmma (approved):
LGTM

joeCarf (approved):

RongtongJin (approved):

PR #7468, opened 2023-10-16
[ISSUE #7437] Add the CRC check of commitlog 

### LINKED ISSUE #7437: [Bug] A bug in the CRC check of commitlog

Linux

### RocketMQ version

5.1.5

### JDK Version

JDK11

### Describe the Bug

A shutdown occurs when the message is not completely written, causing all characters at the end of the message to be unwritten.
image

recover uses magicCode and bodyCrc to verify whether the writing is successful, but properties do not participate in the verification and belong to the end of the entire message.

Therefore, it cannot be verified that the data of this message is wrong. If you continue to use this error message, unpredictable consequences will occur, such as blocking timerMessage distribution.

### Steps to Reproduce

When there is an abnormal downtime, the body of the last message is written successfully, but the properties are not written successfully.

During recover, data abnormalities in any part of the entire message can be verified.

### What Did You See Instead?

Exceptions other than the body field cannot be found during recover.

### Additional Context

_No response_

guyinyou:

Fixes #7437


store/src/main/java/org/apache/rocketmq/store/CommitLog.java
```diff
@@ -61,7 +61,6 @@
 import org.apache.rocketmq.store.logfile.MappedFile;
 import org.apache.rocketmq.store.util.LibC;
 import org.rocksdb.RocksDBException;
-
```
joeCarf, line 64: maybe no need to reformat the import in this file and `MessageConfig.java`

RongtongJin (commented):
Add some Uts

store/src/main/java/org/apache/rocketmq/store/CommitLog.java
guyinyou: done

guyinyou:
> Add some UTs

done

joeCarf (approved):
lgtm

RongtongJin (approved):

guyinyou (commented):
lgtm

PR #7471, opened 2023-10-17
Optimize ConsumeMessageOrderlyService enumerated types of comparison

fqgsoftwareengineer:
Optimized performance

Optimized performance：
The equals of enumeration types can be replaced by ==, which improves performance and is safer at compile and run time.
see:
1. https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.9
2. https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals
Just compare the pressure measurements with the previous ones。
Space formatting has been added


joeCarf (approved):
LGTM

PR #7498, opened 2023-10-24
[ISSUE #7497] Extract the frequency of calling updateNamesrvAddr into a configuration

### LINKED ISSUE #7497: [Enhancement] Extract the frequency of calling updateNamesrvAddr into a configuration

### Summary

Extract the frequency of calling updateNamesrvAddr into a configuration

### Motivation

Currently, the NameServerAddressList is updated every 2 minutes by default, which means that when there is a change in the nameserver address, the broker cannot quickly perceive it.

Extract the frequency of calling updateNamesrvAddr into a configuration

None

### Additional Context

_No response_

xdkxlk:

Fixes #7497


fuyou001 (approved):
LGTM

PR #7436, opened 2023-10-10
Fix tiered store README.md error about Configuration

golden-yang:
### Which Issue(s) This PR Fixes
ISSUE #5923 add README.md file to explain how to use tiered storage.  

There was a simple configuration error

Issue #5923, opened 2023-01-24
[RIP-57] Add example tiered storage backend service provider


ShadowySpirits:
link to #5754 

As stated in the title, this issue aims to add a POSIX provider for tiered storage. It transfers message data from the default store directory to the mount point of storage medium for tiered storage. This provider is fully functional but for demonstration purposes only.

lizhimins:
There is a widely use of filePath in the rocketmq project. The rename operation which not only brings compatibility issues but also conflicts with the naming style of the rocketmq project. It is recommended to revert this change.

PR #7557, opened 2023-11-14
[ISSUE #5923] Revert "Fix tiered store README.md error about Configuration (#7436)"

lizhimins:
[ISSUE #5923] Revert "Fix tiered store README.md error about Configuration (#7436)"

This reverts commit 70dc93abbcb9bf161378d66fcaca55bedc78b905.

for issue #5923

There is a widely use of filePath in the rocketmq project. The rename operation which not only brings compatibility issues but also conflicts with the naming style of the rocketmq project. It is recommended to revert this change.


lollipopjin (approved):
LGTM， if we accept "filepath" as the default naming pattern, we should replace all the places in RocketMQ, not just tiered storage module.

lollipopjin (approved):

RongtongJin (approved):

PR #7598, opened 2023-12-01
using Maps.newHashMapWithExpectedSize method

shenjianeng:
### Brief Description

Creates a HashMap instance, with a high enough "initial capacity" that it should hold expectedSize elements without growth


shenjianeng:
Is anyone there?

PR #7603, opened 2023-12-01
[ISSUE #7601] Fix slave acting master bug

### LINKED ISSUE #7601: [Bug] Slave thrown NPE and consumer receive duplicate messages when enable slaveActingMaster and remoteEscape

OS: CentOS 6.9

### RocketMQ version

branch: (develop|tag 5.1.4) version: 5.1.4

### JDK Version

JDK: 1.8.0_202

### Describe the Bug

When I enable slaveActingMaster and remoteEscape，send some messages to master，then close it，slave throw exceptions as follows:
```
2023-11-28 17:56:25 INFO TimerDequeuePutMessageService - Unknown error
java.lang.NullPointerException: null
        at org.apache.rocketmq.store.timer.TimerMessageStore.doPut(TimerMessageStore.java:1088)
        at org.apache.rocketmq.store.timer.TimerMessageStore.access$1900(TimerMessageStore.java:71)
        at org.apache.rocketmq.store.timer.TimerMessageStore$TimerDequeuePutMessageService.run(TimerMessageStore.java:1487)
        at java.lang.Thread.run(Thread.java:748)
2023-11-28 17:56:25 INFO TimerDequeuePutMessageService - Unknown error
java.lang.NumberFormatException: null
        at java.lang.Integer.parseInt(Integer.java:542)
        at java.lang.Integer.parseInt(Integer.java:615)
        at org.apache.rocketmq.store.timer.TimerMessageStore.convertMessage(TimerMessageStore.java:1142)
        at org.apache.rocketmq.store.timer.TimerMessageStore.convert(TimerMessageStore.java:1059)
        at org.apache.rocketmq.store.timer.TimerMessageStore.access$1800(TimerMessageStore.java:71)
        at org.apache.rocketmq.store.timer.TimerMessageStore$TimerDequeuePutMessageService.run(TimerMessageStore.java:1486)
        at java.lang.Thread.run(Thread.java:748)
```
I fixed the exception，and repeat previous steps，after I restarted master, the consumer received **repeated messages**。

### Steps to Reproduce

**The first problem reproduce steps:**
1. Enable such configs as follows:
   1. NameServer:
          supportActingMaster=true
   2. Broker:
          enableSlaveActingMaster=true
          enableRemoteEscape=true
          totalReplicas=2
2. Deploy one NameServer and tow group of brokers, such as:
   1. master: broker-a，slave: broker-a-s
   2. master: broker-b，slave: broker-b-s
3. Run consumer and send some timer messages，such as:
  ```message.setDeliverTimeMs(System.currentTimeMillis() + 5 * 60_000L);```
4. Close the master, slave will throw NPE and NumberFormatException when timer message escape。

**The second problem reproduce steps:**

1. After I fix the exception and repeat previous steps。
2. Restart master, the consumer will receive **repeated messages**。
For example, the producer sent 10 messages and received 10 messages before the master restarted. After the master restarted, it repeatedly received the previously received messages.

1. Slave not throw exception。
2. Consumer not revieve repeated messages.

### What Did You See Instead?

1. Slave throw exception。
2. Consumer revieve repeated messages.

### Additional Context

_No response_

gaoyf:

Fixes #7601

### Brief Description

Fix #7601's exception and master use slave's TimerCheckPoint。

Use #7601's reproduce step to test.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -1142,8 +1142,8 @@ public class TimerMessageStore {
         } else {
             msgInner.setTopic(msgInner.getProperty(MessageConst.PROPERTY_REAL_TOPIC));
             msgInner.setQueueId(Integer.parseInt(msgInner.getProperty(MessageConst.PROPERTY_REAL_QUEUE_ID)));
-            MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_TOPIC);
-            MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_QUEUE_ID);
+            // MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_TOPIC);
+            // MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_QUEUE_ID);
         }
         return msgInner;
     }
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -602,6 +602,10 @@ public class TimerMessageStore {
         this.shouldRunningDequeue = shouldRunningDequeue;
     }
 
+    public boolean isShouldRunningDequeue() {
+        return shouldRunningDequeue;
+    }
+
     public void addMetric(MessageExt msg, int value) {
         try {
             if (null == msg || null == msg.getProperty(MessageConst.PROPERTY_REAL_TOPIC)) {
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -1084,8 +1084,10 @@ public class TimerMessageStore {
                     case PUT_OK:
                         if (brokerStatsManager != null) {
                             this.brokerStatsManager.incTopicPutNums(message.getTopic(), 1, 1);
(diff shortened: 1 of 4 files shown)


store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -1140,8 +1146,8 @@ public MessageExtBrokerInner convertMessage(MessageExt msgExt, boolean needRoll)
         } else {
             msgInner.setTopic(msgInner.getProperty(MessageConst.PROPERTY_REAL_TOPIC));
             msgInner.setQueueId(Integer.parseInt(msgInner.getProperty(MessageConst.PROPERTY_REAL_QUEUE_ID)));
-            MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_TOPIC);
-            MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_QUEUE_ID);
+            // MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_REAL_TOPIC);
```
joeCarf, line 1149: could you explain why comment these lines and how the second bug occurs?

joeCarf (approved):

store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
RongtongJin: Alternatively, we should add a messageAccessior.deepCopyProperties instead of using the same reference.

gaoyf: > Alternatively, we should add a messageAccessior.deepCopyProperties instead of using the same reference.

Your idea looks better, although I don't think commenting out this code will have any impact，so I modified the code as you suggested.

RongtongJin (approved):

PR #7654, opened 2023-12-13
[ISSUE #7585] Always return duplicate buffer when filter message and fix log format

### LINKED ISSUE #7585: [Feature] Support message filtering in rocketmq tiered storage

In the current implementation of rocketmq tiered storage, message filtering is not supported.
Messages that do not meet the requirements being returned to the upper processor.

image

Support message filtering in rocketmq tiered storage

Messages that do not meet the conditions should be filtered out

### Additional Context

This filtering capability should support both tag and SQL filtering mode.

lizhimins:
…format

Fixes #7585

When pull messages from the cache of tiered storage, always return a copy of the byte buffer instead of the buffer itself.

从分级存储的缓存获取消息时，将返回消息 buffer 的副本而不是 buffer 本身。


lollipopjin (approved):
LGTM

PR #7690, opened 2023-12-20
[ISSUE #7689] In Controller mode, messages may lost due to sharing the same cq offset

### LINKED ISSUE #7689: [Bug] In Controller mode, messages may be lost due to sharing the same cq offset

During my testing of the controller mode, I encountered a scenario that could lead to message loss:

1. The leader controller and master broker are killed simultaneously.
2. Both are restarted at the same time, with the master broker completing recovery and immediately having messages written to it.

### RocketMQ version

5.1.4

### JDK Version

_No response_

### Describe the Bug

Since the master only recovers up to the position of the confirmOffset upon recovery, there may still be some messages that have not been synchronized by the slave. As a result, the data in the cqOffsetTable might be inaccurate. At this time, newly written messages could share the CQ offset with messages after the confirmOffset—this could lead to message loss.

### Steps to Reproduce

using openchaos, major-kill.
Killing brokers and controllers in the same time.

no message lost.

### What Did You See Instead?

message lost.

### Additional Context

_No response_

GenerousMan:

Fixes #7689 

Upon restart, it is necessary to compensate for messages after the confirmOffset by moving the cq offset backwards. It must be ensured that these messages remain unconsumable before the confirmOffset is reached.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/queue/ConsumeQueueStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/queue/ConsumeQueueStore.java
@@ -455,7 +455,7 @@ public class ConsumeQueueStore extends AbstractConsumeQueueStore {
         }
 
         // Correct unSubmit consumeOffset
-        if (messageStoreConfig.isDuplicationEnable()) {
+        if (messageStoreConfig.isDuplicationEnable() || messageStore.getBrokerConfig().isEnableControllerMode()) {
             SelectMappedBufferResult lastBuffer = null;
             long startReadOffset = messageStore.getCommitLog().getConfirmOffset() == -1 ? 0 : messageStore.getCommitLog().getConfirmOffset();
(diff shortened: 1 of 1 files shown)


RongtongJin (approved):

RongtongJin:
LGTM～

store/src/main/java/org/apache/rocketmq/store/queue/ConsumeQueueStore.java
```diff
@@ -444,6 +444,7 @@ public void recoverOffsetTable(long minPhyOffset) {
                 String key = logic.getTopic() + "-" + logic.getQueueId();
 
                 long maxOffsetInQueue = logic.getMaxOffsetInQueue();
+                log.info("Recover offset table, key: {}, maxOffsetInQueue:{}", key, maxOffsetInQueue);
```
RongtongJin, line 447: 这个日志去掉吧，万一队列太多了打印很多，而且也不是最终的位点

RongtongJin (approved):

PR #7694, opened 2023-12-23
[ISSUE #7757] Use `CompositeByteBuf` to prevent memory copy.

dao-jun:
### Brief Description

Use `CompositeByteBuf` instead of `DirectByteBuf` to prevent memory copy.

Existing tests could cover the case.


RongtongJin (dismissed):
It would be better to provide some performance comparison data.

RongtongJin (approved):

RongtongJin (approved):

PR #7732, opened 2024-01-08
[Issue #7730] Add a RocketMQ BOM artifact

### LINKED ISSUE #7730: [Enhancement] Create a `rocketmq-bom` artifact

### Summary

Since RocketMQ provides many different artifacts, as Bill-of-Material artifact would be useful to manage the version of all of them at once (cf. [Maven BOM](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms)).

### Motivation

Projects that use RocketMQ like EventMesh need dozens of line to specify the version of RocketMQ artifacts they use:

https://github.com/apache/eventmesh/blob/1ad5fdf573666516ab9ef279db3b166caa980830/eventmesh-connectors/eventmesh-connector-rocketmq/build.gradle#L18-L30

These could be easily condensed to a single line if RocketMQ had a BOM file.

A nice pattern to introduce BOMs in Maven projects is described in [Garret Wilson's blog](https://www.garretwilson.com/blog/2023/06/14/improve-maven-bom-pattern).

The BOM can be introduced in the POM hierachy between the ASF parent POM and `rocketmq-all`: e.g. the hierarchy of `rocketmq-common` will be: `apache:apache` -> `rocketmq-bom` -> `rocketmq-all` -> `rocketmq-common`.

Since Maven's module inheritance (the `<parent>` POM tag) and aggregation (the `<modules>` POM tag) are separate concepts, the directory structure does not have to change and can be:

```
pom.xml        <- contains `rocketmq-bom`
parent/pom.xml <- contains `rocketmq-all`
common/pom.xml <- contains `rocketmq-common`
...
```

The `rocketmq-all` artifact is a **poor** replacement for a BOM artifact, since it manages the version of a lot of artifacts that are not part of Apache RocketMQ.

### Additional Context

_No response_

ppkarwasz:

Fixes #7730.

### Brief Description

This PR implements the suggestion in [Garret Wilson's blog](https://www.garretwilson.com/blog/2023/06/14/improve-maven-bom-pattern) on how to structure repositories containing a BOM artifact:

 * the `rocketmq-all` POM file is moved to `parent/pom.xml` and the `relativePath` attributes of all the other artifacts are updated,
 * a new `rocketmq-bom` artifact is created as parent POM for `rocketmq-all`. It contains:
     - all the common metadata about the project (description, URL, mailing lists, developer list, license and issue management),
     - it manages the version of all `rocketmq-*` artifacts in **this** repo and also `rocketmq-proto`,
 * since the URL inheritance model described in [Maven Model Builder](https://maven.apache.org/ref/3.9.6/maven-model-builder/) doesn't make sense for RocketMQ (e.g. the URL of each module is `https://rocketmq.org` not `https://rocketmq.org/rocketmq-all/rocketmq-common`), all the `child.*.inherit.append.path` options have been set to `false`.

The change was tested with a `mvn verify` step and also by running:

```
mvn help:effective-pom
```

to check that the metadata of each artifact is correct.


ppkarwasz:
I still believe that having a **`rocketmq-bom`** artifact could provide real value for users by simplifying dependency management and ensuring consistent versions across RocketMQ modules.

There are many possible ways to introduce a BOM, but this implementation has the advantage that all other `rocketmq-*` artifacts directly **consume the BOM**. If you think this approach would be useful for the project, I’m happy to rebase this PR onto the current codebase and address any feedback.

If the BOM concept doesn’t align with the current RocketMQ goals or priorities, feel free to close this PR.

PR #7838, opened 2024-02-20
Fix start() and shutdown() of DefaultMessagingProcessor

### LINKED ISSUE #7837: [Bug] Receipt handle auto-renew scheduled task is not started

all

### RocketMQ version

develop

### JDK Version

_No response_

### Describe the Bug

The receipt handle auto-renew scheduled task is not started, resulting in duplicate consumption after handle expired.

image

### Steps to Reproduce

/

/

### What Did You See Instead?

/

### Additional Context

_No response_

redlsz:

Fixes #7837 

### Brief Description

1. Complete ReceiptHandleProcessor's start() and shutdown()
2. Complete DefaultMessagingProcessor‘s start() and shutdown()

UT

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/DefaultMessagingProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/DefaultMessagingProcessor.java
@@ -124,6 +124,12 @@ public class DefaultMessagingProcessor extends AbstractStartAndShutdown implemen
 
     protected void init() {
         this.appendStartAndShutdown(this.serviceManager);
+        this.appendStartAndShutdown(producerProcessor);
+        this.appendStartAndShutdown(consumerProcessor);
+        this.appendStartAndShutdown(transactionProcessor);
+        this.appendStartAndShutdown(clientProcessor);
+        this.appendStartAndShutdown(requestBrokerProcessor);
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/processor/DefaultMessagingProcessor.java
```diff
@@ -124,6 +124,12 @@ public static DefaultMessagingProcessor createForClusterMode(RPCHook rpcHook) {
 
     protected void init() {
         this.appendStartAndShutdown(this.serviceManager);
+        this.appendStartAndShutdown(producerProcessor);
```
drpmma, line 127: It seems that it's not necessary to start processors because this issue is only related to receiptHandleManager.

redlsz: > It seems that it's not necessary to start processors because this issue is only related to receiptHandleManager.

Indeed, this issue is only related to receiptHandleProcessor. But all these processors extend AbstractStartAndShutdown and there is no caller, which may be error-prone over time?

leizhiyuan:
any update?

yuz10 (approved):

PR #7885, opened 2024-03-05
Inconsistent subscriptions

xzwb:
Inconsistent subscriptions can lead to rebalancing，so why dont Solve this problem when subscribing


drpmma:
It is recommended to discuss changes in an issue before submitting a pull request, particularly when modifying existing behavior.

PR #7901, opened 2024-03-12
[ISSUE #7790] Master broker in Dledger mode didn't sync consume offset in time after reboot, which caused consume repeatedly

### LINKED ISSUE #7790: [Bug] Master broker in Dledger mode didn't sync consume offset in time after reboot, which caused consume repeatedly

CentOS Linux release 7.3.1611

### RocketMQ version

4.8.0

### JDK Version

1.8

### Describe the Bug

In dledger mode, master broker which was shut down and reboot would not synchornize consume offset with the previous temporary master broker. The problem described may caused repeatedly consumption.
After reading code, one bug may exist in the synchornize logic:
BrokerController.java: 
image
The oneway flag is set true, which means that following code in BrokerOuterAPI@registerBrokerAll will return null without waiting for the result: 
image
which caused the following marked code didn't run actually:
image
Masteraddr which is null leads to a result that SlaveSynchronize@syncAll didnt't run either, including function body of SlaveSynchronize@syncConsumerOffset: 
image
The Synchronization task was set to run every 10 seconds , but before next sync happens, the reboot broker was elected to be master again already, which means it stopped to try to get the newest consume offset, but start to push messsages from the offset before.
未命名文件

### Steps to Reproduce

* Run a cluster in dledger mode.
* Shut down master broker gracefully.
* Reboot previous master broker, be sure that consume offset increases during downtime.
* Repeat consumption should be observed.

Rebooted broker should synchornize the newest consume offset and consume from it correctly.

### What Did You See Instead?

Abnormal repeated consumption.

### Additional Context

_No response_

bxfjb:

Fixes #7790

### Brief Description

1. Unset oneway flag in brokerController.registerBrokerAll() parameter list
2. Explicitly call syncAll() after registerBrokerAll() to synchronize metadata immediately

Run DLedgerRoleChangeHandlerTest

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/dledger/DLedgerRoleChangeHandler.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/dledger/DLedgerRoleChangeHandler.java
@@ -146,10 +146,11 @@ public class DLedgerRoleChangeHandler implements DLedgerLeaderElector.RoleChange
         handleSlaveSynchronize(BrokerRole.SLAVE);
 
         try {
-            this.brokerController.registerBrokerAll(true, true, this.brokerController.getBrokerConfig().isForceRegister());
+            this.brokerController.registerBrokerAll(true, false, this.brokerController.getBrokerConfig().isForceRegister());
         } catch (Throwable ignored) {
 
         }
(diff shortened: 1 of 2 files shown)


cserwen (approved):
LGTM

PR #7934, opened 2024-03-17
Fix some ambiguous logs

caizhaomin1:

I only modified the log printed, the modification isn't complicated, so I didn't submit an issue.

### Brief Description

I was reading the code about namesrv and broker, and i found some ambiguous logs, this pr is to fix them.

Nonthing logical is modified, i think there is no need for a test.


RongtongJin:
@caizhaomin1 It would be better to submit a relevant issue.

RongtongJin (approved):
LGTM

Issue #7979, opened 2024-03-28
[Bug] Timing message metric will be inaccurate due to incorrect deletion behavior


GenerousMan:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

Ubuntu

### RocketMQ version

5.2.0

### Describe the Bug

When a timing message is deleted, it is implemented by sending a delete message. Since the timerlog of the timing message is actually an index, it cannot be verified in real time whether the deletion is valid, so -1 will be directly set to the metric when received.
This will lead to a problem: if the deletion behavior is invalid, the -1 operation will not be compensated, and may even cause the metric to become a negative number - which is unacceptable.

### Steps to Reproduce

according to the description of the bug.

### What Did You Expect to See?

the abnormal -1 should be compensated.

### What Did You See Instead?

-1 will always be there.

3424672656:
I am interested in this problem, please assign it to me

3424672656:
[https://github.com/apache/rocketmq/pull/7980](url)

PR #7980, opened 2024-03-28
[ISSUE #7979]Fix timerWheel message metric

3424672656:
In normal cases, the number of messages of the deletion type is equal to the number of messages that should be deleted. When out of queue, add one for the deletion type and subtract one for the message that should be deleted. This ensures that the metric is accurate when deletion fails


RongtongJin (approved):

RongtongJin:
@GenerousMan PTAL～

GenerousMan (approved):
LGTM

PR #7987, opened 2024-03-29
fix fail test

3424672656:
fix tail test


RongtongJin (commented):
I found that the original test could also pass on my computer. Is the test unstable？

3424672656:
Yes, this problem occurred in pr 7980

PR #8042, opened 2024-04-20
[ISSUE #8041] Add error log when build index fails

### LINKED ISSUE #8041: [Enhancement] Message index update failed, but no error log

### Summary

Message index update failed, but no error log. I would like to add the error log to to help locate the problem

### Motivation

When deploying my MQ cluster, due to myself problems,  commitlog path will change every time when  MQ is restarted, so there will be inconsistency between indexFile and commitlog. After the inconsistency, I found that produce and consume can proceed normally, but new produce messages cannot update the index file. I could not find the build index error log . So I parsed the index file and found the problem after reading the source code.
I would like to add the error log to to help locate the problem.
image

I'll add the error log. 

.

### Additional Context

_No response_

biningo:

Fixes #8041

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
@@ -215,6 +215,8 @@ public class IndexService {
             String topic = msg.getTopic();
             String keys = msg.getKeys();
             if (msg.getCommitLogOffset() < endPhyOffset) {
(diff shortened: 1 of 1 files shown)


humkum (approved):

mxsm (commented):
@biningo  It might be better to print the CommitLogOffset and endPhyOffset to the log.

biningo:
> @biningo It might be better to print the CommitLogOffset and endPhyOffset to the log.

I agree with you. I have added.
Thank you for your suggestion!

PR #5845, opened 2023-01-10
[ISSUE #5838] retry to send when broker returns SYSTEM_BUSY

cserwen:
**Make sure set the target branch to `develop`**

## What is the purpose of the change

fix #5838 

## Brief changelog

XX

## Verifying this change

XXXX

Follow this checklist to help us incorporate your contribution quickly and easily. Notice, `it would be helpful if you could finish the following 5 checklist(the last one is not necessary)before request the community to review your PR`.

- [x] Make sure there is a [Github issue](https://github.com/apache/rocketmq/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test(over 80% coverage) to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/rocketmq/tree/master/test).
- [x] Run `mvn -B clean apache-rat:check findbugs:findbugs checkstyle:checkstyle` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

PR #8054, opened 2024-04-23
[ISSUE #8053] Return SYSTEM_BUSY if PutMessageStatus is OS_PAGE_CACHE_BUSY

### LINKED ISSUE #8053: [Bug] Producer should be retry put message if OS_PAGE_CACHE_BUSY

Ubuntu 20.04

### RocketMQ version

develop

### JDK Version

JDK 8

### Describe the Bug

If the broker is overloaded or disk IO is busy, it will clean up some expired requests and return with SYSTEM_BUSY, at which point the producer will not retry to put the message  into the broker. So I think PutMessageStatus is OS_PAGE_CACHE_BUSY should also return SYSTEM_BUSY.

reference: https://github.com/apache/rocketmq/blob/develop/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java#L432

### Steps to Reproduce

1. mock PutMessageStatus=OS_PAGE_CACHE_BUSY
2. put the message into broker
3. observe the number of produce retries

The producer does not retry putting the message when the PutMessageStatus is OS_PAGE_CACHE_BUSY.

### What Did You See Instead?

The producer retries putting the message when the PutMessageStatus is OS_PAGE_CACHE_BUSY.

### Additional Context

_No response_

biningo:

Fixes #8053

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/SendMessageProcessor.java
@@ -430,7 +430,7 @@ public class SendMessageProcessor extends AbstractSendMessageProcessor implement
                         "the broker's disk is full [" + diskUtil() + "], messages are put to the slave, message store has been shut down, etc.");
                 break;
             case OS_PAGE_CACHE_BUSY:
-                response.setCode(ResponseCode.SYSTEM_ERROR);
+                response.setCode(ResponseCode.SYSTEM_BUSY);
                 response.setRemark("[PC_SYNCHRONIZED]broker busy, start flow control for a while");
                 break;
             case LMQ_CONSUME_QUEUE_NUM_EXCEEDED:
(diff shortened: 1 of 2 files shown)


RongtongJin (approved):
From another perspective, semantically returning busy is indeed more reasonable than error.

yuz10 (changes requested):
It is correct to change the response code, but we need more discussion on whether client should retry on this scenario. Personally I think the client should retry by default to avoid producing failure.

biningo:
> It is correct to change the response code, but we need more discussion on whether client should retry on this scenario. Personally I think the client should retry by default to avoid producing failure.

I agree with you

cserwen:
> It is correct to change the response code, but we need more discussion on whether client should retry on this scenario. Personally I think the client should retry by default to avoid producing failure.

There has been a lot of discussion about whether system_busy should be retried. I personally support retrying. For details, please refer to https://github.com/apache/rocketmq/pull/5845 @yuz10

cserwen:
@biningo merge develop to fix the ut

cserwen (approved):

biningo:
> @biningo merge develop to fix the ut

okay, I have merged.

biningo:
Hi @yuz10 
#5845 already merged. Please review again.

yuz10 (approved):
May affect old clients.

mxsm (approved):

lizhimins (approved):

Issue #8062, opened 2024-04-24
Enhancement Request[Enhancement] Virtual thread support for consumer thread


He-Pin:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Add virtual thread support for `ConsumeMessageService`

### Motivation

There are many `ConsumeMessageThread_` threads at work, so I would like to leverage the virtual thread in Java 21.

### Describe the Solution You'd Like

Add virtual thread support when user turns it on and the virtual thread is supported.

### Describe Alternatives You've Considered

delegate to a virtual thread pool.

### Additional Context

at $Work

PR #8063, opened 2024-04-24
[ISSUE #8062] feat: Add virtual thread support for consumers

He-Pin:

Fixes https://github.com/apache/rocketmq/issues/8062

### Brief Description

1. Add config for preferring the virtual thread, `"rocketmq.client.consumer.virtualThread"` default to `false`
2. When config is `true` and the Virtual thread is enabled, Make use of the virtual thread.

with unit tests

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
@@ -50,6 +50,8 @@ import org.apache.rocketmq.logging.org.slf4j.LoggerFactory;
 
 public class ConsumeMessageConcurrentlyService implements ConsumeMessageService {
     private static final Logger log = LoggerFactory.getLogger(ConsumeMessageConcurrentlyService.class);
+    private static final boolean PREFER_VIRTUAL = Boolean.parseBoolean(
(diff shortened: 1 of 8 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageConcurrentlyService.java
```diff
@@ -70,13 +72,14 @@ public ConsumeMessageConcurrentlyService(DefaultMQPushConsumerImpl defaultMQPush
...
         String consumerGroupTag = (consumerGroup.length() > 100 ? consumerGroup.substring(0, 100) : consumerGroup) + "_";
+        final String threadNamePrefix = "ConsumeMessageThread_" + consumerGroupTag;
         this.consumeExecutor = new ThreadPoolExecutor(
             this.defaultMQPushConsumer.getConsumeThreadMin(),
             this.defaultMQPushConsumer.getConsumeThreadMax(),
             1000 * 60,
             TimeUnit.MILLISECONDS,
             this.consumeRequestQueue,
-            new ThreadFactoryImpl("ConsumeMessageThread_" + consumerGroupTag));
+            new ThreadFactoryImpl(threadNamePrefix, false, null, PREFER_VIRTUAL));
```
He-Pin, line 82: This is not using the one task one executor pattern for Virtual Thread, this works well too, otherwise, needs much adoption.

Issue #7111, opened 2023-08-02
[Enhancement] When an exception is thrown, remove the responseFuture in a timely manner


RapperCL:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

org.apache.rocketmq.remoting.netty.NettyRemotingAbstract#invokeAsyncImpl
![image](https://github.com/apache/rocketmq/assets/44110731/91dd1117-d05b-4847-b863-1c9cd9fae5b7)

In this method, When an exception is thrown, remove the responseFuture in a timely manner，Even if there is a scheduled scan（scanResponseTable） for expired tasks, it is most appropriate to remove them as soon as possible.
Moreover, when an exception occurs during sending, there is no need to consider InvokeCallback. just remove it directly
### Motivation

During the process of sending a request, there may be exceptions caused by network or other reasons, which will result in the above-mentioned situation. 

### Describe the Solution You'd Like

In the exception block, remove responseFuture

### Describe Alternatives You've Considered

Not yet

cserwen:
Has been fixed by #7112, so closed

PR #8153, opened 2024-05-16
Remove responseFuture from the responseTable when except

im47cn:
See #7111


cserwen:
It has been fixed in develop and only some important security fixes or vulnerabilities will be merged into 4.9.x, so this PR will not be merged @im47cn

PR #8167, opened 2024-05-20
[ISSUE #8166]optimize: make compression type configurable in producer clinet level

### LINKED ISSUE #8166: [Enhancement] Configure production compression type at the producer instance level instead of at the process level

### Summary

The current way to configure the producer compression type is through `System.setProperty`, which is process level.
When there are multiple topics in one process that need to produce messages, all topics will use the same type of compression method.

### Motivation

When there are multiple topics in a process that need to produce messages, all topics will use the same type of compression method.So we can change compression configure at the  producer instance level, user can create different producer instance to enable different type of compression.

Configure production compression at the producer instance level instead of at the process level.

None.

### Additional Context

_No response_

humkum:

Fixes #8166 

### Brief Description

Configure production compression at the producer instance level instead of at the process level.

Producer configure the compression type with "ZSTD", and send message who's body over 4KB, debug the compression type producer use is ZSTDCompressor.

image

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -70,9 +70,6 @@ import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.ServiceState;
 import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.UtilAll;
-import org.apache.rocketmq.common.compression.CompressionType;
-import org.apache.rocketmq.common.compression.Compressor;
-import org.apache.rocketmq.common.compression.CompressorFactory;
 import org.apache.rocketmq.common.help.FAQUrl;
 import org.apache.rocketmq.common.message.Message;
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -118,11 +115,6 @@ public class DefaultMQProducerImpl implements MQProducerInner {
     private MQFaultStrategy mqFaultStrategy;
     private ExecutorService asyncSenderExecutor;
 
-    // compression related
-    private int compressLevel = Integer.parseInt(System.getProperty(MixAll.MESSAGE_COMPRESS_LEVEL, "5"));
-    private CompressionType compressType = CompressionType.of(System.getProperty(MixAll.MESSAGE_COMPRESS_TYPE, "ZLIB"));
-    private final Compressor compressor = CompressorFactory.getCompressor(compressType);
-
```
RongtongJin, line 125: There may be compatibility issues.

humkum: OK，I will resolve the compatibility issues.

cserwen (approved):

PR #8169, opened 2024-05-20
[ISSUE #8168] fix: There's no need to retry when async produce already timeout

### LINKED ISSUE #8168: [Bug] Produce async no need retry when the request is already timeout.

CentOS7.3

### RocketMQ version

5.2.0

### JDK Version

Open_JDK 1.8.0_202

### Describe the Bug

When producer send a message in async mode, if the request is timeout with throwing a RemotingTooMuchException,  the producer would retry with a negative value with timeoutMillis, which is not necessary.
image
image

### Steps to Reproduce

Producer add a RPCHook
image
Try sleep 4s in in the doBeforeRequest function，to make the request cost time out.

No retry when the request is already timeout.

### What Did You See Instead?

The producer retry with a negative value of timeout, which will must failed, and it's unnecessary.

### Additional Context

_No response_

humkum:

Fixes #8168

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -704,9 +704,10 @@ public class MQClientAPIImpl implements NameServerUpdateCallback {
                         onExceptionImpl(brokerName, msg, timeoutMillis - cost, request, sendCallback, topicPublishInfo, instance,
                             retryTimesWhenSendFailed, times, ex, context, true, producer);
                     } else {
-                        MQClientException ex = new MQClientException("unknow reseaon", throwable);
+                        MQClientException ex = new MQClientException("unknown reason", throwable);
+                        boolean needRetry = !(throwable instanceof RemotingTooMuchRequestException);
                         onExceptionImpl(brokerName, msg, timeoutMillis - cost, request, sendCallback, topicPublishInfo, instance,
-                            retryTimesWhenSendFailed, times, ex, context, true, producer);
+                            retryTimesWhenSendFailed, times, ex, context, needRetry, producer);
                     }
                 }
             });


client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -713,8 +713,10 @@ public void operationFail(Throwable throwable) {
         } catch (Exception ex) {
             long cost = System.currentTimeMillis() - beginStartTime;
             producer.updateFaultItem(brokerName, cost, true, false);
-            onExceptionImpl(brokerName, msg, timeoutMillis - cost, request, sendCallback, topicPublishInfo, instance,
-                retryTimesWhenSendFailed, times, ex, context, true, producer);
+            if (!(ex instanceof RemotingTooMuchRequestException)) {
+                onExceptionImpl(brokerName, msg, timeoutMillis - cost, request, sendCallback, topicPublishInfo, instance,
+                        retryTimesWhenSendFailed, times, ex, context, true, producer);
+            }
```
RongtongJin, line 719: Even if not retried, the onExceptionImpl method should still be executed, with the retry parameter set to false.

humkum: Got，I will make some change for this.

RongtongJin (approved):

RongtongJin:
PTAL～@guyinyou

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -704,7 +704,7 @@ public void operationFail(Throwable throwable) {
                         onExceptionImpl(brokerName, msg, timeoutMillis - cost, request, sendCallback, topicPublishInfo, instance,
                             retryTimesWhenSendFailed, times, ex, context, true, producer);
                     } else {
-                        MQClientException ex = new MQClientException("unknow reseaon", throwable);
+                        MQClientException ex = new MQClientException("unknown reason", throwable);
```
Willhow-Gao, line 707: If the exception here is a RemotingTooMuchRequestException, you should also stop retrying.

Willhow-Gao (changes requested):

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
humkum: You are right, and the invokeAsync won't throw RemotingTooMuchRequestException, I will make some changes.

Willhow-Gao (approved):

guyinyou:
lgtm

guyinyou (approved):

RongtongJin (approved):

cserwen (approved):

lizhimins (approved):

PR #8172, opened 2024-05-20
[ISSUE #8171] Reduce try catch scope in ProcessQueue#removeMessage

### LINKED ISSUE #8171: [Enhancement] Reduce try catch scope in ProcessQueue#removeMessage

### Summary

the try catch only needs to catch the lockInterruptibly, the other place won't throw a exception
image

### Motivation

clean code

just catching up on lockInterruptibly

none

### Additional Context

_No response_

miles-ton:

Fixes #8171


humkum:
IMO, this can be simplified to one try catch.

PR #8180, opened 2024-05-22
[ISSUE #6873]If dns resolve controller address exception will update controllerAddresses to null

weihubeats:
https://github.com/apache/rocketmq/issues/6873


weihubeats:
Let me give you an example, if the normal ip address for dns resolution is 192.168.1.1
After the dns parsing exception, the controller's address should not be updated, should continue to maintain 192.168.1.1, the original return will return an empty List to update the controller address to cause the controller address is empty, this is not reasonable!

weihubeats:
@RongtongJin @ferrirW

weihubeats:
I'm very sorry that the previous pr https://github.com/apache/rocketmq/issues/6873 was closed by me, so please take a look at this latest pr

RongtongJin (approved):

PR #8240, opened 2024-06-01
[ISSUE #8239] Fix the issue of potential message loss after a crash under synchronous disk flushing configuration.

### LINKED ISSUE #8239: [Bug] In the case of single-replica synchronous disk flushing, messages may be lost during a crash

All platforms

### RocketMQ version

All versions

### JDK Version

Not related to the JDK version

### Describe the Bug

image
As shown in the image, the consumer might read unflushed messages, leading to premature consumption and a forward shift of the consumption offset. After a crash, the consumeQueue may roll back, and new messages could have a chance of not being consumed.

### Steps to Reproduce

In a single-replica scenario with synchronous disk flushing configured, send a message and perform a crash test to check for message loss.

No Messages lost.

### What Did You See Instead?

Messages lost.

### Additional Context

Under the condition of synchronous disk flushing, unflushed messages should be unacknowledged messages. Such messages should not be consumed prematurely, as this would lead to the consumption offset moving forward, resulting in potential phantom reads or message loss on the consumer side.

RongtongJin:

Fixes #8239 

### Brief Description

Under the condition of synchronous disk flushing, unflushed messages should be unacknowledged messages. Such messages should not be consumed prematurely, as this would lead to the consumption offset moving forward, resulting in potential phantom reads and message loss on the consumer side.

Therefore, in this PR, the dispatch offset of the ConsumeQueue is restricted to be before the disk flush offset to prevent consumers from experiencing phantom reads.

Add a UT

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -651,7 +651,7 @@ public class CommitLog implements Swappable {
         } else if (this.defaultMessageStore.getMessageStoreConfig().isDuplicationEnable()) {
             return this.confirmOffset;
         } else {
-            return getMaxOffset();
+            return this.defaultMessageStore.isSyncDiskFlush() ? getFlushedWhere() : getMaxOffset();
         }
     }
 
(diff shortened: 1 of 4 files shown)


yx9o (approved):

yuz10 (approved):

lizhimins (dismissed):

guyinyou (approved):
lgtm

lizhimins (approved):

PR #8253, opened 2024-06-04
Add data block configuration and simplify code

lby1119:
Add data block configuration and simplify code


yx9o:
Hi @lby1119 , it would be better to submit a relevant issue first.

lby1119:
@yx9o Okay, I'll add it later.

PR #8267, opened 2024-06-06
[ISSUE #8265] Implement Batch Creation of Topics in RocketMQ Admin

### LINKED ISSUE #8265: [Enhancement] Implement Batch Creation of Topics in RocketMQ Admin

### Summary

The RocketMQ project currently supports topic creation through the MQAdmin API. However, the existing API only supports the creation of one topic at a time, which can be inefficient and time-consuming when there is a need to create a large number of topics simultaneously. To address this limitation, I propose an enhancement to the MQAdmin API to include a new UPDATE_AND_CREATE_TOPIC_LIST interface that enables batch creation of topics, significantly improving efficiency for large-scale topic management tasks.

### Motivation

The motivation for this enhancement arises from scenarios in distributed systems where multiple topics need to be created for different services or instances. These scenarios are common in large-scale deployments, cloud environments, and multi-tenant platforms where topics are dynamically assigned and managed. The batch creation feature will not only decrease the time required to set up such environments but also minimize the potential for errors during the manual creation process. It aligns with the demands of modern distributed architectures, offering a significant benefit by streamlining the management workflow, reducing operational overhead, and improving user experience. Specific use cases include setting up test environments, migrating or replicating environments, and scaling out services that require multiple topics to be set up in parallel.

The proposed solution is to introduce a new API interface UPDATE_AND_CREATE_TOPIC_LIST within the MQAdmin toolset. This interface will accept a list of topic configurations and process the batch creation in a single operation. The change will involve enhancing the current administrative command handlers to process multiple topics in one go, leveraging existing infrastructure and ensuring backward compatibility with existing single-topic creation APIs. The implementation will also include appropriate validation checks to ensure atomicity, where all topics need to be successfully created, or none at all, to maintain consistency.

none

### Additional Context

_No response_

guyinyou:

Fixes #8265

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -116,6 +116,7 @@ import org.apache.rocketmq.remoting.protocol.body.Connection;
 import org.apache.rocketmq.remoting.protocol.body.ConsumeQueueData;
 import org.apache.rocketmq.remoting.protocol.body.ConsumeStatsList;
 import org.apache.rocketmq.remoting.protocol.body.ConsumerConnection;
+import org.apache.rocketmq.remoting.protocol.body.CreateTopicListRequestBody;
 import org.apache.rocketmq.remoting.protocol.body.EpochEntryCache;
 import org.apache.rocketmq.remoting.protocol.body.GroupList;
 import org.apache.rocketmq.remoting.protocol.body.HARuntimeInfo;
@@ -243,6 +244,8 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         switch (request.getCode()) {
             case RequestCode.UPDATE_AND_CREATE_TOPIC:
                 return this.updateAndCreateTopic(ctx, request);
+            case RequestCode.UPDATE_AND_CREATE_TOPIC_LIST:
+                return this.updateAndCreateTopicList(ctx, request);
             case RequestCode.DELETE_TOPIC_IN_BROKER:
                 return this.deleteTopic(ctx, request);
             case RequestCode.GET_ALL_TOPIC_CONFIG:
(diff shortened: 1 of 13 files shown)


RongtongJin (approved):

fuyou001 (approved):

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -536,6 +539,84 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
...
+                    && !brokerController.getBrokerConfig().isEnableMixedMessageType()) {
+                    response.setCode(ResponseCode.SYSTEM_ERROR);
+                    response.setRemark("MIXED message type is not supported.");
+                    return response;
+                }
+                if (topicConfig.equals(this.brokerController.getTopicConfigManager().getTopicConfigTable().get(topic))) {
+                    LOGGER.info("Broker receive request to update or create topic={}, but topicConfig has  no changes , so idempotent, caller address={}",
+                        topic, RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
+                    response.setCode(ResponseCode.SUCCESS);
+                    return response;
```
absolute8511, line 587: we should continue here instead return, otherwise we will left the other topics uncreated after return success.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -536,6 +539,84 @@ private synchronized RemotingCommand updateAndCreateTopic(ChannelHandlerContext
...
+
+        long executionTime;
+
+        try {
+            // Valid topics
+            for (TopicConfig topicConfig : topicConfigList) {
+                String topic = topicConfig.getTopicName();
+                TopicValidator.ValidateTopicResult result = TopicValidator.validateTopic(topic);
+                if (!result.isValid()) {
+                    response.setCode(ResponseCode.SYSTEM_ERROR);
```
absolute8511, line 566: `SYSTEM_ERROR` has been used in many different server error cases, maybe we should add some more code to indicate the client error

PR #8271, opened 2024-06-06
[ISSUE #8269] Support pop consumption filter in long polling service

### LINKED ISSUE #8269: [Enhancement] Support pop consumption filter in long polling service

### Summary

Support pop consumption filter in long polling service

### Motivation

When a message is written, it wakes up long polling. In the current design, for pull message operations, only clients with matching filters are woken up, while for pop operations, all subscribed groups with pending long polling are woken up. This issue and pull request optimizes the notify mechanism, reducing CPU cost in scenarios where both pop consumption and tag filtering are present.

Support pop consumption filter in long polling service

Support pop consumption filter in long polling service

### Additional Context

_No response_

lizhimins:

Fixes #8269


drpmma (approved):
LGTM

PR #8272, opened 2024-06-06
[ISSUE #8261]Avoid unnecessary waiting when a response is successfully returned

### LINKED ISSUE #8261: [Enhancement] Avoid unnecessary waiting when a response is successfully returned

### Summary

image
public Message request(final Message msg, long timeout) Internal logic
image
image
image

Use countDownLatch to implement synchronization logic, but the thread is forced to wake up only after a timeout, and to wait until the timeout returns after a successful response，Therefore, after a successful response, the countDown() function should be called to wake up the thread to return, and do not do ineffective sleep

### Motivation

 Avoid unnecessary waiting when a response is successfully returned

 after a successful response, the countDown() function should be called to wake up the thread to return, and do not do ineffective sleep

null

### Additional Context

_No response_

3424672656:

Fixes #8261

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -969,7 +969,7 @@ public class DefaultMQProducerImpl implements MQProducerInner {
                         boolean messageCloned = false;
                         if (msgBodyCompressed) {
                             //If msg body was compressed, msgbody should be reset using prevBody.
-                            //Clone new message using commpressed message body and recover origin massage.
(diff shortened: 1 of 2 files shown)


GenerousMan (approved):
LGTM

GenerousMan:
LGTM, but would it be better to add `countDown` in this piece of code?

```java
public void setSendRequestOk(boolean sendRequestOk) {
    this.sendRequestOk = sendRequestOk;
}
```

3424672656:
> `countDown`LGTM，但是添加代码会更好吗？
> 
> ```java
> public void setSendRequestOk(boolean sendRequestOk) {
>     this.sendRequestOk = sendRequestOk;
> }
> ```

But exception callbacks are also used
image

RongtongJin (approved):

PR #7687, opened 2023-12-20
[ISSUE #7686] The bornTime is not set when using the popMessage API in cluster mode.

dingshuangxi888:
### Which Issue(s) This PR Fixes

Fixes #7686

### Brief Description

The bornTime is not set when using the popMessage API in cluster mode, So it will cause "timeout too much" error. And will never receive any messages.

### How Did You Test This Change?

Fix before:
Use gRPC client to receive message by Proxy, you can not receive any messages.

Fix after:
Use gRPC client to receive message by Proxy, you can receive all messages.

Issue #7686, opened 2023-12-20
[Bug] The bornTime is not set when using the popMessage API in cluster mode.


dingshuangxi888:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

Linux

### RocketMQ version

develop

### JDK Version

Java 11

### Describe the Bug

The bornTime is not set when using the popMessage API in cluster mode, So it will cause "timeout too much" error. And will never receive any messages.

![image](https://github.com/apache/rocketmq/assets/5010921/5d698a5c-b9d4-499e-9c14-757720fbce67)

### Steps to Reproduce

Use gRPC client to receive  message by Proxy, you can not receive any messages.

### What Did You Expect to See?

When using the gRPC client to receive messages, it will return correctly.

### What Did You See Instead?

When using the gRPC client to receive messages, it will return correctly.

YalikWang:
Why was this commit reverted?

PR #8331, opened 2024-06-26
Revert "[ISSUE #7686] The bornTime is not set when using the popMessage API in cluster mode."

drpmma:
Reverts apache/rocketmq#7687

In cluster mode, when the proxy sets borntime, and there is a clock synchronization issue between the proxy and the broker, it could cause the requests sent by the proxy to the broker to return POLLING_TIMEOUT, ultimately leading to request failures.


lollipopjin (approved):
LGTM

fuyou001 (approved):
LGTM

Issue #8468, opened 2024-07-31
[Enhancement] Add message gray strategy solution, compatible with RocketMQ 4.x and RocketMQ 5.x


syhleo:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

Refer [RIP-72](https://docs.google.com/document/d/18-8kTnqLZWOF5IHs5AEKK2FuvNRo0_yCLzuJk4NfU7M/edit?usp=sharing)
detail link: [语雀](https://www.yuque.com/u2210974/zrf1wg/bgqmutg83wyeyzfq)
### Summary

The solution provides an extensible message grayscale solution for implementing message grayscale publishing that supports **`RocketMQ 4.x`** and **`RocketMQ 5.x`** . It is compatible with **`POP`** consumption mode and **`Push`** consumption mode. No matter the client-rebalance or server-rebalance, the solution can realize the message gray level in a lightweight way.
The solution has been widely applied and verified in our projects, which confirms its reliability, security, and stability.

### Motivation

    At present, many business systems control the impact of version release through the gray scale release process, and almost all service releases will strictly go gray scale release first, and then go online after verification. gRPC and HTTP calls during GRay-scale publishing can be precisely controlled through microservice governance and gateway request headers. 
    However, if the message gray scale scheme is not implemented, even if the current link is a gray scale link, the message transmission is still uncontrollable, may go to the gray cluster monitoring partition, may also enter the online cluster monitoring partition. Therefore, when it comes to consuming logic changes, developers need to incorporate a lot of compatibility logic into their code. Even so, these logics only ensure that new traffic does not affect the line, but they do not guarantee that gray-scale traffic precisely enters the gray-scale consumer client, so that strict gray-scale verification cannot be performed. 
    The industry lacks perfect message gray scale solution, the existing MQ gray scale solution can not completely solve the problem of message isolation and switching interface. Therefore, it is particularly important to design a lightweight message gray scale scheme that is easy to access and controllable to upgrade.

### Describe the Solution You'd Like

1. **Add client configuration:** Add `enableGraySwitch` and `grayTag`. `enableGraySwitch` Indicates whether the grayscale message is enabled. grayTag indicates whether the client is a grayscale client. `grayTag` takes effect only when `enableGraySwitch` is enabled. The solution adds two configuration items that make it easy for the RocketMQ consumer to determine whether a client is a grayscale client. 
[...]

PR #8469, opened 2024-07-31
[RIP-72] : Add a message gray solution , compatible with RocketMQ 4.x and RocketMQ 5.x

syhleo:
[Enhancement] Add message gray strategy solution #8468

### Brief Description
a solution for message gray.
The solution provides an extensible message grayscale solution for implementing message grayscale publishing that supports **`RocketMQ 4.x`** and **`RocketMQ 5.x`** . It is compatible with **`POP`** consumption mode and **`Push`** consumption mode. No matter the client-rebalance or server-rebalance, the solution can realize the message gray level in a lightweight way.
The solution has been widely applied and verified in our projects, which confirms its reliability, security, and stability.

image
image
image

image

image

image

image

Easy access. Business parties control only a few client configurations, such as `enableGraySwitch` and `grayTag`, to seamlessly access RocketMQ's gray-scale publishing capabilities, enabling full-link gray-scale publishing.Specific can refer to `org.apache.rocketmq.example.gray`

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -1941,7 +1941,7 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         final RemotingCommand response = RemotingCommand.createResponseCommand(null);
 
         String content = this.brokerController.getQueryAssignmentProcessor().getMessageRequestModeManager().encode();
-        if (content != null && content.length() > 0) {
+        if (content != null && !content.isEmpty()) {
             try {
                 response.setBody(content.getBytes(MixAll.DEFAULT_CHARSET));
             } catch (UnsupportedEncodingException e) {
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/QueryAssignmentProcessor.java
@@ -32,6 +32,7 @@ import org.apache.rocketmq.broker.topic.TopicRouteInfoManager;
 import org.apache.rocketmq.client.consumer.AllocateMessageQueueStrategy;
 import org.apache.rocketmq.client.consumer.rebalance.AllocateMessageQueueAveragely;
 import org.apache.rocketmq.client.consumer.rebalance.AllocateMessageQueueAveragelyByCircle;
+import org.apache.rocketmq.client.consumer.rebalance.AllocateMessageQueueByGray;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.common.message.MessageQueue;
@@ -68,6 +69,9 @@ public class QueryAssignmentProcessor implements NettyRequestProcessor {
         name2LoadStrategy.put(allocateMessageQueueAveragely.getName(), allocateMessageQueueAveragely);
         AllocateMessageQueueAveragelyByCircle allocateMessageQueueAveragelyByCircle = new AllocateMessageQueueAveragelyByCircle();
         name2LoadStrategy.put(allocateMessageQueueAveragelyByCircle.getName(), allocateMessageQueueAveragelyByCircle);
+        // Add the gray-scale partitioning strategy: AllocateMessageQueueByGray
+        AllocateMessageQueueByGray allocateMessageQueueByGray = new AllocateMessageQueueByGray();
+        name2LoadStrategy.put(allocateMessageQueueByGray.getName(), allocateMessageQueueByGray);
 
         this.messageRequestModeManager = new MessageRequestModeManager(brokerController);
         this.messageRequestModeManager.load();
@@ -234,9 +238,14 @@ public class QueryAssignmentProcessor implements NettyRequestProcessor {
     }
 
     public List<MessageQueue> allocate4Pop(AllocateMessageQueueStrategy allocateMessageQueueStrategy,
-        final String consumerGroup, final String clientId, List<MessageQueue> mqAll, List<String> cidAll,
-        int popShareQueueNum) {
-
+                                           final String consumerGroup, final String clientId, List<MessageQueue> mqAll, List<String> cidAll,
+                                           int popShareQueueNum) {
+        if (allocateMessageQueueStrategy instanceof AllocateMessageQueueByGray) {
+            AllocateMessageQueueByGray allocateMessageQueueByGray = (AllocateMessageQueueByGray) allocateMessageQueueStrategy;
+            if (AllocateMessageQueueByGray.hasGrayTag(cidAll)) {
+                return allocateMessageQueueByGray.allocate4Pop(consumerGroup, clientId, mqAll, cidAll, popShareQueueNum);
+            }
+        }
         List<MessageQueue> allocateResult;
         if (popShareQueueNum <= 0 || popShareQueueNum >= cidAll.size() - 1) {
             //each client pop all messagequeue
@@ -270,7 +279,7 @@ public class QueryAssignmentProcessor implements NettyRequestProcessor {
     }
 
     private List<MessageQueue> allocate(String consumerGroup, String currentCID, List<MessageQueue> mqAll,
-        List<String> cidAll) {
+                                        List<String> cidAll) {
         if (StringUtils.isBlank(currentCID)) {
             throw new IllegalArgumentException("currentCID is empty");
         }
--- a/broker/src/test/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessorTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessorTest.java
@@ -20,21 +20,6 @@ import com.alibaba.fastjson.JSON;
 import com.google.common.collect.Sets;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
-import java.net.InetSocketAddress;
-import java.net.SocketAddress;
-import java.net.UnknownHostException;
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.LongAdder;
 import org.apache.rocketmq.auth.authentication.enums.UserType;
 import org.apache.rocketmq.auth.authentication.manager.AuthenticationMetadataManager;
 import org.apache.rocketmq.auth.authentication.model.Subject;
@@ -45,8 +30,10 @@ import org.apache.rocketmq.auth.authorization.model.Acl;
 import org.apache.rocketmq.auth.authorization.model.Environment;
 import org.apache.rocketmq.auth.authorization.model.Resource;
 import org.apache.rocketmq.broker.BrokerController;
+import org.apache.rocketmq.broker.client.ClientChannelInfo;
 import org.apache.rocketmq.broker.client.ConsumerGroupInfo;
 import org.apache.rocketmq.broker.client.ConsumerManager;
+import org.apache.rocketmq.broker.client.net.Broker2Client;
 import org.apache.rocketmq.broker.offset.ConsumerOffsetManager;
 import org.apache.rocketmq.broker.schedule.ScheduleMessageService;
 import org.apache.rocketmq.broker.subscription.RocksDBSubscriptionGroupManager;
@@ -55,11 +42,13 @@ import org.apache.rocketmq.broker.topic.TopicConfigManager;
 import org.apache.rocketmq.common.BoundaryType;
 import org.apache.rocketmq.common.BrokerConfig;
 import org.apache.rocketmq.common.KeyBuilder;
+import org.apache.rocketmq.common.MQVersion;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.TopicConfig;
 import org.apache.rocketmq.common.TopicFilterType;
 import org.apache.rocketmq.common.TopicQueueId;
 import org.apache.rocketmq.common.action.Action;
+import org.apache.rocketmq.common.constant.FIleReadaheadMode;
 import org.apache.rocketmq.common.constant.PermName;
 import org.apache.rocketmq.common.consumer.ConsumeFromWhere;
 import org.apache.rocketmq.common.message.MessageAccessor;
@@ -67,13 +56,20 @@ import org.apache.rocketmq.common.message.MessageConst;
 import org.apache.rocketmq.common.message.MessageExt;
 import org.apache.rocketmq.common.topic.TopicValidator;
 import org.apache.rocketmq.remoting.exception.RemotingCommandException;
+import org.apache.rocketmq.remoting.exception.RemotingSendRequestException;
+import org.apache.rocketmq.remoting.exception.RemotingTimeoutException;
 import org.apache.rocketmq.remoting.netty.NettyClientConfig;
 import org.apache.rocketmq.remoting.netty.NettyServerConfig;
 import org.apache.rocketmq.remoting.protocol.RemotingCommand;
+import org.apache.rocketmq.remoting.protocol.RemotingSerializable;
 import org.apache.rocketmq.remoting.protocol.RequestCode;
 import org.apache.rocketmq.remoting.protocol.ResponseCode;
 import org.apache.rocketmq.remoting.protocol.body.AclInfo;
+import org.apache.rocketmq.remoting.protocol.body.CreateTopicListRequestBody;
+import org.apache.rocketmq.remoting.protocol.body.GroupList;
+import org.apache.rocketmq.remoting.protocol.body.HARuntimeInfo;
 import org.apache.rocketmq.remoting.protocol.body.LockBatchRequestBody;
+import org.apache.rocketmq.remoting.protocol.body.QueryCorrectionOffsetBody;
 import org.apache.rocketmq.remoting.protocol.body.UnlockBatchRequestBody;
 import org.apache.rocketmq.remoting.protocol.body.UserInfo;
 import org.apache.rocketmq.remoting.protocol.header.CreateAclRequestHeader;
@@ -82,8 +78,11 @@ import org.apache.rocketmq.remoting.protocol.header.CreateUserRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.DeleteAclRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.DeleteTopicRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.DeleteUserRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.ExchangeHAInfoResponseHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetAclRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetAllTopicConfigResponseHeader;
+import org.apache.rocketmq.remoting.protocol.header.GetConsumerRunningInfoRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.GetConsumerStatusRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetEarliestMsgStoretimeRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetMaxOffsetRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetMinOffsetRequestHeader;
@@ -91,6 +90,13 @@ import org.apache.rocketmq.remoting.protocol.header.GetTopicConfigRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.GetUserRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.ListAclsRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.ListUsersRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.NotifyMinBrokerIdChangeRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.QueryCorrectionOffsetHeader;
+import org.apache.rocketmq.remoting.protocol.header.QuerySubscriptionByConsumerRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.QueryTopicConsumeByWhoRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.QueryTopicsByConsumerRequestHeader;
+import org.apache.rocketmq.remoting.protocol.header.ResetMasterFlushOffsetHeader;
+import org.apache.rocketmq.remoting.protocol.header.ResetOffsetRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.ResumeCheckHalfMessageRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.SearchOffsetRequestHeader;
 import org.apache.rocketmq.remoting.protocol.header.UpdateAclRequestHeader;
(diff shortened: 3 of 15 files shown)


HelloGitbin:
perfect

yx9o:
@syhleo  Hello, I have a question. If the producer is configured with grayscale, but the consumer is not configured with grayscale, then the data in the grayscale queue will also be consumed by normal consumers. The same is true in reverse. Is this in line with expectations?

Qoozm:
> @syhleo Hello, I have a question. If the producer is configured with grayscale, but the consumer is not configured with grayscale, then the data in the grayscale queue will also be consumed by normal consumers. The same is true in reverse. Is this in line with expectations?

hi, in the issue's Describe the Solution, i found a note about this, it seems within expectations
image

syhleo:
> @syhleo Hello, I have a question. If the producer is configured with grayscale, but the consumer is not configured with grayscale, then the data in the grayscale queue will also be consumed by normal consumers. The same is true in reverse. Is this in line with expectations?

of course,
RocketMQ triggers the rebalancing process when all grayscale consumer clients do not exist (for example, the consumer goes online after the grayscale verification passes or the grayscale consumer goes offline abnormally). During this process, it is detected that all consumer ClientiDs do not contain the @gray identifier, so other normal consumer clients will immediately take over the consumption of messages, and any messages left in the gray scale queue will also be immediately taken over by normal consumer clients to ensure that messages are not lost.

You can see the related documentation:[语雀](https://www.yuque.com/u2210974/zrf1wg/bgqmutg83wyeyzfq)

syhleo:
> If the grayscale message and normal message data formats are inconsistent, and grayscale is enabled for production but not for consumption, will the grayscale data consumed normally fail to parse or cannot be processed normally?

Yes. If only the production is gray, the consumption is not gray, that is, the gray consumption does not exist at all, that is, all the messages are consumed by the normal consumer end. However, first of all, the change of MQ message data format, the production side and the consumer side are to be logically adjusted together, if your business scene inside the production side of the grayscale message and the data format of the ordinary message is inconsistent, the consumer side needs to follow together with the grayscale, so that the grayscale consumer side consumes the grayscale message (data format A). The normal consumer consumes normal messages (data format B). This is also the full-link gray scale release that we have been advocating.

HelloGitbin:
I tried it. It works.

yx9o:
> > If the grayscale message and normal message data formats are inconsistent, and grayscale is enabled for production but not for consumption, will the grayscale data consumed normally fail to parse or cannot be processed normally?
> 
> Yes. If only the production is gray, the consumption is not gray, that is, the gray consumption does not exist at all, that is, all the messages are consumed by the normal consumer end. However, first of all, the change of MQ message data format, the production side and the consumer side are to be logically adjusted together, if your business scene inside the production side of the grayscale message and the data format of the ordinary message is inconsistent, the consumer side needs to follow together with the grayscale, so that the grayscale consumer side consumes the grayscale message (data format A). The normal consumer consumes normal messages (data format B). This is also the full-link gray scale release that we have been advocating.

What I want is that the grayscale queue can only be consumed by grayscale consumers. When there is no grayscale consumer, I don’t want the data of both the grayscale queue and the ordinary queue to be consumed by normal consumers, so that the data is strung together.

syhleo:
> > > If the grayscale message and normal message data formats are inconsistent, and grayscale is enabled for production but not for consumption, will the grayscale data consumed normally fail to parse or cannot be processed normally?
> > 
> > 
> > Yes. If only the production is gray, the consumption is not gray, that is, the gray consumption does not exist at all, that is, all the messages are consumed by the normal consumer end. However, first of all, the change of MQ message data format, the production side and the consumer side are to be logically adjusted together, if your business scene inside the production side of the grayscale message and the data format of the ordinary message is inconsistent, the consumer side needs to follow together with the grayscale, so that the grayscale consumer side consumes the grayscale message (data format A). The normal consumer consumes normal messages (data format B). This is also the full-link gray scale release that we have been advocating.
> 
> What I want is that the grayscale queue can only be consumed by grayscale consumers. When there is no grayscale consumer, I don’t want the data of both the grayscale queue and the ordinary queue to be consumed by normal consumers, so that the data is strung together.

When implemented this way, there will be problems. Imagine this: If your producer service is in a grayscale environment, at this time a steady stream of grayscale messages are sent, and your consumer service is not in grayscale (because the consumer service has no distribution requirements at all, it has been a non-grayscale state), if the grayscale queue can only be used by grayscale consumers, in this case, the grayscale queue will cause a large number of messages to accumulate. During grayscale, grayscale queues can only be used by grayscale consumers, provided that there are grayscale consumers. If there are grayscale consumers, it is absolutely guaranteed that the grayscale queue can only be used by grayscale consumers. What we need to do is, during the full link gray scale release, the messages sent by gray scale producers can be accurately consumed by gray scale consumers, and the messages sent by non-gray scale producers can be accurately consumed by non-gray scale consumers, so that the gray scale verification involving MQ message changes can be preserved in the business. In order to ensure the normal switching between gray level and gray level, the same group of consumers perceive each other whether there are consumers with gray level identification to determine whether non-gray level consumers need to take over the gray level queue message.

syhleo:
This solution has been implemented in our production environment, and the access method is also low-cost, which confirms its stability and security.

yx9o:
> > > > If the grayscale message and normal message data formats are inconsistent, and grayscale is enabled for production but not for consumption, will the grayscale data consumed normally fail to parse or cannot be processed normally?
> > > 
> > > 
> > > Yes. If only the production is gray, the consumption is not gray, that is, the gray consumption does not exist at all, that is, all the messages are consumed by the normal consumer end. However, first of all, the change of MQ message data format, the production side and the consumer side are to be logically adjusted together, if your business scene inside the production side of the grayscale message and the data format of the ordinary message is inconsistent, the consumer side needs to follow together with the grayscale, so that the grayscale consumer side consumes the grayscale message (data format A). The normal consumer consumes normal messages (data format B). This is also the full-link gray scale release that we have been advocating.
> > 
> > 
> > What I want is that the grayscale queue can only be consumed by grayscale consumers. When there is no grayscale consumer, I don’t want the data of both the grayscale queue and the ordinary queue to be consumed by normal consumers, so that the data is strung together.
> 
> When implemented this way, there will be problems. Imagine this: If your producer service is in a grayscale environment, at this time a steady stream of grayscale messages are sent, and your consumer service is not in grayscale (because the consumer service has no distribution requirements at all, it has been a non-grayscale state), if the grayscale queue can only be used by grayscale consumers, in this case, the grayscale queue will cause a large number of messages to accumulate. During grayscale, grayscale queues can only be used by grayscale consumers, provided that there are grayscale consumers. If there are grayscale consumers, it is absolutely guaranteed that the grayscale queue can only be used by grayscale consumers. What we need to do is, during the full link gray scale release, the messages sent by gray scale producers can be accurately consumed by gray scale consumers, and the messages sent by non-gray scale producers can be accurately consumed by non-gray scale consumers, so that the gray scale verification involving MQ message changes can be preserved in the business. In order to ensure the normal switching between gray level and gray level, the same group of consumers perceive each other whether there are consumers with gray level identification to determine whether non-gray level consumers need to take over the gray level queue message.

Our usage scenarios are different. In our production implementation, grayscale can only be consumed by grayscale, and grayscale and ordinary messages are separated. Otherwise, they will be mixed and the meaning of grayscale will be lost.

imzs:
Good idea. I have a question: how will this solution do when we need more than one gray environment? 

Like Git branch, we may have several active branches and each of them runs in its independent gray environment.

syhleo:
First of all, thank you very much for your response and for asking the question.


This solution also effectively addresses this issue when multiple grayscale environments are required.
In project development, it is true that there are multiple active branches or environments. For this, the proposed solution is:
1. different environment processing: in different environments (e.g. dev/test/stage/prod), the environment suffix can be added for the same topic or group. This kind of processing has been widely used. For example, the topic name can be topic-{environment}, such as pt-order-completed-stage, and the group name can be GID-open-service-stage.
2. Multiple branches in the same environment: In the same environment (e.g., test environment), there may also exist multiple branch environments, such as test1/test2/test3.
In conjunction with this scenario, a grayscale partitioning approach can be used during grayscale release in a given environment. This approach ensures that messages sent by grayscale producers are accurately consumed by grayscale consumers, while messages from non-grayscale producers are processed by non-grayscale consumers. In this way, it ensures both grayscale validation of MQ message changes, as well as a friendly solution to the switchover interface from grayscale to the normal environment once the grayscale validation is passed.


In fact, most enterprises want to perform gray-scale validation of MQ messages in a low-cost and convenient way in an all-link gray-scale publishing scenario. However, when MQ does not support grayscale messages, it involves a change in consumption logic, and developers often need to add a lot of compatibility logic to the code. Nonetheless, these logics can only ensure that the new business does not affect the online, but cannot ensure that the grayscale traffic accurately enters the grayscale consumption client, thus preventing strict grayscale validation. With this solution, these problems can be effectively solved.

&nbsp;


宋燕辉
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人: ***@***.***&gt;; 
发送时间: 2024年8月12日(星期一) 下午2:22
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [apache/rocketmq] [RIP-69] : Add a message gray solution , compatible with POP consumption mode and Push consumption mode (PR #8469)














 
Good idea. I have a question: how will this solution do when we need more than one gray environment?
 
Like Git branch, we may have several active branches and each of them runs in its independent gray environment.

syhleo:
> Good idea. I have a question: how will this solution do when we need more than one gray environment?
> 
> Like Git branch, we may have several active branches and each of them runs in its independent gray environment.

First of all, thank you very much for your response and for asking the question.

This solution also effectively addresses this issue when multiple grayscale environments are required.
In project development, it is true that there are multiple active branches or environments. For this, the proposed solution is:
1. different environment processing: in different environments (e.g. dev/test/stage/prod), the environment suffix can be added for the same topic or group. This kind of processing has been widely used. For example, the topic name can be topic-{environment}, such as pt-order-completed-stage, and the group name can be GID-open-service-stage.
2. Multiple branches in the same environment: In the same environment (e.g., test environment), there may also exist multiple branch environments, such as test1/test2/test3.
In conjunction with this scenario, a grayscale partitioning approach can be used during grayscale release in a given environment. This approach ensures that messages sent by grayscale producers are accurately consumed by grayscale consumers, while messages from non-grayscale producers are processed by non-grayscale consumers. In this way, it ensures both grayscale validation of MQ message changes, as well as a friendly solution to the switchover interface from grayscale to the normal environment once the grayscale validation is passed.


In fact, most enterprises want to perform gray-scale validation of MQ messages in a low-cost and convenient way in an all-link gray-scale publishing scenario. However, when MQ does not support grayscale messages, it involves a change in consumption logic, and developers often need to add a lot of compatibility logic to the code. Nonetheless, these logics can only ensure that the new business does not affect the online, but cannot ensure that the grayscale traffic accurately enters the grayscale consumption client, thus preventing strict grayscale validation. With this solution, these problems can be effectively solved.

makabakaboom:
https://github.com/apache/rocketmq/issues/3265 21年我提交过相同的方案

syhleo:
> #3265 21年我提交过相同的方案

Thank you for your response. While comparing our solutions, I noticed the following differences between this solution and your previous one:

1.MQ version support: This solution is compatible with RocketMQ 4.x and RocketMQ 5.x.
2. Consumption mode: This program supports POP consumption mode and Push consumption mode.
3. Implementation details: This solution utilizes the clientId generation mechanism. By adding public client configurations, it allows MQ users to decide whether or not to enable grayscale messaging. These configurations allow RocketMQ users to easily determine if a client is a grayscale client.
4. Low cost and easy access.

francisoliverlee:
very nice to see gray solution, and i sort some cases that mq's users can be met, if meet, how to solve or extend those cases?
case 1: producers are all gray, consumers are all gray
case 5: producers are both gray and normal, consumers are all gray
others do same up.

image

qianye1001:
In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."

1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.

2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?

3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.

4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

syhleo:
> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.



> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

Thank you very much for your feedback.

Regarding the startup sequence of gray consumers: Yes, the PR recommends that gray consumers be started before gray producers to ensure that gray messages are correctly processed.

Regarding the issue of subscription inconsistency: This PR does not address scenarios with inconsistent subscriptions, as this issue is not caused by the gray partitioning scheme. Even without the gray scheme, when gray consumers (gray pods) and base consumers (base pods) have different subscription relationships, the same problem will occur. However, the issue of inconsistent subscriptions can be resolved through other means.

Regarding rewriting client logic: The Remoting Protocol client can indeed have its logic rewritten. Apologies, I haven't yet found the entry point for this. However, to ensure compatibility with RocketMQ 5.x server-side load balancing and to support both Pop and Push consumption modes, the PR introduces additional client configurations, providing a simple and flexible way to integrate the gray feature.

Multi-environment and gray release: The gray partitioning scheme and binding of different environments to corresponding queues do not conflict; rather, they complement each other. This design ensures that each environment can carry out gray releases in an orderly and independent manner.

> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

Thank you very much for your feedback.

Regarding the startup sequence of gray consumers: Yes, the PR recommends that gray consumers be started before gray producers to ensure that gray messages are correctly processed.

Regarding the issue of subscription inconsistency: This PR does not address scenarios with inconsistent subscriptions, as this issue is not caused by the gray partitioning scheme. Even without the gray scheme, when gray consumers (gray pods) and base consumers (base pods) have different subscription relationships, the same problem will occur. However, the issue of inconsistent subscriptions can be resolved through other means.

Regarding rewriting client logic: The Remoting Protocol client can indeed have its logic rewritten. Apologies, I haven't yet found the entry point for this. However, to ensure compatibility with RocketMQ 5.x server-side load balancing and to support both Pop and Push consumption modes, the PR introduces additional client configurations, providing a simple and flexible way to integrate the gray feature.

Multi-environment and gray release: The gray partitioning scheme and binding of different environments to corresponding queues do not conflict; rather, they complement each other. This design ensures that each environment can carry out gray releases in an orderly and independent manner.

> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

syhleo:
> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

Thank you very much for your feedback.

1.**Regarding the startup sequence of gray consumers**: Yes, the PR recommends that gray consumers be started before gray producers to ensure that gray messages are correctly processed.

2.**Regarding the issue of subscription inconsistency**: This PR does not address scenarios with inconsistent subscriptions, as this issue is not caused by the gray partitioning scheme. Even without the gray scheme, when gray consumers (gray pods) and base consumers (base pods) have different subscription relationships, the same problem will occur. However, the issue of inconsistent subscriptions can be resolved through other means.

3.**Regarding rewriting client logic**: The Remoting Protocol client can indeed have its logic rewritten. Apologies, I haven't yet found the entry point for this. However, to ensure compatibility with RocketMQ 5.x server-side load balancing and to support both Pop and Push consumption modes, the PR introduces additional client configurations, providing a simple and flexible way to integrate the gray feature.

4.**Multi-environment and gray release**: The gray partitioning scheme and binding of different environments to corresponding queues do not conflict; rather, they complement each other. This design ensures that each environment can carry out gray releases in an orderly and independent manner.

syhleo:
> very nice to see gray solution, and i sort some cases that mq's users can be met, if meet, how to solve or extend those cases? case 1: producers are all gray, consumers are all gray case 5: producers are both gray and normal, consumers are all gray others do same up.
> 
> image

MQ灰度 (2)

灰度_正常生命周期管理

Share the link
https://www.processon.com/view/link/66bdc5020ee74e516c89def5?cid=66bd5df5fb35b76d02da1e4c
https://www.processon.com/view/link/66bdc5252515964e0bbc2f63?cid=66bda70c0ee74e516c898985

Case Scenario Description:
In practice, the whole link gray-scale release scenario, almost all the services issued version will be strictly first go gray-scale release, verify ok and then online.
1, environment division and gray-scale release: in the daily development process, will be divided into dev/test/stage/prod and other environments, test environment will be due to a variety of project branches and so on, derived from testA, testB, testC and other project environments. We can decide whether to add a new topic/group with a “grayscale” logo according to the actual needs of each environment, so as to realize the grayscale capability. However, when the business scale is large, this approach may cause two problems: first, the criticality problem, i.e., when the grayscale validation passes and switches to prod (e.g., the pod of the grayscale service is taken offline normally), the grayscale messages may be consumed by omission, which is unacceptable; and second, the cost problem, since doubling of each topic and group will bring significant cost overheads.
For specific environments, complete isolation by adding new topics/groups and combining the ability to partition grayscale during grayscale in online environments are never in conflict and are complementary to each other.
2, the gray-scale partitioning used in this program is, more often than not, by not adding a new topic/group, quasi-environment-specific (eg: prod environment) for gray-scale release, you can make MQ low-cost, convenient way to also have the ability to gray-scale. (ps: full-link grayscale scenario, http, gRPC and other interface calls can be gray-scale traffic forwarding, mq also need to have ...)
This design ensures that each environment can perform grayscale publishing in an orderly and independent manner.

francisoliverlee:
> In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> 
> 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.

nice question 1, 2.
### question 1
no users can guarantee it, this canbe a BIG problem

### question 2
consumers in different-subscriptions, gray and normal consumers could be in-a-mess,  how to make sure consume offset's right

### my question
how gray consumer change into normal consumer. from the source code to see, it need to update consumer config and restart, can do it in admin tools?

syhleo:
> > In the following description, the current environment is referred to as the "base environment," to distinguish it from the "gray environment."
> > 
> > 1. The PR requires that the "gray consumer" be started before the "gray producer." If the gray consumer fails to start, it can lead to "gray messages" being sent to the base environment, may causing issues. In other words, the PR assumes that consumers in the base environment have the capability to consume gray messages, but this cannot be guaranteed.
> > 2. The plan does not account for scenarios where subscription are different. When gray consumers with different subscription and base consumers are running concurrently, how can we ensure that if client crashes or other exceptions, consumer offset is not skipped?
> > 3. The remoting protocol client  can be rewrite load balancing rules and select message queue function by users ; it is not a difficult task. Thus, the question arises whether it is necessary to merge this PR.
> > 4. This PR handles multiple environments poorly, as each environment must be bound to corresponding queues, thereby limiting the number of environments that can be supported.
> 
> @qianye1001 nice question 1, 2
> 
> @syhleo
> 
> ### question 1
> no users can guarantee it, this canbe a BIG problem, can we unbind the dependency on producer and consumer?
> 
> ### question 2
> consumers in different-subscriptions, gray and normal consumers could be in-a-mess, how to make sure consume offset's right
> 
> ### my question
> how gray consumer change into normal consumer. from the source code to see, it need to update consumer config and restart, can do it in admin tools?

Thank you for your feedback.
Subscription relationship inconsistency issue:
The subscription inconsistency issue is not triggered by the grayscale partitioning scheme of this PR. Even without the grayscale partitioning scheme, the same problem is faced when a grayscale consumer (grayscale pod) has a different subscription relationship than a normal consumer (normal pod). Solving the MQ subscription inconsistency problem requires another PR to deal with, and the commit scheme can be further explored if there is time, so we won't expand on it here.

Gray-to-normal bridging issue:
When the grayscale validation passes, it triggers the normal downtime of the grayscale service and triggers the rebalancing logic. By way of admin tools, I have not tried this. Either way, it's essentially modifying the client's grayTag, clientId configurations so that all consumer instances under the same consumer group can sense each other's grayscale status.

syhleo:
Thank you very much for all the great replies. I have summarized the above discussion in the [语雀](https://www.yuque.com/u2210974/zrf1wg/bgqmutg83wyeyzfq#GQruQ), so feel free to share and discuss the grayscale scenarios or MQ grayscale solution that your respective companies are experiencing.  🌹

PR #8484, opened 2024-08-03
[ISSUE #8483] Optimize unnecessary broker reverse notification (notifyConsumerIdsChanged) in broadcast mode

### LINKED ISSUE #8483: [Enhancement] Unnecessary broker back notification (notifyConsumerIdsChanged) in broadcast mode

### Summary

Hi, community, recently I found an unnecessary broker reverse notification (notifyConsumerIdsChanged) in broadcast mode.

Background: In one of our online incident investigations, there were thousands of single group broadcast clients, and it was found that the broker had been sending notifyConsumerIdsChanged requests to the broadcast clients.

image

企业微信截图_8e88e985-7986-494f-8971-8a4aafacbd2f

企业微信截图_b5c15368-849d-4cd1-bf6d-7bbdc80bbd29

企业微信截图_b0d6dd68-a323-4025-9b26-3acc99074024

### Motivation

Analysis: In broadcast mode, each Consumer consumes complete data, and its rebalance operation only updates the messagequeue information, which is also updated by a scheduled task and has nothing to do with the change in the number of Consumers. So I think that in broadcast mode, if there is any change in Consumer, there is no need to send notifyConsumerIdsChanged requests to all broadcast clients.

Modification: When sending ConsumerGroupEvent.CHANGE events in ConsumerManager, it is sent only when it is determined to be in non-broadcast mode.

No

### Additional Context

_No response_

yx9o:
Fixes #8483 .

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java
@@ -145,8 +145,9 @@ public class ConsumerManager {
                         callConsumerIdsChangeListener(ConsumerGroupEvent.UNREGISTER, next.getKey());
                     }
                 }
-
-                callConsumerIdsChangeListener(ConsumerGroupEvent.CHANGE, next.getKey(), info.getAllChannel());
+                if (!isBroadcastMode(info.getMessageModel())) {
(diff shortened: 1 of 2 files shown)


lizhanhui (approved):
Looks good

PR #8506, opened 2024-08-08
l think this default value  -1    fllow if case can run

liqinliang:
1723079207787_2A88375D-C25A-4613-B5CB-A85D94BF5E08


yx9o:
Hi @liqinliang , here we are checking the offset of consumequeue according to commitlog. It should not be -1, but the minimum of commitlog, so it is correct.

PR #8593, opened 2024-08-28
[ISSUE #8592] Not notify long polling request when pop orderly consume blocked

### LINKED ISSUE #8592: [Enhancement] Not notify long polling request when pop orderly consume blocked

### Summary

当顺序消息消费阻塞时，相同的队列又有新的消息写入，此时 rest num 大于 0 造成了 pop message processor 又重新唤醒长轮询。此时长轮询在 process 逻辑反复处理，造成有堆积消息但无法消费的情况。由于此时是顺序消费，服务端没有收到之前消息的 ack，因此消费延迟是符合预期的，只是 cpu 使用率会偏高，可以通过优化减少反复唤起长轮询。

When order message consumption is blocked, new messages are written to the same queue, and the rest num is greater than 0, causing the pop message processor to wake up the long polling again. Long polling is repeatedly processed in the process logic, resulting in a situation where there are lag messages that cannot be consumed. Due to the order consumption at this time, the broker did not receive the ACK of the previous pop message, so the consumption delay is expected. However, the CPU usage will be high, and optimization can be used to reduce the repeated triggering of long polling.

### Motivation

减少反复唤起长轮询造成的 CPU 使用率高。

Reduce the high CPU usage caused by repeatedly invoking long polling

Reduce the high CPU usage caused by repeatedly invoking long polling

not add extra rest num here

### Additional Context

_No response_

lizhimins:

Fixes #8592

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
@@ -551,18 +551,24 @@ public class PopMessageProcessor implements NettyRequestProcessor {
             future.whenComplete((result, throwable) -> queueLockManager.unLock(lockKey));
             offset = getPopOffset(topic, requestHeader.getConsumerGroup(), queueId, requestHeader.getInitMode(),
                 true, lockKey, true);
-            if (isOrder && brokerController.getConsumerOrderInfoManager().checkBlock(attemptId, topic,
-                requestHeader.getConsumerGroup(), queueId, requestHeader.getInvisibleTime())) {
-                future.complete(this.brokerController.getMessageStore().getMaxOffsetInQueue(topic, queueId) - offset + restNum);
-                return future;
-            }
 
+            // Current requests would calculate the total number of messages
+            // waiting to be filtered for new message arrival notifications in
+            // the long-polling service, need disregarding the backlog in order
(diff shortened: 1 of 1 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -551,18 +551,24 @@ private CompletableFuture<Long> popMsgFromQueue(String topic, String attemptId,
...
+            // the long-polling service, need disregarding the backlog in order
+            // consumption scenario. If rest message num including the blocked
+            // queue accumulation would lead to frequent unnecessary wake-ups
+            // of long-polling requests, resulting unnecessary CPU usage.
+            // When client ack message, long-polling request would be notifications
+            // by AckMessageProcessor.ackOrderly() and message will not be delayed.
             if (isOrder) {
+                if (brokerController.getConsumerOrderInfoManager().checkBlock(
+                    attemptId, topic, requestHeader.getConsumerGroup(), queueId, requestHeader.getInvisibleTime())) {
+                    // should not add accumulation(max offset - consumer offset) here
```
lollipopjin, line 566: LGTM，Means return the original restNum if lock failed.

lollipopjin (approved):
LGTM

PR #8619, opened 2024-09-02
[ISSUE #8618] Supports getting consumer group information for all pop patterns on the broker

### LINKED ISSUE #8618: [Enhancement] Supports getting consumer group information for all pop patterns on the broker

### Summary

 Supports getting consumer group information for all pop patterns on the broker

### Motivation

Sometimes it is necessary to invoke the interface to view or synchronize consumption patterns

null

null

### Additional Context

_No response_

3424672656:

Fixes #8618


client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -3080,6 +3081,21 @@ public void setMessageRequestMode(final String brokerAddr, final String topic, f
         }
     }
 
+    public MessageRequestModeSerializeWrapper getAllMessageRequestMode(final String brokerAddr) throws MQBrokerException,
+            RemotingConnectException, RemotingSendRequestException, RemotingTimeoutException, InterruptedException {
+        RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.GET_ALL_MESSAGE_REQUEST_MODE, null);
+        RemotingCommand response = this.remotingClient.invokeSync(brokerAddr, request, 3000);
```
lizhimins, line 3087: may timeout when execute get all

3424672656: The local test does not time out, I can increase the timeout parameter if necessary

Issue #8509, opened 2024-08-09
the slave nodes don't synchronize the LMQ consumer-offset


huyoule:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

ubuntu20.04.1

### RocketMQ version

5.2.0-release

### Describe the Bug

the slave nodes don't synchronize the LMQ consumer-offset, and then a slave becomes the master, the LMQ consumer can't get it's offset
![20240809-110320](https://github.com/user-attachments/assets/1a543f61-06df-4d55-9749-8f6f6671f127)
![20240809-110538](https://github.com/user-attachments/assets/26cbe6d9-c746-4640-bd1f-0bb531e3bdba)
![20240809-110543](https://github.com/user-attachments/assets/52eb54cd-b5f0-4c90-af09-93fff205bf4e)

### Steps to Reproduce

Just switch to the slave node

### What Did You Expect to See?

todo

### What Did You See Instead?

todo

yx9o:
Hi @huyoule , you are right, would you like to submit a PR to fix it?

huyoule:
> Hi @huyoule , you are right, would you like to submit a PR to fix it?

Do you have a plan to fix it?

huyoule:
> Hi @huyoule , you are right, would you like to submit a PR to fix it?

when will the next release version?
If your plan is too late, then let me do it

yx9o:
Hi @huyoule , you fixed it, I didn't see any PR related to it.

humkum:
@huyoule Hello, excuse me and when would you like to fix this issue?

huyoule:
> @huyoule Hello, excuse me and when would you like to fix this issue?

i'll commit it today

huyoule:
> Hi @huyoule , you fixed it, I didn't see any PR related to it.

please make me to be a collaborator

PR #8709, opened 2024-09-18
#8509 lmq offset sync master to slave 

mongolhoo:
### #https://github.com/apache/rocketmq/issues/8509

Fixes #issue_id


lizhanhui:
This pull request does fix something that is broken. However, this is a design error to replicate bits here and there from one node to the other.  

WAL should the the ONLY data structure that is replicated among nodes and it contains source of truth for everything, including offset advancement of each consumer that is managed in the sever side.

fuyou001 (changes requested):
CR need

domyway:
> CR need

How is it going

PR #8713, opened 2024-09-19
[ISSUE #8712] Set namesrvAddrChoosed to null if choosed addr is not exist.

### LINKED ISSUE #8712: [Enhancement] Ensure that when selected Namesrv addr is no longer in use, a new channel will be created.

### Summary

Ensure that when the currently selected Name Server address by the client is no longer in use, a new channel connection will be created.

### Motivation

Closing channel fail may acuse client using old addr 

set choosedNamesrvAddr  to null when addr is no longer in use

null

### Additional Context

_No response_

qianye1001:
…ient is no longer in use, a new channel connection will be created.

Fixes #8712


fuyou001 (changes requested):
CR need

fuyou001 (changes requested):
CR need

fuyou001 (approved):
LGTM

lizhimins (approved):

PR #8727, opened 2024-09-21
[ISSUE #8780] Implement asynchronous storage of ack/ck messages in pop consume to enhance performance

### LINKED ISSUE #8780: [Enhancement] Implement asynchronous storage of ack/ck messages in pop consume to enhance performance

### Summary

Implement asynchronous storage of ack/ck messages in pop consume to enhance performance 

### Motivation

When the PopBufferMerge of the offset cache is compromised, the ck/ack messages consumed by Pop need to be flushed to disk. Currently, both the Pop and Ack threads are synchronously blocking when storing ck/ack messages, waiting for the flush to complete before processing the results and returning to the client. Moreover, the Brokers in the messaging product line utilize synchronous flushing methods. Therefore, when the PopBufferMerge of the offset cache is compromised, a large number of Pop and Ack threads will be waiting during the flush. For example, assuming there are only three ack threads, if all three threads are blocked waiting for the flush, the ack threads cannot fully utilize the CPU for request processing, leading to ack queuing. This queuing and failure of ack will result in message retry write amplification, creating a vicious cycle.

Thus, this proposal aims to optimize performance through asynchronous processing. 

The core idea is that once ck/ack is written to the page cache, it can immediately return, allowing the Pop and Ack threads to handle the next pop/ack request without synchronously blocking on the flush. After the flush is successful, the results can be handled asynchronously, thus enabling Pop/Ack threads to fully utilize the CPU for request processing

Continuously adjusting the number of threads

### Additional Context

No

RongtongJin:

Fixes #8780

### Brief Description

Implement asynchronous storage of ack/ck messages in pop consume to enhance performance

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java
@@ -137,7 +137,7 @@ public class EscapeBridge {
             brokerNameToSend = mqSelected.getBrokerName();
             if (this.brokerController.getBrokerConfig().getBrokerName().equals(brokerNameToSend)) {
                 LOG.warn("putMessageToRemoteBroker failed, remote broker not found. Topic: {}, MsgId: {}, Broker: {}",
-                        messageExt.getTopic(), messageExt.getMsgId(), brokerNameToSend);
+                    messageExt.getTopic(), messageExt.getMsgId(), brokerNameToSend);
                 return null;
             }
         } else {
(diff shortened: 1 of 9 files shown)


broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java
```diff
@@ -264,6 +262,44 @@ public PutMessageResult putMessageToSpecificQueue(MessageExtBrokerInner messageE
         }
     }
 
+    public CompletableFuture<PutMessageResult> asyncPutMessageToSpecificQueue(MessageExtBrokerInner messageExt) {
+        BrokerController masterBroker = this.brokerController.peekMasterBroker();
+        if (masterBroker != null) {
```
fuyou001, line 267: There is a lot of code duplication with putMessageToSpecificQueue, which can be refactored so that both methods share a single implementation

fuyou001 (approved):
LGTM

guyinyou (approved):
lgtm

GenerousMan (approved):
LGTM

ShannonDing (approved):
LGTM

PR #8739, opened 2024-09-24
[Enhancement] Remove kvCQ batch write

### LINKED ISSUE #8698: [Bug] rocksdbcq batch dispatch should be removed

all

### RocketMQ version

all

### JDK Version

all

### Describe the Bug

RocksDBConsumeQueueStore 中在处理dispatchRequest时候会有一个攒批的动作，这里的初衷是为了提高效率，但是实际上并不会起到什么作用，因为rocksdb首先写入的memtable，这是纯内存操作，不会有多少性能损耗。而且这里还有导致一个bug，就是消息比较少的情况，攒批可能会比较久，这样会导致消息消费延迟，严重的话会导致最后15条消息一直消费不到。

In RocksDBConsumeQueueStore, there is a batching action when handling dispatchRequest. The original intention here is to improve efficiency, but in reality, it doesn't have much effect because RocksDB first writes to the memtable, which is a pure memory operation and doesn't incur much performance overhead. Additionally, this can lead to a bug where, in cases of fewer messages, batching might take a long time, resulting in message consumption delays. In severe cases, the last 15 messages might never be consumed.

### Steps to Reproduce

对于攒批我们一般的处理方式是设置超时时间，当时间超过最大超市时间则强制刷新，但是putMessagePositionInfoWrapper没有专门的线程去定时刷新，而是由dispatchRequest触发的，当没有消息的时候，设置超时时间也没有用，所以我们在这里应该直接将这个batch去掉

For batching, our usual approach is to set a timeout, forcing a flush when the time exceeds the maximum timeout. However, putMessagePositionInfoWrapper does not have a dedicated thread for periodic flushing; instead, it is triggered by dispatchRequest. When there are no messages, setting a timeout is ineffective. Therefore, we should remove the batch directly in this case.

nothing

### What Did You See Instead?

nothing

### Additional Context

nothing

LetLetMe:

Fixes #8698

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
@@ -55,7 +55,7 @@ public class RocksDBConsumeQueueStore extends AbstractConsumeQueueStore {
     public static final byte CTRL_1 = '\u0001';
     public static final byte CTRL_2 = '\u0002';
 
-    private static final int BATCH_SIZE = 16;
+    private static final int BATCH_SIZE = 1;
     public static final int MAX_KEY_LEN = 300;
 
     private final ScheduledExecutorService scheduledExecutorService;
@@ -168,8 +168,6 @@ public class RocksDBConsumeQueueStore extends AbstractConsumeQueueStore {
     public void putMessagePositionInfoWrapper(DispatchRequest request) throws RocksDBException {
         if (request != null) {
             this.bufferDRList.add(request);
-        }
-        if (request == null || !enableBatchWriteKvCq || this.bufferDRList.size() >= BATCH_SIZE) {
             putMessagePosition();
         }
     }


store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
```diff
@@ -55,7 +55,7 @@ public class RocksDBConsumeQueueStore extends AbstractConsumeQueueStore {
     public static final byte CTRL_1 = '\u0001';
     public static final byte CTRL_2 = '\u0002';
 
-    private static final int BATCH_SIZE = 16;
+    private static final int BATCH_SIZE = 1;
```
fuyou001, line 58: The impact of performance？

LetLetMe: This cannot be avoided; we must do it this way even if it impacts performance

LetLetMe:
In order to minimize changes, we have opted to set the batch_size to 1 this time. However, this solution is not perfect, and we will need to use a more rigorous approach in the future to completely resolve this issue.

为了尽量少的变更，我们此次采用了batch_size改为1的方式，但这个方案不完美，后续还是要用更严谨的方式，彻底修复这个问题

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueStore.java
RongtongJin: 可以考虑的小建议：可以通过当前DispatchRequest中commitLogOffset与commitLogMaxOffset来估算是否有剩余需要dispatch的，如果有则攒批加速（但赞批要有一个保底时间），否则及时分发。

LetLetMe:
To avoid changing the default logic of the currently deployed cluster, we will make the batch_size configurable and consider a complete optimization later. @RongtongJin

为了不改变现有已部署集群的默认逻辑，我们把batch_size改为可配置，后续在考虑彻底优化 @RongtongJin

LetLetMe:
image
image
I made some modifications to the tool, and the actual effect of the tool is shown in the image.
修改了下工具，工具实际效果如图所示

lizhimins (approved):

PR #8766, opened 2024-09-27
[ISSUE #8765] fix low performance of delay message when enable rocksdb consume queue

### LINKED ISSUE #8765: [Bug] Bad performace of delay message in rocksdb consumequeue

Ubuntu 24.04

### RocketMQ version

5.3.1

### JDK Version

1.8

### Describe the Bug

Bad performace of delay message in rocksdb consumequeue

### Steps to Reproduce

1. set `storeType=defaultRocksDB` in broker.conf
2. send delay messages with delaylever=2 using `cd benchmark;sh producer.sh -n localhost:9876 -t testTopic -d true -e 2`
3. watch delay message count using `mqadmin topicstatus -n localhost:9876  -t SCHEDULE_TOPIC_XXXX`
4. watch delay message deliver count using `cat ${storeRootPath}/config/delayOffset.json`

delay message are deliverd quick.

### What Did You See Instead?

delay message are deliverd slow.

### Additional Context

_No response_

yuz10:

Fixes #8765


lizhanhui:
@yuz10 Is there profiling metrics verifying that prefetch actually improves perf?

yuz10:
> @yuz10 Is there profiling metrics verifying that prefetch actually improves perf?

The performance loss is not related to prefetching. The schedule message deliver speed is 160/s because every time the iterator only returns 16 messages, and the deliver thread will sleep 100ms

yuz10:
@lizhanhui I found no difference between batch and single get key from rocksdb. I will remove prefetch code.
Batch:
QueryCQ iter 10489877 cost 20527
QueryCQ iter 10489877 cost 19496
QueryCQ iter 10489877 cost 19395

Single:
QueryCQ iter 10489877 cost 20313
QueryCQ iter 10489877 cost 19196
QueryCQ iter 10489877 cost 18945

lizhanhui:
@yuz10 Got your update and review it tomorrow.

lizhanhui:
> The performance loss is not related to prefetching. The schedule message deliver speed is 160/s because every time the iterator only returns 16 messages, and the deliver thread will sleep 100ms after iterate finish. See org.apache.rocketmq.broker.schedule.ScheduleMessageService.DeliverDelayedMessageTimerTask#executeOnTimeUp

1. The original implementation uses one-shot(at most 16 results) multi-get to simulate iterator; The outcome iterator fails to return all results, thus, does not fit well for the mentioned use case;
2. You change is to use lazy single get to iterate; and use potential pre-fetch to accelerate;
3. A third option is to directly wrap RocksIterator with prefix;

It would be best to make further comparisons in terms of performance(why multi-get at present), code maintenance, ...
After all pros and cons are clarified, we may finalize this pull request.

Another issue is option 2, aka, this pull request, changes original behavior. We need to verify the change does not impact semantics of upper layer code bases.

yuz10:
> 1. The original implementation uses one-shot(at most 16 results) multi-get to simulate iterator; The outcome iterator fails to return all results, thus, does not fit well for the mentioned use case;
> 2. You change is to use lazy single get to iterate; and use potential pre-fetch to accelerate;
> 3. A third option is to directly wrap RocksIterator with prefix;
> 
> It would be best to make further comparisons in terms of performance(why multi-get at present), code maintenance, ... After all pros and cons are clarified, we may finalize this pull request.
> 
> Another issue is option 2, aka, this pull request, changes original behavior. We need to verify the change does not impact semantics of upper layer code bases.

I did not compare the performance of RocksIterator with current solution, It can be optimized later, the current solution just deals with the issue of delay message. Another solustion is not to sleep 100ms after each iteration.
As for the behavior, the default ConsumeQueue only iters one file, the RocksDBConsumeQueue only iters at most 16 items. so I think the behavior is not defined about how many items the iteration returns. and will not impact upper layer code.

lizhanhui (approved):
Looks good for this pull request. We shall investigate and compare RocksDB Iterator and simulated one soon.

lizhanhui:
> @lizhanhui I found no difference between batch and single get key from rocksdb. I will remove prefetch code. Batch: QueryCQ iter 10489877 cost 20527 QueryCQ iter 10489877 cost 19496 QueryCQ iter 10489877 cost 19395
> 
> Single: QueryCQ iter 10489877 cost 20313 QueryCQ iter 10489877 cost 19196 QueryCQ iter 10489877 cost 18945

Take a look at this blog, MultiGet has quite a few optimizations available, we shall have two options to investigate:
1, Simulated manual iterator with multi-get in the background, to prefetch accordingly;
2, Use RocksDB embedded Iterator;
https://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html

PR #8807, opened 2024-10-10
[ISSUE #8806] Fix autoBatch bug when connecting multiple RocketMQ clusters.

### LINKED ISSUE #8806: [Bug] AutoBatch bug when connecting multiple RocketMQ clusters.

macos 11.7.10

### RocketMQ version

5.3.0

### JDK Version

jdk 1.8

### Describe the Bug

If multiple RocketMQ clusters are connected and autoBatch is enabled, messages may be sent to the wrong cluster if multiple clusters happen to have the same theme.

### Steps to Reproduce

If I connect to multiple RocketMQ clusters on the same client and enable autoBatch, the topic name happens to be the same.
image

Messages may be sent to the wrong cluster
image

The message should be sent to the correct cluster

### What Did You See Instead?

When creating a Default MQProducer on the client side, an internal Producer Accumulator is initialized
Put the information into the accumulator table map, where key is the client ID and value is the created Producer Accumulator.
When initializing, it will first determine whether the Producer Accumulator has been created based on the client ID. But at this time, the user has not had the opportunity to set InstanceName or UnitName. So at this point, the client IDs within the same service are all the same.
So currently, multiple Default MQProducers created on the same client internally use the same produceAccumulator.
If the topic names of different clusters happen to be the same, it will be sent to the wrong cluster.

### Additional Context

_No response_

luozongle01:

Fixes #8806

### Brief Description
1. Delay the creation of the ProducerAccumulator until startup. Since instanceName will change at startup, and unitName should already be set at this time.
2. Due to delaying the ProducerAccumulator to create, in order to allow users to set the parameters of batchMaxDelayMs, batchMaxbytes, and totaled BatchMaxbytes in the Producer Accumulator before executing the start method (Because I am worried that some users may set these parameters before executing start), three properties have been added to the DefaultMQProducer. After the ProducerAccumulator is initialized, it is assigned to the ProducerAccumulator.

I have tested locally and confirmed that it will not be sent to other RocketMQ clusters, and the parameter settings can take effect.
image
image

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -250,6 +250,8 @@ public class DefaultMQProducerImpl implements MQProducerInner {
 
                 this.mQClientFactory = MQClientManager.getInstance().getOrCreateMQClientInstance(this.defaultMQProducer, rpcHook);
 
+                defaultMQProducer.initProduceAccumulator();
+
                 boolean registerOK = mQClientFactory.registerProducer(this.defaultMQProducer.getProducerGroup(), this);
                 if (!registerOK) {
                     this.serviceState = ServiceState.CREATE_JUST;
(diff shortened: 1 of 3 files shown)


guyinyou (approved):
lgtm

RongtongJin (approved):

Issue #8829, opened 2024-10-17
[Feature] LMQ Optimization: Revamp on top of RocksDB


lizhanhui:
### Is Your Feature Request Related to a Problem?

LMQ is supposed to be light-weight in terms of resource cost and lifecycle. Current implementation is still based on file queue and does not scale as expected

### Describe the Solution You'd Like

Provide a key-value based implementation

### Describe Alternatives You've Considered

In-place optimization is not a viable option in the long run.

PR #8830, opened 2024-10-17
fix: atomic flush incorrect use and clean up code

lizhanhui:

to #8829

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java
+++ b/store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java
@@ -16,7 +16,7 @@
  */
 package org.apache.rocketmq.store.rocksdb;
 
-import org.apache.rocketmq.common.config.ConfigRocksDBStorage;
+import org.apache.rocketmq.common.config.ConfigHelper;
 import org.apache.rocketmq.store.MessageStore;
 import org.rocksdb.BlockBasedTableConfig;
 import org.rocksdb.BloomFilter;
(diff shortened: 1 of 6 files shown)


store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java
```diff
@@ -71,7 +71,7 @@ public static ColumnFamilyOptions createCQCFOptions(final MessageStore messageSt
                 setTableFormatConfig(blockBasedTableConfig).
                 setMemTableConfig(new SkipListMemTableConfig()).
                 setCompressionType(CompressionType.LZ4_COMPRESSION).
-                setBottommostCompressionType(CompressionType.ZSTD_COMPRESSION).
+                setBottommostCompressionType(CompressionType.LZ4_COMPRESSION).
```
tianliuliu, line 74: 这里为啥修改为LZ4，相对于iops，cpu更多是瓶颈吗？

RongtongJin: Does modifying the compression type impact in-place upgrade compatibility?

RongtongJin (approved):

store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java
fuyou001, line 74: The default parameters remain unchanged
 may be parameterized is better

tianliuliu (approved):

fuyou001 (approved):

store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java
lizhanhui: Column Family for CQ uses Universal Compaction,  implying each SST will compact to last level. Unlike level compaction where ZSTD makes sense because it achieves better compression ratio at cost of more CPU usage.

LZ4 achieves more balanced CPU/compression ratio, thus is a better fit for  universal compaction.

lizhanhui: OK, let's parameterize this option; However, I prefer to let it default to an option that makes more sense instead of keeping it unchanged. Will this preference make sense for you? @fuyou001

fuyou001: We have used ZSTD_COMPRESSION

lizhanhui: All right, let's make it configurable in the next PR

PR #8838, opened 2024-10-17
[ISSUE #8837] Remove netty codec threads for better performance

### LINKED ISSUE #8837: [Enhancement] Remove netty codec threads for better performance

### Summary

In the current implementation, a codec thread pool is set up for Netty, which is used to perform tasks such as encoding/decoding, and statistics.

We found that removing the codec thread pool and directly assigning these computational tasks to the NIO selector threads results in performance improvement.

Since these tasks do not involve IO operations, they will not block the NIO selector threads.

### Motivation

Our stress test results show that removing the codec thread pool can achieve higher TPS (5% increase) and also reduce CPU usage (4% reduction).
#### [TPS Comparison] Before:
TPS Before
#### [TPS Comparison] After:
TPS After

----

#### [CPU Comparison] Before:
image
#### [CPU Comparison] After:
image

Remove the codec thread pool `NettyRemotingServer #defaultEventExecutorGroup`

No

### Additional Context

_No response_

f1amingo:

Fixes #8837

### Brief Description
In the current implementation, a codec thread pool is set up for Netty, which is used to perform tasks such as encoding/decoding, and statistics.

We found that removing the codec thread pool and directly assigning these computational tasks to the NIO selector threads results in performance improvement.

Since these tasks do not involve IO operations, they will not block the NIO selector threads.

Our stress test results show that removing the codec thread pool can achieve higher TPS (5% increase) and also reduce CPU usage (4% reduction).

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/remoting/MultiProtocolRemotingServer.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/remoting/MultiProtocolRemotingServer.java
@@ -78,8 +78,8 @@ public class MultiProtocolRemotingServer extends NettyRemotingServer {
     @Override
     protected ChannelPipeline configChannel(SocketChannel ch) {
         return ch.pipeline()
-            .addLast(this.getDefaultEventExecutorGroup(), HANDSHAKE_HANDLER_NAME, new HandshakeHandler())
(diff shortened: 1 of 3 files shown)


f1amingo:
A switch has already been added.
Close issue

PR #8842, opened 2024-10-18
Support LMQ dispatch in case if Consume Queue Store is RocksDB-based

lizhanhui:

to #8829


RongtongJin:
@LetLetMe PTAL～

lizhanhui:
@redlsz @leizhiyuan  Please also take a look

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueOffsetTable.java
```diff
@@ -111,69 +117,147 @@ public class RocksDBConsumeQueueOffsetTable {
...
+
+            int keyBufferCapacity = 256;
+            iterator.seekToFirst();
+            ByteBuffer keyBuffer = ByteBuffer.allocateDirect(keyBufferCapacity);
+            ByteBuffer valueBuffer = ByteBuffer.allocateDirect(16);
+            while (iterator.isValid()) {
+                // parse key buffer according to key layout
+                keyBuffer.clear(); // clear position and limit before reuse
+                int total = iterator.key(keyBuffer);
+                if (total > keyBufferCapacity) {
```
fuyou001, line 170: The key length of CQ needs to be reconsidered or should have the same restrictions as the file version on both sides

topic v2  topic's lenght  may be greater 300  


cqKey  lenght only 300 Byte
org.apache.rocketmq.store.queue.RocksDBConsumeQueueTable#getCQByteBufferPair

store/src/main/java/org/apache/rocketmq/store/MessageExtEncoder.java
```diff
@@ -175,11 +176,11 @@ public PutMessageResult encodeWithoutProperties(MessageExtBrokerInner msgInner)
     public PutMessageResult encode(MessageExtBrokerInner msgInner) {
         this.byteBuf.clear();
 
-        if (CommitLog.isMultiDispatchMsg(messageStoreConfig, msgInner)) {
+        if (messageStoreConfig.isEnableLmq() && MixAll.topicAllowsLMQ(msgInner.getTopic())) {
```
RongtongJin, line 179: It should use `msgInner.needDispatchLMQ()` instead of `MixAll.topicAllowsLMQ(msgInner.getTopic())`, otherwise non-lmq messages will enter this branch and cause issues during writing.

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueOffsetTable.java
lizhanhui: @fuyou001 This method does not impose further restriction on topic length. 
`
int total = iterator.key(keyBuffer);
if (total > keyBufferCapacity) {
    keyBufferCapacity = total;
    PlatformDependent.freeDirectBuffer(keyBuffer);
    keyBuffer = ByteBuffer.allocateDirect(keyBufferCapacity);
    continue;
}
`
This code blocks roughly means, try to read key into the pre-allocated direct buffer, if the key length is greater than the pre-allocated key buffer capacity, free the pre-allocated direct buffer(by default, 256 bytes) and re-allocate a larger one and then retry again.

store/src/main/java/org/apache/rocketmq/store/MessageExtEncoder.java
lizhanhui: Exactly

store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueOffsetTable.java
fuyou001: I get it

fuyou001: I get it

fuyou001 (approved):
LGTM

tianliuliu:
LGTM

Issue #8879, opened 2024-10-30
[Enhancement] reduce seach key times in method,to import performance


ChineseTony:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

 reduce seach key times in method,to import performance

### Motivation

 reduce seach key times in method,to import performance

### Describe the Solution You'd Like

 reduce seach key times in method,to import performance

### Describe Alternatives You've Considered

 reduce seach key times in method,to import performance

PR #8880, opened 2024-10-30
[ISSUE #8879] reduce seach key times in method,to import performance

ChineseTony:

Fixes #issue_id

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RemotingCommand.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RemotingCommand.java
@@ -360,13 +360,15 @@ public class RemotingCommand {
     }
 
     private boolean isFieldNullable(Field field) {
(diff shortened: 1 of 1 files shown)


qianye1001:
If refactoring this part of the code, it might be simpler to use ConcurrentHashMap instead of HashMap.

ChineseTony:
ok, I will chage this code 

```java

   private boolean isFieldNullable(Field field) {
        return ConcurrentHashMapUtils.computeIfAbsent(NULLABLE_FIELD_CACHE, field, input -> {
            Annotation annotation = input.getAnnotation(CFNotNull.class);
            return annotation == null;
        });
    }

```

ChineseTony:
i think is't using HashMap ok ,it just lock in put value, but in more times  just get value

RongtongJin (commented):
Could you provide specific performance improvement data using tests? It would be better to use tests to also verify correctness.

ChineseTony:
ok,thanks ,but isFieldNullable function can use temp variable to save and reture the value ,Don't need search key twice

PR #8891, opened 2024-11-05
[ISSUE #8829] Keep data version while reload and XXXConfigManagerV2 turns off sync

lizhanhui:
Two minor changes are included:
1, no need to update data version while reload; 
2, config-storage fdatasync on write proves to be too expensive, turn it off

To #8829

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/config/v2/SubscriptionGroupManagerV2.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/config/v2/SubscriptionGroupManagerV2.java
@@ -72,7 +72,7 @@ public class SubscriptionGroupManagerV2 extends SubscriptionGroupManager {
             while (iterator.isValid()) {
                 SubscriptionGroupConfig subscriptionGroupConfig = parseSubscription(iterator.key(), iterator.value());
                 if (null != subscriptionGroupConfig) {
-                    super.updateSubscriptionGroupConfigWithoutPersist(subscriptionGroupConfig);
+                    super.putSubscriptionGroupConfig(subscriptionGroupConfig);
                 }
                 iterator.next();
             }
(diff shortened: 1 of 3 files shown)


broker/src/main/java/org/apache/rocketmq/broker/config/v2/SubscriptionGroupManagerV2.java
```diff
@@ -72,7 +72,7 @@ private boolean loadSubscriptions() {
             while (iterator.isValid()) {
                 SubscriptionGroupConfig subscriptionGroupConfig = parseSubscription(iterator.key(), iterator.value());
                 if (null != subscriptionGroupConfig) {
-                    super.updateSubscriptionGroupConfigWithoutPersist(subscriptionGroupConfig);
+                    super.putSubscriptionGroupConfig(subscriptionGroupConfig);
```
tianliuliu, line 75: image
after update then print log?

broker/src/main/java/org/apache/rocketmq/broker/config/v2/TopicConfigManagerV2.java
```diff
@@ -76,7 +76,7 @@ private boolean loadTopicConfig() {
                 byte[] value = iterator.value();
                 TopicConfig topicConfig = parseTopicConfig(key, value);
                 if (null != topicConfig) {
-                    super.updateSingleTopicConfigWithoutPersist(topicConfig);
+                    super.putTopicConfig(topicConfig);
```
tianliuliu, line 79: same as above

broker/src/main/java/org/apache/rocketmq/broker/config/v2/SubscriptionGroupManagerV2.java
lizhanhui: We are actually loading existing config from persistence kv store, not updating.

Given there may be a lot of items while reboot loading, log this may flood the log files unnecessarily.

broker/src/main/java/org/apache/rocketmq/broker/config/v2/TopicConfigManagerV2.java
lizhanhui: The same rationale applies.

PR #8897, opened 2024-11-06
[#8859] Fix duplicate mapped file in mutil commitlog store path mode.

### LINKED ISSUE #8859: [Bug] Bug title

centos8
local ssd

### RocketMQ version

5.1.4

### JDK Version

1.8

### Describe the Bug

The configured directories are:
/data, /data2, /data3, /data4.

After a long period of pressure testing, we found that in different directories, there are commitlog files with the same name. One of these commitlog files appears to be empty, but the filenames are identical. Is this a bug, and is there any risk associated with it?

If the process restarts, and duplicate commitlog files are loaded, could this also pose a risk?

As for code analysis, we're unsure if it's correct:

In MultiPathMappedFileQueue.java, this file is used for creating files across multiple paths. In the tryCreateMappedFile implementation, paths where the disk usage is high (85%) will be filtered out, and new commitlog files will not be created in those paths.

However, it is possible that during a previous create map file operation, the path was already added to nextNextFilePath for pre-creation. If the paths change in the next cycle, a new commitlog will be created in a different path as the actual commitlog, while the previously pre-created file will be unused and essentially a "fake" file.

### Steps to Reproduce

set storePathCommitLog=/data,/data1,/data2,/data3
and just continue produce

commitlog will not be the same name

### What Did You See Instead?

none

### Additional Context

_No response_

hexueyuan:

Fixes #8859 

### Brief Description

I changed the key of the requestTable in the AllocateMappedFileService class from the FilePath to the FileName(this must be the only one), and then added a getCachedMappedFilePath method to provide to the MultiPathMappedFileQueue to check whether the MappedFile of a specified offset MappedFile create request has been submitted, and if so, obtain the cached FilePath, otherwise a new path is generated based on the previous rules.

I make a unit test to recurrence this case in testUniqueNextNextMappedFile().

In the previous implementation, the results obtained were as shown in the picture below, the MappedFile 00000000000000001024 repeatedly appear in directory a and directory b.
9daf3e5014d7447a1328fbb5da7a1382

This problem was solved in the new implementation and the result is as follows:
45446178a8a361c316316b62997e64eb

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/AllocateMappedFileService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/AllocateMappedFileService.java
@@ -39,7 +39,7 @@ import org.apache.rocketmq.store.logfile.MappedFile;
 public class AllocateMappedFileService extends ServiceThread {
     private static final Logger log = LoggerFactory.getLogger(LoggerName.STORE_LOGGER_NAME);
     private static int waitTimeOut = 1000 * 5;
-    private ConcurrentMap<String, AllocateRequest> requestTable =
+    private ConcurrentMap<String/*FileName*/, AllocateRequest> requestTable =
         new ConcurrentHashMap<>();
     private PriorityBlockingQueue<AllocateRequest> requestQueue =
         new PriorityBlockingQueue<>();
@@ -48,6 +48,11 @@ public class AllocateMappedFileService extends ServiceThread {
 
     public AllocateMappedFileService(DefaultMessageStore messageStore) {
         this.messageStore = messageStore;
+    }
+
+    public String getCachedMappedFilePath(long createOffset) {
+        AllocateRequest req = requestTable.get(UtilAll.offset2FileName(createOffset));
+        return req == null ? null : req.getFilePath();
     }
 
     public MappedFile putRequestAndReturnMappedFile(String nextFilePath, String nextNextFilePath, int fileSize) {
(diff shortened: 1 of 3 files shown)


store/src/main/java/org/apache/rocketmq/store/AllocateMappedFileService.java
```diff
@@ -76,13 +82,14 @@ public MappedFile putRequestAndReturnMappedFile(String nextFilePath, String next
             canSubmitRequests--;
         }
 
+        String nextNextFileName = nextNextFilePath.substring(nextNextFilePath.lastIndexOf("/") + 1);
```
RongtongJin, line 85: It is necessary to use File.separator, as the file path separator on other OS may not be '/'.

hexueyuan: Okay, fixed now.

RongtongJin: The UTs also need to be modified accordingly.

hexueyuan: Sorry, it was my oversight. I have modified it as required. Please review it again. Thank you.

RongtongJin: > The UTs also need to be modified accordingly.

By the way, is there any compatibility issue here? From the code, it seems that there isn't.

RongtongJin: > Sorry, it was my oversight. I have modified it as required. Please review it again. Thank you.

By the way, is there any compatibility issue here? From the code, it seems that there isn't.

hexueyuan: Do you mean file path separator in test case? I also have this doubts but the other use cases I refer to in this file are like this.

hexueyuan: Sorry, there are indeed some compatibility issues in my UTs,  I have located the problem in the workflows. I will provide a new patch after completing the test in my repository later.

hexueyuan: > a new patch after completing the test in my repository 
Fixed compatibility issue.

hexueyuan: It seems like some errors blocked the workflow checks, but I don't have permission to stop or rerun it.

hexueyuan:
This pr may need someone review it again, thanks ~

RongtongJin (approved):

hexueyuan:
@RongtongJin Hello and sorry to bother you, this is my first contribution, is there any necessary work should I do before this pr merged? Or just wait is OK?

PR #8923, opened 2024-11-14
[ISSUE #8912] Rebalancing logic shift in POP consumption mode

### LINKED ISSUE #8912: [Feature] A shift in the rebalancing logic

In the pull/push consumption mode, the rebalancing logic is carried out on the client side.

However, after switching to the pop consumption mode, all topics on the same client will be rebalanced on the server side, which brings unnecessary losses. Therefore, I want to transfer the rebalancing logic to the client side, which will have a great impact on resource utilization

transfer the rebalancing logic to the client side

null

### Additional Context

_No response_

3424672656:

Fixes #8912 

### Brief Description

https://shimo.im/docs/Ee32mZB85xcK87A2


3424672656:
> Why not reduce the computational load on the server side by adding a cache to the broker, instead of adopting a client-side computation approach?

Do you mean the results of the distribution cached on the broker side? When POP consumption mode is enabled, I think there are also a large number of network requests, and each topic subscribed by the corresponding client is a network request, which does not seem to solve the problem completely. Transferring to the client side can unify the rebalancing logic. Because it doesn't seem easy to completely isolate POP and PULL rebalancing logic in the original version, and the initial client execution is ideally local rebalancing

3424672656:
> Why not reduce the computational load on the server side by adding a cache to the broker, instead of adopting a client-side computation approach?

And the server cache should also consider the problem of sensing and updating cache in time, and the delay of cache update and rebalancing at the same time

PR #8856, opened 2024-10-23
[ISSUE #8829] feat: provide ConfigManagerV2 to make best uses of RocksDB

lizhanhui:
### Which Issue(s) This PR Fixes

To #8829

PR #8939, opened 2024-11-16
[ISSUE #8829] Support for Persisting LMQ Consumer Offsets in Config V1 Using RocksDB

### LINKED ISSUE #8829: [Feature] LMQ Optimization: Revamp on top of RocksDB

RongtongJin:

Fixes #8829

In the current v1, lmq consumer offsets are only stored in memory, which means they will be lost if the broker restarts or crashes. To address this, #8856 @Lizhanhui developed v2 to enable storing lmq consumer offsets in RocksDB. However, upgrading from v1 to v2 with backward compatibility is not feasible. Therefore, it's crucial to introduce support for persisting lmq consumer offsets in v1.

### Brief Description

To streamline the implementation, we will not use a separate class for lmq consumer offset storage. Instead, we will leverage the existing RocksDBConsumerOffsetManager. Under the v1 metadata configuration, this manager will handle both lmq and non-lmq consumer offsets collectively.


fuyou001 (approved):

broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -352,7 +351,7 @@ public BrokerController(
         } else if (this.messageStoreConfig.isEnableRocksDBStore()) {
             this.topicConfigManager = messageStoreConfig.isEnableLmq() ? new RocksDBLmqTopicConfigManager(this) : new RocksDBTopicConfigManager(this);
             this.subscriptionGroupManager = messageStoreConfig.isEnableLmq() ? new RocksDBLmqSubscriptionGroupManager(this) : new RocksDBSubscriptionGroupManager(this);
-            this.consumerOffsetManager = messageStoreConfig.isEnableLmq() ? new RocksDBLmqConsumerOffsetManager(this) : new RocksDBConsumerOffsetManager(this);
+            this.consumerOffsetManager = new RocksDBConsumerOffsetManager(this);
```
DongyuanPan, line 354: Maybe it's better to make a feature switch

lizhimins (approved):

PR #8958, opened 2024-11-20
[ISSUE #8957] Remove excess traffic and fix cache inconsistencies

### LINKED ISSUE #8686: [Bug] Network jitter encountered while rebalancing results in repeated consumption or errors in pop mode

centos

### RocketMQ version

develop

### JDK Version

1.8

### Describe the Bug

When pop consumption mode is enabled for topic-a, the client sets isRebanlance=false
293b0204-1fda-4d5f-80f1-14042da21dc2
1501d98d-a6d2-4ca3-86dd-aedc51d70e3e
39ae3686-1817-4c6a-855b-20ab8f7723b6
5f8f2576-dd0b-4b52-92d4-b4dbdb78aecd
Client rebalancing does not make pop consumption pattern judgment

### Steps to Reproduce

The setMessageRequestMode interface of MQAdmin is called, the broker is network isolated from the client, the client is set isRebanlance=false, the network is restored after some time, and it is found that there is no pop consumption on the topic

Topics are consumed in pop mode

### What Did You See Instead?

Topics are not consumed in pop mode

### Additional Context

_No response_

### LINKED ISSUE #8957: [Enhancement] Avoid unnecessary network communication when rebalance in POP

### Summary

Set clientRebalance = false to POP consumption, which means that all topics need to be rebalanced by the broker, so there is no need to obtain a consumption pattern for each topic, which will cause some repeated consumption problems

### Motivation

#8686

When setting ClientRebalance, go directly to the broker to get the result without network communication

Save money while avoiding cache inconsistencies caused by issues such as broker downtime

### Additional Context

_No response_

3424672656:

Fixes #8957 
Fixes #8686

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalanceImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalanceImpl.java
@@ -241,7 +241,7 @@ public abstract class RebalanceImpl {
             for (final Map.Entry<String, SubscriptionData> entry : subTable.entrySet()) {
                 final String topic = entry.getKey();
                 try {
-                    if (!clientRebalance(topic) && tryQueryAssignment(topic)) {
+                    if (!clientRebalance(topic)) {
                         boolean result = this.getRebalanceResultFromBroker(topic, isOrder);
                         if (!result) {
                             balanced = false;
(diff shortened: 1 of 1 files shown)


humkum:
Why is this deleted? IMO, only the first startup will make a request to the server，will not bring more network traffic.

3424672656:
> Why is this deleted? IMO, only the first startup will make a request to the server，will not bring more network traffic.

Normally, if rebalance=false is enabled, all topics will be rebalanced on the server as they are added to the topicBrokerRebalance collection, but when a network issue occurs, Some topics will be added topicClientRebalance, and this part of the client will be rebalanced in PULL.

Since it should be carried out on the server side, there is no need for this network communication, which not only brings overhead but also causes cache errors. It is better to remove this part

qianye1001:
LGTM

lizhimins (approved):

Issue #8974, opened 2024-11-22
[Feature] Support recalling of delay message


imzs:
### Is Your Feature Request Related to a Problem?

.

### Describe the Solution You'd Like

As the timer wheel implementation of delay message, we've reserved a cancellation ability: recalling scheduled messages by sending a corresponding delete message. However, we have not exposed an API to the upper layer. Some developers may construct and send a special message to achieve this, but this is not a friendly and sustainable solution because:

- There is no clear API semantics.
- It depends on internal implementation details and may become invalid due to system updates.
- Sending and recalling are probably not in the same context, but the required information needs to be saved and depends on the details at the time of sending. 
- Grpc is not supported.

Therefore, the first step is to provide a user-friendly API to implement this feature.

1. A new rpc request and processor.
2. Client side API definition: `recallMessage(topic, handle)`, recall handle from sending result is a bit like ReceiptHandle, containing information such as ID, broker, and timestamp.
3. Still based on time wheel, only support delay message for now, and perform limited parameter validation in the internal implementation, that is
    - can not accurately determine whether the message truly exists
    - recall may fail when it's too close to the delivery time
4. Regard the recall operation as a PUB operation for authorization.
5. Support Remoting and Grpc protocol, proxy and proxy-less mode.

### Describe Alternatives You've Considered

.

### Additional Context

.

PR #8975, opened 2024-11-22
[ISSUE #8974] Support recalling of delay message

imzs:
[ISSUE #8974] Support recalling of delay message


qianye1001:
Returning an additional handle in all send Response operations introduces overhead. If this feature is not required, there should be an option to disable it.

imzs:
> Returning an additional handle in all send Response operations introduces overhead. If this feature is not required, there should be an option to disable it.

This field is somewhat similar to `transactionId`. As previously mentioned, the API defines a common semantic. In cases where unsupported message types are encountered, it returns null values, which also implies that the current message does not support recall operation, this may be considered as a switch.

RongtongJin (approved):

RongtongJin:
It seems that the newly added integration test are failing.
image

imzs:
> It seems that the newly added integration test are failing. image

The tests passed locally, it maybe timeout issue; I'll try to fix it.

lollipopjin (approved):
LGTM

PR #8981, opened 2024-11-25
[ISSUE #8460] Set default broker name when revive found ack without broker name field

### LINKED ISSUE #8460: [Enhancement] Improve the pop revive process when reading biz messages from a remote broker

### Summary

In [RIP-32](https://github.com/apache/rocketmq/wiki/RIP-32-Slave-Acting-Master-Mode)
we put forward a new proposal, slave acting master mode, as an upgrade of master-slave deployment mode. 
To prevent consumption stagnation of delayed message, transaction message and pop mode, RIP-32 provides two basic abilities:
- write to remote brokers (ck, delay message, op)
- read from remote brokers (biz message)

but it lacks robustness for the pop revive process when reading messages from remote.

### Motivation

chain
1.  As the basis for the mechanism, let's see the definition of lower and upper api
```java
CompletableFuture<PullResult> BrokerOutAPI.pullMessageFromSpecificBrokerAsync
CompletableFuture<Pair<GetMessageStatus, MessageExt>> EscapeBridge.getMessageAsync
```
 PullResult & GetMessageStatus means and only means FOUND or not, but when reading from a remote broker, it's not appropriate to use the same semantics as local store, because it can't deal with unexpected errors such as RPC problems.
The api invoker has no extra information to decide whether to retry or not.

2. Then talk about pop revive process, when we find a message to be revived, read biz message from local or remote and write to retry topic, but rewrite CK if read failed. In the current implementation, there will be no CK reentry because of the issue above, and even if we fix it, in some condition we'll face an endless CK rewrite because it enters into a loop of failed reading result and CK rewrite, so we need an exponential backoff and ending condition.

1. extra information to the call chain of remote message reading
2. an exponential backoff and ending condition of CK rewrite

See Above

### Additional Context

_No response_

lizhimins:
Set default broker name when revive found ack without broker name field

POP CK 新旧格式兼容，为了尽可能的减少消费重复，当没有 brokerName 在 ack 信息里的时候自动补全

Fixes #8460

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.NavigableMap;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Triple;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.metrics.BrokerMetricsManager;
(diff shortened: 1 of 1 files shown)


GenerousMan (approved):
LGTM

PR #8989, opened 2024-11-26
[ISSUE #8988] Support dispatchBehindMilliseconds

### LINKED ISSUE #8988: [Enhancement] support dispatchBehindMilliseconds

### Summary

Add a new concern standard, dispatchBehindMilliseconds, for the time delay of fast index building tasks relative to the latest news. This new metric will next provide units with more direct time delays, helping users better understand and optimize the real-time nature of the system.

### Motivation

While the existing dispatchBehindBytes metric provides information about index build progress, it relies on bytes as a metric, which is not intuitive for understanding actual latency. Especially when the traffic size changes greatly, it is difficult to accurately reflect the true service level only by the number of bytes. The introduction of dispatchBehindMilliseconds can provide a unified and easy-to-understand time unit to evaluate the speed of index construction and help quickly identify potential performance bottlenecks or resource allocation issues.

Define new attention standards: add dispatchBehindMilliseconds to the existing monitoring system, which is used to represent the time difference between the latest message generation and the successful construction of the current index.

Continue using dispatchBehindBytes without making any changes. However, as mentioned earlier, this approach may not fully reflect the actual situation, especially when performance needs to be compared across different environments.
Try to derive the time delay indirectly through other means, such as combining information such as message rate. But doing so is not only complicated, but also difficult to guarantee accuracy.

### Additional Context

_No response_

guyinyou:

Fixes #8988

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -1556,6 +1556,10 @@ public class DefaultMessageStore implements MessageStore {
     public long dispatchBehindBytes() {
         return this.reputMessageService.behind();
     }
+    @Override
+    public long dispatchBehindMilliseconds() {
+        return this.reputMessageService.behindMs();
+    }
 
     public long flushBehindBytes() {
         if (this.messageStoreConfig.isTransientStorePoolEnable()) {
@@ -2793,6 +2797,7 @@ public class DefaultMessageStore implements MessageStore {
     class ReputMessageService extends ServiceThread {
 
         protected volatile long reputFromOffset = 0;
+        protected volatile long currentReputTimestamp = 0;
 
         public long getReputFromOffset() {
             return reputFromOffset;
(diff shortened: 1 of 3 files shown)


GenerousMan (approved):
LGTM

store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
```diff
@@ -2793,6 +2797,7 @@ public void notifyMessageArriveIfNecessary(DispatchRequest dispatchRequest) {
     class ReputMessageService extends ServiceThread {
 
         protected volatile long reputFromOffset = 0;
+        protected volatile long currentReputTimestamp = 0;
```
RongtongJin, line 2800: Should the initial value be set to the current time? If it's set to 0, would it retrieve an unexpectedly large value before being assigned?

guyinyou: it is ok

RongtongJin (approved):

caigy (commented):
`System.currentTimeMillis()` can be affected by changes to the system time, such as those caused by daylight saving time or NTP time server sync. Should we use `System.nanoTime()` instead?

guyinyou:
> `System.currentTimeMillis()` can be affected by changes to the system time, such as those caused by daylight saving time or NTP time server sync. Should we use `System.nanoTime()` instead?

Your concern is correct, but the calculation of dispatchBehindMilliseconds needs to be calculated based on the message storage time stored in the commitlog and cannot be converted to nanoTime.

PR #9048, opened 2024-12-12
[ISSUE #9025] [RIP-73] Pop Consumption Improvement Based on RocksDB

### LINKED ISSUE #9025: [Feature] [RIP‐73] Pop Consumption Improvement Based on RocksDB

Pop consumption is a new consumption model with the advantages of being lightweight, stateless on the client side, and non-exclusive queueing in load balance. However, the current pop implementation has a large codebase, high disk write traffic when the buffer is closed, and more complex state maintenance when the buffer is enabled. Therefore, we propose a new implementation based on rocksdb storage to elegantly solve the compatibility issues between pop retry topic v1 and v2 formats and the probabilistic retry access problem. Other improvements include full asynchrony operation, optimized cache design, accelerated encoding and decoding, and pop message fairness probleam. Performance testing with a read-write ratio of 1:1 in the same environment compared to the current mainline (Version 5.3.1), the new pop implementation saves 28% of the overall cpu usage of the broker when the buffer is closed, and reduces the overall cpu usage by 4.5% when the Buffer is enabled, with the pop part showing a reduction of 17%.

image

As mentioned above, the proposal aims to eliminate the dependency of pop consumption on scheduled messages, reduce the codebase and implementation complexity, and improve performance to some extent.

image

In the new implementation, we use RocksDB to store Pop CheckPoints, and the introduction of the KV model significantly reduces the complexity of state storage. Common operations in Pop consumption have been converted into KV model operations:

- Write CK => Put KV: The Key must include group+ topic + queueId + offset and pop time.
- Write AK => Delete by Key: There is no need to reverse-parse the content of the Value.
- Revive => Prefix Scan using seek to first: Utilize the seek to first feature provided by the KV model to perform a prefix scan.

No

### Additional Context

More details: [RIP-73](https://github.com/apache/rocketmq/wiki/%5BRIP%E2%80%9073%5D-Pop-Consumption-Improvement-Based-on-RocksDB)

lizhimins:

Fixes #9025

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -181,6 +181,15 @@ public class PopConsumerService extends ServiceThread {
         if (!context.isFifo() && result.getNextBeginOffset() > OFFSET_NOT_EXIST) {
             this.brokerController.getConsumerOffsetManager().commitPullOffset(
                 context.getClientHost(), context.getGroupId(), topicId, queueId, result.getNextBeginOffset());
+            long commitOffset = result.getStatus() == GetMessageStatus.FOUND ? offset : result.getNextBeginOffset();
+            if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
+                long minOffset = popConsumerCache.getMinOffsetInCache(context.getGroupId(), topicId, queueId);
+                if (minOffset != OFFSET_NOT_EXIST) {
+                    commitOffset = minOffset;
+                }
+            }
+            this.brokerController.getConsumerOffsetManager().commitOffset(
+                context.getClientHost(), context.getGroupId(), topicId, queueId, commitOffset);
         }
 
         return context;
@@ -308,9 +317,6 @@ public class PopConsumerService extends ServiceThread {
             CompletableFuture.completedFuture(popConsumerContext);
 
         try {
-            GetMessageResult finalGetMessageResult = new GetMessageResult();
-            finalGetMessageResult.setStatus(GetMessageStatus.NO_MESSAGE_IN_QUEUE);
-
             if (!fifo && preferRetry) {
                 if (brokerConfig.isRetrieveMessageFromPopRetryTopicV1()) {
                     getMessageFuture = this.getMessageAsync(getMessageFuture, clientHost, groupId,
@@ -493,7 +499,7 @@ public class PopConsumerService extends ServiceThread {
                         record.setInvisibleTime(record.getInvisibleTime() + backoffInterval);
                         record.setAttemptTimes(record.getAttemptTimes() + 1);
                         failureList.add(record);
-                        log.error("PopConsumerService revive backoff retry, record={}", record);
+                        log.warn("PopConsumerService revive backoff retry, record={}", record);
                     } else {
                         log.error("PopConsumerService drop record, message may be lost, record={}", record);
                     }
@@ -505,9 +511,14 @@ public class PopConsumerService extends ServiceThread {
         this.popConsumerStore.writeRecords(new ArrayList<>(failureList));
         this.popConsumerStore.deleteRecords(consumerRecords);
 
-        log.info("PopConsumerService, key size={}, cache size={}, revive count={}, failure count={}, cost={}ms",
-            popConsumerCache.getCacheKeySize(), popConsumerCache.getCacheSize(), consumerRecords.size(),
-            failureList.size(), stopwatch.elapsed(TimeUnit.MILLISECONDS));
+        if (brokerConfig.isEnablePopBufferMerge()) {
+            log.info("PopConsumerService, key size={}, cache size={}, revive count={}, failure count={}, cost={}ms",
+                popConsumerCache.getCacheKeySize(), popConsumerCache.getCacheSize(), consumerRecords.size(),
+                failureList.size(), stopwatch.elapsed(TimeUnit.MILLISECONDS));
+        } else {
+            log.info("PopConsumerService, revive count={}, failure count={}, cost={}ms",
+                consumerRecords.size(), failureList.size(), stopwatch.elapsed(TimeUnit.MILLISECONDS));
+        }
 
         return consumerRecords.size();
     }
@@ -695,6 +706,7 @@ public class PopConsumerService extends ServiceThread {
                 }
             } catch (Exception e) {
                 log.error("PopConsumerService revive error", e);
+                this.waitForRunning(500);
             }
         }
         this.consumerRunning.set(false);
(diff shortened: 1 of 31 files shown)


broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -0,0 +1,699 @@
...
+            CompletableFuture.completedFuture(popConsumerContext);
+
+        try {
+            GetMessageResult finalGetMessageResult = new GetMessageResult();
+            finalGetMessageResult.setStatus(GetMessageStatus.NO_MESSAGE_IN_QUEUE);
+
+            if (!fifo && preferRetry) {
+                if (brokerConfig.isRetrieveMessageFromPopRetryTopicV1()) {
+                    getMessageFuture = this.getMessageAsync(getMessageFuture, clientHost, groupId,
+                        retryTopicV1, 0, batchSize, filter, PopConsumerRecord.RetryType.RETRY_TOPIC_V1);
```
leizhiyuan, line 317: Here isRetrieveMessageFromPopRetryTopicV1 and isEnableRetryTopicV2 may be established at the same time in compatible scenarios. At this time, when processing getMessageFuture here, will it always use isEnableRetryTopicV2?

broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
```diff
@@ -296,6 +308,76 @@ private void appendAck(final AckMessageRequestHeader requestHeader, final BatchA
         }
     }
 
+    private void appendAckNew(final AckMessageRequestHeader requestHeader, final BatchAck batchAck,
+        final RemotingCommand response, final Channel channel, String brokerName) throws RemotingCommandException {
+
+        if (requestHeader != null && batchAck == null) {
+            String[] extraInfo = ExtraInfoUtil.split(requestHeader.getExtraInfo());
+            brokerName = ExtraInfoUtil.getBrokerName(extraInfo);
```
leizhiyuan, line 316: long startOffset = ExtraInfoUtil.getCkQueueOffset(extraInfo);
            brokerName = ExtraInfoUtil.getBrokerName(extraInfo);

seems no used by any place

broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
```diff
@@ -296,6 +308,76 @@ private void appendAck(final AckMessageRequestHeader requestHeader, final BatchA
...
+                        this.brokerController.getPopConsumerService().ackAsync(
+                            popTime, invisibleTime, groupId, topicId, queueId, offset);
+                    }
+                    ackCount++;
+                }
+
+                this.brokerController.getBrokerStatsManager().incBrokerAckNums(ackCount);
+                this.brokerController.getBrokerStatsManager().incGroupAckNums(groupId, topicId, ackCount);
+            } catch (ConsumeQueueException e) {
+                throw new RemotingCommandException("Failed to get max offset in queue", e);
```
leizhiyuan, line 376: This exception message should not be confirmed to be get max offset, right?

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -0,0 +1,699 @@
...
+                if (throwable != null) {
+                    log.error("PopConsumerService popAsync get message error",
+                        throwable instanceof CompletionException ? throwable.getCause() : throwable);
+                }
+                if (result.getMessageCount() > 0) {
+                    log.debug("PopConsumerService popAsync result, found={}, groupId={}, topicId={}, queueId={}, " +
+                            "batchSize={}, invisibleTime={}, fifo={}, attemptId={}, filter={}", result.getMessageCount(),
+                        groupId, topicId, queueId, batchSize, invisibleTime, fifo, attemptId, filter);
+                }
+                consumerLockService.unlock(groupId, topicId);
```
leizhiyuan, line 384: In some scenarios, it seems that this unlock may not be able to go? Whether to put it in final to unlock.

broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
```diff
@@ -213,27 +218,26 @@ public void notifyMessageArriving(final String topic, final int queueId, final S
     @Override
     public RemotingCommand processRequest(final ChannelHandlerContext ctx, RemotingCommand request)
         throws RemotingCommandException {
+
         final long beginTimeMills = this.brokerController.getMessageStore().now();
+
+        // fill bron time to properties if not exist, why we need this?
         request.addExtFieldIfNotExist(BORN_TIME, String.valueOf(System.currentTimeMillis()));
         if (Objects.equals(request.getExtFields().get(BORN_TIME), "0")) {
```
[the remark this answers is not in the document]
lizhimins: I use put if not exist method here. This code is equivalent to the original. Maybe there was originally a value of 0?

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
lizhimins: In the final version, the v2 format is used only. During the retry topic format transfer phase, there may be unconsumed retry messages in the v1 topic.

broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
lizhimins: Copied and pasted from the old method [doge]... I have already made the modifications.

broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
lizhimins: removed

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
lizhimins: I wrapped it with try-finally.

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -0,0 +1,702 @@
...
+    public CompletableFuture<PopConsumerContext> popAsync(String clientHost, long popTime, long invisibleTime,
+        String groupId, String topicId, int queueId, int batchSize, boolean fifo, String attemptId,
+        MessageFilter filter) {
+
+        PopConsumerContext popConsumerContext =
+            new PopConsumerContext(clientHost, popTime, invisibleTime, groupId, fifo, attemptId);
+
+        TopicConfig topicConfig = brokerController.getTopicConfigManager().selectTopicConfig(topicId);
+        if (topicConfig == null || !consumerLockService.tryLock(groupId, topicId)) {
+            return CompletableFuture.completedFuture(popConsumerContext);
```
redlsz, line 293: Can this lock granularity be optimized to queue id level?  Don't see the need for a larger lock.

lizhimins: In most scenarios, it is recommended to use pop broker(-1), and there is a unnessary cost of repeatedly locking and unlocking when there are a large number of queues

fuyou001 (approved):

PR #9113, opened 2025-01-08
[ISSUE #9112] Speedup revive scan in Pop Consumption and support server side reset offset

### LINKED ISSUE #9112: [Bug] Speedup revive scan in Pop Consumption and support server side reset offset

Linux 5.10

### RocketMQ version

only develop branch about version 5.3.1

### JDK Version

JDK11

### Describe the Bug

The current development branch uses the rocksdb seek first API, which has certain performance issues. The related issues are as follows:

1. https://github.com/aptos-labs/aptos-core/pull/3032
2. https://github.com/facebook/rocksdb/issues/10300
3. https://github.com/facebook/rocksdb/issues/10510

Solution:
> Actually, our application receives a delete request every time it is invoked by the user, and this is not something we can control. In some scenarios, the user will delete all keys under one prefix one by one, and the order of deletion is not consecutive, so we cannot convert the point deletes into range deletes.

Rocksdb official description: 
https://github.com/facebook/rocksdb/wiki/Prefix-Seek

When the server-side offset feature is enabled, the Pop KV implementation does not effectively reset the offset.

### Steps to Reproduce

I have written a test case, We can observe that using the seek first API results in higher latency.

image

1. The seek API has stable latency and fast
2. We can reset offset when server offset feature is enable

### What Did You See Instead?

None

### Additional Context

_No response_

lizhimins:
…er side reset offset

Fixes #9112

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerKVStore.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerKVStore.java
@@ -50,9 +50,8 @@ public interface PopConsumerKVStore {
 
     /**
      * Scans and returns a list of expired consumer records before the current time.
-     * @param currentTime The current revive checkpoint timestamp.
      * @param maxCount The maximum number of records to return.
      * @return A list of expired consumer records.
      */
-    List<PopConsumerRecord> scanExpiredRecords(long currentTime, int maxCount);
+    List<PopConsumerRecord> scanExpiredRecords(long lowerTime, long upperTime, int maxCount);
 }
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -75,6 +75,7 @@ public class PopConsumerService extends ServiceThread {
     private final AtomicBoolean consumerRunning;
     private final BrokerConfig brokerConfig;
     private final BrokerController brokerController;
+    private final AtomicLong currentTime;
     private final AtomicLong lastCleanupLockTime;
     private final PopConsumerCache popConsumerCache;
     private final PopConsumerKVStore popConsumerStore;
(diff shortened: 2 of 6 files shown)


lizhimins:
@redlsz Please help review this pr

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerKVStore.java
```diff
@@ -50,9 +50,8 @@ public interface PopConsumerKVStore {
 
     /**
      * Scans and returns a list of expired consumer records before the current time.
```
lollipopjin, line 52: Should update the java doc here.

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -513,14 +532,18 @@ public long revive(long currentTime, int maxCount) {
...
         this.popConsumerStore.writeRecords(new ArrayList<>(failureList));
         this.popConsumerStore.deleteRecords(consumerRecords);
+        currentTime.set(consumerRecords.isEmpty() ?
+            upperTime : consumerRecords.get(consumerRecords.size() - 1).getVisibilityTimeout());
 
         if (brokerConfig.isEnablePopBufferMerge()) {
-            log.info("PopConsumerService, key size={}, cache size={}, revive count={}, failure count={}, cost={}ms",
-                popConsumerCache.getCacheKeySize(), popConsumerCache.getCacheSize(), consumerRecords.size(),
-                failureList.size(), stopwatch.elapsed(TimeUnit.MILLISECONDS));
+            log.info("PopConsumerService, key size={}, cache size={}, revive count={}, failure count={}, behind={}ms, scan={}ms, cost={}ms",
```
lollipopjin, line 539: remove the ms unit,  and use timeBehindInMillis instead.

lollipopjin (approved):
LGTM

PR #9116, opened 2025-01-09
[ISSUE #9115] Optimize the broker's reverse notification for consumerId change

### LINKED ISSUE #9115: [Enhancement] Optimize the broker's reverse notification for consumerId change

### Summary

Hi, community
This is a reverse notification issue triggered by broker for consumerId change.

### Motivation

We encountered such a problem in the online environment:
An application has 100+ clients, and each client has multiple group subscriptions. When the application is released, any node will trigger a large number of reverse notifications. When the threshold is exceeded, most of the notifications fail, and we only need the latest notification.
企业微信截图_0a642809-d3a5-446c-bc65-9d0f4ae4a500

Our idea is to improve the real-time notification part. If there are new channels that need to be notified in the same group, we will stop the last notification and always notify with the latest channels.

Stop the broker's reverse notification and downgrade to the client's own balance logic processing.

### Additional Context

_No response_

yx9o:
Fixes #9115 .

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/DefaultConsumerIdsChangeListener.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/DefaultConsumerIdsChangeListener.java
@@ -23,6 +23,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.common.AbstractBrokerRunnable;
 import org.apache.rocketmq.common.constant.LoggerName;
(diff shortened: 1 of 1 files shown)


yx9o:
> 1. task1 has already done notification
> 2. task2 try interrupt task1 and update map
> 3. task1 remove group in map
> 
> if so, task2 will never be interrupted, but it is better than now

Hi, thank you for your review. I added the atomic check here. Please help review it again. Thank you.

qianye1001:
LGTM

qianye1001 (approved):

broker/src/main/java/org/apache/rocketmq/broker/client/DefaultConsumerIdsChangeListener.java
```diff
@@ -125,4 +145,27 @@ private void notifyConsumerChange() {
     public void shutdown() {
         this.scheduledExecutorService.shutdown();
     }
+
+    private static class NotifyTaskControl {
+
+        private final AtomicBoolean interrupted = new AtomicBoolean(false);
+
+        private final List<Channel> channels;
```
Koado, line 153: What is the role of channels in NotifyTaskControl?

yx9o: Ensure that when a new notification task arrives, the channels in the old notification task are no longer notified.

Koado: I understand that this is controlled by the interrupted flag? once the interrupted flag is true, the for loop will be break.

lollipopjin (approved):
LGTM

PR #9131, opened 2025-01-14
[Enhancement] Load balancing cache in POP consumption mode

### LINKED ISSUE #9123: [Enhancement] Load balancing cache in POP consumption mode

### Summary

In the POP consumption mode, the client will obtain the result from the server during each rebalancing, which may involve a lot of computing resource consumption. Should we consider adding a layer of cache on the server

### Motivation

Server computing resources are wasted

Caching can reduce computing resource consumption

null

### Additional Context

_No response_

YoWuwuuuw:

Add cache for each client request broker end rebalancing in the pop consumption mode to avoid wasting computing resources of broker end due to repeated calculation of load balancing without changes. The effect is more obvious in theory when there are a large number of queues or consumers

Fixes #9123

### Brief Description
1. **Cache Location**: The cache is strategically placed on the broker side, executed before the load balancing calculation.
2. **Cache Logic**:
- Topic serves as the primary cache key.
- Within each topic's cache entry, a `ConcurrentHashMap` stores the assigned queues, keyed by a combination of `clientId, strategyName, and popShareQueueNum`.
- Consumer group and message queue information are also stored to act as a verification basis for cache validity.
3. **Cache Hit**: A cache hit occurs when the topic exists in the cache, and the current consumer group and message queue information matches the cached data.
4. **Cache Update**: If a cache miss occurs (i.e., the topic isn't in the cache, or the consumer group/queue information has changed), the load balancing calculation is performed, and the cache is then updated with the new result.
Concurrency and Efficiency: A `ReentrantLock` ensures data consistency and efficient access to the cache.
5. **Cache Expiration**: Each cache entry includes a timestamp. An entry is considered expired if it hasn't been accessed for 24 hours. A dedicated cleaning mechanism periodically removes expired entries at fixed intervals, maintaining cache effectiveness and preventing excessive resource consumption.

1. **Existing Tests**: All 5 original test methods within `QueryAssignmentProcessorTest` passed, ensuring no regressions.
2. **New Cache Test**: A new test method, `testDoLoadBalanceWithCache()`, was added specifically to validate the caching mechanism. This test simulates multiple consumers and queues to verify:
- The first call to `doLoadBalance` correctly performs the calculation and populates the cache.
- Subsequent calls with identical parameters successfully retrieve results from the cache.
- When topic route information changes (e.g., adding more queues), the cache correctly invalidates, triggering a recalculation.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopRebalanceCacheManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopRebalanceCacheManager.java
@@ -1,14 +1,14 @@
 package org.apache.rocketmq.broker.processor;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.logging.org.slf4j.Logger;
 import org.apache.rocketmq.logging.org.slf4j.LoggerFactory;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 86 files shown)


YoWuwuuuw:
> Hello, please fix the code running test and output the test document

okok

Koado (approved):

3424672656:
The cache should be updated when the client is offline or the number of queues changes, but I don't see the logic for that

broker/src/main/java/org/apache/rocketmq/broker/processor/PopRebalanceCacheManager.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+public class PopRebalanceCacheManager {
+
+    private static final long LOCK_TIMEOUT_MILLIS = 3000L;
+
+    private static final Logger log = LoggerFactory.getLogger(LoggerName.BROKER_LOGGER_NAME);
+
+    // Cache with pop mode rebalancing under the condition that consumer groups and queues remain unchanged
+    private final ConcurrentMap<String, ConcurrentHashMap<String, Set<MessageQueue>>> loadBalanceDateTable = new ConcurrentHashMap<>();
```
3424672656, line 22: When the number of clients changes or the number of queues changes.

3424672656 (changes requested):

broker/src/main/java/org/apache/rocketmq/broker/processor/PopRebalanceCacheManager.java
YoWuwuuuw: Ok, I've added a cleanup thread for PopCacheManager and tested it to completely clear the cache every 24 hours (the cache expiration only happens when the topic goes offline, and I'll modify "Brief Description" to simply explain the cache logic). This cleanup interval can be added to the configuration if needed later

YoWuwuuuw:
@RongtongJin @lollipopjin @3424672656 PTAL

PR #9135, opened 2025-01-15
6398: adding a small change

yuktidave:
made a dummy change for test purpose/ im not sure if this is correct or not.


client/src/main/java/org/apache/rocketmq/client/impl/producer/TopicPublishInfo.java
```diff
@@ -72,50 +66,16 @@ public void setHaveTopicRouterInfo(boolean haveTopicRouterInfo) {
         this.haveTopicRouterInfo = haveTopicRouterInfo;
     }
 
-    public MessageQueue selectOneMessageQueue(QueueFilter ...filter) {
```
yx9o, line 75: Hi, why are they deleted, these methods are referenced.

qianye1001 (changes requested):
do not delete selectOneMessageQueue method

PR #9153, opened 2025-01-22
[ISSUE #9152] Broker getConsumeStats supports inputting multiple topics

### LINKED ISSUE #9152: [Enhancement] Supports multiple topics as input when send getConsumeStats request at the same time

### Summary

Supports multiple topics as input when send getConsumeStats request at the same time

### Motivation

getConsumeStats only can input one topic

add field

null

### Additional Context

_No response_

qianye1001:

Fixes #9152

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -1947,16 +1947,14 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         final RemotingCommand response = RemotingCommand.createResponseCommand(null);
         try {
             final GetConsumeStatsRequestHeader requestHeader = request.decodeCommandCustomHeader(GetConsumeStatsRequestHeader.class);
+            List<String> topicListProvided = requestHeader.fetchTopicList();
+            String topicProvided = requestHeader.getTopic();
+            String group = requestHeader.getConsumerGroup();
+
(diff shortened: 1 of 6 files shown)


broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -2038,6 +2022,37 @@ private RemotingCommand getConsumeStats(ChannelHandlerContext ctx,
...
+                String topic = iterator.next();
+                SubscriptionData findSubscriptionData = this.brokerController.getConsumerManager().findSubscriptionData(group, topic);
+                if (findSubscriptionData == null && subscriptionCount > 0) {
+                    LOGGER.warn(
+                        "AdminBrokerProcessor#getConsumeStats: topic does not exist in consumer group's subscription, topic={}, consumer group={}",
+                        topic, group);
+                    iterator.remove();
+                }
+            }
+        }
```
Koado, line 2052: why no need to enter this filtering logic when providing topicList?

qianye1001: When topic or topicList is provided, the request sponsor has clearly explicitly needed topics information, so it is unnecessary to check the subscription relationship, but in order to compatible the old logic, it was checked when providing topic.

Koado (approved):
LGTM

RongtongJin (dismissed):

lollipopjin (approved):
LGTM

PR #9165, opened 2025-02-04
add producer stats manager

### LINKED ISSUE #9164: [Enhancement] Add producer stats manager for 4.x java client

### Summary

Add a producer stats manager to the 4.x Java client to periodically print produce stats in the client logs.

### Motivation

Currently, we have a consumer stats manager in the 4. x Java client that periodically logs consumption stats in the log, but we don't have a producer stats manager to print production stats.

Add a producer stats manager similar to the consumer stats manager to periodically print producer stats in the client log.

When sending a message, increase the number of producer stats manager.

-

### Additional Context

_No response_

HScarb:

Fixes #9164 

### Brief Description

Add a producer stats manager to the 4.x Java client to periodically print produce stats in the client logs.

Running 4.x java client's producer benchmark and checking for the logs.

image

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/stat/ProducerStatsManager.java
+++ b/client/src/main/java/org/apache/rocketmq/client/stat/ProducerStatsManager.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


yx9o:
Hi, I think it is enough to monitor the time taken to call the sending API directly in the application. There is no need to integrate it into the sending client.

client/src/main/java/org/apache/rocketmq/client/stat/ProducerStatsManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private static final String TOPIC_SEND_OK_TPS = "SEND_OK_TPS";
+    private static final String TOPIC_SEND_FAILED_TPS = "SEND_FAILED_TPS";
+    private static final String TOPIC_SEND_RT = "SEND_RT";
+
+    private final StatsItemSet topicSendOKTPS;
+    private final StatsItemSet topicSendFailedTPS;
+    private final StatsItemSet topicSendRT;
+
+    public ProducerStatsManager(final ScheduledExecutorService scheduledExecutorService) {
```
humkum, line 36: Maybe you should add some unit tests for this feature

PR #9194, opened 2025-02-19
[ISSUE #9193]Optimized get sequential message lock logic

### LINKED ISSUE #9193: [Enhancement] Remove invalid judgments and use atomicity functions

### Summary

Image
Remove invalid judgments and use atomicity functions

### Motivation

Remove invalid judgments and use atomicity functions

null

null

### Additional Context

_No response_

3424672656:

Fixes #9193

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/rebalance/RebalanceLockManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/rebalance/RebalanceLockManager.java
@@ -55,24 +55,18 @@ public class RebalanceLockManager {
             try {
                 this.lock.lockInterruptibly();
                 try {
-                    ConcurrentHashMap<MessageQueue, LockEntry> groupValue = this.mqLockTable.get(group);
-                    if (null == groupValue) {
-                        groupValue = new ConcurrentHashMap<>(32);
-                        this.mqLockTable.put(group, groupValue);
-                    }
+                    ConcurrentHashMap<MessageQueue, LockEntry> groupValue = this.mqLockTable.computeIfAbsent(group,
+                        k -> new ConcurrentHashMap<>(32));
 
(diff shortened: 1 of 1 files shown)


broker/src/main/java/org/apache/rocketmq/broker/client/rebalance/RebalanceLockManager.java
```diff
@@ -55,24 +55,18 @@ public boolean tryLock(final String group, final MessageQueue mq, final String c
...
-                        this.mqLockTable.put(group, groupValue);
-                    }
+                    ConcurrentHashMap<MessageQueue, LockEntry> groupValue = this.mqLockTable.computeIfAbsent(group,
+                        k -> new ConcurrentHashMap<>(32));
 
                     LockEntry lockEntry = groupValue.get(mq);
                     if (null == lockEntry) {
                         lockEntry = new LockEntry();
                         lockEntry.setClientId(clientId);
+                        lockEntry.setLastUpdateTimestamp(System.currentTimeMillis());
```
caigy, line 65: It seem that `lastUpdateTimestamp` has been initialized when `LockEntry` is created.

caigy (changes requested):

broker/src/main/java/org/apache/rocketmq/broker/client/rebalance/RebalanceLockManager.java
3424672656: Since the lock should have been acquired by the current thread during initialization, setting updatetime means that the lock was acquired

3424672656: In older versions, updatetime was not initialized here，After changing the logic, you can get the lock by default here and return success directly

Issue #9176, opened 2025-02-12
[Bug] POP Msg failed when enable Authorization of ACL 2.0


zhiliatom:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

mac os

### RocketMQ version

5.3.x 

### JDK Version

jdk 1.8

### Describe the Bug

PopMessagingProcessors.processRequest() will failed when configure `authorizationEnabled=true`, So  client can't consume message. The root cause is that incorrect caching was applied during the decoding process of the RequestHeader,  lack of the required `bornTime` field

![Image](https://github.com/user-attachments/assets/ce6457ac-3950-4d59-9bb4-103b3fadde2a)

![Image](https://github.com/user-attachments/assets/b13c97a6-5e05-462f-8077-5ec1ca47440c)

### Steps to Reproduce

1. enable authorization
2. start a grpc client 

### What Did You Expect to See?

pop and consume message 

### What Did You See Instead?

can't POP message

PR #9211, opened 2025-02-27
Revert "[ISSUE #9176] Fix pop message header missing fields when enable ACL 2.0"

lizhimins:
When there is a time difference between the client and the server, it can cause issues with suspending long-polling POP requests.


lollipopjin (commented):
LGTM

lollipopjin (approved):

PR #9251, opened 2025-03-16
Optimized ConsumerOffsetManager for Faster Topic & Group Lookups

### LINKED ISSUE #9240: [Enhancement] Improving Broker Performance: Optimizing the whichTopicByConsumer Method Query Efficiency

### Summary

The current implementation of the method `whichTopicByConsumer` is as follows:

```java
public Set<String> whichTopicByConsumer(final String group) {
    Set<String> topics = new HashSet<>();

    Iterator<Entry<String, ConcurrentMap<Integer, Long>>> it = this.offsetTable.entrySet().iterator();
    while (it.hasNext()) {
        Entry<String, ConcurrentMap<Integer, Long>> next = it.next();
        String topicAtGroup = next.getKey();
        String[] arrays = topicAtGroup.split(TOPIC_GROUP_SEPARATOR);
        if (arrays.length == 2) {
            if (group.equals(arrays[1])) {
                topics.add(arrays[0]);
            }
        }
    }

    return topics;
}
```
This method is used to query all topics under a specified group. However, it requires iterating through the entire offsetTable, and the advantages of the hash table are not utilized at all. The query efficiency is far too slow! As a result, a large portion of my CPU resources is being consumed by String.split.

Below is the CPU flame graph. Although it is from an older version of RocketMQ, I have reviewed the code in the latest version, and this part still hasn't been optimized.
Image

### Motivation

I need to periodically collect the offset progress of groups every minute, which involves calling the inefficient `whichTopicByConsumer` method. Such an inefficient query can be completely avoided. Simply querying the topics corresponding to a single group shouldn't require traversing all the data—this is unreasonable.

I would like the `whichTopicByConsumer` method to be optimized to avoid the need for a full traversal of the `offsetTable` when querying the topics corresponding to a specific group. A more efficient approach would involve introducing a pre-built mapping (e.g., a hash table or a similar indexed data structure) that directly maps consumer groups to their associated topics. This mapping can be maintained dynamically and updated whenever changes occur in the `offsetTable`, ensuring consistency.

With this solution, the query logic can simply perform a direct lookup in the mapping, reducing the complexity from **O(N)** (full traversal) to **O(1)** or **O(K)**, where K is the number of topics associated with the group. This would significantly improve query efficiency and reduce CPU usage caused by repetitive and expensive operations like `String.split`.

By implementing this optimization, the system would handle periodic offset progress collection much more efficiently, especially in scenarios where the number of groups and topics grows large.

One approach is to manually pass in the specified topic when calling the `getConsumeStats` method. This way, the execution of `whichTopicByConsumer` can be avoided. However, the inefficiency of `whichTopicByConsumer` still remains as an issue.

### Additional Context

_No response_

DivyanshIITB:
This PR enhances the performance of ConsumerOffsetManager by reducing unnecessary iterations and optimizing key lookups in offset-related methods.

Key Optimizations :

1) Faster Lookups
- Replaced O(N) iterations in whichTopicByConsumer() and whichGroupByTopic() with O(1) HashMap lookups using precomputed mappings (groupToTopics & topicToGroups).

2) Efficient Cleanup Methods
- cleanOffset() and cleanOffsetByTopic() now use temporary HashSets to track keys to remove, preventing concurrent modification issues.
- Eliminated incorrect matches caused by contains() checks.

3)Optimized Offset Scanning
- scanUnsubscribedTopic() avoids unnecessary split() calls and uses bulk removal instead of modifying offsetTable inside the loop.

4)Maintains Offset-Topic Consistency
- Added updateOffsetTable() to ensure groupToTopics & topicToGroups mappings stay in sync with offsetTable.

This results in faster topic-group lookups, reduced memory overhead, and improved efficiency when managing consumer offsets. 

Fixes #9240

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
@@ -60,6 +60,10 @@ public class ConsumerOffsetManager extends ConfigManager {
 
     private final transient AtomicLong versionChangeCounter = new AtomicLong(0);
 
+    // Optimized HashMap for faster lookups
+    private final ConcurrentMap<String, Set<String>> groupToTopics = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, Set<String>> topicToGroups = new ConcurrentHashMap<>();
+
     public ConsumerOffsetManager() {
     }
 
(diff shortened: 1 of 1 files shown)


DivyanshIITB:
Just a gentle reminder

lizhimins:
Need to pass all test cases first

DivyanshIITB:
I have made the necessary changes.
@lizhimins

PR #9321, opened 2025-04-10
[ISSUE #9320] Ignore extra put when checkImmunityTime exceeds the maximum check time range

### LINKED ISSUE #9320: [Enhancement] ignore extra put when checkImmunityTime exceeds  the maximum check time range

### Summary

For a half message, if its CHECK_IMMUNITY_TIME_IN_SECONDS property is lager than checkMax * checkInterval, then it will never be checked, but still has been put for checkMax times. It's a waste of put.
If we have a lot of trans messages, the put metrics will by nearly 15 times high.

### Motivation

Transactional message's property checkImmunityTime has no upper limit, everyone can set a large value, for example 1800 seconds.

Continue loop and log warn when CHECK_IMMUNITY_TIME_IN_SECONDS property is lager than checkMax * checkInterval.

Do not use checkImmunityTime property, but that's not why it is built for.

### Additional Context

_No response_

WJ66880:
…imum check time range #9320

Fixes #9320

send transactional messages with checkImmunityTime 1800.


lizhimins (approved):

WJ66880:
repository removed, submit a PR again

PR #9334, opened 2025-04-15
[#ISSUE 9333] Use fastjson2 in broker module

yx9o:
Fixes #9333 .

PR #9387, opened 2025-05-06
Revert "[ISSUE #9333] Use fastjson2 in broker module"

lizhimins:
Reverts apache/rocketmq#9334

In consideration of compatibility for minor versions, this commit will be temporarily reverted.
Once all changes related to fastjson2 are consolidated, we will release them in a new major version.


lollipopjin (approved):
LGTM

PR #9429, opened 2025-05-23
[ISSUE#9411]: ensure broker re-registers after namesrv restart to avoid route loss

nero0912:
### Motivation

When the NameServer restarts, all registered broker route information is lost. However, brokers do not automatically detect and recover from this, leading to potential message routing failures.

This PR introduces a periodic detection mechanism in the `BrokerController`. When a NameServer restart is detected (e.g., due to socket timeout or address list changes), the broker will proactively re-register itself to each NameServer individually to restore routing data.

### Modifications

- **New Method**: Added `registerBrokerSingle()` in `BrokerOuterAPI` to encapsulate logic for registering with a single NameServer.
- **Code Cleanup**: Extracted common logic from `registerBrokerToSingleNameserver` to reduce duplication by reusing `registerBroker`.
- **Heartbeat Mechanism**: Added a scheduled task in `BrokerController` to periodically detect NameServer restarts.
    - Detection logic includes socket connection timeouts and address list changes.
    - If a restart is detected, the broker calls `registerBrokerSingle()` for each NameServer address.
- **Logging**: Improved log messages for better observability of re-registration events.
- **Testability**: Made behavior easier to verify and observe in broker logs.

### Verifying this change

Tested locally:
- Launched broker and NameServer.
- Restarted NameServer.
- Verified that broker automatically re-registered with each NameServer and restored route information.
- No regression observed in normal broker startup and routing behavior.

### Does this PR introduce any user-facing change?

Yes. This improves system robustness and availability during NameServer restarts by automatically recovering broker routing information.

### Documentation

No user-facing documentation update is required, as this behavior is internal and backward compatible.

### DIFF
--- a/.github/workflows/license-checker.yaml
+++ b/.github/workflows/license-checker.yaml
@@ -19,7 +19,8 @@ name: License checker
 on:
   pull_request:
     branches:
-      - release-4.9.8
+      - develop
+      - master
 
 jobs:
   check-license:
--- a/.github/workflows/license-checker.yaml
+++ b/.github/workflows/license-checker.yaml
@@ -19,7 +19,7 @@ name: License checker
 on:
   pull_request:
     branches:
-      - '*'
+      - release-4.9.8
 
 jobs:
   check-license:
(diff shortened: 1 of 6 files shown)


RongtongJin (commented):
When a nameserver goes offline, it seems that re-registering other brokers doesn't make much sense.

PR #9437, opened 2025-05-27
[ISSUE #9436] Add new requestcode to return only retrytopic

### LINKED ISSUE #9436: [Enhancement] Add new requestcode to return only retrytopic

### Summary

Add code to get retrytopic

### Motivation

There are several places in the code to get the full topic. In actual use, only retrytopic is used. If the cluster's topic reaches 10,000 or 100,000, the interface returns a lot of unnecessary data. Therefore, I want to add a new requestcode to return only retrytopic.

Image

Image

Add new requestcode to return only retrytopic.

Add new requestcode to return only retrytopic.

### Additional Context

_No response_

yx9o:
Fixes #9436 .


lizhimins:
To reduce the total number of RPC, it also seems reasonable to implement this feature only in the tools.

PR #9440, opened 2025-05-30
[ISSUE #9439] Add escape for win in the method returning broker configuration

### LINKED ISSUE #9439: [Bug] getBrokerConfig return null and log displayed "Malformed \uxxxx encoding."

windows10

### RocketMQ version

5.3.3 release

### JDK Version

java 8

### Describe the Bug

getBrokerConfig return null and log displayed "Malformed \uxxxx encoding."

### Steps to Reproduce

This happens when, under Windows, a file path such as ROCKETMQ_HOME contains \u, because the path is not escaped.

Normal return configuration

### What Did You See Instead?

return

Image

### Additional Context

_No response_

Crazylychee:

Fixes #9439

### Brief Description

utilize MixAll.isWindows for platform-specific escaping

Affected files:
- `MixAll.java`: utilize MixAll.isWindows for platform-specific escaping

image
The above is before the modification, you can see the bug. The following is after the modification, you can see the program returns normally.
image

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/MixAll.java
+++ b/common/src/main/java/org/apache/rocketmq/common/MixAll.java
@@ -557,4 +557,13 @@ public class MixAll {
             && !topic.startsWith(TopicValidator.SYSTEM_TOPIC_PREFIX)
             && !topic.equals(TopicValidator.RMQ_SYS_SCHEDULE_TOPIC);
     }
+
+    public static String adjustConfigForPlatform(String config) {
+        if (StringUtils.isNotBlank(config)) {
+            if (isWindows()) {
+                config = StringUtils.replace(config, "\\", "\\\\");
+            }
+        }
+        return config;
+    }
 }
--- a/common/src/test/java/org/apache/rocketmq/common/MixAllTest.java
+++ b/common/src/test/java/org/apache/rocketmq/common/MixAllTest.java
@@ -85,4 +85,43 @@ public class MixAllTest {
         testLmq = "%LMQ%GID_TEST";
         assertThat(MixAll.isLmq(testLmq)).isTrue();
     }
+
+    @Test
+    public void testAdjustConfigForPlatform_OnWindows() {
+        if (MixAll.isWindows()) {
+            String configWithSingleBackslash = "data\\path\\config\\file.properties";
+            String adjusted = MixAll.adjustConfigForPlatform(configWithSingleBackslash);
+            assertThat(adjusted).isEqualTo("data\\\\path\\\\config\\\\file.properties");
+
+            String configWithMultipleBackslashes = "C:\\\\RocketMQ\\\\logs\\\\broker.log";
+            adjusted = MixAll.adjustConfigForPlatform(configWithMultipleBackslashes);
+            assertThat(adjusted).isEqualTo("C:\\\\\\\\RocketMQ\\\\\\\\logs\\\\\\\\broker.log");
+
+            String configWithoutBackslash = "listenPort=10911";
+            adjusted = MixAll.adjustConfigForPlatform(configWithoutBackslash);
+            assertThat(adjusted).isEqualTo("listenPort=10911");
+
+            String emptyConfig = "";
+            adjusted = MixAll.adjustConfigForPlatform(emptyConfig);
+            assertThat(adjusted).isEqualTo("");
+
+            adjusted = MixAll.adjustConfigForPlatform(null);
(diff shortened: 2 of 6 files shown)


Crazylychee:
> It is recommended to perform escaping on the client side (at the admin tools code), and handle it according to the platform type (can be done using MixAll.isWindows).

Thanks for the feedback! I've implemented the client-side escaping in the admin tools code as suggested. The logic now uses `MixAll.isWindows` to handle the escaping differently based on the platform type.

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -334,11 +334,17 @@ public static String properties2String(final Properties properties, final boolea
...
+        }
         try {
-            InputStream in = new ByteArrayInputStream(str.getBytes(DEFAULT_CHARSET));
+            InputStream in = new ByteArrayInputStream(processedStr.getBytes(DEFAULT_CHARSET));
             properties.load(in);
         } catch (Exception e) {
-            log.error("Failed to handle properties", e);
+            log.error("Failed to handle properties from string: '{}'", processedStr, e);
             return null;
         }
```
RongtongJin, line 349: This is still used in many places on the server side, so it would be better to only modify the section of getBrokerConfig returned by admin tools

Crazylychee: Changes made: The scope now covers server-side config return methods. A new system-type-based escaping method has been encapsulated and integrated into these methods. Corresponding unit tests have also been added.

common/src/main/java/org/apache/rocketmq/common/MixAll.java
```diff
@@ -557,4 +557,13 @@ public static boolean topicAllowsLMQ(String topic) {
             && !topic.startsWith(TopicValidator.SYSTEM_TOPIC_PREFIX)
             && !topic.equals(TopicValidator.RMQ_SYS_SCHEDULE_TOPIC);
     }
+
+    public static String adjustConfigForPlatform(String config) {
+        if (StringUtils.isNotBlank(config)) {
+            if (isWindows()) {
+                config = config.replace("\\", "\\\\");
```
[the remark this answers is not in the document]
Crazylychee: Thank you, this is a good suggestion, I have corrected it as you said.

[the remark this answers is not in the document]
xiaoyifang: paramater should also be changed.
```
config.replace("\\","\\\\")
```
equals to 
```
StringUtils.replace(config,"\","\\");
```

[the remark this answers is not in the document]
xiaoyifang: you are right. maybe need more test.

common/src/test/java/org/apache/rocketmq/common/MixAllTest.java
```diff
@@ -85,4 +85,43 @@ public void testIsLmq() {
         testLmq = "%LMQ%GID_TEST";
         assertThat(MixAll.isLmq(testLmq)).isTrue();
     }
+
+    @Test
+    public void testAdjustConfigForPlatform_OnWindows() {
+        if (MixAll.isWindows()) {
+            String configWithSingleBackslash = "data\\path\\config\\file.properties";
+            String adjusted = MixAll.adjustConfigForPlatform(configWithSingleBackslash);
+            assertThat(adjusted).isEqualTo("data\\\\path\\\\config\\\\file.properties");
```
xiaoyifang, line 94: seems not correct.  the point is that the file must existed. after replace the `\` ,does the file can be still found.

Crazylychee: It is like this: the characters are displayed normally only after replacement. If they are not replaced, \ will be interpreted as an escape character, resulting in an error and failure to obtain the broker configuration.

Crazylychee: Moreover, this configuration acquisition does not really need to find the file with the corresponding path, but is converted into a string for display on the management side. I verified the effect and the result is correct after escaping.
image

Crazylychee: Oh, I understand what you mean, that is, I don't need to add the escape character for the configuration that already has two escape characters like \\.

GenerousMan (approved):
LGTM

PR #9468, opened 2025-06-12
[ISSUE #9467]Implement paged retrieval of Topic and SubscriptionGroup information

### LINKED ISSUE #9467: [Enhancement] Implement paged retrieval of topic and subscription group information

### Summary

By introducing sequence-based pagination, you can safely handle large metadata sets like Topic and SubscriptionGroup configurations, ensuring compatibility with Netty’s frame limits and improving robustness across environments.

### Motivation

When retrieving topic or subscription group information via getAllTopicConfig and getAllSubscriptionGroup, if the number of topics or groups is too large, the serialized data may exceed Netty's maxFrameLength, leading to invocation failures.

Modify the request logic so that clients can specify:
- A starting sequence number (topicSeq or groupSeq)
- Maximum number of items to return per request (maxTopicNum or maxGroupNum)

This allows the client to fetch data in smaller, manageable chunks.

none

### Additional Context

_No response_

ltamber:

Fixes #9467

### Brief Description

- Implemented pagination logic to support data retrieval by sequence number and maximum count.
- Added data version checking to ensure the retrieved data is the latest.
- Optimized the result structure to include total count and current page data.
- Added unit tests

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
@@ -778,20 +778,28 @@ public class BrokerOuterAPI {
 
         DataVersion topicConfigDataVersion = null;
         DataVersion mappingDataVersion = null;
+        long timeoutMills = getTimeoutMillis();
         int topicSeq = 0;
+        long beginTime = System.nanoTime();
         ConcurrentHashMap<String, TopicConfig> topicConfigTable = new ConcurrentHashMap<>();
         Map<String, TopicQueueMappingDetail> topicQueueMappingDetailMap = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
```diff
@@ -111,6 +111,8 @@ public class ClientConfig {
      */
     protected String traceTopic;
 
+    protected int maxPageSizeInGetMetadata = 2000;
```
RongtongJin, line 114: It may not be necessary to add this parameter to clientConfig; perhaps it would be better to allow the caller to set it based on their specific needs.

RongtongJin (approved):

caigy (approved):

fuyou001 (approved):

PR #9472, opened 2025-06-13
[ISSUE #9471]  load kv store before service start in PopConsumerService

### LINKED ISSUE #9471: [Enhancement] load kv store before service start in PopConsumerService

### Summary

Loading the kv database should be performed before initializing the service so that the information of the kv database can be statistically analyzed.

### Motivation

The indicators and other data of the kv database can be obtained for statistical analysis

load kv store before service start

Load the database when the service is initialized

### Additional Context

null

3424672656:

Fixes #9471

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -883,6 +883,10 @@ public class BrokerController {
             result = this.messageStore.load();
         }
 
+        if (brokerConfig.isPopConsumerKVServiceInit()) {
+            result = result && popConsumerService.load();
+        }
+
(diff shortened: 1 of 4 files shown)


broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
```diff
@@ -883,6 +883,10 @@ public boolean recoverAndInitService() throws CloneNotSupportedException {
             result = this.messageStore.load();
         }
 
+        if (brokerConfig.isPopConsumerKVServiceInit()) {
+            result = result && popConsumerService.load();
```
lizhimins, line 887: init method in BrokerController#startBasicService

lizhimins (changes requested):

PR #9480, opened 2025-06-16
Relax topic length constraint when creating retry and dlq topics

### LINKED ISSUE #9479: [Enhancement] Relax topic length constraint when creating retry and dlq topics

### Summary

Currently, the max topic length is set to be 127 bytes and this length constraint is for all topics as shown in the screenshots below:

Image

Image

We want to enhance the capabilities of RocketMQ by allowing the users to create topics and groups of longer length, from 60 bytes previously to 120 bytes. To achieve this, we need to relax the topic length constraint mentioned above for dlq and retry topics because dlq and retry topics' names are created by concatenating the string "%DLQ%" or "%RETRY%" with the corresponding group name and/or topic name in subscription. The resulting length of the concatenated string maybe longer than 127 if topic/group is too long. 

### Motivation

Enhance the capabilities of RocketMQ by relaxing the topic/group length constraint. 

Add two more conditions in the if statement of validateTopic method to additionally check whether the validating topic is dlq or retry topic and if it is then skip the topic length validation. 

Image

N/A

### Additional Context

_No response_

Allenhuangrui:

Fixes #9479 

### Brief Description

Add two more conditions in the if statement of validateTopic method to additionally check whether the validating topic is dlq or retry topic and if it is then skip the topic length validation.
image

Built a snapshot of this change and tested creating dlq and retry topics of length greater than 127.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
+++ b/common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.common.topic;
 
 import java.util.HashSet;
 import java.util.Set;
+import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.UtilAll;
 
 public class TopicValidator {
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/rocketmq/common/topic/TopicValidator.java
```diff
@@ -111,7 +112,9 @@ public static ValidateTopicResult validateTopic(String topic) {
             return new ValidateTopicResult(false, "The specified topic contains illegal characters, allowing only ^[%|a-zA-Z0-9_-]+$");
         }
 
-        if (topic.length() > TOPIC_MAX_LENGTH) {
+        if (!topic.startsWith(MixAll.RETRY_GROUP_TOPIC_PREFIX) &&
+            !topic.startsWith(MixAll.DLQ_GROUP_TOPIC_PREFIX) &&
```
lizhimins, line 116: 那么 retry 和 dlq 的限制呢

lizhimins (changes requested):
那么 retry 和 dlq 的限制呢

qianye1001 (approved):
LGTM

lizhimins (approved):

PR #9493, opened 2025-06-21
[#ISSUE 9453] Remove bazel workflow

### LINKED ISSUE #9453: [Enhancement] Removing the Bazel workflow

### Summary

Previously, we introduced the Bazel workflow in our community, which can quickly provide test results, but it also introduced some issues:

- The maintenance cost is relatively high. This includes timely updates to configuration files when new packages are introduced. The workflow often fails due to configuration issues, which can be confusing even when there are no actual problems.

- Although it can quickly provide test results, multiple workflows run simultaneously, and due to the bucket effect (wooden bucket principle), the final test results still require a long waiting time. Moreover, the test cases in other workflows can cover those in the Bazel workflow.

Therefore, I would like to propose removing this workflow. If you have any suggestions, please feel free to reply below.

### Motivation

Removing the Bazel workflow

Removing the Bazel workflow 

No

### Additional Context

_No response_

OmCheeLin:

Fixes #9453

### Brief Description

Removing the Bazel workflow


RongtongJin:
@OmCheeLin Could you help remove some extra files from the code as well?

RongtongJin (approved):

RongtongJin:
> > @OmCheeLin Could you help remove some extra files from the code as well?
> 
> Remove all code related to Bazel？

Yes, we can do this in next pull request

lizhanhui (changes requested):
Removing the workflow is hiding problems

lizhanhui:
Let's discuss the issue first, before rushing to removal

PR #9513, opened 2025-07-01
[ISSUE #8920] Make TLS certificate watch interval configurable

### LINKED ISSUE #8920: [Enhancement] Refactor SSL context loading process to support multiple protocols dynamic loading.

### Summary

Refactor SSL context loading process to support multiple protocols dynamic loading.

### Motivation

There are currently multiple implementations of SSL context loading, which should be consolidated. Furthermore, the current system requires a service restart when certificates need to be replaced. It's necessary to implement dynamic loading to avoid this downtime.

Refactor SSL context loading process to support multiple protocols dynamic loading.

null

### Additional Context

_No response_

EnableAsync:
This PR makes the watch interval for TLS certificate and key files configurable in the Proxy module.

Previously, the interval was hardcoded to 1 hour, which is not flexible for different production environments. A new configuration property tlsCertWatchIntervalMs has been introduced in ProxyConfig with a default value of 1 hour (60 * 60 * 1000ms), which is more reasonable for most use cases.

Fixes #8920


lollipopjin (approved):
LGTM

PR #9608, opened 2025-08-11
[ISSUE #9600] fix: remove wrong logic in the callback for sending messages in rpc

3424672656:

Fixes #9600
the logic should be executed in ClientRemotingProcessor#processReplyMessage


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -1626,7 +1626,6 @@ public Message request(final Message msg,
                 @Override
                 public void onSuccess(SendResult sendResult) {
                     requestResponseFuture.setSendRequestOk(true);
-                    requestResponseFuture.acquireCountDownLatch();
```
tianliuliu, line 1629: LGTM

tianliuliu (approved):

PR #9627, opened 2025-08-21
[ISSUE #9626] Prevent premature offset commit before consumer record flush

### LINKED ISSUE #9626: [Bug] Prevent premature offset commit before consumer record flush

Linux 5.10

### RocketMQ version

5.2-5.3.3

### JDK Version

JDK11

### Describe the Bug

It is a concurrency issue: when thread A tries to remove data from the cache, the records have not yet been flushed to disk. If thread B attempts to fetch the minimum offset from the cache and commit it at the same time, any unflushed records may be lost in the event of a crash.

### Steps to Reproduce

It is a concurrency issue: when thread A tries to remove data from the cache, the records have not yet been flushed to disk. If thread B attempts to fetch the minimum offset from the cache and commit it at the same time, any unflushed records may be lost in the event of a crash.

Prevent premature offset commit before consumer record flush

### What Did You See Instead?

consumer record flush then commit min offset

### Additional Context

_No response_

lizhimins:

Fixes #9626

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
@@ -20,13 +20,11 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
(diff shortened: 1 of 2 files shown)


yx9o (approved):

lollipopjin (approved):
LGTM

PR #9665, opened 2025-09-03
[ISSUE #9663] Adopt Maven's revision mechanism (requires Maven >= 3.5.0)

### LINKED ISSUE #9663: [Enhancement] it should use revision in pom.xml to manage dependency

### Summary

We should introduce revision in pom.xml. 

### Motivation

In this way, we don't have to modify the version in every pom.xml

we should introduce revision, refer to https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html

Nothing

### Additional Context

_No response_

cvictory:

Fixes #9663

### Brief Description

We should introduce revision in pom.xml. In this way, we don't have to modify the version in every pom.xml

just run `mvn clean install`

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -13,7 +13,7 @@
     <parent>
         <groupId>org.apache.rocketmq</groupId>
         <artifactId>rocketmq-all</artifactId>
-        <version>5.3.4-SNAPSHOT</version>
+        <version>${revision}</version>
     </parent>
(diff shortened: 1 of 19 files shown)


lizhimins:
Maven's revision mechanism (officially called CI Friendly Versions) was introduced in Maven 3.5.0.

lizhimins (approved):

PR #9736, opened 2025-09-25
[ISSUE #9735] Add TimerWheel snapshot mechanism for reliable recovery

### LINKED ISSUE #9735: [Enhancement] Add TimerWheel snapshot mechanism for reliable recovery

### Summary

Add a snapshot backup and recovery mechanism for TimerWheel to ensure reliable recovery of timer message state after broker restarts. Currently, the discrete TimerWheel state cannot be fully recovered from TimerLog alone, leading to potential data inconsistency and performance issues during recovery.

### Motivation

The current implementation has a critical limitation where TimerWheel state is not persisted, making it impossible to fully recover the timer message scheduling state after a broker restart. This creates several problems:
Data Recovery Issues: TimerWheel contains crucial scheduling information that cannot be reconstructed from TimerLog alone
Performance Impact: During recovery, the entire TimerWheel needs to be rebuilt from scratch, which is time-consuming and resource-intensive
Data Consistency: Without proper state recovery, some timer messages may be lost or incorrectly scheduled
Operational Reliability: Production environments require reliable recovery mechanisms to ensure business continuity
This enhancement is particularly important for:
High-availability deployments where broker restarts are common
Large-scale timer message processing scenarios
Production environments requiring consistent timer message delivery

Implement a comprehensive snapshot mechanism for TimerWheel with the following components:
1. Configuration Options
timerWheelSnapshotFlush: Enable/disable snapshot functionality (default: false for backward compatibility)
timerWheelDefaultFlush: Control default flush behavior (default: true)
timerWheelSnapshotIntervalMs: Configure snapshot creation interval (default: 10 seconds)
2. Snapshot Management
Snapshot Creation: Periodically backup TimerWheel state to snapshot files
Snapshot Recovery: Support recovery from the latest snapshot file during initialization
Snapshot Cleanup: Automatically clean up expired snapshot files to manage disk space
Atomic Operations: Ensure snapshot operations are atomic to prevent corruption
3. Implementation Details
Add backup() method to TimerWheel for creating snapshots
Modify TimerWheel constructor to support recovery from snapshots
Implement snapshot file selection logic based on flush positions
Add synchronization locks to ensure flush operations are atomic
Integrate snapshot functionality into TimerFlushService
4. Backward Compatibility
Default configuration maintains existing behavior
Existing deployments continue to work without changes
Gradual migration path for enabling snapshot functionality

Alternative 1: Enhanced TimerLog Recovery
Approach: Improve TimerLog to include all necessary TimerWheel state information
Why Rejected: This would require significant changes to the existing TimerLog format and could break backward compatibility. It would also increase the complexity of the logging mechanism.
Alternative 2: In-Memory State Persistence
Approach: Persist TimerWheel state to a separate state file during normal operations
Why Rejected: This approach would require frequent disk I/O operations during normal message processing, which could significantly impact performance.
Alternative 3: Database-Based State Storage
Approach: Store TimerWheel state in an external database
Why Rejected: This would introduce external dependencies and complexity, making the system harder to deploy and maintain. It would also add network latency and potential failure points.
Alternative 4: Checkpoint-Based Recovery
Approach: Use existing checkpoint mechanism to store TimerWheel state
Why Rejected: The current checkpoint mechanism is not designed for complex data structures like TimerWheel, and modifying it would affect other parts of the system.
The proposed snapshot mechanism provides the best balance of:
Performance: Minimal impact on normal operations
Reliability: Atomic operations ensure data consistency
Compatibility: Backward compatible with existing deployments
Simplicity: Clean separation of concerns without affecting existing code paths

### Additional Context

_No response_

guyinyou:
- Add snapshot backup and recovery functionality to TimerWheel
- Add configuration options for snapshot control with backward compatibility
- Implement atomic flush operations with synchronization locks
- Add snapshot file management (creation, cleanup, selection)
- Support recovery from snapshot files to avoid rebuilding TimerWheel
- Add TimerFlushService.flush() method for manual flush operations

This addresses the issue where discrete TimerWheel state cannot be recovered from TimerLog alone, ensuring data consistency and improving recovery performance.

Configuration changes:
- timerWheelSnapshotFlush: enable/disable snapshot functionality (default: false)
- timerWheelDefaultFlush: enable/disable default flush behavior (default: true)
- timerWheelSnapshotIntervalMs: snapshot creation interval (default: 10s)

Change-Id: I801ce7ca6330d02aefcae905e59e1e2b7642ebf5

Fixes #9735

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -179,9 +179,13 @@ public class TimerMessageStore {
         long snapOffset = -1;
         if (storeConfig.isTimerWheelSnapshotFlush()) {
             snapOffset = TimerWheel.getMaxSnapshotFlag(timerWheelPath);
-            // correct recover offset
-            timerCheckpoint.setLastTimerLogFlushPos(snapOffset);
-            LOGGER.info("found timerWheel snapshot offset {}", snapOffset);
+            if (snapOffset > 0) {
+                // correct recover offset
+                timerCheckpoint.setLastTimerLogFlushPos(snapOffset);
+                LOGGER.info("found timerWheel snapshot offset {}", snapOffset);
+            } else {
+                LOGGER.info("not found timerWheel snapshot", snapOffset);
+            }
         }
         this.timerWheel = new TimerWheel(
             timerWheelPath, this.slotsTotal, precisionMs, snapOffset);
@@ -1836,14 +1840,11 @@ public class TimerMessageStore {
             synchronized (lockWhenFlush) {
                 prepareTimerCheckPoint();
                 timerLog.getMappedFileQueue().flush(0);
-                if (storeConfig.isTimerWheelDefaultFlush()) {
-                    timerWheel.flush();
+                if (System.currentTimeMillis() - lastSnapshotTime > storeConfig.getTimerWheelSnapshotIntervalMs()) {
+                    lastSnapshotTime = System.currentTimeMillis();
+                    timerWheel.backup(timerLog.getMappedFileQueue().getFlushedWhere());
+                    timerCheckpoint.flush();
                 }
-            }
-            if (System.currentTimeMillis() - lastSnapshotTime > storeConfig.getTimerWheelSnapshotIntervalMs()) {
-                lastSnapshotTime = System.currentTimeMillis();
-                timerWheel.backup(timerLog.getMappedFileQueue().getFlushedWhere());
-                timerCheckpoint.flush();
             }
             if (System.currentTimeMillis() - start > storeConfig.getTimerProgressLogIntervalMs()) {
                 start = System.currentTimeMillis();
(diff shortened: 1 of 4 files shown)


store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
```diff
@@ -1805,33 +1828,48 @@ private String format(long time) {
...
         }
+
+        long start = System.currentTimeMillis();
+        long lastSnapshotTime = System.currentTimeMillis();
+
+        public void flush() throws IOException {
+            synchronized (lockWhenFlush) {
+                prepareTimerCheckPoint();
+                timerLog.getMappedFileQueue().flush(0);
+                if (storeConfig.isTimerWheelSnapshotFlush()) {
```
RongtongJin: It would be better to acquire the lock only when the timerWheelSnapshotFlush is true.

RongtongJin (dismissed):

RongtongJin (approved):

PR #9785, opened 2025-10-27
[ISSUE #9784] Fix the policy comparator to prioritize DENY over ALLOW

### LINKED ISSUE #9784: [Bug] Fix the policy comparator to prioritize DENY over ALLOW

All platform

### RocketMQ version

develop and 5.3.3 (the branch using ACL 2.0)

### JDK Version

All

### Describe the Bug

In the `AclAuthorizationHandler#comparePolicyEntries` method, if the priority of the policy cannot be distinguished by resource type or matching mode, the priority of deny should be higher than that of allow.

However, in the current code, if o1's policy is DENY and o2's policy is ALLOW, the method returns `1`. This causes o1 to be sorted after o2, resulting in incorrect logic that ALLOW takes precedence over DENY.

Image

### Steps to Reproduce

At this stage, since multiple decisions cannot be configured for the same resource, this problem does not occur in practice. We can use test code to verify this problem.

Need make `AclAuthorizationHandler#comparePolicyEntries` to public for test.
```
    AclAuthorizationHandler handler = new AclAuthorizationHandler(new AuthConfig());

    Resource resource = Resource.of(ResourceType.TOPIC, null, ResourcePattern.LITERAL);
    PolicyEntry allow = PolicyEntry.of(resource, Arrays.asList(Action.PUB), null, Decision.ALLOW);
    PolicyEntry deny = PolicyEntry.of(resource, Arrays.asList(Action.PUB), null, Decision.DENY);

    List<PolicyEntry> policyEntries = new ArrayList<>(Arrays.asList(deny, allow));
    policyEntries.sort(handler::comparePolicyEntries);
    PolicyEntry policyEntry = policyEntries.get(0);
    System.out.printf(policyEntry.getDecision().toString());
}
```

Image

Use correct policy judgment logic to give DENY a higher priority than ALLOW.

### What Did You See Instead?

If o1 is DENY and o2 is ALLOW, then ALLOW takes precedence over DENY.

### Additional Context

_No response_

majialoong:

Fixes #9784 

### Brief Description

Use correct policy judgment logic to give DENY a higher priority than ALLOW.

Verification using unit tests.

### DIFF
--- a/auth/src/main/java/org/apache/rocketmq/auth/authorization/chain/AclAuthorizationHandler.java
+++ b/auth/src/main/java/org/apache/rocketmq/auth/authorization/chain/AclAuthorizationHandler.java
@@ -154,7 +154,11 @@ public class AclAuthorizationHandler implements Handler<DefaultAuthorizationCont
         // the decision deny has higher priority
         Decision d1 = o1.getDecision();
         Decision d2 = o2.getDecision();
-        return d1 == Decision.DENY ? 1 : d2 == Decision.DENY ? -1 : 0;
+
+        if (d1 != d2) {
+            return d1 == Decision.DENY ? -1 : 1;
+        }
+        return 0;
     }
 
     private static void throwException(DefaultAuthorizationContext context, String detail) {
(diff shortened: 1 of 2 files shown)


majialoong:
Hi, @dingshuangxi888 @lizhimins. PTAL. Thanks !

dingshuangxi888 (approved):
LGTM

lizhimins (approved):

PR #9788, opened 2025-10-29
Correct method name from updateDateVersion to updateDataVersion

majialoong:
Correct method name from `updateDateVersion` to `updateDataVersion` in TimerCheckpoint.


majialoong:
Fix method name typo. Since this method is internal and not part of the public API, the typo can be corrected directly without deprecating the original method.

Hi, @RongtongJin , would you mind take a look when you are available? Thanks!

RongtongJin (approved):

PR #9810, opened 2025-11-04
[ISSUE #9809] Fix NPE in getAcl when subject is null

### LINKED ISSUE #9809: [Enhancement] Fix NPE in getAcl when subject is null

### Summary

When the getAcl method receives a null subject (for example, due to incorrect user input or a non-existent subjectType), it currently returns a raw NPE error message. This is not user-friendly. We should avoid NPE and provide a clear, helpful error message instead.

Image

### Motivation

Avoid returning raw NPE error messages directly.

Check whether the subject is null and handle the exception accordingly.This is also similar to other methods in `AuthorizationMetadataManagerImpl`, such as `createAcl` and `updateAcl`.

The improved message prompt is as follows:

Image

No

### Additional Context

_No response_

majialoong:

Fixes #9809 

### Brief Description

Provide more user-friendly error messages.

And in the `subjectFuture#thenCompose` method, use the correct `sub`, not `subject`.

unit test and local test

Image

### DIFF
--- a/auth/src/main/java/org/apache/rocketmq/auth/authorization/manager/AuthorizationMetadataManagerImpl.java
+++ b/auth/src/main/java/org/apache/rocketmq/auth/authorization/manager/AuthorizationMetadataManagerImpl.java
@@ -176,19 +176,26 @@ public class AuthorizationMetadataManagerImpl implements AuthorizationMetadataMa
 
     @Override
     public CompletableFuture<Acl> getAcl(Subject subject) {
-        CompletableFuture<? extends Subject> subjectFuture;
-        if (subject.isSubject(SubjectType.USER)) {
-            User user = (User) subject;
-            subjectFuture = this.getAuthenticationMetadataProvider().getUser(user.getUsername());
-        } else {
-            subjectFuture = CompletableFuture.completedFuture(subject);
-        }
-        return subjectFuture.thenCompose(sub -> {
-            if (sub == null) {
-                throw new AuthorizationException("The subject is not exist.");
+        try {
+            if (subject == null) {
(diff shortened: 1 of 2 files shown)


auth/src/main/java/org/apache/rocketmq/auth/authorization/manager/AuthorizationMetadataManagerImpl.java
```diff
@@ -176,19 +176,26 @@ public CompletableFuture<Void> deleteAcl(Subject subject, PolicyType policyType,
...
+                User user = (User) subject;
+                subjectFuture = this.getAuthenticationMetadataProvider().getUser(user.getUsername());
+            } else {
+                subjectFuture = CompletableFuture.completedFuture(subject);
+            }
+            return subjectFuture.thenCompose(sub -> {
+                if (sub == null) {
+                    throw new AuthorizationException("The subject is not exist.");
+                }
+                return this.getAuthorizationMetadataProvider().getAcl(sub);
```
sinberCS: It seems that when `subject` is `SubjectType.USER`, the previous logic would directly return the `subject`'s ACL (instead of the ACL retrieved twice by `getAuthenticationMetadataProvider().getUser()`), and you've modified that.

I'm not sure if this is a design flaw or just another minor bug ( from before the update ).

majialoong:
Hi, @dingshuangxi888 @lizhimins @RongtongJin When you have a moment, please take a look at this PR. Thanks !

RongtongJin (approved):

lizhimins (approved):

PR #9840, opened 2025-11-11
Fix resource leak in IndexStoreFile shutdown

RongtongJin:

Fixes #issue_id

### Brief Description

Add cleanResources() calls after shutdown() for both mappedFile and compactMappedFile to ensure proper cleanup of memory-mapped buffers and file channels.


fuyou001 (dismissed):

ShannonDing (approved):
LGTM

PR #9844, opened 2025-11-11
Removed unnecessary class casts

TheRealHaui:
Unnecessary class casts removed.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/filter/impl/PolishExpr.java
+++ b/common/src/main/java/org/apache/rocketmq/common/filter/impl/PolishExpr.java
@@ -169,11 +169,11 @@ public class PolishExpr {
     }
 
     public static boolean isLeftParenthesis(Op token) {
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/apache/rocketmq/common/filter/impl/PolishExpr.java
```diff
@@ -169,11 +169,11 @@ public static boolean isOperand(Op token) {
     }
 
     public static boolean isLeftParenthesis(Op token) {
-        return token instanceof Operator && LEFTPARENTHESIS == (Operator) token;
+        return token instanceof Operator && LEFTPARENTHESIS == token;
```
yuz10, line 172: this file is auto generated, should not change it.

RockteMQ-AI (approved):
LGTM — clean refactoring. Removing redundant explicit casts from `decodeCommandCustomHeader()` calls is safe since the method's generic return type already provides the correct type. No behavioral change.

---

PR #9866, opened 2025-11-20
[ISSUE #9663] Add flatten-maven-plugin plugin

### LINKED ISSUE #9663: [Enhancement] it should use revision in pom.xml to manage dependency

lizhimins:
Change-Id: I5cc3bcd985d75bcd70f38ce9a1b59fafa26476ef

Fixes #9663

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,6 +155,7 @@
         <rocketmq-client-java.version>5.0.5</rocketmq-client-java.version>
 
         <!-- Build plugin dependencies -->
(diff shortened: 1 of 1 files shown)


dingshuangxi888 (approved):
LGTM

lollipopjin (approved):
LGTM

PR #9878, opened 2025-11-28
add config to disable namespace stripping to support topics containing '%'

1782935682:
feat(proxy): add switch to disable automatic namespace stripping

When a topic is named in the format `Namespace%Topic` (e.g., `111%222`), the Proxy previously truncated it to just `222`. This caused message send/receive errors and ACL authentication failures due to resource name mismatches.

This commit introduces a configuration option to disable this automatic namespace parsing, allowing topics to retain their full name structure.


lizhimins:
@qianye1001 @RongtongJin Please consider

PR #9924, opened 2025-12-13
[ISSUE #9923] Transactional messages should not send custom delayed messages

### LINKED ISSUE #9923: [Enhancement] Transactional messages should not send custom delayed messages

### Summary

Transactional messages should not send custom delayed messages.

### Motivation

Transactional messages should not send custom delayed messages.

Transactional messages should not send custom delayed messages.

No

### Additional Context

_No response_

yx9o:
Fixes #9923 .

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -1442,6 +1442,15 @@ public class DefaultMQProducerImpl implements MQProducerInner {
         if (msg.getDelayTimeLevel() != 0) {
             MessageAccessor.clearProperty(msg, MessageConst.PROPERTY_DELAY_TIME_LEVEL);
         }
+        if (msg.getDelayTimeMs() != 0) {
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
```diff
@@ -1442,6 +1442,15 @@ public TransactionSendResult sendMessageInTransaction(final Message msg,
         if (msg.getDelayTimeLevel() != 0) {
             MessageAccessor.clearProperty(msg, MessageConst.PROPERTY_DELAY_TIME_LEVEL);
         }
+        if (msg.getDelayTimeMs() != 0) {
+            MessageAccessor.clearProperty(msg, MessageConst.PROPERTY_TIMER_DELAY_MS);
```
humkum: In this situation, it is recommended to throw an exception directly, so that the user can perceive the usage error.

yx9o: I've thought about that too. Currently, the processing is consistent with the existing level. But if an exception is thrown, will there be semantic differences after the user upgrades?

RongtongJin (approved):

PR #9967, opened 2025-12-26
Fix CID_DefaultHeartBeatSyncerTopic not created when autoCreateSubscriptionGroup is false.

1782935682:

The HeartbeatSyncer in Proxy relies on the Broker to automatically create the internal consumer group (CID_DefaultHeartBeatSyncerTopic). If autoCreateSubscriptionGroup is set to false on the Broker, the Syncer fails to start because the group does not exist.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/service/admin/AdminService.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/service/admin/AdminService.java
@@ -19,6 +19,7 @@ package org.apache.rocketmq.proxy.service.admin;
 
 import java.util.List;
 import org.apache.rocketmq.remoting.protocol.route.BrokerData;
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes the issue where `CID_DefaultHeartBeatSyncerTopic` is not created when `autoCreateSubscriptionGroup` is set to `false` on the Broker, causing the HeartbeatSyncer in Proxy to fail at startup.

The fix adds a `createSubscriptionGroupIfNotExist()` method to `AdminService` and calls `createSysConsumerGroup()` in `AbstractSystemMessageSyncer.start()` before creating the consumer. The implementation follows the existing pattern used for system topic creation (`createSysTopic()`), with proper retry logic and error handling.

LGTM — clean fix that ensures system consumer groups are always available regardless of the auto-creation setting.

---

Issue #9930, opened 2025-12-17
[Enhancement] In RocketMQ 5.3.4, the rejection policy of the thread pool is set redundantly in the init method of the GrpcMessagingApplication class


709572460:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

In RocketMQ 5.3.4, the rejection policy of the thread pool is set redundantly in the init method of the GrpcMessagingApplication class，the code is as follows :
protected void init() {
        GrpcTaskRejectedExecutionHandler rejectedExecutionHandler = new GrpcTaskRejectedExecutionHandler();
        this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.producerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.consumerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.clientManagerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.transactionThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
    }

### Motivation

this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler)执行了两次，没有必要。

### Describe the Solution You'd Like

the improved code is as follows :
protected void init() {
        GrpcTaskRejectedExecutionHandler rejectedExecutionHandler = new GrpcTaskRejectedExecutionHandler();
        this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.producerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.consumerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.clientManagerThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
        this.transactionThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
    }

### Describe Alternatives You've Considered

the improved code is as follows :
protected void init() {
        GrpcTaskRejectedExecutionHandler rejectedExecutionHandler = new GrpcTaskRejectedExecutionHandler();
[...]

PR #9968, opened 2025-12-27
[ISSUE #9930] Solve the problem of duplicate assignment in the init method of the GrpcMessagingApplication class

709572460:
Solve the problem of duplicate assignment in the init method of the GrpcMessagingApplication class

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/GrpcMessagingApplication.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/GrpcMessagingApplication.java
@@ -136,10 +136,6 @@ public class GrpcMessagingApplication extends MessagingServiceGrpc.MessagingServ
 
     protected void init() {
         GrpcTaskRejectedExecutionHandler rejectedExecutionHandler = new GrpcTaskRejectedExecutionHandler();
-
-        //Repeated assignment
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/GrpcMessagingApplication.java
```diff
@@ -136,7 +136,10 @@ protected GrpcMessagingApplication(GrpcMessagingActivity grpcMessagingActivity,
 
     protected void init() {
         GrpcTaskRejectedExecutionHandler rejectedExecutionHandler = new GrpcTaskRejectedExecutionHandler();
-        this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
+
+        //Repeated assignment
+        //this.routeThreadPoolExecutor.setRejectedExecutionHandler(rejectedExecutionHandler);
```
lizhimins: You can directly remove the duplicate parts.

lizhimins (commented):
You can directly remove the duplicate parts.

709572460:
I've deleted all the redundant code

proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/GrpcMessagingApplication.java
709572460: OK

PR #9981, opened 2026-01-04
[ISSUE #9980] Skip invalid records when the group is absent in Pop

### LINKED ISSUE #9980: [Enhancement] Skip invalid records when the group is absent in Pop

### Summary

Skip invalid records when the group is absent in Pop

### Motivation

Avoid generating too many invalid retry messages when the Group does not exist.

Add switch

Avoid generating a large number of invalid retry messages.

### Additional Context

Image

lizhimins:

Fixes #9980


lollipopjin (approved):
LGTM

humkum (approved):

PR #9996, opened 2026-01-08
 [ISSUE #9984] Fix: slave delete all topic config and subGroup config when sync config from master

### LINKED ISSUE #9984: [Bug] auto delete topics when in Master-Slave Automatic Failover Mode

OS: Alibaba Cloud Linux release 3 (OpenAnolis Edition). 
2C8G
1 master 2 slave

### RocketMQ version

rocketmq-all-5.3.4

### JDK Version

_No response_

### Describe the Bug

In 5.3.4, using Master-Slave Automatic Failover Mode, I have 1 master and 2 slave, I find rocketmq broker delete topics automatically. Here is my monitor graph below:

Image

The broker log writes:
2026-01-05 14:59:21 INFO BrokerControllerScheduledThread1 - delete topic config OK, topic: TopicConfig [topicName=OFFSET_MOVED_EVENT, readQueueNums=1, writeQueueNums=1, perm=RW-, topicFilterType=SINGLE_TAG, topicSysFlag=0, order=false, attributes={}]

Is this a expected behavior?How can I know if the delete behavior will not cause error?And how can I know if there is an error when syncing topics?

### Steps to Reproduce

I deploy 3 nameserver, and the configs are the same, as below:
```
#Namesrv config
listenPort = 9876
enableControllerInNamesrv = true

#controller config
controllerDLegerGroup = sg
controllerDLegerPeers = n0-10.0.3.82:9878;n1-10.1.2.219:9878;n2-10.1.1.249:9878
controllerDLegerSelfId = n0
metricsExporterType = PROM
metricsPromExporterPort = 5558
metricsPromExporterHost = 0.0.0.0
```

also deploy 3 brokers , the confgs are same  as below:
```
brokerClusterName = sg
brokerName = broker-a
brokerId = -1
deleteWhen = 04
fileReservedTime = 48
brokerRole = SLAVE
namesrvAddr = 10.0.3.82:9876;10.1.2.219:9876;10.1.1.249:9876
enableControllerMode = true
controllerAddr = 10.0.3.82:9878;10.1.2.219:9878;10.1.1.249:9878
allAckInSyncStateSet = true
listenPort = 30911
abortFile = /data/rocketmq/store/abort
storePathRootDir = /data/rocketmq/store
storePathCommitLog = /data/rocketmq/store/commitlog
storePathConsumeQueue = /data/rocketmq/store/consumequeue
storePathIndex = /data/rocketmq/store/index
storeCheckpoint = /data/rocketmq/store/checkpoint
autoCreateTopicEnable = true
enablePropertyFilter = true
messageDelayLevel=1s 5s 10s 30s 1m 2m 3m 4m 5m 6m 7m 8m 9m 10m 20m 30m 1h 2h 6h 12h 1d
metricsExporterType = PROM
metricsPromExporterPort = 5557
metricsPromExporterHost = 0.0.0.0
```

topics are not deleted automatically

### What Did You See Instead?

delete topics automatically, unexpectedly and unregularly

### Additional Context

_No response_

humkum:

Fixes #9984

### Brief Description
slave broker delete all topic config and subGroup config when sync meta data from master. The deletion operation should only be performed when the topic no longer exists on the master node and the corresponding deletion logic already exists.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/slave/SlaveSynchronize.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/slave/SlaveSynchronize.java
@@ -17,7 +17,8 @@
 package org.apache.rocketmq.broker.slave;
 
 import java.io.IOException;
-import java.util.Iterator;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -87,14 +88,13 @@ public class SlaveSynchronize {
                     ConcurrentMap<String, TopicConfig> topicConfigTable = topicConfigManager.getTopicConfigTable();
 
                     //delete
-                    Iterator<Map.Entry<String, TopicConfig>> iterator = topicConfigTable.entrySet().iterator();
-                    while (iterator.hasNext()) {
-                        Map.Entry<String, TopicConfig> entry = iterator.next();
+                    List<String> toRemoveTopic = new ArrayList<>();
+                    for (Map.Entry<String, TopicConfig> entry : topicConfigTable.entrySet()) {
                         if (!newTopicConfigTable.containsKey(entry.getKey())) {
-                            iterator.remove();
-                            topicConfigManager.deleteTopicConfig(entry.getKey());
+                            toRemoveTopic.add(entry.getKey());
                         }
                     }
+                    toRemoveTopic.forEach(topicConfigManager::deleteTopicConfig);
 
                     //update
                     newTopicConfigTable.values().forEach(topicConfigManager::putTopicConfig);
@@ -181,14 +181,14 @@ public class SlaveSynchronize {
                     ConcurrentMap<String, SubscriptionGroupConfig> newSubscriptionGroupTable =
                             subscriptionWrapper.getSubscriptionGroupTable();
                     // delete
-                    Iterator<Map.Entry<String, SubscriptionGroupConfig>> iterator = curSubscriptionGroupTable.entrySet().iterator();
-                    while (iterator.hasNext()) {
-                        Map.Entry<String, SubscriptionGroupConfig> configEntry = iterator.next();
+                    List<String> toRemoveSubscriptionGroup = new ArrayList<>();
+                    for (Map.Entry<String, SubscriptionGroupConfig> configEntry : curSubscriptionGroupTable.entrySet()) {
                         if (!newSubscriptionGroupTable.containsKey(configEntry.getKey())) {
-                            iterator.remove();
-                            subscriptionGroupManager.deleteSubscriptionGroupConfig(configEntry.getKey());
+                            toRemoveSubscriptionGroup.add(configEntry.getKey());
                         }
                     }
+                    toRemoveSubscriptionGroup.forEach(subscriptionGroupManager::deleteSubscriptionGroupConfig);
+
                     // update
                     newSubscriptionGroupTable.values().forEach(subscriptionGroupManager::putSubscriptionGroupConfig);
                     subscriptionGroupManager.setDataVersion(subscriptionWrapper.getDataVersion());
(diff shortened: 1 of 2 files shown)


broker/src/main/java/org/apache/rocketmq/broker/slave/SlaveSynchronize.java
```diff
@@ -95,7 +95,6 @@ private void syncTopicConfig() {
                         if (!newTopicConfigTable.containsKey(entry.getKey())) {
                             iterator.remove();
                         }
-                        topicConfigManager.deleteTopicConfig(entry.getKey());
```
RongtongJin: Putting deleteTopicConfig outside the if does indeed reduce efficiency because it clears all data before re-adding it (especially when there’s a large amount of metadata). However, removing it directly would prevent topicConfig entries of the RocksDB type from being deleted during synchronization. We may need to come up with a better approach to address this issue.

RongtongJin (changes requested):

broker/src/main/java/org/apache/rocketmq/broker/slave/SlaveSynchronize.java
humkum: I got that. I've revised it again. This implementation should be able to delete the deleted topics' topicConfig instances in RocksDB. Please help to review it again.

RongtongJin: > I got that. I've revised it again. This implementation should be able to delete the deleted topics' topicConfig instances in RocksDB. Please help to review it again.

Not sure whether deleting while iterating could cause any conflicts. Would it be better to first collect the items to be deleted into a list and then remove them?

humkum: > > I got that. I've revised it again. This implementation should be able to delete the deleted topics' topicConfig instances in RocksDB. Please help to review it again.
> 
> Not sure whether deleting while iterating could cause any conflicts. Would it be better to first collect the items to be deleted into a list and then remove them?

Yes, you're right, I change this again, please help to review.

RongtongJin (dismissed):

RongtongJin:
@humkum Could you resolve the conflict？

humkum:
> @humkum Could you resolve the conflict？

All conflict is resolved, please help to review again, thanks. @RongtongJin

RockteMQ-AI (commented):

### Summary
This PR fixes a bug in `SlaveSynchronize` where the slave broker could incorrectly delete topic and subscription group configs during metadata synchronization from the master. The fix replaces the `Iterator.remove()` pattern with a safer two-phase approach: first collect keys to remove, then delete them in a separate pass.

### Findings

- **[Info]** `SlaveSynchronize.java:88-99` — The two-phase removal pattern (collect-then-delete) is a good improvement over `Iterator.remove()` on `ConcurrentHashMap`. It avoids potential `ConcurrentModificationException` and makes the intent clearer.

- **[Info]** `SlaveSynchronize.java:181-193` — Same safe pattern applied to subscription group config sync. Consistent approach across both sync methods.

- **[Info]** `SlaveSynchronizeTest.java` — Two new test cases added:
  - `testSyncAllIncludesTopicConfig`: Verifies topic config is properly synced
  - `testSyncTopicConfigWithTopicDeletion`: Verifies that local-only topics are removed while new topics from master are added
  
  The deletion test is particularly valuable as it directly validates the bug fix scenario.

- **[Warning]** `SlaveSynchronize.java:96` — `toRemoveTopic.forEach(topicConfigManager::deleteTopicConfig)` does not also call `topicConfigTable.remove(key)`. The original code had `iterator.remove()` which removed from the map directly. If `deleteTopicConfig()` does not internally remove from `topicConfigTable`, stale entries may remain in memory until the next full sync. Please verify that `deleteTopicConfig()` handles the in-memory map cleanup, or add an explicit `topicConfigTable.remove(key)` call.

- **[Warning]** Same concern applies to `toRemoveSubscriptionGroup.forEach(subscriptionGroupManager::deleteSubscriptionGroupConfig)` at line ~193 — verify the in-memory map is also cleaned up.

### Suggestions

1. Consider adding a brief comment explaining why the two-phase pattern is used (to prevent `ConcurrentModificationException`), so future maintainers understand the intent.
2. The test `testSyncTopicConfigWithTopicDeletion` could be strengthened by also asserting the `DataVersion` is properly incremented after deletion, to ensure downstream consumers detect the change.

Overall the fix is sound and well-tested. The two warnings about in-memory cleanup should be verified.

---

PR #10014, opened 2026-01-16
[ISSUE #10013] Prevent AsyncLearner from being elected as Master during initialization

### LINKED ISSUE #10013: [Bug] Prevent AsyncLearner from being elected as Master during initialization

centos7

### RocketMQ version

5.3.3

### JDK Version

1.8

### Describe the Bug

Prevent AsyncLearner from being elected as Master during initialization

### Steps to Reproduce

Prevent AsyncLearner from being elected as Master during initialization

Prevent AsyncLearner from being elected as Master during initialization

### What Did You See Instead?

Prevent AsyncLearner from being elected as Master during initialization

### Additional Context

_No response_

1782935682:

Fixes #10013 

### Brief Description

AsyncLearner nodes are designed as read-only backups and should never assume the Master role. Currently, if an AsyncLearner starts and registers before other nodes, it triggers the election process and becomes Master. 

 In the brokerElect() method:
1. Check if the broker is an AsyncLearner.
2. If true, skip the active election request.
3. Instead, proactively fetch replica info from the Controller.
4. Only transition to SLAVE status when a valid Master is found; otherwise, wait/retry.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/controller/ReplicasManager.java
@@ -52,6 +52,7 @@ import org.apache.rocketmq.remoting.protocol.header.controller.register.ApplyBro
 import org.apache.rocketmq.remoting.protocol.header.controller.register.GetNextBrokerIdResponseHeader;
 import org.apache.rocketmq.remoting.protocol.header.controller.register.RegisterBrokerToControllerResponseHeader;
 import org.apache.rocketmq.store.config.BrokerRole;
+import org.apache.rocketmq.store.config.MessageStoreConfig;
 import org.apache.rocketmq.store.ha.autoswitch.AutoSwitchHAService;
 import org.apache.rocketmq.store.ha.autoswitch.BrokerMetadata;
 import org.apache.rocketmq.store.ha.autoswitch.TempBrokerMetadata;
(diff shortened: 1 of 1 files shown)


Wubabalala (commented):
Thanks for the fix — the direction is correct: preventing AsyncLearner from triggering `brokerElect()` on the broker side. I verified the bug independently and confirmed the root cause matches your analysis.

## Bug Verification

I wrote two unit tests against `DefaultElectPolicy` to confirm the controller has zero AsyncLearner awareness:

**Test 1**: syncStateSet empty + AsyncLearner has higher epoch → AsyncLearner wins election
**Test 2**: syncStateSet empty + AsyncLearner is the only candidate → AsyncLearner elected as master

Both pass on current `develop`, confirming the bug exists. (Tests available if helpful.)

## Suggestions

**1. Simplify the guard — remove the `getReplicaInfo` + `changeToSlave` logic (P1)**

The added 26 lines inside the `if (isAsyncLearner)` block duplicate what `schedulingSyncBrokerMetadata()` already does on a timer. When `brokerElect()` returns `false`, the startup path in line 200 handles it. The entire guard can be reduced to:

```java
private boolean brokerElect() {
    if (this.messageStoreConfig.isAsyncLearner()) {
        LOGGER.info("AsyncLearner broker skips election");
        return false;
    }
    // ... existing code
}
```

This avoids introducing a second code path for master discovery that could diverge from the main one.

**2. Add a unit test (P1)**

Codecov shows 5.88% patch coverage. A test that mocks `messageStoreConfig.isAsyncLearner() == true` and verifies `brokerElect()` returns `false` without calling `brokerOuterAPI.brokerElect()` would make this PR more convincing.

**3. Cover the `schedulingSyncBrokerMetadata` path (P2)**

`brokerElect()` is also called at line 663 inside `schedulingSyncBrokerMetadata()`. With the guard in place it's already covered, but worth mentioning in the PR description for completeness.

Issue #10050, opened 2026-01-28
[Enhancement] Support ChangeInvisibleTime without incrementing message reconsume times


ltamber:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Support NACK (negative acknowledgment) without incrementing message reconsume times. By adding a suspend flag, consumers can requeue messages for retry without increasing the message's reconsumeTimes counter.

### Motivation

In RocketMQ's Pop consumption mode, any NACK operation increments the message's reconsumeTimes counter. When the retry count reaches the maximum value, the message enters the Dead Letter Queue (DLQ). However, in certain business scenarios, consumers may need to
 temporarily delay processing a message, but this delay should not be considered a true retry failure.

### Describe the Solution You'd Like

Add a suspend boolean flag to the message processing pipeline. When this flag is true, message retry does not increment reconsumeTimes.

### Describe Alternatives You've Considered

No

PR #10051, opened 2026-01-28
[ISSUE #10050] Support ChangeInvisibleTime without incrementing message reconsume times

ltamber:
Change-Id: Ia37f63dd42adca940f7b452d251a604e9e5235b2

for #10050 

### Brief Description

Support NACK (negative acknowledgment) without incrementing message reconsume times. By adding a suspend flag, consumers can requeue messages for retry without increasing the message's reconsumeTimes counter.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRecord.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRecord.java
@@ -70,12 +70,20 @@ public class PopConsumerRecord {
     @JSONField(ordinal = 8)
     private String attemptId;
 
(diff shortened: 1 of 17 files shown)


qianye1001:
LGTM

RongtongJin (commented):
Saying it’s called “nack” may not be accurate. In some message queues such as RabbitMQ, “nack” refers to Negative Acknowledgement. It means the consumer is telling the broker: “I received the message, but I’m not acknowledging it as successfully processed.” Here it’s more like confirming that a retry is needed, but without incrementing the retry count.

RongtongJin (approved):

lizhimins (approved):

PR #10074, opened 2026-02-05
[ISSUE #10073] Fix StoreCheckpoint logicsMsgTimestamp to reflect flushed consume queue storetime

### LINKED ISSUE #10073: [Enhancement] Fix StoreCheckpoint logicsMsgTimestamp to reflect flushed consume queue storetime

### Summary

Make logicsMsgTimestamp in StoreCheckpoint represent the storetime of the last message whose consume queue entry has been flushed to disk, instead of the storetime of the latest entry written to consume queue in memory. Introduce logicsMsgTempTimestamp for the in-memory latest storetime and update logicsMsgTimestamp only when consume queue is actually flushed.

### Motivation

Today, logicsMsgTimestamp is updated in putMessagePositionInfoWrapper whenever a message is written to consume queue (in memory). That makes it the “latest CQ storetime in memory,” not the “latest CQ storetime that has been flushed.”
That leads to:
Recovery: Code that uses getMinTimestamp() (e.g. dispatchFromStoreTimestamp) assumes logicsMsgTimestamp reflects persisted progress. If it is ahead of what is actually on disk, recovery or time-based logic can be wrong.
Semantic misuse: Any logic that treats logicsMsgTimestamp as “last flushed CQ storetime” is incorrect.
This enhancement is necessary so that checkpoint timestamps match real persistence and recovery/time-based features behave correctly.

StoreCheckpoint
Add a volatile field logicsMsgTempTimestamp (in-memory only, not persisted in the checkpoint file).
Add getters/setters for it.
Keep logicsMsgTimestamp as the “last flushed CQ storetime” and continue to read/write it in the checkpoint file.
Write path (ConsumeQueue & BatchConsumeQueue)
In putMessagePositionInfoWrapper, after successfully appending to consume queue, call setLogicsMsgTempTimestamp(request.getStoreTimestamp()) instead of setLogicsMsgTimestamp(...).
Flush path (ConsumeQueueStore.FlushConsumeQueueService)
In doFlush, when a “thorough” flush is triggered (e.g. when flushConsumeQueueThoroughInterval has elapsed), read getLogicsMsgTempTimestamp() at the start of the flush into a local variable.
After flushing all consume queues (and any related stores), if this run was a thorough flush (flushConsumeQueueLeastPages == 0) and the captured value &gt; 0, call setLogicsMsgTimestamp(capturedValue) and then storeCheckpoint.flush() so the checkpoint file is updated.
Result: logicsMsgTimestamp is only advanced after CQ data is flushed, and it correctly represents the storetime of the last flushed consume queue entry.

Update logicsMsgTimestamp on every CQ flush (e.g. per-MappedFile)
Rejected: Would require tracking “last flushed storetime” per file or per flush batch and merging into a single checkpoint field, adding complexity without clear benefit over the current “thorough flush” approach.
Keep current behavior and document that logicsMsgTimestamp is “best effort”
Rejected: Recovery and time-based logic assume it reflects persisted state; documenting it as best-effort does not fix incorrect behavior.
Persist logicsMsgTempTimestamp in the checkpoint file
Rejected: The checkpoint is meant to record persisted progress. Persisting the in-memory temp value would blur the meaning of the file and could confuse recovery if the process crashes before a flush.

### Additional Context

_No response_

guyinyou:
…logicsMsgTempTimestamp for in-memory storetime

Change-Id: I6085bf6efaef84168ece31d080481717465f2b13

Fixes #10073

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/StoreCheckpoint.java
+++ b/store/src/main/java/org/apache/rocketmq/store/StoreCheckpoint.java
@@ -33,6 +33,7 @@ public class StoreCheckpoint {
     private final RandomAccessFile randomAccessFile;
     private final FileChannel fileChannel;
     private final MappedByteBuffer mappedByteBuffer;
+    private volatile long logicsMsgTempTimestamp = 0;
     private volatile long physicMsgTimestamp = 0;
     private volatile long logicsMsgTimestamp = 0;
     private volatile long indexMsgTimestamp = 0;
@@ -110,6 +111,14 @@ public class StoreCheckpoint {
 
     public void setLogicsMsgTimestamp(long logicsMsgTimestamp) {
         this.logicsMsgTimestamp = logicsMsgTimestamp;
+    }
+
+    public long getLogicsMsgTempTimestamp() {
+        return logicsMsgTempTimestamp;
+    }
+
+    public void setLogicsMsgTempTimestamp(long logicsMsgTempTimestamp) {
+        this.logicsMsgTempTimestamp = logicsMsgTempTimestamp;
     }
 
     public long getConfirmPhyOffset() {
--- a/store/src/main/java/org/apache/rocketmq/store/StoreCheckpoint.java
+++ b/store/src/main/java/org/apache/rocketmq/store/StoreCheckpoint.java
@@ -33,7 +33,7 @@ public class StoreCheckpoint {
     private final RandomAccessFile randomAccessFile;
     private final FileChannel fileChannel;
     private final MappedByteBuffer mappedByteBuffer;
-    private volatile long logicsMsgTempTimestamp = 0;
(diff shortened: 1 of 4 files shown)


RongtongJin (dismissed):

store/src/main/java/org/apache/rocketmq/store/StoreCheckpoint.java
```diff
@@ -33,6 +33,7 @@ public class StoreCheckpoint {
     private final RandomAccessFile randomAccessFile;
     private final FileChannel fileChannel;
     private final MappedByteBuffer mappedByteBuffer;
+    private volatile long logicsMsgTempTimestamp = 0;
```
RongtongJin: tmpLogicsMsgTimestamp maybe better

RongtongJin (approved):

PR #10115, opened 2026-02-27
[ISSUE #10114] Support delete consumer offset by consumer group and t…

### LINKED ISSUE #10114: [Enhancement] Support delete consumer offset by consumer group and topic

### Summary

Support delete consumer offset by consumer group and topic.

### Motivation

When a consumer group subscribes to an unintended topic by mistake, the client will continuously report and update the consumer offset to the broker, causing the broker to persist the offset for this erroneous subscription. Since deleting the topic or the consumer group entirely is not desirable. other subscriptions under the same group or other groups consuming the same topic should not be affected. users need the ability to precisely clean up the consumer offset for a specific topic under a specific consumer group, without removing the topic or the group itself.

1. add a new request code.
2. provide a new sub command to delete consumer offset.

No

### Additional Context

_No response_

echooymxq:

Fixes #10114

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -41,7 +41,7 @@ maven_install(
     artifacts = [
         "junit:junit:4.13.2",
         "com.alibaba:fastjson:1.2.76",
-        "com.alibaba.fastjson2:fastjson2:2.0.43",
+        "com.alibaba.fastjson2:fastjson2:2.0.59",
         "org.hamcrest:hamcrest-library:1.3",
         "io.netty:netty-all:4.1.65.Final",
         "org.assertj:assertj-core:3.22.0",
@@ -112,7 +112,7 @@ maven_install(
         "com.alipay.sofa:hessian:3.3.6",
         "io.netty:netty-tcnative-boringssl-static:2.0.48.Final",
         "org.mockito:mockito-junit-jupiter:4.11.0",
-        "com.alibaba.fastjson2:fastjson2:2.0.43",
+        "com.alibaba.fastjson2:fastjson2:2.0.59",
         "org.junit.jupiter:junit-jupiter-api:5.9.1",
     ],
     fetch_sources = True,
--- a/pom.xml
+++ b/pom.xml
@@ -105,7 +105,7 @@
         <netty.tcnative.version>2.0.53.Final</netty.tcnative.version>
         <bcpkix-jdk18on.version>1.83</bcpkix-jdk18on.version>
         <fastjson.version>1.2.83</fastjson.version>
-        <fastjson2.version>2.0.43</fastjson2.version>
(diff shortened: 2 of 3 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -2257,6 +2258,28 @@ public void deleteSubscriptionGroup(final String addr, final String groupName, f
...
+        final long timeoutMillis)
+        throws RemotingException, InterruptedException, MQClientException {
+        DeleteConsumerOffsetRequestHeader requestHeader = new DeleteConsumerOffsetRequestHeader();
+        requestHeader.setConsumerGroup(group);
+        requestHeader.setTopic(topic);
+        RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.DELETE_CONSUMER_OFFSET, requestHeader);
+
+        RemotingCommand response = this.remotingClient.invokeSync(MixAll.brokerVIPChannel(this.clientConfig.isVipChannelEnabled(), addr),
+            request, timeoutMillis);
+        assert response != null;
```
Kris20030907, line 2271: Using `assert` to check for null values ​​is not ideal. Users must add the JVM startup parameter `-ea` during configuration for it to take effect; otherwise, there is a risk of NPE errors.

echooymxq: @Kris20030907 There are many similar issues in MQClientAPIImpl. We may need to open a separate issue for discussion.

Kris20030907: Yep, I think this is necessary; other projects have encountered similar issues in production environments.
cc.@RongtongJin

echooymxq: @Kris20030907 From your perspective, i'd like to ask whether this features is necessary for users. we have encountered many users with similar demands in the production environment.

Kris20030907: This is just an optimization point. Without adding the `-ea` parameter, the assert logic here will not take effect. People who are unaware of this feature might be misled when troubleshooting, because IDEA automatically adds this parameter, which may differ from the actual environment.

echooymxq: @Kris20030907 The feature I'm referring to specifically is `delete consumer offset`.😂

Kris20030907: got it, No users have raised this request yet, but I think it could be a good option.

PR #10153, opened 2026-03-03
[refactor] Use isEmpty() instead of size() == 0 in PopBufferMergeService

Senrian:
Use isEmpty() method for better readability and performance.

Signed-off-by: Senrian <senrian@github.com>


RongtongJin:
Hi @Senrian, thank you very much for submitting the pull requests. Could you squash these refactoring PRs into a single PR? I’ll close some of the current ones first—there are too many and it’s flooding the feed.

PR #10175, opened 2026-03-19
[ISSUE #10076] Make orderly resetOffset wait on consume lock while preserving timeout semantics

### LINKED ISSUE #10076: [Enhancement] Consumer may wait shorter time after receive reset offset request

### Summary

```
    public synchronized void resetOffset(String topic, String group, Map<MessageQueue, Long> offsetTable) {
        DefaultMQPushConsumerImpl consumer = null;
        try {
            ...

            ConcurrentMap<MessageQueue, ProcessQueue> processQueueTable = consumer.getRebalanceImpl().getProcessQueueTable();
            for (Map.Entry<MessageQueue, ProcessQueue> entry : processQueueTable.entrySet()) {
                MessageQueue mq = entry.getKey();
                if (topic.equals(mq.getTopic()) && offsetTable.containsKey(mq)) {
                    ProcessQueue pq = entry.getValue();
                    pq.setDropped(true);
                    pq.clear();
                }
            }

            try {
                TimeUnit.SECONDS.sleep(10);
            } catch (InterruptedException ignored) {
            }
            ...
```
Consumer need to wait fixed 10 seconds after recieve broker's reset offset request, which is neither elegant nor efficient. The current implementation could cause badly consumption pause with multiple and dense request. We may check each process queue's status with a max waiting time until they are all clear.

### Motivation

To polish code implementation and avoid perhaps performance side effect.

Implement a neat polling for every process queue.

More complicated logic like `CountDownLatch` may help too.

### Additional Context

_No response_

LystranG:

- Fixes #10076 

### Brief Description

This change mainly optimizes the execution of resetOffset on the client side for orderly consumption to reduce the waiting time (during this period, the consumer is in a suspended state and should be resumed as soon as possible).

For orderly consumption, resetOffset no longer enforces a mandatory 10-second wait. Instead, it reuses the writeLock of the ProcessQueue. If the lock is successfully acquired, it indicates that no other thread is currently holding the ProcessQueue, and subsequent operations can be safely executed. Otherwise, it waits for the lock to be released (up to a maximum of 10 seconds, which is consistent with the original logic).

For concurrent consumption, the logic currently remains the same as before, i.e., waiting for 10 seconds. The current idea is to define a counter variable in the ProcessQueue to track the number of threads currently holding it. However, this approach would be intrusive to the codebase and require modifications to ConsumeMessageConcurrentlyService.

image

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -96,6 +96,7 @@ import static org.apache.rocketmq.remoting.rpc.ClientMetadata.topicRouteData2End
 
 public class MQClientInstance {
     private final static long LOCK_TIMEOUT_MILLIS = 3000;
+    private final static long RESET_OFFSET_MAX_WAIT = 10;
     private final static Logger log = LoggerFactory.getLogger(MQClientInstance.class);
     private final ClientConfig clientConfig;
     private final String clientId;
@@ -1380,9 +1381,11 @@ public class MQClientInstance {
                 }
             }
 
-            try {
-                TimeUnit.SECONDS.sleep(10);
-            } catch (InterruptedException ignored) {
+            if (!consumer.isConsumeOrderly()) {
+                try {
+                    TimeUnit.SECONDS.sleep(RESET_OFFSET_MAX_WAIT);
+                } catch (InterruptedException ignored) {
+                }
             }
 
             Iterator<MessageQueue> iterator = processQueueTable.keySet().iterator();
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
```diff
@@ -1391,8 +1394,10 @@ public synchronized void resetOffset(String topic, String group, Map<MessageQueu
                 Long offset = offsetTable.get(mq);
                 if (topic.equals(mq.getTopic()) && offset != null) {
                     try {
+                        ProcessQueue pq = processQueueTable.get(mq);
+                        waitResetOffsetReady(consumer, pq);
```
lizhimins, line 1398: Note: The purpose of acquiring a write lock is to wait for the currently ongoing sequential consumption process to complete.

lizhimins (approved):

PR #10179, opened 2026-03-20
[ISSUE #xxxx] Fix thread-safe issue in ConsumerManager.topicGroupTable

cherryMJY:
Brief Description
The topicGroupTable in ConsumerManager uses HashSet to store consumer groups, which is not thread-safe. When multiple consumers concurrently register with the same topic via heartbeat requests, HashSet.add() may lose entries due to race conditions.

Concurrent scenarios that trigger this bug:

Multiple consumers start up simultaneously and send heartbeat requests
Network reconnect triggers batch consumer re-registration
Proxy syncs consumer info to broker concurrently
Impact:

Consumer groups may not be recorded in topicGroupTable, causing message routing failures
In extreme cases, HashSet internal structure may be corrupted, leading to infinite loops during iteration
Solution:
Replace HashSet with ConcurrentHashMap.newKeySet() which provides thread-safe add/remove operations using CAS + synchronized mechanism. This pattern is already used in other RocketMQ components:

LiteSubscriptionRegistryImpl.liteTopic2Group
TopicList.topicList
LiteSubscription.liteTopicSet
Modified files:

broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java
@@ -236,12 +236,10 @@ public class ConsumerManager {
         }
 
         for (SubscriptionData subscriptionData : subList) {
-            Set<String> groups = this.topicGroupTable.get(subscriptionData.getTopic());
-            if (groups == null) {
-                Set<String> tmp = new HashSet<>();
-                Set<String> prev = this.topicGroupTable.putIfAbsent(subscriptionData.getTopic(), tmp);
-                groups = prev != null ? prev : tmp;
-            }
+            Set<String> groups = this.topicGroupTable.computeIfAbsent(
+                subscriptionData.getTopic(),
+                k -> ConcurrentHashMap.newKeySet()
+            );
             groups.add(group);
         }
 
@@ -287,12 +285,10 @@ public class ConsumerManager {
         }
 
         for (SubscriptionData subscriptionData : consumerGroupInfo.getSubscriptionTable().values()) {
-            Set<String> groups = this.topicGroupTable.get(subscriptionData.getTopic());
-            if (groups == null) {
-                Set<String> tmp = new HashSet<>();
-                Set<String> prev = this.topicGroupTable.putIfAbsent(subscriptionData.getTopic(), tmp);
-                groups = prev != null ? prev : tmp;
-            }
+            Set<String> groups = this.topicGroupTable.computeIfAbsent(
+                subscriptionData.getTopic(),
+                k -> ConcurrentHashMap.newKeySet()
+            );
             groups.add(group);
         }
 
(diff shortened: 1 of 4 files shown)


lizhimins (dismissed):

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -298,9 +300,30 @@ public void check(long transactionTimeout, int transactionCheckMax,
 
                         if (isNeedCheck) {
 
-                            if (!putBackHalfMsgQueue(msgExt, i)) {
+                            int retryTimes = 0;
+                            boolean putBackSuccess = false;
+                            while (retryTimes < PUT_BACK_RETRY_TIMES) {
```
lizhimins, line 305: It seems to be no need to introduce retries here.

cherryMJY: ok

RockteMQ-AI, line 305: **[Warning]** The retry logic for `putBackHalfMsgQueue` was questioned by @lizhimins ("It seems to be no need to introduce retries here") and you replied "ok", but the retry code is still present in this PR.

Please either:
1. Remove the retry logic as suggested, or
2. Explain why retries are necessary with data/evidence (e.g., observed transient failures in production)

Leaving unresolved reviewer feedback without action may delay merge.

broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
```diff
@@ -298,9 +300,30 @@ public void check(long transactionTimeout, int transactionCheckMax,
...
+                                if (putBackSuccess) {
+                                    break;
+                                }
+                                retryTimes++;
+                                if (retryTimes < PUT_BACK_RETRY_TIMES) {
+                                    try {
+                                        Thread.sleep(100L * retryTimes);
+                                    } catch (InterruptedException ignored) {
+                                    }
+                                }
```
RockteMQ-AI, line 316: **[Info]** The `catch (InterruptedException e)` block swallows the interrupt without restoring the interrupt status. Consider adding `Thread.currentThread().interrupt()` before the log statement to preserve the interrupt signal for upstream callers.

broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerManager.java
```diff
@@ -236,12 +236,10 @@ public boolean registerConsumer(final String group, final ClientChannelInfo clie
         }
 
         for (SubscriptionData subscriptionData : subList) {
-            Set<String> groups = this.topicGroupTable.get(subscriptionData.getTopic());
-            if (groups == null) {
-                Set<String> tmp = new HashSet<>();
-                Set<String> prev = this.topicGroupTable.putIfAbsent(subscriptionData.getTopic(), tmp);
-                groups = prev != null ? prev : tmp;
-            }
+            Set<String> groups = this.topicGroupTable.computeIfAbsent(
```
RockteMQ-AI, line 239: **[Info]** Good fix — `ConcurrentHashMap.newKeySet()` is the correct replacement for `HashSet` in concurrent contexts. The use of `computeIfAbsent` also simplifies the code nicely. This pattern is consistent with `LiteSubscriptionRegistryImpl.liteTopic2Group` and `TopicList.topicList` as noted in the PR description.

PR #10194, opened 2026-03-22
[ISSUE #10193] incorrect min offset returned from TieredMessageStore when local store returns -1

### LINKED ISSUE #10193: [Bug] Incorrect min offset returned from TieredMessageStore when local store returns -1

MacOS

### RocketMQ version

develop

### JDK Version

jdk8

### Describe the Bug

Minimum read offset boundary issue.

`TieredMessageStore#getMinOffsetInQueue` may return an incorrect result when the local store does not have a valid min offset but tiered storage still has readable data.

At present, the method merges the min offset from the next store and tiered store like this:

- return local min offset if tiered min offset is invalid
- otherwise return `Math.min(localMinOffset, tieredMinOffset)`

The problem is that the local store may return `-1` as a sentinel value when no valid local min offset is available. In that case, `-1` is not a real queue offset and should not participate in the final comparison.

### Steps to Reproduce

null

need to return when the tiered store offset is valid and local store don't have valid offset

### What Did You See Instead?

The effective offset of a tiered store will be ignored, and the method returns -1

### Additional Context

_No response_

Kris20030907:
… store returns -1

- Fixes #10193

### DIFF
--- a/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/TieredMessageStore.java
+++ b/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/TieredMessageStore.java
@@ -310,6 +310,9 @@ public class TieredMessageStore extends AbstractPluginMessageStore {
         if (minOffsetInTieredStore < 0) {
             return minOffsetInNextStore;
         }
+        if (minOffsetInNextStore < 0) {
+            return minOffsetInTieredStore;
+        }
         return Math.min(minOffsetInNextStore, minOffsetInTieredStore);
     }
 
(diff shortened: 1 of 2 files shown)


lizhimins:
Could you rebase the develop branch and check the CI?

Kris20030907:
> Could you rebase the develop branch and check the CI?

done. Thanks for pointing it out, plz check😺.

lizhimins:
It seems CI has encountered some issues on the Windows environment.

2026-04-03 11:29:24.074  WARN [                main] Destroy Posix FileSegment error, filePath: 212d6b50_DefaultCluster\brokerName\MessageStoreTest\0\CONSUME_QUEUE\08f90c1a00000000000000002000

Kris20030907:
> It seems CI has encountered some issues on the Windows environment.
> 
> 2026-04-03 11:29:24.074 WARN [ main] Destroy Posix FileSegment error, filePath: 212d6b50_DefaultCluster\brokerName\MessageStoreTest\0\CONSUME_QUEUE\08f90c1a00000000000000002000

The reason is that the previous unit test had an issue with handles not being released on Windows. I re-checked it and found that we just need to adjust the unit test.

Also, the latest failed CI/CD doesn't seem to be related to my changes. Please trigger it again.

PR #10200, opened 2026-03-24
[ISSUE #10199] Fix stale write timestamp leads to duplicate transaction commit

### LINKED ISSUE #10199: [Bug] Fix stale write timestamp leads to duplicate transaction commit

Linux 5.10

### RocketMQ version

before 5.4.0

### JDK Version

JDK11

### Describe the Bug

OP消息落盘延迟导致事务回查机制异常：当OP消息写入耗时≥1ms且首次发送时间落在特定时间窗口时，会因lastWriteTimestamp计算问题导致两个3秒间隔叠加，造成6秒延迟，使得已提交消息被错误回查并重复提交。

Transaction check mechanism anomaly caused by OP message flush delay: When OP message write latency is ≥1ms and the first write timestamp falls within a specific time window, a calculation issue with lastWriteTimestamp causes two 3-second intervals to be stacked, resulting in a 6-second delay that leads to already-committed messages being incorrectly rechecked and duplicate commits.

### Steps to Reproduce

Image

Transaction check mechanism anomaly caused by OP message flush delay: When OP message write latency is ≥1ms and the first write timestamp falls within a specific time window, a calculation issue with lastWriteTimestamp causes two 3-second intervals to be stacked, resulting in a 6-second delay that leads to already-committed messages being incorrectly rechecked and duplicate commits.

### What Did You See Instead?

No repeat

### Additional Context

_No response_

lizhimins:

- Fixes #10199

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
@@ -710,6 +710,7 @@ public class TransactionalMessageServiceImpl implements TransactionalMessageServ
                         // wait for the interval
                         mqContext.getTotalSize().get() < maxSize &&
                                 startTime - mqContext.getLastWriteTimestamp() < interval) {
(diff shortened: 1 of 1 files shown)


qianye1001 (approved):
LGTM

RongtongJin (approved):

PR #10226, opened 2026-03-27
[ISSUE #10225] Optimize HandleData locking mechanism to avoid unnecessary blocking

### LINKED ISSUE #10225: Throughput of in-flight message renewal degrades while some handle data locks have expired

### Summary

In ReceiptHandleGroup.HandleData.lock(long timeoutMs), the current implementation always blocks for the full timeoutMs (default 3 seconds) on semaphore.tryAcquire(timeoutMs, ...) before checking whether the lock has expired.

This means even when the lock holder has been stuck for well over the expiration threshold (9 seconds), the calling thread still wastes up to 3 seconds waiting on the semaphore before it discovers it can force-acquire via the expiration path.

This unnecessary blocking impacts proxy throughput in scenarios like message renewal and could lead to significant duplicate consumptions due to untimely renewal.

### Motivation

Optimize in-flight message renewal throughput in lock expired scenario.

Add a non-blocking tryAcquire() as the fast path, and check the expiration condition before falling through to the blocking tryAcquire(timeout, unit)

/

### Additional Context

_No response_

redlsz:

- Fixes #10225 

### Brief Description

Add a non-blocking tryAcquire() as the fast path, and check the expiration condition before falling through to the blocking tryAcquire(timeout, unit)

Add a unit test to cover diff.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/ReceiptHandleGroup.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/ReceiptHandleGroup.java
@@ -113,22 +113,40 @@ public class ReceiptHandleGroup {
 
         public Long lock(long timeoutMs) {
             try {
-                boolean result = this.semaphore.tryAcquire(timeoutMs, TimeUnit.MILLISECONDS);
+                long expiredTimeMs = ConfigurationManager.getProxyConfig().getLockTimeoutMsInHandleGroup() * 3;
                 long currentTimeMs = System.currentTimeMillis();
+                // Try non-blocking acquire first
+                if (this.semaphore.tryAcquire()) {
+                    this.lastLockTimeMs.set(currentTimeMs);
+                    return currentTimeMs;
+                }
+                // Check if lock is already expired before blocking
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/common/ReceiptHandleGroup.java
```diff
@@ -113,22 +113,40 @@ public HandleData(MessageReceiptHandle messageReceiptHandle) {
 
         public Long lock(long timeoutMs) {
             try {
-                boolean result = this.semaphore.tryAcquire(timeoutMs, TimeUnit.MILLISECONDS);
+                long expiredTimeMs = ConfigurationManager.getProxyConfig().getLockTimeoutMsInHandleGroup() * 3;
```
qianye1001: Put it below tryAcquire

redlsz: > Put it below tryAcquire

Fixed

Issue #10218, opened 2026-03-26
[Bug] Resource leak in DumpCompactionLogCommand - RandomAccessFile and FileChannel never closed


daguimu:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

All platforms

### RocketMQ version

develop branch (latest)

### JDK Version

JDK 8+

### Describe the Bug

In `DumpCompactionLogCommand.java`, a `RandomAccessFile` is created to obtain a `FileChannel`, but neither the `RandomAccessFile` nor the `FileChannel` is ever closed:

```java
FileChannel fileChannel = new RandomAccessFile(fileName, "rw").getChannel();
ByteBuffer buf = fileChannel.map(MapMode.READ_WRITE, 0, fileSize);
// ... use buf ...
UtilAll.cleanBuffer(buf);
// fileChannel and RandomAccessFile are never closed
```

The anonymous `RandomAccessFile` instance is immediately discarded after calling `.getChannel()`, making it impossible to close later. The `FileChannel` obtained from it is also never closed. This leaks file descriptors.

### Steps to Reproduce

Run the `DumpCompactionLogCommand` tool to dump a compaction log file. After execution, the file descriptor remains open.

### What Did You Expect to See?

The `RandomAccessFile` and `FileChannel` should be properly closed after use, ideally using try-with-resources.

### What Did You See Instead?

File descriptors are leaked because neither resource is closed.

### Additional Context

The fix should use try-with-resources to ensure both the `RandomAccessFile` and `FileChannel` are properly closed.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The resource leak in `DumpCompactionLogCommand` is valid. The `RandomAccessFile` is created inline and immediately discarded after `.getChannel()`, making it impossible to close. The `FileChannel` is also never closed.

**Root Cause:** Missing try-with-resources for `RandomAccessFile` and `FileChannel`.
**Impact:** File descriptor leak each time the compaction log dump tool is invoked.
[...]

PR #10231, opened 2026-03-30
[Bug] Fix resource leak: properly close RandomAccessFile and FileChannel using try-with-resources

Senrian:
## Bug Description

In `DumpCompactionLogCommand.java`, a `RandomAccessFile` is created to obtain a `FileChannel`, but neither the `RandomAccessFile` nor the `FileChannel` is ever closed. This causes a file descriptor leak.

**Issue:** #10218

```java
// Before (buggy):
FileChannel fileChannel = new RandomAccessFile(fileName, "rw").getChannel();
ByteBuffer buf = fileChannel.map(MapMode.READ_WRITE, 0, fileSize);
// ...
UtilAll.cleanBuffer(buf);
// fileChannel and RandomAccessFile are never closed
```

## Fix

Use try-with-resources to ensure both `RandomAccessFile` and `FileChannel` are properly closed:

```java
try (RandomAccessFile raf = new RandomAccessFile(fileName, "rw");
     FileChannel fileChannel = raf.getChannel()) {
    // ...
}
```

This follows the recommended Java practice for resource management.

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/message/DumpCompactionLogCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/message/DumpCompactionLogCommand.java
@@ -69,9 +69,9 @@ public class DumpCompactionLogCommand implements SubCommand {
                 throw new SubCommandException("file " + fileName + " is a directory.");
             }
 
-            try {
+            try (RandomAccessFile raf = new RandomAccessFile(fileName, "rw");
+                 FileChannel fileChannel = raf.getChannel()) {
                 long fileSize = Files.size(filePath);
-                FileChannel fileChannel = new RandomAccessFile(fileName, "rw").getChannel();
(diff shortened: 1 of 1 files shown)


lizhimins (commented):
use read only mode

tools/src/main/java/org/apache/rocketmq/tools/command/message/DumpCompactionLogCommand.java
```diff
@@ -69,9 +69,9 @@ public void execute(CommandLine commandLine, Options options, RPCHook rpcHook)
                 throw new SubCommandException("file " + fileName + " is a directory.");
             }
 
-            try {
+            try (RandomAccessFile raf = new RandomAccessFile(fileName, "rw");
```
lizhimins: use read only mode

Issue #10214, opened 2026-03-26
[Bug] TOCTOU race condition in MQClientInstance.brokerVersionTable access


daguimu:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

All platforms

### RocketMQ version

develop branch (latest)

### JDK Version

JDK 8+

### Describe the Bug

In `MQClientInstance.java`, the `brokerVersionTable` (a `ConcurrentHashMap`) is accessed using a non-atomic check-then-act pattern in multiple locations:

**Location 1 - `sendHeartbeatToBroker()` (line ~657):**
```java
if (!this.brokerVersionTable.containsKey(brokerName)) {
    this.brokerVersionTable.put(brokerName, new ConcurrentHashMap<>(4));
}
this.brokerVersionTable.get(brokerName).put(addr, version);
```

**Location 2 - `sendHeartbeatToAllBrokerV2()` (line ~737):**
Same pattern as Location 1.

**Location 3 - `findBrokerVersion()` (line ~1304):**
```java
if (this.brokerVersionTable.containsKey(brokerName)) {
    if (this.brokerVersionTable.get(brokerName).containsKey(brokerAddr)) {
        return this.brokerVersionTable.get(brokerName).get(brokerAddr);
    }
}
```

Between `containsKey()` and `get()`, another thread could remove the entry, causing a NullPointerException when the result of `get()` is dereferenced.

Notably, a similar operation in the same class at line ~814 already uses the correct pattern:
```java
ConcurrentHashMap<String, Integer> inner = MQClientInstance.this.brokerVersionTable
    .computeIfAbsent(brokerName, k -> new ConcurrentHashMap<>(4));
inner.put(brokerAddr, version);
```

### Steps to Reproduce

The race condition occurs under concurrent heartbeat sending and broker disconnection. When a broker is removed from the version table concurrently with a heartbeat send or version lookup, a NullPointerException can occur.

### What Did You Expect to See?

Thread-safe access to `brokerVersionTable` using atomic operations like `computeIfAbsent` and local variable caching.

[...]

PR #10235, opened 2026-03-30
[Bug] Fix TOCTOU race condition in MQClientInstance.brokerVersionTable access

Senrian:
## Bug Description

Issue: #10214

In `MQClientInstance.java`, `brokerVersionTable` (a `ConcurrentHashMap`) is accessed using a non-atomic check-then-act pattern that can cause NullPointerException under concurrent access.

### Locations Fixed

1. **sendHeartbeatToBroker()** - Race between containsKey() and get()
2. **sendHeartbeatToBrokerV2()** - Same issue
3. **findBrokerVersion()** - Triple get() calls with potential NPE

### Fix Pattern

Use `computeIfAbsent` for atomic get-or-create:
```java
// Before (race condition):
if (!this.brokerVersionTable.containsKey(brokerName)) {
    this.brokerVersionTable.put(brokerName, new ConcurrentHashMap<>(4));
}
this.brokerVersionTable.get(brokerName).put(addr, version);

// After (thread-safe):
this.brokerVersionTable.computeIfAbsent(brokerName, k -> new ConcurrentHashMap<>(4))
        .put(addr, version);
```

For findBrokerVersion(), use a local variable to avoid NPE if entry is removed between calls.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -654,10 +654,8 @@ public class MQClientInstance {
     private boolean sendHeartbeatToBroker(long id, String brokerName, String addr, HeartbeatData heartbeatData) {
         try {
             int version = this.mQClientAPIImpl.sendHeartbeat(addr, heartbeatData, clientConfig.getMqClientApiTimeout());
-            if (!this.brokerVersionTable.containsKey(brokerName)) {
-                this.brokerVersionTable.put(brokerName, new ConcurrentHashMap<>(4));
-            }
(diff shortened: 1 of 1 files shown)


Kris20030907:
@Senrian This change is not recommended as it may cause performance issues. Please refer to the documentation：https://juejin.cn/post/7094561581631012878

Senrian:
Hi @Kris20030907, thank you for the review. Could you clarify which specific change you are concerned about? The `computeIfAbsent` pattern used here is a standard ConcurrentHashMap atomic operation and is actually the recommended approach for this type of race condition (check-then-act on a map entry). It avoids the performance overhead of synchronized blocks while ensuring thread-safety. Would you prefer a different approach, such as using `synchronized` blocks or accepting the race condition?

PR #10236, opened 2026-03-31
[ISSUE #10193] Fix incorrect min offset when next store returns -1

### LINKED ISSUE #10193: [Bug] Incorrect min offset returned from TieredMessageStore when local store returns -1

Wubabalala:

- Fixes #10193

### Brief Description

`TieredMessageStore.getMinOffsetInQueue()` merges the min offset from the next store and tiered store using `Math.min()`. However, when the next store has no valid data, it returns -1 as a sentinel value. This -1 is not a real offset but still participates in the `Math.min(-1, tieredOffset)` comparison, which always returns -1 and silently discards a valid tiered store offset.

The fix adds a guard for `minOffsetInNextStore < 0` before the `Math.min` call, mirroring the existing `minOffsetInTieredStore < 0` guard above it. When the next store is invalid, the tiered store offset is returned directly.

Added a test case in `TieredMessageStoreTest.testGetMinOffsetInQueue()`: set the next store to return -1 while the tiered store holds a valid offset, and verify the tiered offset is returned instead of -1. All 111 tieredstore module tests pass.

### DIFF
--- a/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/TieredMessageStore.java
+++ b/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/TieredMessageStore.java
@@ -310,6 +310,9 @@ public class TieredMessageStore extends AbstractPluginMessageStore {
         if (minOffsetInTieredStore < 0) {
             return minOffsetInNextStore;
         }
+        if (minOffsetInNextStore < 0) {
+            return minOffsetInTieredStore;
+        }
         return Math.min(minOffsetInNextStore, minOffsetInTieredStore);
     }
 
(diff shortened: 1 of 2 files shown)


Wubabalala:
Sorry, I didn't notice #10194 was already submitted — the PR wasn't linked on the issue when I checked. Our fixes are identical. Closing this in favor of #10194.

PR #10257, opened 2026-04-13
[ISSUE #10256] Fix shared mutable batchAckIndexList causing toStoreBits update failure in PopBufferMergeService

### LINKED ISSUE #10256: [Bug] Shared mutable batchAckIndexList in PopBufferMergeService causes toStoreBits update failure when enablePopBatchAck and appendAckAsync are both enabled

All platforms

### RocketMQ version

develop branch

### JDK Version

JDK 8+

### Describe the Bug

In `PopBufferMergeService.scan()`, when `enablePopBatchAck` is `true`, the method collects ACK indices into a class-level shared variable `batchAckIndexList`, then passes this same list reference to `putBatchAckToStore()`. Immediately afterwards, it calls `indexList.clear()` in the `finally `block.

When `appendAckAsync `is also `true`, `putBatchAckToStore()` performs an asynchronous write. The async callback `handleBatchAckPutMessageResult `captures the same list reference. By the time the callback executes, the list has already been cleared (or contains stale data from the next loop iteration).

As a result, `pointWrapper.getToStoreBits()` is never updated correctly. This causes `isCkDoneForFinish()` to always return `false`, preventing `PopCheckPointWrapper `objects from ever being removed from the `commitOffsets `queue (an unbounded `LinkedBlockingDeque`). Over time, this leads to unbounded memory growth and eventually OOM.

### Steps to Reproduce

Set `enablePopBatchAck=true` and `appendAckAsync=true`
Run a Pop consumer with moderate traffic
Observe that `commitOffsets `queue size grows indefinitely

`toStoreBits `should be correctly updated after async IO completes. `commitOffsets `queue should be drained normally.

### What Did You See Instead?

`toStoreBits `remains 0. `commitOffsets `queue grows without bound.

### Additional Context

Fix: Pass a defensive copy `new ArrayList<>(indexList)` instead of the shared reference.

LiyunZhang10:

- Fixes #10256 

### Brief Description

When `enablePopBatchAck `and `appendAckAsync `are both enabled, `PopBufferMergeService.scan()` passes a class-level shared `batchAckIndexList `to the async `putBatchAckToStore()`, then immediately clears the list in the `finally `block. The async callback captures the same (now-empty) list reference, causing `toStoreBits `to never be updated. This prevents checkpoint wrappers from being removed from the unbounded `commitOffsets `queue, leading to memory leak and potential OOM.

### Changes

Pass a defensive copy `new ArrayList<>(indexList)` to `putBatchAckToStore()` so the async callback captures an independent snapshot
Added unit test `testBatchAckAsyncShouldUpdateToStoreBitsAfterFix `to verify `toStoreBits `is correctly updated after async callback completes


qianye1001 (approved):
LGTM

RongtongJin (dismissed):

RongtongJin:
Could you please merge the latest `develop` branch into this PR and let CI run again? The fix is focused and already has a relevant regression test; once it is up to date and CI is green, this should be ready to merge.

PR #10259, opened 2026-04-13
[ISSUE #9583] Fix ack not working after changeInvisibleDuration in Proxy

### LINKED ISSUE #9583: [Bug] ack doesn't work after changeInvisibleDuration

RocketMQ Nameserver, Broker, Proxy ran in linux container with Podman Desktop on MacOS.
Client ran on MacOS.

### RocketMQ version

Broker 5.3.3, Client Java SDK 5.0.8.

### JDK Version

JDK 21 Temurin-21.0.4+7

### Describe the Bug

In particular sequence of operations, `ack` would not take effect nor throw exception. 

### Steps to Reproduce

1. Create a topic that has 1 read queue and 1 write queue.
2. Send 10 messages into the topic.
3. Read the messages, for the 5th message, change its invisible duration  and acknowledge it.
4. Read the topic again, with the same consumer group.

The acknowledged message would not be seen in the second read.

### What Did You See Instead?

The acknowledged message was seen in the second read.

### Additional Context

Test case:

[rocketmq-test.zip](https://github.com/user-attachments/files/21500983/rocketmq-test.zip)

LiyunZhang10:

- Fixes #9583

## Brief Description

After `changeInvisibleDuration` succeeds, `ChangeInvisibleDurationActivity` removes the old receipt handle from `ReceiptHandleManager` (line 53) but never re-registers the new handle returned by the broker. This causes two problems:

1. **Ack fails silently**: When the client sends ack using the new handle, the Proxy's `AckMessageActivity.getHandleString()` calls `removeReceiptHandle()` which returns `null` (the new handle was never stored). The ack then goes directly to the broker with only the client-side handle, bypassing the Proxy's managed handle path.

2. **Auto-renew stops**: The `ReceiptHandleManager.scheduleRenewTask()` mechanism loses track of the message entirely, since the handle was removed but never replaced. If the new invisible duration expires before the ack reaches the broker through the revive topic, the message gets redelivered — causing the duplicate consumption reported in #9583.

## Changes

- In `ChangeInvisibleDurationActivity.convertToChangeInvisibleDurationResponse()`, after a successful response, decode the new receipt handle and re-register it into `ReceiptHandleManager` via `addReceiptHandle()`
- Added unit test `testNewHandleIsReRegisteredAfterSuccess` to verify the new handle is properly registered after a successful changeInvisibleDuration

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ChangeInvisibleDurationActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ChangeInvisibleDurationActivity.java
@@ -65,7 +65,7 @@ public class ChangeInvisibleDurationActivity extends AbstractMessagingActivity {
                 MessagingProcessor.DEFAULT_TIMEOUT_MILLS,
                 request.getSuspend()
             ).thenApply(
-                ackResult -> convertToChangeInvisibleDurationResponse(ctx, request, ackResult));
+                ackResult -> convertToChangeInvisibleDurationResponse(ctx, request, ackResult, messageReceiptHandle != null));
         } catch (Throwable t) {
             future.completeExceptionally(t);
         }
@@ -73,10 +73,10 @@ public class ChangeInvisibleDurationActivity extends AbstractMessagingActivity {
     }
 
     protected ChangeInvisibleDurationResponse convertToChangeInvisibleDurationResponse(ProxyContext ctx,
-        ChangeInvisibleDurationRequest request, AckResult ackResult) {
+        ChangeInvisibleDurationRequest request, AckResult ackResult, boolean isAutoRenew) {
         if (AckStatus.OK.equals(ackResult.getStatus())) {
             String newHandleStr = ackResult.getExtraInfo();
-            if (newHandleStr != null) {
+            if (newHandleStr != null && isAutoRenew) {
                 try {
                     ReceiptHandle newHandle = ReceiptHandle.decode(newHandleStr);
                     String group = request.getGroup().getName();
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ChangeInvisibleDurationActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ChangeInvisibleDurationActivity.java
@@ -75,9 +75,25 @@ public class ChangeInvisibleDurationActivity extends AbstractMessagingActivity {
     protected ChangeInvisibleDurationResponse convertToChangeInvisibleDurationResponse(ProxyContext ctx,
         ChangeInvisibleDurationRequest request, AckResult ackResult) {
         if (AckStatus.OK.equals(ackResult.getStatus())) {
+            String newHandleStr = ackResult.getExtraInfo();
+            if (newHandleStr != null) {
+                try {
+                    ReceiptHandle newHandle = ReceiptHandle.decode(newHandleStr);
+                    String group = request.getGroup().getName();
+                    String topic = request.getTopic().getName();
+                    MessageReceiptHandle newMessageReceiptHandle = new MessageReceiptHandle(
+                        group, topic, newHandle.getQueueId(), newHandleStr,
+                        request.getMessageId(), newHandle.getOffset(), 0);
+                    messagingProcessor.addReceiptHandle(ctx,
+                        grpcChannelManager.getChannel(ctx.getClientID()),
+                        group, request.getMessageId(), newMessageReceiptHandle);
+                } catch (Exception e) {
+                    // log but do not fail the response
+                }
+            }
             return ChangeInvisibleDurationResponse.newBuilder()
                 .setStatus(ResponseBuilder.getInstance().buildStatus(Code.OK, Code.OK.name()))
-                .setReceiptHandle(ackResult.getExtraInfo())
+                .setReceiptHandle(newHandleStr)
                 .build();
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ChangeInvisibleDurationActivity.java
```diff
@@ -75,9 +75,25 @@ public CompletableFuture<ChangeInvisibleDurationResponse> changeInvisibleDuratio
...
+            String newHandleStr = ackResult.getExtraInfo();
+            if (newHandleStr != null) {
+                try {
+                    ReceiptHandle newHandle = ReceiptHandle.decode(newHandleStr);
+                    String group = request.getGroup().getName();
+                    String topic = request.getTopic().getName();
+                    MessageReceiptHandle newMessageReceiptHandle = new MessageReceiptHandle(
+                        group, topic, newHandle.getQueueId(), newHandleStr,
+                        request.getMessageId(), newHandle.getOffset(), 0);
+                    messagingProcessor.addReceiptHandle(ctx,
```
absolute8511, line 87: should only add this when auto renew is turn on

LiyunZhang10: 您好，已修复

qianye1001:
Thanks for the PR. I have a question about the underlying assumption here.

Could you elaborate on which gRPC client scenario would lead to a `changeInvisibleDuration` call followed by an `ack` on the same message? As far as I understand the current client SDK designs (both Java and other language implementations), this sequence isn't a well-established pattern:

- For **PushConsumer**, the SDK handles ack/nack internally after the message listener returns. There's no public API for the user to call `changeInvisibleDuration` mid-processing and then explicitly `ack`.
- For **SimpleConsumer**, while the user does have access to both `changeInvisibleDuration` and `ack` APIs, the typical intended use of `changeInvisibleDuration` is to extend processing time — the subsequent `ack` would be the natural follow-up, but it's worth noting that `changeInvisibleDuration` on the broker side essentially performs an ack-then-re-enqueue (creating a new CK timer message). The new receipt handle returned is meant for *further* `changeInvisibleDuration` calls or a final `ack`, but is there a documented contract guaranteeing that the proxy must maintain handle continuity across this transition?

My concern is: if no existing client SDK actually relies on calling `changeInvisibleDuration` followed by `ack` through the proxy's receipt handle management path, then this fix might be papering over a gap that should instead be addressed at the protocol/contract level. Could you provide a concrete client-side code path (from one of the official SDKs) that demonstrates this sequence passing through the proxy's `ReceiptHandleManager`?

LiyunZhang10:
> Thanks for the PR. I have a question about the underlying assumption here.
> 
> Could you elaborate on which gRPC client scenario would lead to a `changeInvisibleDuration` call followed by an `ack` on the same message? As far as I understand the current client SDK designs (both Java and other language implementations), this sequence isn't a well-established pattern:
> 
> * For **PushConsumer**, the SDK handles ack/nack internally after the message listener returns. There's no public API for the user to call `changeInvisibleDuration` mid-processing and then explicitly `ack`.
> * For **SimpleConsumer**, while the user does have access to both `changeInvisibleDuration` and `ack` APIs, the typical intended use of `changeInvisibleDuration` is to extend processing time — the subsequent `ack` would be the natural follow-up, but it's worth noting that `changeInvisibleDuration` on the broker side essentially performs an ack-then-re-enqueue (creating a new CK timer message). The new receipt handle returned is meant for _further_ `changeInvisibleDuration` calls or a final `ack`, but is there a documented contract guaranteeing that the proxy must maintain handle continuity across this transition?
> 
> My concern is: if no existing client SDK actually relies on calling `changeInvisibleDuration` followed by `ack` through the proxy's receipt handle management path, then this fix might be papering over a gap that should instead be addressed at the protocol/contract level. Could you provide a concrete client-side code path (from one of the official SDKs) that demonstrates this sequence passing through the proxy's `ReceiptHandleManager`?

和 ack 无关，CID 后新 handle 没写回 manager，scheduler 停续期，broker 就重投（#9583），和客户端调不调 ack 无关。Proxy 内部续期路径（DefaultReceiptHandleManager.java:215-219）本来就把新 handle 写回 manager，这个 PR 只是补齐用户入口这条同样的路径。

qianye1001:
SimpleConsumer 不会发起带 AutoRenew 的 receive，所以proxy不会保存handle

PushConsumer 的 receive 有 AutoRenew，所以 proxy 会保存handle，但是同时，PushConsumer 只有在希望消息进入重试时会发起 changeInversibleTime，所以不应该继续执行 AutoRenew

所以你提到的问题不存在，如果修改，反而会影响正常的客户端的重试行为。

此外，我们正在评估关闭 Proxy 对 grpc 客户端的请求 AutoRenew 的能力。这个能力的副作用往往大于收益

LiyunZhang10:
> SimpleConsumer 不会发起带 AutoRenew 的 receive，所以proxy不会保存handle
> 
> PushConsumer 的 receive 有 AutoRenew，所以 proxy 会保存handle，但是同时，PushConsumer 只有在希望消息进入重试时会发起 changeInversibleTime，所以不应该继续执行 AutoRenew
> 
> 所以你提到的问题不存在，如果修改，反而会影响正常的客户端的重试行为。
> 
> 此外，我们正在评估关闭 Proxy 对 grpc 客户端的请求 AutoRenew 的能力。这个能力的副作用往往大于收益

感谢指正，我关掉pr了先。

PR #10271, opened 2026-04-21
[ISSUE #10270] Make Pop RocksDB BlockCache size configurable via MessageStoreConfig

### LINKED ISSUE #10270: [Enhancement] Make Pop RocksDB BlockCache size configurable via MessageStoreConfig

### Summary

Make the Pop Consumer RocksDB BlockCache size configurable through `MessageStoreConfig` instead of being hardcoded, allowing operators to tune it based on their deployment's memory budget.

### Motivation

The Pop Consumer RocksDB store (`PopConsumerRocksdbStore`) creates two ColumnFamilies (default and popState), each with its own independent BlockCache. Previously, the BlockCache was hardcoded at 1024MB per ColumnFamily in `RocksDBOptionsFactory.createPopCFOptions()`, consuming ~2GB of memory even though the Pop workload primarily performs sequential reads where OS PageCache handles data block caching effectively.

For containers with limited memory budgets (e.g., 8GB pods with 4GB JVM heap), this hardcoded allocation leaves insufficient room for RocketMQ's PageCache requirements (CommitLog MappedFile, ConsumeQueue, etc.) and JVM off-heap overhead.

1. Add `popRocksdbBlockCacheSize` (default 256MB) to `MessageStoreConfig` as a configurable property.
2. Update `RocksDBOptionsFactory.createPopCFOptions(long blockCacheSize)` to accept the cache size as a parameter.
3. `PopConsumerRocksdbStore` reads the value from `MessageStoreConfig` during initialization.

- **Disabling BlockCache entirely**: RocksDB requires a non-null block cache, so this is not possible.
- **Hardcoding a smaller value (e.g., 16MB)**: This would not allow operators to adjust based on their specific workload or memory capacity.
- **Using BrokerConfig instead**: `MessageStoreConfig` is the existing home for RocksDB-related configuration and is already passed to the Pop Consumer initialization chain.

### Additional Context

The change touches the following files:
- `store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java`
- `store/src/main/java/org/apache/rocketmq/store/rocksdb/RocksDBOptionsFactory.java`
- `broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java`
- `broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java`

lizhimins:
Add popRocksdbBlockCacheSize to MessageStoreConfig with a default of 256MB, replacing the hardcoded value in RocksDBOptionsFactory.createPopCFOptions(). This allows operators to tune Pop RocksDB memory usage based on container memory budgets.

- Fixes #10270

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
@@ -46,9 +46,11 @@ public class PopConsumerRocksdbStore extends AbstractRocksDBStorage implements P
     private WriteOptions writeOptions;
     private WriteOptions deleteOptions;
     protected ColumnFamilyHandle columnFamilyHandle;
+    private final long blockCacheSize;
 
-    public PopConsumerRocksdbStore(String filePath) {
+    public PopConsumerRocksdbStore(String filePath, long blockCacheSize) {
         super(filePath);
+        this.blockCacheSize = blockCacheSize;
     }
 
     // https://www.cnblogs.com/renjc/p/rocksdb-class-db.html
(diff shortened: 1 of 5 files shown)


qianye1001:
Thanks for making the Pop RocksDB BlockCache size configurable — this is a good step toward better memory tuning in container environments.

One thing worth considering: `blockCacheSize` is only part of the memory footprint. In `createPopCFOptions()`, `writeBufferSize` (128 MB) and `maxWriteBufferNumber` (4) are still hardcoded, which means the memtable alone can consume up to 512 MB. For memory-constrained containers, this is just as significant as the block cache.

It would be great to also expose `writeBufferSize` and `maxWriteBufferNumber` (and possibly `minWriteBufferNumberToMerge`) as configurable fields in `MessageStoreConfig`, so operators can holistically tune the Pop RocksDB memory budget. Otherwise, someone who carefully tunes the block cache down may still be surprised by the fixed memtable overhead.

Happy to discuss further if needed.

qianye1001:
LGTM, thanks for the quick update!

lollipopjin (approved):
LGTM

Issue #10244, opened 2026-04-03
[Bug] Clean MemoryConsumerOrderInfo if group does not exist


imzs:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

*

### RocketMQ version

*

### JDK Version

*

### Describe the Bug

MemoryConsumerOrderInfoManager, as an internal manager, lacks external autoClean invocation and cannot perform fallback cleanup when the group is missing.

### Steps to Reproduce

.

### What Did You Expect to See?

Execute autoClean periodically.

### What Did You See Instead?

.

### Additional Context

.

salmaloel657-dotcom:
Describe the Bug
MemoryConsumerOrderInfoManager doesn’t have any external trigger for
autoClean, so it can’t clean up properly when a consumer group is missing.

What Did You Expect to See?
autoClean being executed periodically to handle stale or missing groups.

What Did You See Instead?
No cleanup happens in those cases.

kobe-yang:
I’d like to work on this issue and submit a PR.

imzs:
Thank you for the suggestion and the fix! This issue has been resolved in commit cb06a0dbce8f6709b97c3a60ac891ccf3aa12475, so we'll go ahead and close it.

PR #10272, opened 2026-04-22
[ISSUE #10244] Trigger MemoryConsumerOrderInfoManager autoClean periodically

kobe-yang:
### Summary

This PR fixes the missing periodic cleanup for `MemoryConsumerOrderInfoManager`.

Previously, the cleanup logic already existed in `autoClean()`, including the case where the subscription group no longer exists, but it was not triggered periodically for the in-memory manager path.

### Changes

- schedule periodic cleanup for memory consumer order info in broker periodic tasks
- expose a cleanup entry from `PopLiteMessageProcessor`
- keep `PopLiteLockManager` focused on lock timeout cleanup
- add unit coverage for the missing-group cleanup case

### Why

The issue is not in the cleanup rule itself, but in the missing trigger path for `MemoryConsumerOrderInfoManager.autoClean()`.

### Tests

- `git diff --check`
- `mvn -pl broker -am -Dmaven.repo.local=/tmp/rocketmq-m2 -Dtest=ConsumerOrderInfoManagerTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false compiler:compile compiler:testCompile surefire:test`

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -719,6 +719,17 @@ public class BrokerController {
             @Override
             public void run() {
                 try {
+                    BrokerController.this.popLiteMessageProcessor.cleanConsumerOrderInfo();
+                } catch (Throwable e) {
+                    LOG.error("BrokerController: failed to clean pop lite consumer order info", e);
+                }
+            }
+        }, 5, 5, TimeUnit.MINUTES);
+
+        this.scheduledExecutorService.scheduleAtFixedRate(new Runnable() {
+            @Override
+            public void run() {
+                try {
                     BrokerController.this.protectBroker();
                 } catch (Throwable e) {
                     LOG.error("BrokerController: failed to protectBroker", e);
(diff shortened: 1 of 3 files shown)


kobe-yang:
Hi maintainers, this is my first contribution here. Could you please help approve the workflows so the required checks can start running? Thanks!

qianye1001:
Thanks for the contribution. A couple of concerns after reviewing the code:

1. **autoClean() already handles missing groups.** The existing `PopLiteLockManager` already calls `autoClean()` every 5 minutes, and the `autoClean()` implementation in `QueueLevelConsumerManager` already removes entries when `containsSubscriptionGroup(group)` returns false. So the cleanup for deleted groups is already working as expected — the issue described in #10244 does not seem to exist in practice.

2. **Potential concurrency issue.** Before this PR, both `lockService.removeTimeout()` and `autoClean()` ran sequentially in the same `PopLiteLockManager` ServiceThread. After this PR, `autoClean()` is moved to `BrokerController.scheduledExecutorService` while `removeTimeout()` stays in the `PopLiteLockManager` thread. This means they now execute concurrently on different threads, which could introduce race conditions between lock timeout cleanup and order info cleanup operating on related data structures simultaneously.

kobe-yang:
Thanks for the review and for pointing this out.

After re-checking the current implementation, I agree that autoClean() is already triggered periodically by PopLiteLockManager, and the missing-group case is already handled in QueueLevelConsumerManager.autoClean().

That means this PR does not fix a real gap, and the scheduling change indeed introduces unnecessary concurrency risk. I will close this PR.

Thanks again for the careful review.

PR #10277, opened 2026-04-28
[ISSUE #10276] Fix PopConsumerService changeInvisibilityDuration losing CK record when visibilityTimeout collision

### LINKED ISSUE #10276: [Bug] PopConsumerService changeInvisibilityDuration may lose CK record when visibilityTimeout collision

All platforms

## RocketMQ version
branch: develop

## Describe the Bug
In `PopConsumerService.changeInvisibilityDuration`, the RocksDB key is composed of `visibilityTimeout(8) + groupId + topicId + queueId + offset`, where `visibilityTimeout = popTime + invisibleTime`.

When `changedPopTime + changedInvisibleTime == popTime + invisibleTime` (i.e., the new CK has the same visibilityTimeout as the old one), both ckRecord and ackRecord produce the same RocksDB key.

The current code flow:
1. `writeRecords(ckRecord)` — writes new CK (overwrites old record at same key)
2. `deleteRecords(ackRecord)` — deletes using the same key, which removes the newly written CK

This results in the message never being retried.

## Steps to Reproduce
1. Pop a message with popTime=1000, invisibleTime=5000 (visibilityTimeout=6000)
2. Call changeInvisibilityDuration with changedPopTime=3000, changedInvisibleTime=3000 (visibilityTimeout=6000)
3. The new CK and old CK have the same RocksDB key
4. After writeRecords + deleteRecords, the CK record is lost

The new CK record should persist in RocksDB so the message can be retried after the new invisibility timeout.

## What Did You See Instead?
The new CK record is deleted because the delete operation uses the same key as the write, causing the message to never be retried.

qianye1001:

- Fixes #10276

### Brief Description

In `PopConsumerService.changeInvisibilityDuration`, the RocksDB key is composed of `visibilityTimeout + groupId + topicId + queueId + offset`, where `visibilityTimeout = popTime + invisibleTime`.

When the new CK's `changedPopTime + changedInvisibleTime` equals the old CK's `popTime + invisibleTime`, both records produce the same RocksDB key. The current code first writes the new CK (which overwrites the old record at the same key), then deletes using the old CK's key — which inadvertently removes the newly written record, causing the message to never be retried.

**Fix:** Added a check before the delete operation. If the write was performed and the new CK has the same `visibilityTimeout` as the old one (same key), the delete is skipped since the write already replaced the old value in place.

- Verified compilation passes successfully
- Verified no new lint or compilation errors in the modified file
- Logic analysis confirms correctness for all scenarios:
  1. Keys differ → write new CK + delete old CK (normal path)
  2. Keys same → write overwrites old record, skip delete (fix path)
  3. Group absent (skipWrite) → skip write + delete old CK (unchanged behavior)

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -530,11 +530,9 @@ public class PopConsumerService extends ServiceThread {
         // If the new CK has the same key as the old CK (same visibilityTimeout),
         // the write already overwrites the old record in RocksDB, skip delete
         // to avoid removing the newly written record.
-        if (!skipWrite && ckRecord.getVisibilityTimeout() == ackRecord.getVisibilityTimeout()) {
-            return;
+        if (skipWrite || ckRecord.getVisibilityTimeout() != ackRecord.getVisibilityTimeout()) {
+            this.popConsumerStore.deleteRecords(Collections.singletonList(ackRecord));
         }
-
-        this.popConsumerStore.deleteRecords(Collections.singletonList(ackRecord));
     }
 
     // Use broker escape bridge to support remote read
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -511,8 +511,10 @@ public class PopConsumerService extends ServiceThread {
 
         // No need to generate new records when the group does not exist,
         // because these retry messages will not be consumed by anyone.
-        if (brokerConfig.isPopReviveSkipIfGroupAbsent() &&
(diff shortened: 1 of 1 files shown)


broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -525,6 +527,13 @@ public void changeInvisibilityDuration(long popTime, long invisibleTime, long ch
             }
         }
 
+        // If the new CK has the same key as the old CK (same visibilityTimeout),
+        // the write already overwrites the old record in RocksDB, skip delete
+        // to avoid removing the newly written record.
+        if (!skipWrite && ckRecord.getVisibilityTimeout() == ackRecord.getVisibilityTimeout()) {
```
lizhimins: 这里和下面合并吧，别取反了

lizhimins (commented):
这里和下面合并吧，别取反了

lizhimins (approved):

PR #10279, opened 2026-04-28
[ISSUE #10278] Add enableGrpcChannelReceiptHandleRenew switch to disable proxy periodic handle renewal for gRPC clients

### LINKED ISSUE #10278: [Feature] Add enableGrpcChannelReceiptHandleRenew switch to disable proxy receipt handle periodic renewal for gRPC clients

Currently, when `enableProxyAutoRenew` is enabled, the proxy uses a short `defaultInvisibleTimeMills` (~60s) to pop messages and then periodically renews the invisible time via `changeInvisibleTime` RPC calls to the broker. This introduces several problems:

1. **Handle mapping complexity**: Every `changeInvisibleTime` call returns a new receipt handle from the broker. The proxy must maintain a mapping from the original handle (known to the client) to the latest renewed handle. This adds significant memory overhead and concurrency control complexity (`ReceiptHandleGroup` with semaphore-based locking).

2. **Proxy crash makes client handles invalid**: Since the proxy pops with a short invisible time (~60s), the client receives a short-lived handle. If the proxy crashes, the renewed handle (stored only in proxy memory) is lost, and the client's original handle has already expired from the broker's perspective. The client **cannot** ack through another proxy, leading to **unavoidable duplicate consumption**.

3. **Unnecessary network overhead**: The periodic `changeInvisibleTime` calls create continuous RPC traffic between proxy and broker for every un-acked message, which is wasteful when consumers are simply processing normally.

Add a dynamic switch `enableGrpcChannelReceiptHandleRenew` (default `true` for backward compatibility) in `ProxyConfig`. This switch specifically targets gRPC (`GrpcClientChannel`) consumers:

- When `enableGrpcChannelReceiptHandleRenew = true`: Behavior remains unchanged (short invisible time + periodic renew) for gRPC clients.
- When `enableGrpcChannelReceiptHandleRenew = false`:
  - The proxy uses the consumer group's `consumeTimeoutMinute` (from `SubscriptionGroupConfig`) as the invisible time during pop for gRPC clients, instead of the short `defaultInvisibleTimeMills`.
  - The proxy **does not** periodically renew handles via `changeInvisibleTime` for `GrpcClientChannel` instances.
  - Non-gRPC channels (e.g., Remoting) are unaffected and continue to renew normally.
  - The proxy still saves handles and monitors client connectivity. When a client disconnects, the proxy uses the saved (unchanged) handle to nack/changeInvisibleTime, enabling fast message re-delivery.
  - The client's original receipt handle remains valid for the full consume timeout duration, so even if the proxy crashes, the client can reconnect to another proxy and ack successfully.

This approach preserves fast crash recovery for client failures (proxy detects disconnect and nacks) while also being resilient to proxy failures (client handle stays valid).

1. **Keep the current renew mechanism as-is**: This works but has the drawbacks described above — particularly the proxy-crash scenario where client handles become invalid.

2. **Persist renewed handles to external storage (e.g., Redis)**: This would survive proxy crashes but adds infrastructure dependency and latency. The proposed solution achieves the same resilience without any external dependency.

3. **Use a fixed long invisible time without saving handles**: This would lose the ability to nack on client disconnect, meaning client crashes would require waiting for the full invisible time to expire before re-delivery.

### Additional Context

**Failure scenario comparison:**

| Scenario | Current (renew=true) | Proposed (renew=false) |
|----------|---------------------|----------------------|
| Client crash, proxy alive | Proxy stops renewing → ~1min recovery | Proxy nacks → fast recovery |
| Proxy crash, client alive | Client's short-lived handle expired, renewed handle lost → **duplicate consumption** | Client's handle still valid → **ack succeeds via another proxy** |
| Both crash | ~1min recovery (short invisible time) | Wait for consumeTimeout (configurable, e.g. 15min) |

The probability of simultaneous proxy + client failure is low, making the proposed approach a better trade-off for most production scenarios.

qianye1001:

- Fixes #10278

### Brief Description

This PR adds a new dynamic switch `enableGrpcChannelReceiptHandleRenew` in `ProxyConfig` to allow disabling the proxy's periodic receipt handle renewal mechanism specifically for gRPC (`GrpcClientChannel`) consumers.

**Background**: The current auto-renew approach uses a short invisible time (~60s) and periodically calls `changeInvisibleTime` to extend it. While this protects against client crashes, it has a critical flaw: if the **proxy** crashes, the client's original handle (issued with short invisible time) has already expired, and the renewed handle is lost in proxy memory. The client cannot ack through another proxy, causing unavoidable duplicate consumption.

**Changes**:
1. **`ProxyConfig.enableGrpcChannelReceiptHandleRenew`** (default `true`): New boolean switch targeting only gRPC channels. When set to `false`:
   - The proxy uses the consumer group's `consumeTimeoutMinute` (from `SubscriptionGroupConfig`) as the invisible time during pop for gRPC clients, giving the client a long-lived handle.
   - The proxy skips periodic `changeInvisibleTime` renewal calls for `GrpcClientChannel` instances only.
   - Non-gRPC channels (e.g., Remoting) continue to renew normally.
   - Handle storage and client-disconnect cleanup (nack) are preserved.

2. **`ReceiveMessageActivity`**: When `enableGrpcChannelReceiptHandleRenew=false`, resolves the invisible time from the group's `consumeTimeoutMinute` instead of the short `defaultInvisibleTimeMills`.

3. **`DefaultReceiptHandleManager`**: When `enableGrpcChannelReceiptHandleRenew=false`, the scheduled task still detects client disconnects (and nacks), but skips the periodic renewal loop only for channels that are `instanceof GrpcClientChannel`.

1. Verified compilation passes (`mvn compile -pl proxy -am`).
2. Verified no IDE errors in all modified files.
3. Existing unit tests in `DefaultReceiptHandleManagerTest` and `ReceiptHandleProcessorTest` continue to pass (they test with default `enableGrpcChannelReceiptHandleRenew=true`).

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/MessageReceiptHandle.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/MessageReceiptHandle.java
@@ -37,6 +37,14 @@ public class MessageReceiptHandle {
     private final long consumeTimestamp;
     private String liteTopic;
     private volatile String receiptHandleStr;
+    /**
+     * Indicates whether this handle needs periodic renewal by the proxy.
+     * When true, the proxy will periodically call changeInvisibleTime to extend the handle's lifetime.
+     * When false, the handle was popped with a long invisibleTime (e.g., consumeTimeoutMinute) and
+     * only needs expiry cleanup. This flag is set at handle creation time based on the
+     * enableGrpcChannelReceiptHandleRenew config, ensuring safe transitions when the config changes.
+     */
+    private final boolean needRenew;
 
     public MessageReceiptHandle(String group, String topic, int queueId, String receiptHandleStr, String messageId,
         long queueOffset, int reconsumeTimes) {
@@ -45,6 +53,11 @@ public class MessageReceiptHandle {
 
     public MessageReceiptHandle(String group, String topic, int queueId, String receiptHandleStr, String messageId,
         long queueOffset, int reconsumeTimes, String liteTopic) {
+        this(group, topic, queueId, receiptHandleStr, messageId, queueOffset, reconsumeTimes, liteTopic, true);
+    }
+
+    public MessageReceiptHandle(String group, String topic, int queueId, String receiptHandleStr, String messageId,
+        long queueOffset, int reconsumeTimes, String liteTopic, boolean needRenew) {
         this.originalReceiptHandle = ReceiptHandle.decode(receiptHandleStr);
         this.group = group;
         this.topic = topic;
@@ -56,6 +69,7 @@ public class MessageReceiptHandle {
         this.reconsumeTimes = reconsumeTimes;
         this.consumeTimestamp = originalReceiptHandle.getRetrieveTime();
         this.liteTopic = liteTopic;
+        this.needRenew = needRenew;
     }
 
     @Override
@@ -168,5 +182,9 @@ public class MessageReceiptHandle {
 
     public void setLiteTopic(String liteTopic) {
         this.liteTopic = liteTopic;
+    }
+
+    public boolean isNeedRenew() {
+        return needRenew;
     }
 }
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/config/ProxyConfig.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/config/ProxyConfig.java
@@ -216,6 +216,18 @@ public class ProxyConfig implements ConfigFile {
 
     private long invisibleTimeMillisWhenClear = 1000L;
     private boolean enableProxyAutoRenew = true;
+    /**
+     * When enableProxyAutoRenew is true, this controls whether the proxy periodically renews
+     * (extends) the invisible time of receipt handles for gRPC (GrpcClientChannel) consumers
+     * via changeInvisibleTime calls to the broker.
+     * <p>
+     * When set to false, the proxy will NOT periodically renew handles for gRPC clients. Instead,
+     * it will use the consumer group's consumeTimeoutMinute as the initial invisible time during
+     * pop, and only save handles for client-disconnect cleanup (nack). This avoids the
+     * handle-mapping complexity and makes the client's original receipt handle remain valid
+     * even if the proxy crashes.
+     */
+    private boolean enableGrpcChannelReceiptHandleRenew = true;
     private int maxRenewRetryTimes = 3;
     private int renewThreadPoolNums = 2;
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (commented):

### Summary
Adds a new dynamic switch `enableGrpcChannelReceiptHandleRenew` in `ProxyConfig` to allow disabling the proxy periodic receipt handle renewal for gRPC consumers. When disabled, uses `consumeTimeoutMinute` as the initial invisible time, avoiding periodic `changeInvisibleTime` calls.

### Findings
- **[Info]** `ProxyConfig.java:219-230` — New config field `enableGrpcChannelReceiptHandleRenew` defaults to `true`, preserving backward compatibility. Well-documented Javadoc explaining the behavior.
- **[Info]** `MessageReceiptHandle.java:40-46` — `needRenew` field is `final`, set at construction time. This is correct — it captures the config state when the handle is created, preventing race conditions if the config changes dynamically.
- **[Warning]** `ReceiveMessageActivity.java:116-121` — The `needRenew` flag is computed **before** the pop call, which is correct (avoids TOCTOU race with dynamic config). However, the `getConsumeTimeoutInvisibleTime()` fallback uses `DEFAULT_CONSUME_TIMEOUT_INVISIBLE_TIME_MILLIS = 15 min` when `SubscriptionGroupConfig` is unavailable. This is reasonable but worth noting that a missing config could silently use a long invisible time.
- **[Info]** `DefaultReceiptHandleManager.java` — Correctly skips periodic renewal when `needRenew` is false, only keeping handles for client-disconnect cleanup (nack).
- **[Info]** Tests cover both renewal modes (needRenew=true/false) and the config-missing fallback. Good coverage.

### Cross-repo Note
This PR is in apache/rocketmq (proxy module). No changes needed in apache/rocketmq-clients since the behavior is transparent to the gRPC protocol.

### Verdict
Well-designed solution with proper backward compatibility. The pre-pop config snapshot prevents race conditions. LGTM.

---

lizhimins:
当前 proxy 的 auto-renew 机制使用一个短 invisibleTime（默认 60s），然后通过周期性 changeInvisibleTime RPC 来续期。这带来三个问题：handle 映射的内存开销与并发复杂度、proxy 崩溃后短 handle 过期导致重复消费、以及持续 RPC 续期的网络开销。本 PR 引入 enableGrpcChannelReceiptHandleRenew 配置开关，关闭后不再周期性续期，改用 consumer group 的 consumeTimeoutMinute 作为初始 invisibleTime，从根本上避免了 proxy crash 导致的重复消费。

PR #10290, opened 2026-05-07
[ISSUE #10289] Fix send message timeout

### LINKED ISSUE #10289: [Bug] Client message sending timeout

centos7

### RocketMQ version

4.x、5.x

### JDK Version

jdk8

### Describe the Bug

When the client enables the VIP port function (vipChannelEnabled=true), the client always sends heartbeats to the broker's non VIP port, while other operations (such as message sending ) use the VIP port. This results in the connection to the VIP port not having heartbeat maintenance, being considered idle by the broker and closed. At the same time, any message sent will timeout due to the connection being closed.

### Steps to Reproduce

producer.setVipChannelEnabled(true); 

Send a message every 2 minutes

When the VIP port function is enabled, the heartbeat should also be sent to the VIP port to maintain the active connection status of the VIP port and avoid message sending timeout caused by idle connection closure.

### What Did You See Instead?

Even if the client uses the VIP port, heartbeats always use the non-VIP port, resulting in the following issues:
No heartbeat maintenance for connections on the VIP port
The broker closes VIP port connections as it regards them as idle
The client encounters a timeout error when sending messages at the moment the connection is closed

broker：close the connection
Image

producer:   is sending a message but timeout

Image

### Additional Context

_No response_

qsrg:
- Fixes #10289


RockteMQ-AI (approved):
LGTM. The fix correctly addresses the heartbeat timeout by routing through `MixAll.brokerVIPChannel()` to derive the proper address, consistent with other `invokeSync` calls in this class. Clean 2-line fix.

---

PR #10335, opened 2026-05-18
[ISSUE #10334] Support native CqCompactionFilter with cross-platform JNI shim

### LINKED ISSUE #10334: [Enhancement] Replace native JNI CqCompactionFilter with built-in Java CompactionFilterFactory
## Before Creating the Enhancement Request

## Summary

Replace the current native C++ JNI-based `CqCompactionFilter` (used for ConsumeQueue compaction in RocksDB mode) with RocksDB Java's built-in `AbstractCompactionFilterFactory` + `RemoveConsumeQueueCompactionFilter` API.

The current implementation requires:
- A custom C++ shim library (`cq_compaction_filter.cpp`) compiled per platform (Linux `.so`, macOS `.dylib`, Windows `.dll`)
- JNI loading with complex platform detection and temp directory extraction
- Reflection to call `ColumnFamilyOptions.setCompactionFilterHandle` (a private method)
- A `NativeCqCompactionFilter` wrapper with `disOwnNativeHandle()` to avoid use-after-free
- Pre-loading the RocksDB native library to resolve `DT_NEEDED` / `LC_LOAD_DYLIB` dependencies

This should all be replaced with a pure Java `ConsumeQueueCompactionFilterFactory` that uses the public RocksDB Java API.

## Motivation

1. **Maintainability**: The native C++ shim must be recompiled for every target platform (Linux x86_64, macOS x86_64/aarch64, Windows x64). Each new platform requires a separate build toolchain and binary artifact. This is a significant maintenance burden for the community.

2. **Correctness / Safety**: The current approach uses reflection to call a private `setCompactionFilterHandle` method, which is fragile across RocksDB Java version upgrades. The `disOwnNativeHandle()` workaround to prevent use-after-free adds complexity and risk.

3. **Build simplicity**: The native library requires C++ compilation with RocksDB headers, which is not part of the standard Maven build. Contributors cannot easily build/test the compaction filter without a C++ toolchain.

4. **RocksDB Java already provides the right API**: `AbstractCompactionFilterFactory` creates a fresh filter per compaction job, which naturally solves the `minPhyOffset` update problem — no need for a JNI `setMinPhyOffset0()` call. The factory's `createCompactionFilter()` method reads the latest `minPhyOffset` from a `LongSupplier` each time compaction starts.

### 1. Add `ConsumeQueueCompactionFilterFactory`

A pure Java class extending `AbstractCompactionFilterFactory<RemoveConsumeQueueCompactionFilter>`, taking a `LongSupplier` for `minPhyOffset`. Each `createCompactionFilter()` call reads the latest offset and creates a new `RemoveConsumeQueueCompactionFilter`:

```java
public class ConsumeQueueCompactionFilterFactory
        extends AbstractCompactionFilterFactory<RemoveConsumeQueueCompactionFilter> {

    private final LongSupplier minPhyOffsetSupplier;

    public ConsumeQueueCompactionFilterFactory(LongSupplier minPhyOffsetSupplier) {
        this.minPhyOffsetSupplier = minPhyOffsetSupplier;
    }

    @Override
    public RemoveConsumeQueueCompactionFilter createCompactionFilter(Context context) {
        return new RemoveConsumeQueueCompactionFilter(minPhyOffsetSupplier.getAsLong());
    }
}
```

### 2. Wire the factory into `RocksDBOptionsFactory`

Call `setCompactionFilterFactory(factory)` on the `ColumnFamilyOptions` for the default (CQ) column family:

```java
public static ColumnFamilyOptions createCQCFOptions(MessageStore messageStore,
        ConsumeQueueCompactionFilterFactory factory) {
    return new ColumnFamilyOptions()
        // ... existing options ...
        .setCompactionFilterFactory(factory);
}
```

### 3. Remove the native C++ code and JNI infrastructure

- Delete `CqCompactionFilterJni.java`, `NativeCqCompactionFilter.java`
- Delete `cq_compaction_filter.cpp` and pre-built binaries (`.so`, `.dylib`)
- Delete `CqCompactionFilterJniTest.java`
- Remove the `rocksdbjni-dev` build dependency from `store/pom.xml`

### 4. Simplify `ConsumeQueueRocksDBStorage`

- Remove all `CqCompactionFilterJni.isLoaded()` checks and conditional logic
- Remove `triggerCompactionSync()`, `flushAll()`, `countEntries()` test-only methods
- Compaction filtering is now always active (no degraded mode when native lib fails to load)

## Describe Alternatives You've Considered

1. **Keep the native C++ approach but improve it**: Add a CI matrix to auto-compile for all platforms, fix the reflection with a RocksDB Java PR to expose the method publicly. This would work but adds CI complexity and couples to RocksDB internals.

2. **Use a Java `AbstractCompactionFilter` (not factory)**: A single filter instance shared across compaction jobs. This was the approach in an intermediate refactoring (the `NativeCqCompactionFilter` wrapper). However, the factory pattern is superior because it avoids shared mutable state (`setMinPhyOffset` updates) and lifecycle issues (filter outliving options).

3. **Do compaction filtering in application code (not in RocksDB)**: Iterate and delete expired entries manually. This is much slower than letting RocksDB filter during compaction (which is a free piggyback on existing I/O) and doesn't clean up tombstones.

## Additional Context

- `RemoveConsumeQueueCompactionFilter` is already available in the RocksDB Java library bundled with RocketMQ — no new dependencies required.
- The `CompactionFilterFactory` pattern is the recommended RocksDB approach for filters that need per-compaction state, as documented in the [RocksDB wiki](https://github.com/facebook/rocksdb/wiki/Compaction-Filter).
- This change reduces the `store` module's native code from ~300 lines of C++ to zero, and removes ~400 lines of Java JNI loading/reflection code.
- Net effect: **-1200 lines** (including native binaries), **+50 lines** of clean Java code.

lizhimins:
about: #10334 

## Summary

- Add native C++ `CqCompactionFilter` JNI shim for ConsumeQueue compaction, supporting Linux, macOS (x86_64/aarch64), and Windows platforms
- Replace the Java-based `ConsumeQueueCompactionFilterFactory` with a native compaction filter that runs entirely in RocksDB's C++ compaction threads, avoiding JNI callback overhead per key
- Enhance `manualCompaction` logging with `estimateNumKeys` before/after and reduction ratio
- Remove unused `minPhyOffset` parameter from `AbstractRocksDBStorage.manualCompaction`

## Motivation

The existing Java `CompactionFilterFactory` creates a `RemoveConsumeQueueCompactionFilter` per compaction job, which incurs JNI cross-boundary calls for every key during compaction. For ConsumeQueue databases with millions of entries, this overhead is significant.

A native C++ compaction filter:
1. Runs entirely in RocksDB's native compaction threads — zero JNI overhead per key
2. Uses atomic `minPhyOffset` threshold updated from Java via a single JNI call
3. Supports all major platforms (Linux, macOS, Windows)

## Changes

1. **`CqCompactionFilterJni.java`** — JNI loader with platform detection, RocksDB native library co-location for `DT_NEEDED`/`LC_LOAD_DYLIB` resolution
2. **`NativeCqCompactionFilter.java`** — Thin wrapper with `disOwnNativeHandle()` to prevent use-after-free when options close before DB
3. **`cq_compaction_filter.cpp`** — Native C++ filter implementation with `std::atomic<int64_t>` threshold and JNI exports
4. **Pre-built binaries** — `libcq_compaction_filter.so` (Linux), `.dylib` (macOS)
5. **`AbstractRocksDBStorage`** — Enhanced compaction logging, removed unused `minPhyOffset` param
6. **`ConsumeQueueRocksDBStorage`** — Integrates native filter on startup with graceful fallback
7. **`docs/en/Native_RocksDB.md`** — Build instructions for all platforms

## Test Plan

- [x] `CqCompactionFilterJniTest` — verifies native library loading, filter creation, minPhyOffset update, and actual compaction filtering
- [x] Existing `ConsumeQueueRocksDBStorage` integration tests pass
- [x] Manual verification on macOS and Windows

Closes #10334


lizhimins:
@lizhanhui @leizhiyuan Please help review this pr

leizhiyuan:
意见 1: loadNativeShim 未捕获 UnsatisfiedLinkError + 缺少 ARM 架构 shim

问题： loadNativeShim() 中 System.load() 调用可能抛出 UnsatisfiedLinkError（例如在 Linux aarch64 上加载 x86-64 的 .so），但当前只捕获了 IOException。由于该方法在 static {} 块中调用，未捕获的 UnsatisfiedLinkError
 会导致 ExceptionInInitializerError，使 CqCompactionFilterJni 类在整个 JVM 生命周期内永久不可用，后续任何对该类的引用都会抛出 NoClassDefFoundError。

 当前打包的原生库只有：
 - .so — 仅 x86-64
 - .dylib — 仅 arm64
 - .dll — 仅 x86-64

 这意味着在 Linux aarch64（云环境中非常常见）加载必然失败，且由于缺少 UnsatisfiedLinkError 的捕获，不会优雅降级，而是直接崩溃。

 建议修复：

 1. 在 loadNativeShim() 中增加对 UnsatisfiedLinkError 的捕获，确保加载失败时优雅降级（loaded 保持 false，系统正常运行但无 compaction filter）：

 ```java
               System.load(tempLib.getAbsolutePath());
               loaded = true;
               log.info("[CqCompactionFilterJni] Native library loaded from classpath: {}", tempLib.getAbsolutePath());
           } catch (IOException e) {
               log.error("[CqCompactionFilterJni] Failed to load native shim", e);
           } catch (UnsatisfiedLinkError e) {
               log.error("[CqCompactionFilterJni] Failed to load native shim library, "
                   + "platform may not be supported: {}", e.getMessage());
           }
 ```

 2. 补充 Linux aarch64 的 .so 文件，并在 Java 侧根据 os.arch 选择对应的库文件名。

 ────────────────────────────────────────────────────────────────────────────────

 ### English

 Issue: System.load() in loadNativeShim() can throw UnsatisfiedLinkError (e.g., when loading an x86-64 .so on Linux aarch64), but only IOException is caught. Since this method is called from a static {} initializer,
 an uncaught UnsatisfiedLinkError will propagate as ExceptionInInitializerError, making the CqCompactionFilterJni class permanently unusable for the entire JVM lifetime — any subsequent reference to the class will
 throw NoClassDefFoundError.

 The bundled native libraries currently only cover:
 - .so — x86-64 only
 - .dylib — arm64 only
 - .dll — x86-64 only

 This means on Linux aarch64 (very common in cloud environments) and macOS x86_64, loading will inevitably fail, and without catching UnsatisfiedLinkError, the system won't degrade gracefully — it will crash.

 Suggested fix:

 1. Catch UnsatisfiedLinkError in loadNativeShim() to ensure graceful degradation when loading fails (loaded stays false, system runs normally without the compaction filter):

 ```java
               System.load(tempLib.getAbsolutePath());
               loaded = true;
               log.info("[CqCompactionFilterJni] Native library loaded from classpath: {}", tempLib.getAbsolutePath());
           } catch (IOException e) {
               log.error("[CqCompactionFilterJni] Failed to load native shim", e);
           } catch (UnsatisfiedLinkError e) {
               log.error("[CqCompactionFilterJni] Failed to load native shim library, "
                   + "platform may not be supported: {}", e.getMessage());
           }
 ```

 2. Bundle a Linux aarch64 .so ， and select the correct library filename based on os.arch in the Java code.

 ────────────────────────────────────────────────────────────────────────────────


 问题： setMinPhyOffset() 方法在每次调用时以 INFO 级别记录日志。该方法由 cleanExpired() 周期性触发（通常每隔几秒到几十秒一次），在生产环境中会产生大量重复日志，增加日志存储压力且淹没真正有价值的信息。

 建议修复： 将日志级别改为 DEBUG：

 ```java
       public static void setMinPhyOffset(long minPhyOffset) {
           long ptr = NATIVE_FILTER_PTR.get();
           if (ptr != 0) {
               setMinPhyOffset0(ptr, minPhyOffset);
               log.debug("CqCompactionFilter setMinPhyOffset={}", minPhyOffset);
           }
       }
 ```

 ────────────────────────────────────────────────────────────────────────────────


 Issue: setMinPhyOffset() logs at INFO level on every invocation. This method is called periodically by cleanExpired() (typically every few seconds), which produces high-volume repetitive log output in production,
 increasing log storage costs and drowning out genuinely useful information.

 Suggested fix: Change the log level to DEBUG:

 ```java
       public static void setMinPhyOffset(long minPhyOffset) {
           long ptr = NATIVE_FILTER_PTR.get();
           if (ptr != 0) {
               setMinPhyOffset0(ptr, minPhyOffset);
               log.debug("CqCompactionFilter setMinPhyOffset={}", minPhyOffset);
           }
       }
 ```

lizhimins:
Re: Already added support for the ARM platform, the size of this log is small.

lollipopjin (approved):
LGTM

PR #10380, opened 2026-05-26
[ISSUE #9926] Avoid redundant list allocation in parsePublishMessageQueues

### LINKED ISSUE #9926: [Enhancement] massive memory allocation in the `parsePublishMessageQueues`

### Summary

During my benchmark (50K LMQ tps), the rocketmq client in my process shows massive memory allocation in the `parsePublishMessageQueues`.

Here is my profiling result about the memory allocation:

Image

### Motivation

This multiple growths of the array list introduce a great influence on the performance under the high throughput workload.

Here is a simple proposal:
- do pre-allocation for the returned list of the `parsePublishMessageQueues`.
- Just return the original `List<MessageQueue>` if the namespace is empty.

There are more complex solutions:
- Upgrade the interface about the `MessageQueueSelector` to use `size of the message queue` and  `accessor by index to MessageQueue` instead of `List<MessageQueue>`, so that no need to create another `List<MessageQueue>`.
- Use thread-local buffer array list for the parsed `List<MessageQueue>` to avoid allocating memory for any to-be-sent message.

### Additional Context

_No response_

wushiyuanmaimob:

- Fixes #9926

### Brief Description

`MQAdminImpl.parsePublishMessageQueues()` is called on every message send (hot path). Under high throughput (50K+ TPS), it creates significant GC pressure by allocating a new `ArrayList` and N new `MessageQueue` objects per call — even when namespace is empty (the common case for most users), where the copied data is identical to the original.

**Changes:**
1. When namespace is null or empty, return the original list directly (zero allocation)
2. When namespace stripping is needed, pre-allocate `ArrayList` with exact capacity to avoid internal array resizing
3. Cache the namespace string to avoid repeated `getNamespace()` calls in the loop

**Impact:** Eliminates per-message object allocation on the send path for the majority of users who don't use namespaces. For users with namespaces, reduces allocation by pre-sizing the list.

- Added unit test `assertParsePublishMessageQueuesReturnsOriginalListWhenNoNamespace` — verifies same list instance is returned when namespace is empty
- Added unit test `assertParsePublishMessageQueuesStripsNamespace` — verifies namespace stripping still works correctly
- Ran full client module test suite: 566 tests passed, 0 failures

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
@@ -157,9 +157,14 @@ public class MQAdminImpl {
     }
 
     public List<MessageQueue> parsePublishMessageQueues(List<MessageQueue> messageQueueList) {
-        List<MessageQueue> resultQueues = new ArrayList<>();
+        String namespace = this.mQClientFactory.getClientConfig().getNamespace();
+        if (namespace == null || namespace.isEmpty()) {
+            return messageQueueList;
+        }
+
+        List<MessageQueue> resultQueues = new ArrayList<>(messageQueueList.size());
         for (MessageQueue queue : messageQueueList) {
-            String userTopic = NamespaceUtil.withoutNamespace(queue.getTopic(), this.mQClientFactory.getClientConfig().getNamespace());
+            String userTopic = NamespaceUtil.withoutNamespace(queue.getTopic(), namespace);
             resultQueues.add(new MessageQueue(userTopic, queue.getBrokerName(), queue.getQueueId()));
         }
 
--- a/client/src/test/java/org/apache/rocketmq/client/impl/MQAdminImplTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/impl/MQAdminImplTest.java
@@ -44,6 +44,7 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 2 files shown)


client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
```diff
@@ -157,9 +157,14 @@ public List<MessageQueue> fetchPublishMessageQueues(String topic) throws MQClien
     }
 
     public List<MessageQueue> parsePublishMessageQueues(List<MessageQueue> messageQueueList) {
-        List<MessageQueue> resultQueues = new ArrayList<>();
+        String namespace = this.mQClientFactory.getClientConfig().getNamespace();
+        if (namespace == null || namespace.isEmpty()) {
+            return messageQueueList;
```
RongtongJin, line 162: One concern about this change is that the method semantics change from “always returning a new list” to “returning the original list when namespace is empty”.

If any caller mutates the returned list, this could accidentally mutate the original message queue list, especially if that list is reused or cached elsewhere in the send path.

Could you confirm that the return value of `parsePublishMessageQueues` is treated as read-only by all callers, or that the original list is not reused as mutable shared state?

RockteMQ-AI (approved):
### Summary
Optimizes `parsePublishMessageQueues()` to avoid unnecessary `ArrayList` and `MessageQueue` allocation on the hot send path when namespace is empty (the common case). When namespace is present, the list is pre-sized to avoid resizing. Both branches are covered by new unit tests.

LGTM — clean, well-scoped performance improvement with solid test coverage.

---

PR #10396, opened 2026-05-28
[ISSUE #10395] Fix native memory leak on TLS certificate hot-reload

qianye1001:
## Summary

Fix native memory leak caused by old `SslContext` not being released during TLS certificate hot-reload when using the OpenSSL (netty-tcnative) provider.

Fixes #10395

## Root Cause

When TLS certificates are dynamically reloaded via `FileWatchService`, the `loadSslContext()` methods in `NettyRemotingServer` and `ProxyAndTlsProtocolNegotiator` directly overwrite the `sslContext` field without releasing the old instance. Since `ReferenceCountedOpenSslContext` allocates native off-heap memory (SSL_CTX, X509 chain, EVP_PKEY), this native memory is leaked on every certificate rotation cycle.

### `remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java`
- Capture old `sslContext` before building the replacement
- Call `ReferenceCountUtil.release(oldSslContext)` after successful assignment
- Added imports for `SslContext` and `ReferenceCountUtil`

### `proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java`
- Same pattern: capture old context, release after new one is assigned
- Made `sslContext` field `volatile` for thread-safe reads from TLS handshake threads
- Added import for shaded `ReferenceCountUtil`

## Safety

- **"Build new, then release old"** ordering ensures `sslContext` is never null or prematurely released
- `ReferenceCountUtil.release()` is safe for JDK provider `SslContext` (no-op for non-ReferenceCounted)
- In-flight connections using the old context are unaffected (refCount managed by Netty channel pipeline)

## Testing

- [ ] Existing TLS tests pass
- [ ] Manual RSS monitoring during certificate rotation shows stable native memory

---
*This PR was automatically generated based on the approved fix proposal.*

PR #10399, opened 2026-05-28
[ISSUE #10398] Fix native memory leak on TLS certificate hot-reload

### LINKED ISSUE #10398: [Bug] TLS certificate hot-reload leaks native memory due to unreleased old SslContext

Linux (observed on CentOS 7 / Ubuntu 22.04)

### RocketMQ version

develop (also affects 5.x releases with TLS hot-reload enabled)

### JDK Version

JDK 8 / JDK 11, using netty-tcnative (OpenSSL provider)

### Describe the Bug

When TLS certificates are dynamically reloaded via `TlsCertificateManager` (file-watch triggered), a new `SslContext` is created but the old one is never explicitly released. Since netty-tcnative's `OpenSslContext` is reference-counted and allocates native (off-heap) memory for the certificate chain, private key, and SSL session cache, simply dereferencing the old context does not free native memory — it relies on GC finalization which may never run under low heap pressure.

This causes native memory (RSS) to grow monotonically with each certificate rotation cycle. In long-running Proxy/Broker deployments with frequent cert rotations (e.g., short-lived certificates rotated every few hours), this eventually leads to OOM kills.

### Steps to Reproduce

1. Enable TLS with OpenSSL provider (`tls.provider=OPENSSL`) on Broker or Proxy
2. Configure certificate hot-reload (`tlsCertWatchIntervalMs`)
3. Repeatedly replace the certificate files to trigger reload cycles
4. Monitor native memory (RSS or `jcmd VM.native_memory`) — it grows on each reload and never reclaims

Native memory should remain stable after certificate rotation. The old `SslContext` should be released promptly when replaced.

### What Did You See Instead?

Native memory grows ~200KB–1MB per rotation cycle (depending on cert chain length and session cache size) and is never reclaimed until process restart.

### Additional Context

The fix should call `ReferenceCountUtil.release(oldSslContext)` after the new context is installed. Care is needed to defer release until in-flight channels using the old context have closed, or use `ReferenceCountUtil.safeRelease()` with proper draining logic.

Related: #10302 (SNI multi-domain support) introduces more `SslContext` instances per domain, making this leak more severe if not addressed.

qianye1001:
## Summary

Fix native memory leak caused by old `SslContext` not being released during TLS certificate hot-reload when using the OpenSSL (netty-tcnative) provider.

Fixes #10398

## Root Cause

When TLS certificates are dynamically reloaded via `FileWatchService`, the `loadSslContext()` methods in `NettyRemotingServer` and `ProxyAndTlsProtocolNegotiator` directly overwrite the `sslContext` field without releasing the old instance. Since `ReferenceCountedOpenSslContext` allocates native off-heap memory for the certificate chain, private key, and SSL session cache, each reload leaks ~100KB-1MB of native memory per rotation cycle.

### File 1: `remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java`
- Added `import io.netty.util.ReferenceCountUtil;`
- Modified `loadSslContext()`: build new context into local variable, save old reference, assign new context to field, then release old context via `ReferenceCountUtil.release(oldSslContext)` in try-catch

### File 2: `proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java`
- Added `import io.grpc.netty.shaded.io.netty.util.ReferenceCountUtil;`
- Changed `sslContext` field to `private static volatile SslContext sslContext;` for thread-safe visibility
- Modified `loadSslContext()`: build new context into local variable, save old reference, assign new context to field, then release old context via `ReferenceCountUtil.release(oldSslContext)` in try-catch

## Fix Strategy

Uses **"build new, then release old"** ordering to ensure `sslContext` is never null or pointing to a released context during the swap:

1. Build new `SslContext` into a local variable
2. Save old `sslContext` reference  
3. Assign new context to the field
4. Release old context via `ReferenceCountUtil.release()` (no-op for non-refcounted JDK SslContext)

## Testing

- Existing TLS integration tests cover handshake correctness
- Manual verification: run broker with TLS enabled, trigger cert reload 100+ times, confirm stable RSS/native memory

## Backward Compatibility

- No public API changes
- No configuration changes  
- No new dependencies (`ReferenceCountUtil` already in Netty transitive deps)
- Existing connections unaffected (SslHandler holds its own reference)

## Risk Assessment

**LOW** — Minimal, well-isolated lifecycle fix following established Netty ReferenceCounted resource management patterns.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java
@@ -141,6 +141,14 @@ public class ProxyAndTlsProtocolNegotiator implements InternalProtocolNegotiator
         SslContext oldSslContext = sslContext;
         sslContext = newSslContext;
         if (oldSslContext != null) {
+            // Release the old SslContext to free native memory (OpenSSL provider only).
+            // ReferenceCountUtil.release() is a no-op for JDK SslContext since it does not
+            // implement ReferenceCounted.
+            // Note: there is a theoretical race where an event-loop thread could read the old
+            // sslContext (volatile) and call newHandler() after release. In practice this is
+            // negligible because cert reload is very infrequent and the window is nanoseconds.
+            // Worst case: the single new connection gets an IllegalReferenceCountException and
+            // the client retries successfully — no pod crash or service disruption.
             try {
                 ReferenceCountUtil.release(oldSslContext);
                 log.info("Old SslContext released for proxy server");
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/ProxyAndTlsProtocolNegotiator.java
@@ -45,6 +45,7 @@ import io.grpc.netty.shaded.io.netty.handler.ssl.util.InsecureTrustManagerFactor
 import io.grpc.netty.shaded.io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.grpc.netty.shaded.io.netty.util.AsciiString;
 import io.grpc.netty.shaded.io.netty.util.CharsetUtil;
+import io.grpc.netty.shaded.io.netty.util.ReferenceCountUtil;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -77,7 +78,7 @@ public class ProxyAndTlsProtocolNegotiator implements InternalProtocolNegotiator
      */
     private static final int SSL_RECORD_HEADER_LENGTH = 5;
 
-    private static SslContext sslContext;
+    private static volatile SslContext sslContext;
 
     public ProxyAndTlsProtocolNegotiator() {
         try {
@@ -113,9 +114,10 @@ public class ProxyAndTlsProtocolNegotiator implements InternalProtocolNegotiator
             provider = SslProvider.JDK;
             log.info("Using JDK SSL provider");
         }
+        SslContext newSslContext;
         if (proxyConfig.isTlsTestModeEnable()) {
             SelfSignedCertificate selfSignedCertificate = new SelfSignedCertificate();
-            sslContext = GrpcSslContexts.forServer(selfSignedCertificate.certificate(), selfSignedCertificate.privateKey())
+            newSslContext = GrpcSslContexts.forServer(selfSignedCertificate.certificate(), selfSignedCertificate.privateKey())
                 .sslProvider(provider)
                 .trustManager(InsecureTrustManagerFactory.INSTANCE)
                 .clientAuth(ClientAuth.NONE)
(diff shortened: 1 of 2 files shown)


qianye1001 (commented):
## Review by rocketmq-reviewer-bot

### Summary
Fix native memory leak in TLS certificate hot-reload by releasing the old `SslContext` via `ReferenceCountUtil.release()` after assigning the new one. Uses a safe "build new → swap → release old" pattern with try-catch protection.

### Comparison with Previous Attempt (PR #10396, closed)
This PR improves upon the earlier closed PR #10396 in two important ways:
1. **Builds new context into a local variable first** — if `buildSslContext()` throws, the old context remains valid (no partial state)
2. **Wraps `release()` in try-catch** — prevents release failures from disrupting the reload flow

### Findings

- **[Info]** `NettyRemotingServer.java:187-196` — Release ordering is correct. Building into `newSslContext` local var, saving `oldSslContext`, assigning new to field, then releasing old. This is the standard Netty `ReferenceCounted` swap pattern. ✅

- **[Info]** `NettyRemotingServer.java` — `sslContext` is declared as `protected volatile` in parent class `NettyRemotingAbstract` (line 121), ensuring visibility across threads. ✅

- **[Info]** `ProxyAndTlsProtocolNegotiator.java:81` — Good addition of `volatile` to the static `sslContext` field. The gRPC TLS handshake threads read this field, so visibility is important. ✅

- **[Info]** `ProxyAndTlsProtocolNegotiator.java:143-148` — Uses `io.grpc.netty.shaded.io.netty.util.ReferenceCountUtil` (shaded package), which is correct since the proxy module depends on the shaded Netty from `grpc-netty-shaded`. ✅

- **[Low]** `NettyRemotingServer.java:56` — Import ordering: `ReferenceCountUtil` is inserted between `HashedWheelTimer` and `Timeout`, breaking alphabetical order. Minor style issue — should be placed after `io.netty.util.HashedWheelTimer` and before `io.netty.util.Timeout` (or at end of the `io.netty.util.*` block depending on project convention). Non-blocking.

- **[Low]** Both files — `loadSslContext()` is not synchronized. If called concurrently (e.g., from multiple `FileWatchService` callbacks), two threads could capture the same `oldSslContext` and both release it → `IllegalReferenceCountException`. In practice, `FileWatchService` dispatches from a single callback thread, so the risk is low. Adding `synchronized` would provide defense-in-depth but is not strictly required.

- **[Suggestion]** Consider adding a unit test that verifies the fix:
  ```java
  // Pseudocode
  server.loadSslContext();  // First call: sslContext is set, oldContext is null
  SslContext first = server.sslContext;
  server.loadSslContext();  // Second call: old context should be released
  // Assert: first.refCnt() == 0 (for OpenSSL) or no exception (for JDK)
  ```
  This would prevent regression and validate the release logic.

### Cross-repo Note
No cross-repo impact. This change is internal to the broker/proxy TLS lifecycle management.

### Verdict
**Approve with minor suggestions** — This is a well-structured fix that improves upon the earlier PR #10396. The "build into local var → swap → release old with try-catch" pattern is the correct approach for `ReferenceCounted` resource management in Netty. Thread safety is adequate with `volatile` on both declarations. The try-catch around `release()` is a good defensive measure. Main improvement: add automated test coverage.

---
*Automated review by rocketmq-reviewer-bot*

lizhimins (approved):

Issue #10383, opened 2026-05-26
[Bug] Bazel dependencies are out of sync with Maven; align Bazel BUILD files to pom.xml


qianye1001:
**Describe the bug**

The project maintains two parallel build systems — Maven (`pom.xml`) and Bazel (`BUILD.bazel` / `WORKSPACE` / `MODULE.bazel`). Over time the dependency graphs of the two have drifted: some artifacts present in Maven are missing, outdated, or differently versioned in Bazel, which leads to inconsistent classpaths, build failures on the Bazel path, and runtime behavior that differs between the two toolchains.

**Current behavior**

- `pom.xml` declares dependencies / versions / scopes that are not reflected in the corresponding `BUILD.bazel` `deps` / `runtime_deps` / `exports`.
- A few third-party libraries have been bumped in Maven but their `maven_install` / `rules_jvm_external` pins in Bazel still point to older versions.
- New modules added under Maven do not yet have matching `java_library` / `java_binary` targets, or the targets exist but reference a stale dep set.
- Some `provided` / `test` scope mappings in Maven are not mirrored in Bazel (`neverlink = True`, `testonly = True`).

**Expected behavior**

Maven (`pom.xml`) is the source of truth. Bazel BUILD files and the external dependency lock should be regenerated/aligned so that:

1. Every Maven dependency (same `group:artifact:version` and scope semantics) is reachable from the equivalent Bazel target.
2. `maven_install_json` / lock file is regenerated and committed.
3. Per-module `BUILD.bazel` `deps` lists exactly mirror the module's `pom.xml` `<dependencies>`, including `provided` → `neverlink`, `test` → `testonly`, `optional` → not exported.
4. `bazel build //...` and `bazel test //...` succeed and produce the same effective classpath as `mvn -pl <module> dependency:tree`.

**Proposed scope of work**

- Diff every module: `mvn dependency:tree` vs `bazel query 'deps(//module:target)'`, output a reconciliation table.
- Update `MODULE.bazel` / `WORKSPACE` artifact list and regenerate the lock.
- Update each module's `BUILD.bazel` `deps` / `runtime_deps` / `exports` to match.
- Add a CI check (script or Bazel test) that fails when `pom.xml` and `BUILD.bazel` drift again.
- Document the alignment rule in the contributor guide so future PRs touching `pom.xml` must also touch the matching `BUILD.bazel`.

**Acceptance criteria**

- [ ] `bazel build //...` green
- [ ] `bazel test //...` green
[...]

PR #10404, opened 2026-05-28
[ISSUE #10383] Align Bazel dependencies with Maven pom.xml

lizhimins:
## Summary
- Align Bazel WORKSPACE dependencies with Maven pom.xml (source of truth)
- Fix the root cause of Bazel CI SIGSEGV crashes (wrong RocksDB artifact)
- Fix ConfigStorage daemon thread use-after-free race condition
- Fix 3 flaky tests that caused intermittent CI failures across platforms

### Bazel dependency alignment

| Category | Change |
|----------|--------|
| **Critical fix** | `org.apache.rocketmq:rocketmq-rocksdb:1.0.6` → `org.rocksdb:rocksdbjni:8.4.4` |
| **Artifact rename** | `org.lz4:lz4-java:1.8.0` → `at.yawk.lz4:lz4-java:1.10.3` |
| **Artifact rename** | `org.bouncycastle:bcpkix-jdk15on:1.69` → `bcpkix-jdk18on:1.83` |
| **Version bumps** | netty-all, guava, commons-lang3, commons-io, gson, okio, etc. |
| **Missing deps** | Added `slf4j-api:2.0.3`, `javassist:3.20.0-GA`, `grpc-protobuf:1.47.0` |
| **Cleanup** | Removed duplicate `fastjson2` entry |
| **Build fix** | Added `--nojava_header_compilation` (netty-all 4.1.130+ is POM-only) |

### Bug fix: ConfigStorage SIGSEGV

Removed `setDaemon(true)` from `ConfigStorage.FlushSyncService` and `scheduledExecutorService`. The daemon flag caused `ServiceThread.shutdown()` to skip `thread.join()`, so the flush thread could still be calling `db.flushWal()` when `AbstractRocksDBStorage.shutdown()` closed the RocksDB native handle — triggering SIGSEGV in `rocksdb::log::Writer::WriteBuffer()`. Core storage flush threads (like CommitLog `GroupCommitService`) are non-daemon by convention.

### Flaky test fixes

| Test | Platform | Root Cause | Fix |
|------|----------|-----------|-----|
| `RocksDBMessageStoreTest` | Ubuntu | `RocksGroupCommitService` async flush not complete before CQ read | Added `Awaitility.await()` to poll until offset is committed |
| `IndexStoreServiceTest` | Windows | `System.currentTimeMillis()` ~15ms resolution causes timestamp collision in `createNewIndexFile` | Increment timestamp until unique key found |
| `ReplicasManagerTest` | macOS | Fixed sleep replaced with Awaitility polling; added `setHaListenPort(0)`; fixed shutdown order |

## Motivation

The Bazel CI had been failing with JVM SIGSEGV crashes on RocksDB-related tests. Root cause: WORKSPACE declared `rocketmq-rocksdb:1.0.6` (an obsolete shim with mismatched native library) while Maven uses `org.rocksdb:rocksdbjni:8.4.4`. The daemon thread issue in ConfigStorage amplified this — even with the correct library, the use-after-free race could still crash.

## Test plan
- [x] `bazel build //...` passes (583 targets)
- [x] `bazel test //...` passes on CI (remote execution)
- [x] All three previously-flaky tests verified locally
- [x] Maven compilation unaffected (no pom.xml changes)


lollipopjin (approved):
LGTM

PR #10408, opened 2026-05-29
[ISSUE #10405] Fix orderly consumer retry message routed to wrong broker via TBW102 fallback

### LINKED ISSUE #10405: [Bug] when orderly consumer consume message fails time reach the maxReconsumeTimes, the %RETRY% topic will create in broker that has  not contain the orignial topic

linux

### RocketMQ version

RocketMQ 5.3.3
proxy cluster mode

### JDK Version

1.8

### Describe the Bug

use proxy cluster mode and 5 master broker(A B C D E),   create topic t in A B C , and create orderly consumer connect proxy to consume message and set maxReconsumeTimes, when consume fails time reach the maxReconsumeTimes, the %RETRY%consumergroup will create in A B C D E broker,  i wish the %RETRY% only create in A B C 。

i read the code,  orderly consumer mode the %RETRY%  topic create depend on TBW102 system topic

### Steps to Reproduce

use proxy cluster mode and 5 master broker(A B C D E),   create topic t in A B C , and create orderly consumer connect proxy to consume message and set maxReconsumeTimes, when consume fails time reach the maxReconsumeTimes, the %RETRY%consumer group will create in A B C D E broker,  i wish the %RETRY% only create in A B C 

 i wish the %RETRY% only create in A B C 。

### What Did You See Instead?

the %RETRY%consumergroup will create in A B C D E broker

### Additional Context

_No response_

qianye1001:
## Summary

- Fix `ConsumeMessageOrderlyService.sendMessageBack()` and `ConsumeMessagePopOrderlyService.sendMessageBack()` to route retry messages to brokers that actually host the original topic, instead of falling back to TBW102 which routes to all brokers in the cluster.

## Root Cause

When an orderly consumer's reconsume times reach `maxReconsumeTimes`, `sendMessageBack()` sends the retry message via `getDefaultMQProducer().send(newMsg)` — a normal producer send. Since the `%RETRY%` topic doesn't exist yet, the producer falls back to TBW102 (a system topic present on ALL brokers) for broker selection. This causes the `%RETRY%` topic to be auto-created on brokers that don't host the original topic.

**Contrast with concurrent consumers**: `ConsumeMessageConcurrentlyService` uses `consumerSendMsgBack` which targets `msg.getBrokerName()` directly — no TBW102 fallback.

## Fix

Added `trySendToTopicBroker()` to both orderly service classes. It uses the consumer's existing `topicSubscribeInfoTable` (route info for the original topic) to select candidate brokers:

1. **Prefer original broker** — the broker where the message was consumed
2. **Failover to other brokers hosting the topic** — if the original broker is down
3. **Final fallback to default send** — only when route info is unavailable (should not happen in practice)

## Impact

| Scenario | Before | After |
|---|---|---|
| Original broker alive | May route to any broker (TBW102) | Routes to original broker ✓ |
| Original broker down | Routes to any broker | Failover to other brokers hosting the topic ✓ |
| All topic brokers down | Routes to any broker | Same (fallback) |

No behavioral change for brokers that DO host the original topic. Backward compatible.

## Test plan

- [ ] Verify `trySendToTopicBroker` prefers the original broker when available
- [ ] Verify failover to another broker when original broker send fails
- [ ] Verify fallback to default send when route info is empty
- [ ] Integration test: proxy cluster mode with multi-broker, orderly consumer hitting maxReconsumeTimes — `%RETRY%` should only appear on brokers hosting the original topic

Closes #10405

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
@@ -404,8 +404,12 @@ public class ConsumeMessageOrderlyService implements ConsumeMessageService {
     }
 
     private boolean trySendToTopicBroker(Message retryMsg, MessageExt originalMsg) {
-        Set<MessageQueue> mqs = this.defaultMQPushConsumerImpl.getRebalanceImpl()
-            .getTopicSubscribeInfoTable().get(originalMsg.getTopic());
+        java.util.concurrent.ConcurrentMap<String, Set<MessageQueue>> table =
+            this.defaultMQPushConsumerImpl.getRebalanceImpl().getTopicSubscribeInfoTable();
+        if (table == null) {
+            return false;
+        }
+        Set<MessageQueue> mqs = table.get(originalMsg.getTopic());
         if (mqs == null || mqs.isEmpty()) {
             return false;
         }
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessageOrderlyService.java
@@ -19,7 +19,9 @@ package org.apache.rocketmq.client.impl.consumer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -390,12 +392,42 @@ public class ConsumeMessageOrderlyService implements ConsumeMessageService {
             MessageAccessor.clearProperty(newMsg, MessageConst.PROPERTY_TRANSACTION_PREPARED);
             newMsg.setDelayTimeLevel(3 + msg.getReconsumeTimes());
 
-            this.defaultMQPushConsumerImpl.getmQClientFactory().getDefaultMQProducer().send(newMsg);
+            if (!trySendToTopicBroker(newMsg, msg)) {
+                this.defaultMQPushConsumerImpl.getmQClientFactory().getDefaultMQProducer().send(newMsg);
+            }
             return true;
         } catch (Exception e) {
             log.error("sendMessageBack exception, group: " + this.consumerGroup + " msg: " + msg.toString(), e);
         }
 
+        return false;
+    }
+
+    private boolean trySendToTopicBroker(Message retryMsg, MessageExt originalMsg) {
+        Set<MessageQueue> mqs = this.defaultMQPushConsumerImpl.getRebalanceImpl()
+            .getTopicSubscribeInfoTable().get(originalMsg.getTopic());
+        if (mqs == null || mqs.isEmpty()) {
+            return false;
+        }
+
+        Set<String> brokerNames = new LinkedHashSet<>();
+        String originalBroker = originalMsg.getBrokerName();
+        if (originalBroker != null) {
+            brokerNames.add(originalBroker);
+        }
+        for (MessageQueue mq : mqs) {
+            brokerNames.add(mq.getBrokerName());
+        }
+
+        for (String brokerName : brokerNames) {
+            try {
+                MessageQueue retryMQ = new MessageQueue(retryMsg.getTopic(), brokerName, 0);
+                this.defaultMQPushConsumerImpl.getmQClientFactory().getDefaultMQProducer().send(retryMsg, retryMQ);
+                return true;
+            } catch (Exception e) {
+                log.warn("Failed to send retry message to broker {}, trying next", brokerName, e);
+            }
+        }
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #10408 — Fix orderly consumer retry message routed to wrong broker via TBW102 fallback
**Type:** Bug fix (2 files, +74/-2)

### Assessment
Well-documented fix for a real routing issue in orderly consumers. The root cause analysis is clear:

**Problem:** When orderly consumers hit `maxReconsumeTimes`, retry messages were routed via `getDefaultMQProducer().send()` which falls back to TBW102 (present on ALL brokers), causing `%RETRY%` topics to be auto-created on brokers that don't host the original topic.

**Solution:** Added `trySendToTopicBroker()` to both `ConsumeMessageOrderlyService` and `ConsumeMessagePopOrderlyService`:
1. Prefer original broker (where message was consumed)
2. Failover to other brokers hosting the topic
3. Final fallback to default send (backward compatible)

### Verdict
✅ Clean, focused fix with proper failover logic. Consistent with how `ConsumeMessageConcurrentlyService` handles this case. Fixes #10405.

---

RockteMQ-AI:

### Summary
Fixes orderly consumer retry message routing by sending to brokers that host the original topic instead of falling back to TBW102, which routes to all brokers.

### Findings
- **[Warning]** ConsumeMessageOrderlyService.java:414 and ConsumeMessagePopOrderlyService.java:328 — `trySendToTopicBroker` modifies `retryMsg` state on each iteration: `send(retryMsg, retryMQ)` may internally set properties on the message (e.g., `setDelayTimeLevel` or other fields). If the first broker send fails and the loop retries on a second broker, the message object may be in a partially-modified state from the failed attempt. This depends on the producer implementation but is worth verifying.
- **[Warning]** ConsumeMessageOrderlyService.java:406 and ConsumeMessagePopOrderlyService.java:320 — The `table` reference is obtained once but `getTopicSubscribeInfoTable()` returns a live concurrent map. The entry for `originalMsg.getTopic()` could be removed between the null/empty check and iteration. This is a minor race — the worst case is a `ConcurrentModificationException` or empty iteration, which would fall through to the default send. Low risk but worth noting.
- **[Info]** ConsumeMessageOrderlyService.java:420 and ConsumeMessagePopOrderlyService.java:334 — `new MessageQueue(retryMsg.getTopic(), brokerName, 0)` uses queueId `0`. For retry topics, this is fine since retry topics typically have a single queue, but the hardcoded `0` could be confusing. A comment or constant would clarify intent.
- **[Info]** Both files — The `trySendToTopicBroker` method is duplicated verbatim across the two classes. Consider extracting it to a shared utility or a base class to reduce code duplication.
- **[Info]** Both files — No unit tests are included. The PR description lists test plan items as unchecked. Given the routing logic is safety-critical for message delivery, tests covering: (1) original broker preferred, (2) failover to other broker, (3) empty route fallback, are strongly recommended.

### Suggestions
1. Verify that `DefaultMQProducer.send(msg, mq)` doesn't mutate the `Message` object in a way that affects retries on subsequent brokers. If it does, consider creating a fresh copy per attempt or resetting any modified state.
2. Extract the identical `trySendToTopicBroker` into a shared method (e.g., in a utility class or via a common base) to avoid the exact duplication across the two service classes.
3. Add unit tests — at minimum, verify the broker ordering preference and fallback behavior. The logic is subtle and hard to verify manually in integration.

---

RockteMQ-AI (commented):

### Summary
This PR fixes a bug where orderly consumer retry messages (when `maxReconsumeTimes` is reached) are routed to the wrong broker. The root cause is that the `%RETRY%` topic lacks explicit route info, causing the producer to fall back to the `TBW102` default topic, which routes to all brokers — potentially sending retry messages to brokers that don't host the original topic.

### Findings

- **[Info]** `ConsumeMessageOrderlyService.java:416-454` and `ConsumeMessagePopOrderlyService.java:418-456` — The new `trySendToTopicBroker()` method is logically sound:
  - Correctly identifies the root cause (TBW102 fallback routing to all brokers).
  - Uses `topicSubscribeInfoTable` to discover which brokers actually host the original topic's queues.
  - `LinkedHashSet` preserves insertion order (original broker first) while deduplicating — good choice.
  - Falls back to the original `send(newMsg)` behavior if all targeted brokers fail or no queue info is available — preserves backward compatibility.

- **[Warning]** Missing unit tests — The new `trySendToTopicBroker()` method has no test coverage. Given this is a bug fix in the consumer retry path (a critical reliability concern), tests should be added to verify:
  - Correct broker selection from `topicSubscribeInfoTable`.
  - Fallback behavior when `topicSubscribeInfoTable` returns null/empty.
  - Fallback behavior when all targeted brokers fail.
  - The `originalMsg` parameter is correctly passed through (used for `sendMessageBack` with the original broker name).

- **[Info]** Queue index hardcoded to 0 — `new MessageQueue(retryMsg.getTopic(), brokerName, 0)` uses queue ID 0. For `%RETRY%` topics this is conventional, but worth noting that this assumes all brokers accept retry messages on queue 0.

- **[Info]** Code duplication — The `trySendToTopicBroker()` method is nearly identical in both `ConsumeMessageOrderlyService` and `ConsumeMessagePopOrderlyService`. This mirrors the existing pattern in RocketMQ (the two services share similar logic), so it's consistent with the codebase style. A future refactoring could extract the common logic.

- **[Info]** Thread safety — `topicSubscribeInfoTable` is a `ConcurrentMap`, so the `get()` call is safe. The `Set<MessageQueue>` values returned by `mqSet()` are iterated without explicit synchronization. This is acceptable if the underlying set is a `CopyOnWriteArraySet` or similar, which is typical for the rebalance impl.

### Suggestions
1. **Add unit tests** for `trySendToTopicBroker()` — this is the most important improvement.
2. Consider adding a brief log message when all targeted brokers fail (before falling back to default send), to aid debugging in production.

---

Issue #8262, opened 2024-06-06
[Enhancement] Increase the code test coverage of the project to more than 50% with the help of Tongyi Lingma's capabilities.


ShannonDing:
### Before Creating the Enhancement Request

- [X] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

提交的PR中包含众多的新功能的实现、旧功能更新迭代、以及代码的优化和接口的变更，新的PR合入后不能影响整体代码的健壮性。

### Motivation

目前整体项目的代码覆盖率在46%左右，增加原模块的单测代码覆盖率，新合入的代码可以被单测覆盖。

### Describe the Solution You'd Like

借助通义灵码的能力，补充项目各个模块的代码测试覆盖率全部达到50%以上。

### Describe Alternatives You've Considered

null

supervate:
Is I should create a new issue and related to here instead of submit commits to this issue?

ShannonDing:
> Is I should create a new issue and related to here instead of submit commits to this issue?

it‘s better to create a new issue and label it with ‘soc’.

ShannonDing:
Reminder from the Tianchi platform: it‘s better to submit your results to the [official website](https://tianchi.aliyun.com/competition/entrance/532215/information) asap.

PR #10409, opened 2026-05-30
[ISSUE #8262] [Enhancement] Add tests and harden UtilAll.isItTimeToDo

ChaoXoX:
Before Creating the Enhancement Request
 I have confirmed that this should be classified as an enhancement rather than a bug/feature.
Summary
为 [UtilAll.isItTimeToDo](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) 增加边界输入的容错处理，并补充对应单测，提高 common 模块覆盖率。

Motivation
当前方法对 null、空白和非法数字输入会抛 NumberFormatException，容易影响定时任务判断流程。补充容错逻辑与单测能提高健壮性并提升覆盖率。

Describe the Solution You'd Like
[UtilAll.isItTimeToDo](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) 对 null/空白/非法 token 安全忽略，跳过越界小时值。
新增单测覆盖：
当前小时命中
含非法 token 的混合输入
null/空串/空白/非法数字/越界小时
Describe Alternatives You've Considered
暂无

Additional Context
关联 #8262（提升项目整体测试覆盖率）。


common/src/main/java/org/apache/rocketmq/common/UtilAll.java
```diff
@@ -117,13 +117,29 @@ public static long computeElapsedTimeMilliseconds(final long beginTime) {
...
+                String trimmed = w.trim();
+                try {
+                    int hour = Integer.parseInt(trimmed);
+                    if (hour < 0 || hour > 23) {
+                        continue;
+                    }
+                    if (hour == nowHour) {
+                        return true;
+                    }
+                } catch (NumberFormatException ignored) {
```
RongtongJin, line 141: Could we avoid silently ignoring invalid `when` tokens here, or at least log a warning when a token cannot be parsed or is outside the 0-23 range? This helper is used by scheduled maintenance paths such as commitlog cleanup and topic queue mapping cleanup. If `deleteWhen` or a similar schedule is misconfigured, returning `false` forever would silently disable the task and make the operator miss the configuration problem. The previous behavior failed loudly; changing that to silent ignore seems risky without some visibility.

common/src/test/java/org/apache/rocketmq/common/UtilAllTest.java
```diff
@@ -151,6 +153,20 @@ public void testSplit() {
...
     }
 
+    @Test
+    public void testIsItTimeToDo() {
+        int currentHour = Calendar.getInstance().get(Calendar.HOUR_OF_DAY);
+        assertTrue(UtilAll.isItTimeToDo(String.valueOf(currentHour)));
+        assertTrue(UtilAll.isItTimeToDo("foo; " + currentHour + " ; 25"));
+        assertTrue(UtilAll.isItTimeToDo("  " + currentHour + "  "));
+
+        assertFalse(UtilAll.isItTimeToDo(null));
```
RongtongJin, line 163: This test depends on the hour read here matching the hour read inside `UtilAll.isItTimeToDo`. It can theoretically become flaky if the test crosses an hour boundary between the two calls. Could we make the positive case independent of wall-clock timing, for example by passing all valid hours (`0;1;...;23`) and asserting it returns true, while keeping the invalid-token cases separate?

RongtongJin (commented):
Thanks for the patch. I left two review comments around the behavior change and test stability.

oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #10409 — [Enhancement] Add tests and harden UtilAll.isItTimeToDo
**Type:** Enhancement + Test coverage (12 files, +49/-17)

### Code Analysis

**UtilAll.isItTimeToDo hardening:**
- ✅ Null/blank guard at method entry — prevents NPE
- ✅ Try-catch around `Integer.parseInt` — gracefully handles non-numeric tokens instead of throwing `NumberFormatException`
- ✅ Hour range validation (0-23) — skips out-of-range values
- ✅ `Calendar.HOUR_OF_DAY` computed once — minor efficiency improvement

**Test coverage:**
- ✅ Current hour match (positive case)
- ✅ Mixed input with invalid tokens (`foo; <hour> ; 25`)
- ✅ Whitespace handling
- ✅ Null, empty, blank, non-numeric, out-of-range (negative cases)

**Documentation changes:**
- Minor doc file renames and link fixes — unrelated but harmless

### Verdict
Clean, well-scoped enhancement that improves robustness of scheduled task timing logic. Test coverage is comprehensive. No concerns.

---

RockteMQ-AI:

### Summary
Harden `UtilAll.isItTimeToDo` to silently skip null/blank/invalid/out-of-range hour tokens instead of throwing `NumberFormatException`, with new unit tests and unrelated doc fixes.

### Findings

- **[Critical]** `common/src/test/java/org/apache/rocketmq/common/UtilAllTest.java:156` — The `@Test` annotation is corrupted: the diff shows `@repos/apache_rocketmq/auth/src/test/java/...` instead of `@Test`. This means `testIsItTimeToDo` will not be recognized as a test method and will never run. The entire test coverage rationale for this PR breaks down.

- **[Warning]** `common/src/main/java/org/apache/rocketmq/common/UtilAll.java:120` — The outer `if (whiles.length > 0)` guard is now dead code: `String.split(";")` on a non-blank string always returns an array of length ≥ 1. Not a bug, but misleading.

- **[Warning]** `common/src/test/java/org/apache/rocketmq/common/UtilAllTest.java:157` — The positive test `assertTrue(UtilAll.isItTimeToDo(String.valueOf(currentHour)))` is time-sensitive and will fail if the hour rolls over between `Calendar.getInstance()` and `isItTimeToDo`'s own `Calendar.getInstance()` call (extremely rare, but flaky by design). Consider passing a fixed mock or accepting the tiny risk explicitly.

- **[Warning]** `common/src/test/java/org/apache/rocketmq/common/UtilAllTest.java:158` — `assertFalse(UtilAll.isItTimeToDo("99"))` is correct for the out-of-range case, but there is no `assertFalse` for a valid hour that is *not* the current hour (e.g., `(currentHour + 1) % 24`). The negative path for a well-formed but non-matching hour is untested.

- **[Info]** `common/src/main/java/org/apache/rocketmq/common/UtilAll.java:130` — The catch block comment ("Ignore invalid hour tokens to avoid breaking scheduled tasks.") violates the project's no-obvious-comment style, but is borderline acceptable given the silently-swallowed exception.

- **[Info]** Doc renames (`Design_LoadBlancing.md` → `Design_LoadBalancing.md`, `Design_Trancation.md` → `Design_Transaction.md`, `Troubleshoopting.md` → `Troubleshooting.md`) fix long-standing typos; these are welcome cleanups but are unrelated to the stated issue scope.

### Suggestions

1. **Fix the `@Test` annotation immediately** — the annotation on `testIsItTimeToDo` is the file path of an unrelated file (`@repos/apache_rocketmq/auth/...`). This is the blocking issue; the PR's primary value (test coverage) is entirely negated without it.
2. Add a `assertFalse(UtilAll.isItTimeToDo(String.valueOf((currentHour + 1) % 24)))` case to cover the valid-but-non-matching hour path.
3. Remove the dead `if (whiles.length > 0)` wrapper, or at minimum flatten the indentation, to keep the method readable.

---

common/src/main/java/org/apache/rocketmq/common/UtilAll.java
RongtongJin: I still consider this blocking. Silently accepting an invalid schedule can hide a bad `deleteWhen`-style configuration and leave scheduled maintenance disabled without any clear signal. Please keep configuration mistakes observable, either by preserving fail-fast behavior or by logging/validating invalid tokens explicitly.

common/src/test/java/org/apache/rocketmq/common/UtilAllTest.java
RongtongJin: Still worth fixing before merge. This can be made deterministic by passing all valid hours (`0;1;...;23`) for the positive case, and by adding a valid-but-non-current hour negative case.

docs/en/Example_Simple.md
```diff
@@ -9,12 +9,12 @@ maven:
 <dependency>
   <groupId>org.apache.rocketmq</groupId>
   <artifactId>rocketmq-client</artifactId>
-  <version>4.3.0</version>
+  <version>5.5.0</version>
```
RongtongJin, line 12: These documentation cleanups are unrelated to the `UtilAll.isItTimeToDo` behavior change and overlap with docs-only PR scope. Please split or drop the doc changes from this PR so the behavioral change can be reviewed independently.

RockteMQ-AI (commented):

### Summary
This PR hardens  to handle null, blank, and invalid numeric inputs gracefully, and adds corresponding unit tests. It also includes unrelated documentation updates (version bumps and typo fixes in doc filenames).

### Findings

- **[Info]**  — The code changes are clean and well-structured:
  - Early return on  is correct.
  - Hoisting  outside the loop is a good micro-optimization.
  - Try-catch around  with silent ignore is appropriate for scheduled-task tokens — invalid tokens should not break the entire schedule.
  - Range check  correctly skips out-of-range values.

- **[Info]**  — Good test coverage for boundary cases: null, empty, whitespace-only, non-numeric, and out-of-range (25, 99). The mixed-input test  nicely validates that valid tokens still match even among invalid ones.

- **[Warning]** Mixed concerns in one PR — The documentation changes (version bumps from 4.x → 5.5.0, filename typo fixes like  → ,  → ) are unrelated to the  enhancement. While these are welcome fixes, mixing them with code changes makes review harder and can complicate backporting. Consider splitting into separate PRs in the future.

- **[Info]**  →  — The rename has 100% similarity (pure rename). Any external links or bookmarks to the old filename will break. Consider adding a redirect or noting this in release notes if the docs are published with stable URLs.

### Suggestions
- The code logic is solid. No functional changes needed.
- For future PRs, consider separating code hardening from documentation housekeeping.

---

PR #10418, opened 2026-06-02
[ISSUE #10417][BUG] Fix stale minOffset after truncate

### LINKED ISSUE #10417: [Bug] Stale minOffset after truncating all ConsumeQueue files

ubuntu 24.04

### RocketMQ version

rocketmq-all-5.5.0

### JDK Version

JDK 8u202

### Describe the Bug

When broker-side commitlog truncation removes all ConsumeQueue files for a topic queue, the queue may retain a stale minOffset instead of resetting it to 0, leaving an inconsistent state where minOffset > maxOffset (with maxOffset == 0).

### Steps to Reproduce

- Enable Jraft controller mode
- Make a broker-slave offline
- Produce some messages to a never-used topic queue on broker-master
- Make the master offline
- Make the slave online and let controller choose it as new master(if the slave was out of ISR, enable electUncleanMaster)
- Produce more messages to the topic queue mentioned above

Messages produced to the topic queue can be consumed

### What Did You See Instead?

Messages produced to the topic queue cannot be consumed until enough messages had been written into the queue(which will make the stale minOffset effective again)

### Additional Context

This problem is caused by method: 

org.apache.rocketmq.store.ConsumeQueue#correctMinOffset

when 'minLogicOffset >= mappedFileQueue.getMaxOffset()' is satisfied, this method did nothing and just return

chenxu80:

- Fixes #10417

## Summary
  - fix a case where broker-side commitlog truncation removes all ConsumeQueue files for a topic queue but leaves a stale
  `minOffset`
  - reset the queue's logical minimum offset when the queue becomes empty, so `minOffset` stays consistent with `maxOffset`
  - avoid exposing an invalid state where `maxOffset == 0` but `minOffset > 0`

  ## Root cause
  When all ConsumeQueue files of a topic queue are deleted during commitlog truncation, the in-memory logical minimum offset may
   not be reset. As a result, later `minOffset` queries can return a stale value even though the queue is already empty.

  ## Test Plan
  - reproduce the scenario where commitlog truncation removes all ConsumeQueue files for a topic queue
  - verify the queue reports `minOffset == 0` and `maxOffset == 0` after truncation
  - run the related store tests

### DIFF
--- a/store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
@@ -24,6 +24,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.net.UnknownHostException;
+import java.nio.file.Files;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -712,5 +713,66 @@ public class ConsumeQueueTest {
 
         consumeQueue.destroy();
         FileUtils.deleteDirectory(tmpDir);
+    }
+
+    @Test
+    public void testCorrectMinOffsetAfterAllFilesDeleted() throws IOException {
+        String topic = "T1";
+        int queueId = 0;
+        MessageStoreConfig storeConfig = new MessageStoreConfig();
+        File tmpDir = Files.createTempDirectory("testCorrectMinOffsetAfterAllFilesDeleted").toFile();
+        storeConfig.setStorePathRootDir(tmpDir.getAbsolutePath());
+        storeConfig.setEnableConsumeQueueExt(false);
+        DefaultMessageStore messageStore = Mockito.mock(DefaultMessageStore.class);
+        Mockito.when(messageStore.getMessageStoreConfig()).thenReturn(storeConfig);
+
+        RunningFlags runningFlags = new RunningFlags();
+        Mockito.when(messageStore.getRunningFlags()).thenReturn(runningFlags);
+
+        StoreCheckpoint storeCheckpoint = Mockito.mock(StoreCheckpoint.class);
+        Mockito.when(messageStore.getStoreCheckpoint()).thenReturn(storeCheckpoint);
+
+        ConsumeQueue consumeQueue = null;
+        ConsumeQueue reloadedConsumeQueue = null;
+        try {
+            consumeQueue = new ConsumeQueue(topic, queueId, storeConfig.getStorePathRootDir(),
+                storeConfig.getMappedFileSizeConsumeQueue(), messageStore);
+
+            int max = 10;
+            int messageSize = 100;
+            for (int i = 0; i < max; ++i) {
+                DispatchRequest dispatchRequest = new DispatchRequest(topic, queueId, messageSize * i, messageSize, 0, 0, i,
+                    null, null, 0, 0, null);
+                consumeQueue.putMessagePositionInfoWrapper(dispatchRequest);
+            }
+
+            File consumeQueueDir = new File(storeConfig.getStorePathRootDir(), topic + File.separator + queueId);
+            Assert.assertTrue(consumeQueueDir.exists());
+
+            consumeQueue.setMinLogicOffset(ConsumeQueue.CQ_STORE_UNIT_SIZE);
+            consumeQueue.setMaxPhysicOffset(max * messageSize);
+            consumeQueue.destroy();
+            consumeQueue = null;
+            FileUtils.deleteQuietly(consumeQueueDir);
+            Assert.assertFalse(consumeQueueDir.exists());
+
+            reloadedConsumeQueue = new ConsumeQueue(topic, queueId, storeConfig.getStorePathRootDir(),
+                storeConfig.getMappedFileSizeConsumeQueue(), messageStore);
+            Assert.assertTrue(reloadedConsumeQueue.load());
+            reloadedConsumeQueue.recover();
+            reloadedConsumeQueue.setMinLogicOffset(ConsumeQueue.CQ_STORE_UNIT_SIZE);
+            reloadedConsumeQueue.setMaxPhysicOffset(max * messageSize);
+            reloadedConsumeQueue.correctMinOffset(0L);
+            Assert.assertEquals(0L, reloadedConsumeQueue.getMinLogicOffset());
+            Assert.assertEquals(-1L, reloadedConsumeQueue.getMaxPhysicOffset());
+        } finally {
+            if (reloadedConsumeQueue != null) {
+                reloadedConsumeQueue.destroy();
+            }
+            if (consumeQueue != null) {
+                consumeQueue.destroy();
+            }
+            FileUtils.deleteQuietly(tmpDir);
+        }
     }
 }
--- a/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
@@ -552,17 +552,18 @@ public class ConsumeQueue implements ConsumeQueueInterface {
      */
     @Override
     public void correctMinOffset(long minCommitLogOffset) {
-        // Check if the consume queue is the state of deprecation.
-        if (minLogicOffset >= mappedFileQueue.getMaxOffset()) {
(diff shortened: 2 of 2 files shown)


lollipopjin:
Since this bug can surface during HA master-slave switch or CommitLog truncation, would you consider adding a test that deletes all ConsumeQueue files and then verifies correctMinOffset properly resets minLogicOffset and maxPhysicOffset? That would help guard against future regressions in this code path.

oss-sentinel-ai (commented):

### Summary
Fixes a bug where `ConsumeQueue#correctMinOffset` fails to reset `minLogicOffset` when commitlog truncation deletes all ConsumeQueue files for a topic queue, leaving an inconsistent state where `minOffset > maxOffset`.

### Findings

- **[Correctness]** ✅ The fix is well-structured and addresses the root cause:

  1. **Reordering checks** — `getLastMappedFile() == null` is now checked first. This is the critical case where all CQ files have been deleted, and the queue must be reset to empty state (`minLogicOffset = 0`, `maxPhysicOffset = -1`). Previously this was checked second, and the first `>=` check would short-circuit and return without any reset.

  2. **`>=` → `>` change** — When `minLogicOffset == maxOffset`, the queue is legitimately empty (min points to the end, which equals the current max). This is not an anomalous state and does not require a reset. The `>` guard correctly targets only the truly invalid case where `minLogicOffset` exceeds `maxOffset`.

  3. **Reset values** — `minLogicOffset = 0` matches the default initialization (line 78). `maxPhysicOffset = -1` matches the default initialization (line 73). This is consistent with `reset()` (line 1087).

- **[Performance]** ✅ No performance concern. The `getLastMappedFile()` call was already present in the original code; it's simply moved earlier in the method.

- **[Compatibility]** ✅ No API or protocol changes. Internal state management only.

- **[Tests]** ⚠️ No unit test added. The scenario (commitlog truncation removing all CQ files) is complex to reproduce in a unit test, but the fix is small enough that a targeted test for `correctMinOffset` with a mocked/empty `mappedFileQueue` would be valuable to prevent regression. Consider adding:

  ```java
  @Test
  public void testCorrectMinOffsetWhenAllFilesDeleted() {
      // Setup: create a ConsumeQueue, write some entries, then delete all mapped files
      // After calling correctMinOffset:
      //   assertEquals(0, consumeQueue.getMinOffsetInQueue() * CQ_STORE_UNIT_SIZE);
      //   assertEquals(-1, consumeQueue.getMaxPhysicOffset());
  }
  ```

### Suggestions

1. **Add a unit test** — As noted above, a test covering the "all CQ files deleted" path would strengthen this fix.

2. **Consider thread safety** — `minLogicOffset` is `volatile` (good), but `maxPhysicOffset` is not. In the reset path, both are written sequentially. If another thread reads `minLogicOffset` after it's reset to 0 but before `maxPhysicOffset` is reset to -1, the intermediate state could be briefly inconsistent. This is likely benign (the queue is empty, no consumer should be iterating), but worth a quick look to confirm no reader depends on both values being consistent simultaneously.

### CI
✅ All 10 checks pass (CodeQL, bazel-compile, maven-compile × 4, coverage, license, misspell).

---

chenxu80:
@lollipopjin Thanks for the suggestion. I have added a regression test in ConsumeQueueTest for the case where all ConsumeQueue files are deleted, to verify that correctMinOffset properly resets minLogicOffset to 0 and maxPhysicOffset to -1.

RockteMQ-AI (approved):

### Summary
Re-review after commit `aca3da2` (June 22). The core fix in `ConsumeQueue#correctMinOffset` remains unchanged from the previous review. The new commit adds a unit test covering the "all CQ files deleted" scenario, addressing the test gap noted in the previous review.

### Changes Since Last Review

- **[Tests]** ✅ New test `testCorrectMinOffsetAfterAllFilesDeleted` in `ConsumeQueueTest.java`:
  - Creates a ConsumeQueue, writes 10 entries, then deletes all CQ files
  - Reloads the queue, sets stale `minLogicOffset`/`maxPhysicOffset`, calls `correctMinOffset(0L)`
  - Asserts `minLogicOffset == 0` and `maxPhysicOffset == -1` after correction
  - Effectively validates the fix path where `getLastMappedFile() == null`

- **[Correctness]** ✅ Core fix unchanged and correct:
  1. Null-check for `lastMappedFile` moved first — handles the "all files deleted" case before offset comparisons
  2. Reset to `minLogicOffset = 0`, `maxPhysicOffset = -1` matches default initialization values (consistent with `reset()` at line 1087)
  3. `>=` → `>` change in the second guard correctly targets only the truly invalid case

- **[Performance]** ✅ No concern. `getLastMappedFile()` was already called in the original code; just reordered.

- **[Compatibility]** ✅ Internal state management only. No API or protocol changes.

### Minor Notes

1. **Test cleanup**: The test deletes `tmpDir` at the start but not at the end. Consider adding `FileUtils.deleteDirectory(tmpDir)` after `destroy()` to avoid leaving temp files. Non-blocking.

2. **CI failures**: `bazel-compile` and `maven-compile (ubuntu-latest, JDK-8)` show test failures. These may be unrelated flaky tests — worth checking if they reproduce on the `develop` branch.

### Verdict
The fix is small, targeted, and well-tested. Approving.

---

RongtongJin:
Results :

Tests in error: 
  ConsumeQueueTest.testCorrectMinOffsetAfterAllFilesDeleted:750 � IOExceptionList

Tests run: 82, Failures: 0, Errors: 1, Skipped: 63

store/src/test/java/org/apache/rocketmq/store/ConsumeQueueTest.java
```diff
@@ -713,4 +713,51 @@ public void correctMinOffsetWithEmptyQueueReadAheadOptimizationTest() throws IOE
...
+            DispatchRequest dispatchRequest = new DispatchRequest(topic, queueId, messageSize * i, messageSize, 0, 0, i, null, null, 0, 0, null);
+            consumeQueue.putMessagePositionInfoWrapper(dispatchRequest);
+        }
+
+        File consumeQueueDir = new File(storeConfig.getStorePathRootDir(), topic + File.separator + queueId);
+        Assert.assertTrue(consumeQueueDir.exists());
+
+        consumeQueue.setMinLogicOffset(ConsumeQueue.CQ_STORE_UNIT_SIZE);
+        consumeQueue.setMaxPhysicOffset(max * messageSize);
+        FileUtils.deleteDirectory(consumeQueueDir);
```
RongtongJin: This is the source of the Windows CI failure. The test deletes the consume-queue directory while the first `ConsumeQueue` instance still owns the mapped file and file channel, so Windows refuses to delete `.../T1/0/00000000000000000000` with "The process cannot access the file because it is being used by another process". Please release the first queue before deleting/reloading, preferably with `try/finally` cleanup, e.g. call `consumeQueue.destroy()` before `FileUtils.deleteDirectory(consumeQueueDir)` (and still destroy `reloadedConsumeQueue` in a finally block).

chenxu80: Thanks for pointing this out. I’ve updated the test to release the ConsumeQueue before deleting the directory, and also use a unique temporary directory with cleanup in finally. This should avoid the IOExceptionList caused by deleting files while resources are still held.

I’ve verified the test locally with Bazel and it passes now.

RockteMQ-AI (approved):
## Review by github-manager-bot (Re-review after squash)

### Summary
PR squashed/rebased — previous commit `aca3da2` replaced by `a47200d` (June 23). The diff is unchanged from the previous review.

### Changes (unchanged from previous review)

- **[Correctness]** ✅ Fix in `ConsumeQueue#correctMinOffset`:
  1. Null-check for `lastMappedFile` moved first — handles the "all CQ files deleted" case before offset comparisons
  2. Reset to `minLogicOffset = 0`, `maxPhysicOffset = -1` matches default initialization values (consistent with `reset()`)
  3. `>=` → `>` change in the second guard correctly targets only the truly invalid case
  4. Log level upgraded to `warn` for the no-valid-entries path

- **[Tests]** ✅ New test `testCorrectMinOffsetAfterAllFilesDeleted` validates the fix where `getLastMappedFile() == null`

- **[Performance]** ✅ No concern — `getLastMappedFile()` was already called; just reordered.

- **[Compatibility]** ✅ Internal state management only. No API or protocol changes.

### Verdict
Squash/rebase only — logic unchanged. Still looks good. Approving.

---

RongtongJin (approved):

guyinyou (approved):
lgtm

PR #10422, opened 2026-06-03
[#10421] Fix Timer message rocksdb use wrong cache key.

### LINKED ISSUE #10421: [Bug] Timer message rocksdb wrong cache key

ubuntu

### RocketMQ version

develop

### JDK Version

1.8

### Describe the Bug

In the timer message RocksDB store, when a delayed message is consumed, a `DELETE` record is written to remove it from RocksDB. Meanwhile, if the same message undergoes a Roll (CommitLog rotation), an `UPDATE `record is written to update its offset. The `DELETE_KEY_CACHE_FOR_TIMER` is used to prevent `UPDATE` from re-inserting an already-deleted key.

the cache uses `byte[]` as its key type. Since `byte[].equals()` compares object references rather than content.

### Steps to Reproduce

as describe.

The UPDATE should be skipped since the key was already deleted, and no record should remain in RocksDB.

### What Did You See Instead?

The UPDATE writes back the deleted record as a ghost record in RocksDB.

### Additional Context

_No response_

echooymxq:
Fix #10421

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorage.java
+++ b/store/src/main/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorage.java
@@ -84,7 +84,7 @@ public class MessageRocksDBStorage extends AbstractRocksDBStorage {
     private volatile ColumnFamilyHandle transCFHandle;
 
     private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
-    private static final Cache<byte[], byte[]> DELETE_KEY_CACHE_FOR_TIMER = CacheBuilder.newBuilder()
+    private static final Cache<String, byte[]> DELETE_KEY_CACHE_FOR_TIMER = CacheBuilder.newBuilder()
         .maximumSize(10000)
         .expireAfterWrite(60, TimeUnit.MINUTES)
         .build();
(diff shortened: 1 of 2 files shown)


echooymxq:
@zk-drizzle help review it.

oss-sentinel-ai (commented):

### Summary
Fixes #10421 — `DELETE_KEY_CACHE_FOR_TIMER` used raw `byte[]` as Guava cache key, causing identity-based lookups to always miss. Switches to `ByteBuffer` for content-equality semantics.

### Findings
- **[Correctness]** ✅ Changing cache key from `byte[]` to `ByteBuffer` properly uses content-based `equals()`/`hashCode()`, which is the correct fix for the ghost record issue.
- **[Correctness]** ⚠️ Line 87-88: The cache declaration `Cache<byte[], byte[]>` appears unchanged in the diff — please verify the cache type is also updated to `Cache<ByteBuffer, byte[]>` to match the `ByteBuffer.wrap()` calls at lines 357 and 359.
- **[Correctness]** ⚠️ `ByteBuffer.wrap(keyBytes).equals()` compares from `position()` to `limit()`. Since `wrap()` sets position=0 and limit=array.length, this works correctly for equal-length arrays. However, if key lengths could differ, consider using `ByteBuffer.wrap(keyBytes, offset, length)` consistently.
- **[Tests]** ⚠️ No new tests visible in this diff. The bug is about cache key identity vs equality — recommend adding a unit test that verifies `DELETE` + `UPDATE` for the same timer key results in the UPDATE being correctly skipped.
- **[Compatibility]** ✅ No public API changes. Internal implementation detail only.

### Suggestions
1. Ensure cache type declaration matches usage: `Cache<ByteBuffer, byte[]>` (not `Cache<byte[], byte[]>`)
2. Add a test case that creates a DELETE record and an UPDATE record for the same timer key, and verifies the UPDATE does not re-insert the deleted key
3. Consider adding a brief comment explaining why `ByteBuffer` is used instead of `byte[]` (identity vs content equality)

---

RongtongJin (approved):

3424672656 (approved):
lgtm

RockteMQ-AI (commented):

### Summary
Fixes a cache key bug in Timer message RocksDB storage where `byte[]` keys were used in a Guava `Cache`, causing cache misses due to Java's reference-based `byte[]` equality.

### Findings
- **[Critical]** `MessageRocksDBStorage.java:87` — Root cause: `Cache<byte[], byte[]>` uses reference equality for `byte[]` keys. `DELETE_KEY_CACHE_FOR_TIMER.getIfPresent(keyBytes)` would never match a previously stored entry because each `byte[]` is a distinct object. Changing to `Cache<String, byte[]>` with a stable `delayTime:uniqKey` string key is the correct fix.
- **[Info]** `MessageRocksDBStorage.java:376` — `getTimerCacheKey()` allocates a new `String` on every call. For the timer write hot path, consider whether the allocation is acceptable. Given the correctness fix is more important, this is fine for now.
- **[Info]** `MessageRocksDBStorageTest.java` — Good test coverage: 3 scenarios (PUT→DELETE, PUT→UPDATE, DELETE→UPDATE) validate the fix. Tests correctly verify record counts after each operation.

### Suggestions
- Consider adding a test case that verifies the cache actually prevents a redundant write (i.e., the UPDATE after DELETE is skipped because the cache hit returns `DELETE_VAL_FLAG`). The existing `testDeleteThenUpdate` already covers this implicitly.

**Verdict: LGTM.** Clean bug fix with correct root cause analysis and good test coverage.

---

PR #10426, opened 2026-06-03
[ISSUE #10425] Revert split registration dataVersion change

### LINKED ISSUE #10425: [Bug] Split registration can skip topic updates after #9521

- [x] I have searched the GitHub Issues and GitHub Discussions of this repository and believe that this is not a duplicate.

All platforms when broker split registration is enabled.

### RocketMQ version

`develop` after #9521 / commit `2accb9797771b9395d427eef1d6a7982f27b8dfc`.

### JDK Version

Not JDK-specific.

### Describe the Bug

#9521 removed the `DataVersion` advancement from `TopicConfigManager#buildSerializeWrapper` when `enableSplitRegistration` is enabled.

During split broker registration, `BrokerController#registerBrokerAll` sends topic configs in multiple `RegisterBrokerBody` requests. `RouteInfoManager#registerBroker` stores the broker `DataVersion` after each request and uses it to decide whether topic queue data needs to be updated. If every split request carries the same `DataVersion`, the later split requests can be treated as unchanged for topics that already exist under the broker, so topic config updates in those later batches may be skipped.

### Steps to Reproduce

1. Enable broker split registration with `enableSplitRegistration=true` and configure enough topics to require multiple registration batches.
2. Let the broker register once so the name server already has route data for those topics.
3. Update topic configs for topics that are not in the first split batch.
4. Trigger broker registration again.

Every split registration batch should be processed as a topic config update, so name server route data reflects updated topic configs from all batches.

### What Did You See Instead?

Only batches that are considered changed by `DataVersion` are guaranteed to update existing topic route data. Later split batches with the same `DataVersion` can be skipped for existing topics.

### Additional Context

Reverting #9521 restores the previous behavior where each split registration wrapper advances the topic config `DataVersion`, so each batch is visible as a distinct update to the name server.

RongtongJin:

- Fixes #10425

### Brief Description

This reverts #9521 and restores the topic config `DataVersion` advancement when `enableSplitRegistration` is enabled.

With split broker registration, a broker sends topic configs to the name server in multiple registration requests. The name server records the broker `DataVersion` after each request and uses it to decide whether existing topic route data should be updated. If all split requests carry the same `DataVersion`, later batches can be treated as unchanged for topics that already exist under the broker.

Restoring the per-wrapper `DataVersion` advancement keeps each split registration batch visible as a distinct update. This PR also adds a focused regression test for `TopicConfigManager#buildSerializeWrapper`.

- `/usr/local/bin/mvn -pl broker -am -Dtest=TopicConfigManagerTest -DfailIfNoTests=false -DskipITs -DskipCheckStyle test`

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
@@ -647,6 +647,9 @@ public class TopicConfigManager extends ConfigManager {
         topicConfigWrapper.setTopicConfigTable(topicConfigTable);
         topicConfigWrapper.setTopicQueueMappingInfoMap(topicQueueMappingInfoMap);
         topicConfigWrapper.setDataVersion(this.getDataVersion());
+        if (this.brokerController.getBrokerConfig().isEnableSplitRegistration()) {
+            this.getDataVersion().nextVersion();
+        }
         return topicConfigWrapper;
     }
 
--- a/broker/src/test/java/org/apache/rocketmq/broker/topic/TopicConfigManagerTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/topic/TopicConfigManagerTest.java
@@ -38,6 +38,7 @@ import org.apache.rocketmq.common.attribute.LongRangeAttribute;
 import org.apache.rocketmq.common.constant.PermName;
 import org.apache.rocketmq.common.utils.QueueTypeUtils;
 import org.apache.rocketmq.remoting.protocol.DataVersion;
+import org.apache.rocketmq.remoting.protocol.body.TopicConfigAndMappingSerializeWrapper;
 import org.apache.rocketmq.store.DefaultMessageStore;
 import org.apache.rocketmq.store.config.MessageStoreConfig;
 import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


fuyou001 (approved):

oss-sentinel-ai (approved):

### Summary
Restores the `DataVersion.nextVersion()` advancement in `TopicConfigManager#buildSerializeWrapper` when `enableSplitRegistration` is enabled, fixing the issue where later split registration batches skip topic config updates on the NameServer.

### Findings

- **[Correctness]** ✅ The fix is minimal and precisely scoped. The `nextVersion()` call is correctly guarded by `isEnableSplitRegistration()`, so non-split registration is unaffected.

- **[Root cause verified]** ✅ Traced through `RouteInfoManager#registerBroker` → `isBrokerTopicConfigChanged` (compares `DataVersion` from each split batch against stored value from `brokerLiveTable`). Without per-batch `DataVersion` advancement, the NameServer stores the version after the first batch and subsequent batches appear unchanged, causing `isTopicConfigChanged` to return `false` for existing topics.

- **[Tests]** ✅ `testBuildSerializeWrapperUpdatesDataVersionWhenSplitRegistrationEnabled` directly verifies the counter increment. Good regression test coverage for the specific fix.

- **[CI]** ✅ All 10 checks pass (CodeQL, bazel-compile, maven-compile × 4 platforms, coverage, license, misspell).

- **[Compatibility]** ✅ No API or protocol changes. No backward compatibility concerns.

### Suggestions

Consider adding a second assertion in the test to also verify the `DataVersion` stored in the returned wrapper matches the counter, confirming the wrapper carries the incremented version — this would more closely mirror the actual runtime check in `RouteInfoManager`:

```java
// Already present, good:
Assert.assertEquals(counterAfter, wrapper.getDataVersion().getCounter().get());
```

Actually this assertion already exists — well done. LGTM.

---

yx9o (approved):

Issue #9998, opened 2026-01-08
[Enhancement] Remove synchronized from deleteTopic to speed up topic deletion.


RongtongJin:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Remove synchronized from deleteTopic to speed up topic deletion.

### Motivation

Remove synchronized from deleteTopic to speed up topic deletion.

### Describe the Solution You'd Like

Remove synchronized from deleteTopic to speed up topic deletion.

### Describe Alternatives You've Considered

No

PR #9997, opened 2026-01-08
[ISSUE #9998] Remove synchronized from deleteTopic.

RongtongJin:
### Which Issue(s) This PR Fixes

Fixes #9998

PR #10448, opened 2026-06-09
[ISSUE #10446] Support batch deletion of topics and subscription groups in broker

### LINKED ISSUE #10446: [Feature] Support batch deletion of topics and subscription groups in broker

### Summary

`AdminBrokerProcessor` today only exposes single-target deletion APIs (`DELETE_TOPIC_IN_BROKER`, `DELETE_SUBSCRIPTIONGROUP`). Bulk metadata cleanup (e.g. tenant decommissioning, retention purge, cluster migration) currently requires N RPCs and triggers N `persist()` calls on the broker config files, which is the dominant source of admin-side latency and metadata pile-up under load.

This issue proposes adding two batch APIs whose names follow the existing `*_LIST` convention (`UPDATE_AND_CREATE_TOPIC_LIST`, `UPDATE_AND_CREATE_SUBSCRIPTIONGROUP_LIST`):

- `DELETE_TOPIC_IN_BROKER_LIST`
- `DELETE_SUBSCRIPTIONGROUP_LIST`

> Note: removing the `synchronized` modifier from `AdminBrokerProcessor#deleteTopic` is **out of scope** here and is already being discussed in #9998 / #9997. This issue/PR intentionally keeps `synchronized` as-is and only focuses on the new batch APIs.

### Motivation

Single-item `DELETE_TOPIC_IN_BROKER` (see [AdminBrokerProcessor.java#L272](https://github.com/apache/rocketmq/blob/82a6a784ee35b36d94b3ab644bd5394834e2c540/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java#L272)) and `DELETE_SUBSCRIPTIONGROUP` require N RPCs to delete N items and trigger N `persist()` calls on the on-disk config files. In metadata cleanup scenarios this leads to user-visible slow admin operations and unnecessary disk I/O.

Each `persist()` involves a full JSON re-serialization of the entire `topicConfigTable` / `subscriptionGroupTable` plus several `fsync`s, so batching N items into a single persist is a significant win on its own.

1. **Add two new request codes** in `RequestCode`:
   - `DELETE_TOPIC_IN_BROKER_LIST`
   - `DELETE_SUBSCRIPTIONGROUP_LIST`

2. **Add corresponding request bodies**:
   - `DeleteTopicListRequestBody { List<String> topicList }`
   - `DeleteSubscriptionGroupListRequestBody { List<String> groupNameList; boolean cleanOffset }`

3. **Add batch entry points** in `AdminBrokerProcessor` (kept `synchronized` to stay consistent with the existing `deleteTopic` / `updateAndCreate*` pattern):
   - `deleteTopicList(...)` — validates each topic, dedupes (preserving input order), optionally collects pop-retry topics, then calls `TopicConfigManager#deleteTopicConfigList(...)` with **one** `persist()` for the entire batch.
   - `deleteSubscriptionGroupList(...)` — same pattern, calling `SubscriptionGroupManager#deleteSubscriptionGroupConfigList(...)`.

4. **Aggregate `persist()` in manager-level helpers**:
   - `TopicConfigManager#deleteTopicConfigList(List<String>)`
   - `SubscriptionGroupManager#deleteSubscriptionGroupConfigList(List<String>)`

5. **Add client-side helpers** in `MQClientAPIImpl`:
   - `deleteTopicInBrokerList(addr, topicList, timeout)`
   - `deleteSubscriptionGroupList(addr, groupNameList, cleanOffset, timeout)`

- **Keep N RPCs but parallelize on the client side** (`DefaultMQAdminExtImpl#deleteTopicInBrokerConcurrent` already exists): still triggers N `persist()` calls per broker, doesn't address the persistence amplification on the broker side.
- **Async dispatch**: changing the contract to async would break callers that rely on synchronous success/failure semantics. Synchronous batch keeps the existing behavior.

### Additional Context

A PR will follow once this issue is created. Unit tests are added in `AdminBrokerProcessorTest`:

- `testDeleteTopicListInBroker`
- `testDeleteTopicListBatchPersist` (verifies single `persist()` and dedup)
- `testDeleteSubscriptionGroupList`

wang-jiahua:

Fixes #10446

### Brief Description

Adds two new batch admin APIs to `AdminBrokerProcessor` whose names follow the existing `*_LIST` convention (e.g. `UPDATE_AND_CREATE_TOPIC_LIST`, `UPDATE_AND_CREATE_SUBSCRIPTIONGROUP_LIST`):

- `DELETE_TOPIC_IN_BROKER_LIST`
- `DELETE_SUBSCRIPTIONGROUP_LIST`

Bulk metadata cleanup (e.g. tenant decommissioning, retention purge, cluster migration) currently requires N round-trips to delete N items and triggers N `persist()` calls on the broker config files. These new APIs allow doing the same in 1 RPC + 1 persist + 1 `messageStore.deleteTopics(Set)` call.

> Note: removing the `synchronized` modifier from `AdminBrokerProcessor#deleteTopic` is **out of scope** here and is already being discussed in #9997 / #9998. This PR keeps `synchronized` as-is for both the existing `deleteTopic` and the new `deleteTopicList`, so the lock semantics match the rest of the existing `updateAndCreate*` family.

#### Changes

- `RequestCode`: add `DELETE_TOPIC_IN_BROKER_LIST` and `DELETE_SUBSCRIPTIONGROUP_LIST`.
- New request bodies:
  - `DeleteTopicListRequestBody { List<String> topicList }`
  - `DeleteSubscriptionGroupListRequestBody { List<String> groupNameList; boolean cleanOffset }`
- `TopicConfigManager#deleteTopicConfigList(List<String>)` and `SubscriptionGroupManager#deleteSubscriptionGroupConfigList(List<String>)`: aggregate `persist()` once per batch.
- `AdminBrokerProcessor`:
  - Extract `collectPopRetryTopics(...)` shared helper used by both single and batch paths.
  - Add `deleteTopicList(...)` and `deleteSubscriptionGroupList(...)` (both `synchronized`, consistent with existing single APIs) with input dedup (preserving order), system-topic / blank validation, one-shot persist, and one-shot `messageStore.deleteTopics(...)` call.
- `MQClientAPIImpl`: add `deleteTopicInBrokerList(...)` and `deleteSubscriptionGroupList(...)` client helpers; both use an explicit null check (instead of `assert`) so callers get a deterministic `MQClientException(SYSTEM_ERROR, ...)` when `invokeSync` returns null.

#### Compatibility

- Single-item `DELETE_TOPIC_IN_BROKER` / `DELETE_SUBSCRIPTIONGROUP` request codes and methods are unchanged; existing clients are not affected.
- New `*_LIST` request codes are additive; older brokers will fall through to `getUnknownCmdResponse`, so older clients/brokers continue to work as today.

New unit tests added in `AdminBrokerProcessorTest`:

- `testDeleteTopicListInBroker` — covers empty input, system-topic rejection, and the success path.
- `testDeleteTopicListBatchPersist` — verifies that:
  - the manager-level `deleteTopicConfigList` is invoked once (so `persist()` runs once for the whole batch),
  - the singular `deleteTopicConfig` is **not** called on the batch path,
  - `messageStore.deleteTopics(...)` is invoked once with the batched set,
  - duplicate inputs are deduplicated.
- `testDeleteSubscriptionGroupList` — covers empty input and the success path with `cleanOffset=true`.

Local verification:

```
mvn -pl remoting,client,broker -Dspotbugs.skip=true validate
# 0 Checkstyle violations across all three modules

mvn -pl broker -Dcheckstyle.skip -Dspotbugs.skip=true -Djacoco.skip=true \
  -Dtest='AdminBrokerProcessorTest#testDeleteTopic+testDeleteTopicListInBroker+testDeleteTopicListBatchPersist+testDeleteSubscriptionGroup+testDeleteSubscriptionGroupList+testDeleteWithPopRetryTopic' \
  test
# Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -862,10 +862,17 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 if (LiteMetadataUtil.isLiteMessageType(topic, brokerController)) {
                     brokerController.getLiteLifecycleManager().cleanByParentTopic(topic);
                 }
-                deleteTopicInBroker(topic);
+                deleteTopicInBroker(topic, false);
             }
         } catch (Throwable t) {
             return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+        } finally {
+            try {
+                this.brokerController.getTopicConfigManager().persist();
+            } catch (Throwable t) {
+                LOGGER.error("Failed to persist topic config after batch delete", t);
+                return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+            }
         }
         response.setCode(ResponseCode.SUCCESS);
         response.setRemark(null);
@@ -886,13 +893,17 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         }
     }
 
-    private void deleteTopicInBroker(String topic) {
-        this.brokerController.getTopicConfigManager().deleteTopicConfig(topic);
+    private void deleteTopicInBroker(String topic, boolean persist) {
+        this.brokerController.getTopicConfigManager().deleteTopicConfig(topic, persist);
         this.brokerController.getTopicQueueMappingManager().delete(topic);
         this.brokerController.getConsumerOffsetManager().cleanOffsetByTopic(topic);
         this.brokerController.getPopInflightMessageCounter().clearInFlightMessageNumByTopicName(topic);
         this.brokerController.getMessageStore().deleteTopics(Sets.newHashSet(topic));
         this.brokerController.getMessageStore().getTimerMessageStore().getTimerMetrics().removeTimingCount(topic);
+    }
+
+    private void deleteTopicInBroker(String topic) {
+        deleteTopicInBroker(topic, true);
     }
 
     private RemotingCommand getUnknownCmdResponse(ChannelHandlerContext ctx, RemotingCommand request) {
@@ -1778,7 +1789,11 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
     }
 
     private void deleteSubscriptionGroupInBroker(String groupName, boolean cleanOffset) {
-        this.brokerController.getSubscriptionGroupManager().deleteSubscriptionGroupConfig(groupName);
+        deleteSubscriptionGroupInBroker(groupName, cleanOffset, true);
+    }
+
+    private void deleteSubscriptionGroupInBroker(String groupName, boolean cleanOffset, boolean persist) {
+        this.brokerController.getSubscriptionGroupManager().deleteSubscriptionGroupConfig(groupName, persist);
         if (cleanOffset) {
             this.brokerController.getConsumerOffsetManager().removeOffset(groupName);
             this.brokerController.getPopInflightMessageCounter().clearInFlightMessageNumByGroupName(groupName);
@@ -1831,10 +1846,17 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 }
                 boolean shouldCleanOffset = cleanOffset
                     || LiteMetadataUtil.isLiteGroupType(groupName, this.brokerController);
-                deleteSubscriptionGroupInBroker(groupName, shouldCleanOffset);
+                deleteSubscriptionGroupInBroker(groupName, shouldCleanOffset, false);
             }
         } catch (Throwable t) {
             return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+        } finally {
+            try {
+                this.brokerController.getSubscriptionGroupManager().persist();
+            } catch (Throwable t) {
+                LOGGER.error("Failed to persist subscription group config after batch delete", t);
+                return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+            }
         }
         response.setCode(ResponseCode.SUCCESS);
         response.setRemark(null);
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -28,9 +28,11 @@ import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -128,6 +130,8 @@ import org.apache.rocketmq.remoting.protocol.body.ConsumeQueueData;
 import org.apache.rocketmq.remoting.protocol.body.ConsumeStatsList;
 import org.apache.rocketmq.remoting.protocol.body.ConsumerConnection;
 import org.apache.rocketmq.remoting.protocol.body.CreateTopicListRequestBody;
+import org.apache.rocketmq.remoting.protocol.body.DeleteSubscriptionGroupListRequestBody;
+import org.apache.rocketmq.remoting.protocol.body.DeleteTopicListRequestBody;
 import org.apache.rocketmq.remoting.protocol.body.EpochEntryCache;
 import org.apache.rocketmq.remoting.protocol.body.GroupList;
 import org.apache.rocketmq.remoting.protocol.body.HARuntimeInfo;
@@ -270,6 +274,8 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 return this.updateAndCreateTopicList(ctx, request);
             case RequestCode.DELETE_TOPIC_IN_BROKER:
                 return this.deleteTopic(ctx, request);
+            case RequestCode.DELETE_TOPIC_IN_BROKER_LIST:
+                return this.deleteTopicList(ctx, request);
             case RequestCode.GET_ALL_TOPIC_CONFIG:
                 return this.getAllTopicConfig(ctx, request);
             case RequestCode.GET_TIMER_CHECK_POINT:
@@ -310,6 +316,8 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 return this.getAllSubscriptionGroup(ctx, request);
             case RequestCode.DELETE_SUBSCRIPTIONGROUP:
                 return this.deleteSubscriptionGroup(ctx, request);
+            case RequestCode.DELETE_SUBSCRIPTION_GROUP_LIST:
+                return this.deleteSubscriptionGroupList(ctx, request);
             case RequestCode.GET_TOPIC_STATS_INFO:
                 return this.getTopicStatsInfo(ctx, request);
             case RequestCode.GET_CONSUMER_CONNECTION_LIST:
@@ -782,17 +790,7 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         topicsToClean.add(topic);
 
         if (brokerController.getBrokerConfig().isClearRetryTopicWhenDeleteTopic()) {
-            final Set<String> groups = this.brokerController.getConsumerOffsetManager().whichGroupByTopic(topic);
-            for (String group : groups) {
-                final String popRetryTopicV2 = KeyBuilder.buildPopRetryTopic(topic, group, true);
-                if (brokerController.getTopicConfigManager().selectTopicConfig(popRetryTopicV2) != null) {
-                    topicsToClean.add(popRetryTopicV2);
-                }
-                final String popRetryTopicV1 = KeyBuilder.buildPopRetryTopicV1(topic, group);
-                if (brokerController.getTopicConfigManager().selectTopicConfig(popRetryTopicV1) != null) {
-                    topicsToClean.add(popRetryTopicV1);
-                }
-            }
+            collectPopRetryTopics(topic, topicsToClean);
         }
 
         try {
@@ -809,6 +807,83 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         response.setCode(ResponseCode.SUCCESS);
         response.setRemark(null);
         return response;
+    }
+
+    private synchronized RemotingCommand deleteTopicList(ChannelHandlerContext ctx,
+        RemotingCommand request) {
+        final RemotingCommand response = RemotingCommand.createResponseCommand(null);
+
+        DeleteTopicListRequestBody requestBody = DeleteTopicListRequestBody.decode(
+            request.getBody(), DeleteTopicListRequestBody.class);
+        List<String> topicList = requestBody == null ? null : requestBody.getTopicList();
+
+        if (CollectionUtils.isEmpty(topicList)) {
+            response.setCode(ResponseCode.INVALID_PARAMETER);
+            response.setRemark("The specified topic list is blank.");
+            return response;
+        }
+
+        LOGGER.info("AdminBrokerProcessor#deleteTopicList: broker receive request to delete topics={}, caller={}",
+            topicList, RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
+
+        boolean validateSystemTopic = brokerController.getBrokerConfig().isValidateSystemTopicWhenUpdateTopic();
+        // dedup while preserving the input order
+        Set<String> topicsToClean = new LinkedHashSet<>();
+        for (String topic : topicList) {
+            if (UtilAll.isBlank(topic)) {
+                response.setCode(ResponseCode.INVALID_PARAMETER);
+                response.setRemark("The specified topic is blank.");
+                return response;
+            }
+            if (validateSystemTopic && TopicValidator.isSystemTopic(topic)) {
+                response.setCode(ResponseCode.INVALID_PARAMETER);
+                response.setRemark("The topic[" + topic + "] is conflict with system topic.");
+                return response;
+            }
+            topicsToClean.add(topic);
+        }
+
+        if (brokerController.getBrokerConfig().isClearRetryTopicWhenDeleteTopic()) {
+            // snapshot the inputs before mutating the set, so retry topics for already-added retry topics are not collected
+            for (String topic : new ArrayList<>(topicsToClean)) {
+                collectPopRetryTopics(topic, topicsToClean);
+            }
+        }
+
+        try {
+            double maxRate = this.brokerController.getBrokerConfig().getBatchDeleteTopicMaxRate();
+            com.google.common.util.concurrent.RateLimiter rateLimiter = maxRate > 0
+                ? com.google.common.util.concurrent.RateLimiter.create(maxRate) : null;
+
+            for (String topic : topicsToClean) {
+                if (rateLimiter != null) {
+                    rateLimiter.acquire();
+                }
+                if (LiteMetadataUtil.isLiteMessageType(topic, brokerController)) {
+                    brokerController.getLiteLifecycleManager().cleanByParentTopic(topic);
+                }
+                deleteTopicInBroker(topic);
+            }
+        } catch (Throwable t) {
+            return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+        }
+        response.setCode(ResponseCode.SUCCESS);
+        response.setRemark(null);
+        return response;
+    }
+
+    private void collectPopRetryTopics(String topic, Collection<String> topicsToClean) {
+        final Set<String> groups = this.brokerController.getConsumerOffsetManager().whichGroupByTopic(topic);
+        for (String group : groups) {
+            final String popRetryTopicV2 = KeyBuilder.buildPopRetryTopic(topic, group, true);
+            if (brokerController.getTopicConfigManager().selectTopicConfig(popRetryTopicV2) != null) {
+                topicsToClean.add(popRetryTopicV2);
+            }
+            final String popRetryTopicV1 = KeyBuilder.buildPopRetryTopicV1(topic, group);
+            if (brokerController.getTopicConfigManager().selectTopicConfig(popRetryTopicV1) != null) {
+                topicsToClean.add(popRetryTopicV1);
+            }
+        }
     }
 
     private void deleteTopicInBroker(String topic) {
@@ -1693,16 +1768,73 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         LOGGER.info("AdminBrokerProcessor#deleteSubscriptionGroup, caller={}",
             RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
 
-        this.brokerController.getSubscriptionGroupManager().deleteSubscriptionGroupConfig(requestHeader.getGroupName());
+        boolean cleanOffset = requestHeader.isCleanOffset()
+            || LiteMetadataUtil.isLiteGroupType(requestHeader.getGroupName(), this.brokerController);
+        deleteSubscriptionGroupInBroker(requestHeader.getGroupName(), cleanOffset);
 
-        if (requestHeader.isCleanOffset()
-            || LiteMetadataUtil.isLiteGroupType(requestHeader.getGroupName(), this.brokerController)) {
-            this.brokerController.getConsumerOffsetManager().removeOffset(requestHeader.getGroupName());
-            this.brokerController.getPopInflightMessageCounter().clearInFlightMessageNumByGroupName(requestHeader.getGroupName());
+        response.setCode(ResponseCode.SUCCESS);
+        response.setRemark(null);
+        return response;
+    }
+
+    private void deleteSubscriptionGroupInBroker(String groupName, boolean cleanOffset) {
+        this.brokerController.getSubscriptionGroupManager().deleteSubscriptionGroupConfig(groupName);
+        if (cleanOffset) {
+            this.brokerController.getConsumerOffsetManager().removeOffset(groupName);
+            this.brokerController.getPopInflightMessageCounter().clearInFlightMessageNumByGroupName(groupName);
+        }
+        if (this.brokerController.getBrokerConfig().isAutoDeleteUnusedStats()) {
+            this.brokerController.getBrokerStatsManager().onGroupDeleted(groupName);
+        }
+    }
+
+    private RemotingCommand deleteSubscriptionGroupList(ChannelHandlerContext ctx,
+        RemotingCommand request) {
+        final RemotingCommand response = RemotingCommand.createResponseCommand(null);
+
+        DeleteSubscriptionGroupListRequestBody requestBody = DeleteSubscriptionGroupListRequestBody.decode(
+            request.getBody(), DeleteSubscriptionGroupListRequestBody.class);
+        List<String> groupNameList = requestBody == null ? null : requestBody.getGroupNameList();
+
+        if (CollectionUtils.isEmpty(groupNameList)) {
+            response.setCode(ResponseCode.INVALID_PARAMETER);
+            response.setRemark("The specified group name list is blank.");
+            return response;
         }
 
-        if (this.brokerController.getBrokerConfig().isAutoDeleteUnusedStats()) {
-            this.brokerController.getBrokerStatsManager().onGroupDeleted(requestHeader.getGroupName());
+        // dedup while preserving the input order
+        Set<String> groupNames = new LinkedHashSet<>();
+        for (String groupName : groupNameList) {
+            if (UtilAll.isBlank(groupName)) {
+                response.setCode(ResponseCode.INVALID_PARAMETER);
+                response.setRemark("The specified group name is blank.");
+                return response;
+            }
+            groupNames.add(groupName);
+        }
+
+        LOGGER.info("AdminBrokerProcessor#deleteSubscriptionGroupList: groupNames={}, caller={}",
+            groupNames, RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
+
+        try {
+            boolean cleanOffset = requestBody.isCleanOffset();
+            double maxRate = this.brokerController.getBrokerConfig().getBatchDeleteSubscriptionGroupMaxRate();
+            com.google.common.util.concurrent.RateLimiter rateLimiter = maxRate > 0
+                ? com.google.common.util.concurrent.RateLimiter.create(maxRate) : null;
+
+            // Check isLiteGroupType before deletion: once the group config is removed,
+            // findSubscriptionGroupConfig(...) may auto-recreate the group when
+            // autoCreateSubscriptionGroup=true.
+            for (String groupName : groupNames) {
+                if (rateLimiter != null) {
+                    rateLimiter.acquire();
+                }
+                boolean shouldCleanOffset = cleanOffset
+                    || LiteMetadataUtil.isLiteGroupType(groupName, this.brokerController);
+                deleteSubscriptionGroupInBroker(groupName, shouldCleanOffset);
+            }
+        } catch (Throwable t) {
+            return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
         }
         response.setCode(ResponseCode.SUCCESS);
         response.setRemark(null);
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -850,6 +850,7 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
             }
         }
 
+        boolean isSuccess = false;
         try {
             double maxRate = this.brokerController.getBrokerConfig().getBatchDeleteTopicMaxRate();
             com.google.common.util.concurrent.RateLimiter rateLimiter = maxRate > 0
@@ -864,18 +865,21 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 }
                 deleteTopicInBroker(topic, false);
             }
+            isSuccess  = true;
         } catch (Throwable t) {
-            return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+            LOGGER.error("Failed to delete topic ", t);
         } finally {
             try {
                 this.brokerController.getTopicConfigManager().persist();
             } catch (Throwable t) {
+                isSuccess = false;
                 LOGGER.error("Failed to persist topic config after batch delete", t);
-                return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
             }
         }
-        response.setCode(ResponseCode.SUCCESS);
+
+        response.setCode(isSuccess ? ResponseCode.SUCCESS : ResponseCode.SYSTEM_ERROR);
         response.setRemark(null);
+
         return response;
     }
 
@@ -1831,6 +1835,7 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
         LOGGER.info("AdminBrokerProcessor#deleteSubscriptionGroupList: groupNames={}, caller={}",
             groupNames, RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
 
+        boolean isSuccess = false;
         try {
             boolean cleanOffset = requestBody.isCleanOffset();
             double maxRate = this.brokerController.getBrokerConfig().getBatchDeleteSubscriptionGroupMaxRate();
@@ -1848,17 +1853,18 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                     || LiteMetadataUtil.isLiteGroupType(groupName, this.brokerController);
                 deleteSubscriptionGroupInBroker(groupName, shouldCleanOffset, false);
             }
+            isSuccess = true;
         } catch (Throwable t) {
-            return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
+            LOGGER.error("Failed to delete subscription group config ", t);
         } finally {
             try {
                 this.brokerController.getSubscriptionGroupManager().persist();
             } catch (Throwable t) {
+                isSuccess = false;
                 LOGGER.error("Failed to persist subscription group config after batch delete", t);
-                return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
             }
         }
-        response.setCode(ResponseCode.SUCCESS);
+        response.setCode(isSuccess ? ResponseCode.SUCCESS : ResponseCode.SYSTEM_ERROR);
         response.setRemark(null);
         return response;
     }
--- a/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
@@ -626,6 +626,15 @@ public class TopicConfigManager extends ConfigManager {
         }
     }
 
+    public void deleteTopicConfigList(final List<String> topics) {
+        if (topics == null || topics.isEmpty()) {
+            return;
+        }
+        for (String topic : topics) {
+            deleteTopicConfig(topic);
+        }
+    }
+
     public TopicConfigSerializeWrapper buildTopicConfigSerializeWrapper() {
         TopicConfigSerializeWrapper topicConfigSerializeWrapper = new TopicConfigSerializeWrapper();
         DataVersion dataVersionCopy = new DataVersion();
--- a/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
@@ -615,15 +615,21 @@ public class TopicConfigManager extends ConfigManager {
         }
     }
 
-    public void deleteTopicConfig(final String topic) {
+    public void deleteTopicConfig(final String topic, boolean persist) {
         TopicConfig old = removeTopicConfig(topic);
         if (old != null) {
             log.info("delete topic config OK, topic: {}", old);
             updateDataVersion();
-            this.persist();
+            if (persist) {
+                this.persist();
+            }
         } else {
             log.warn("delete topic config failed, topic: {} not exists", topic);
         }
+    }
+
+    public void deleteTopicConfig(final String topic) {
+        deleteTopicConfig(topic, true);
     }
 
     public void deleteTopicConfigList(final List<String> topics) {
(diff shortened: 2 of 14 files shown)


oss-sentinel-ai (approved):
﻿## Review by github-manager-bot

### Summary
This PR addresses two related performance issues in `AdminBrokerProcessor`:
1. Removes `synchronized` from `deleteTopic` since `topicConfigTable` is already a `ConcurrentHashMap`
2. Adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST` and `DELETE_SUBSCRIPTIONGROUP_LIST`) to reduce network round-trips

### Findings
- **[Good]** AdminBrokerProcessor.java:769 — Removed `synchronized` from `deleteTopic`. The underlying `topicConfigTable` is a `ConcurrentHashMap`, so per-key `remove` is atomic. This eliminates unnecessary serialization of concurrent admin delete requests.
- **[Good]** AdminBrokerProcessor.java — Added `deleteTopicList()` and `deleteSubscriptionGroupList()` handlers following the existing `*_LIST` convention (consistent with `UPDATE_AND_CREATE_TOPIC_LIST`).
- **[Good]** DeleteTopicListRequestBody.java / DeleteSubscriptionGroupListRequestBody.java — Clean request body classes extending `RemotingSerializable`, following existing patterns.
- **[Good]** MQClientAPIImpl.java:2202-2220 / 2281-2301 — Client-side batch deletion methods properly implemented with timeout handling.
- **[Good]** RequestCode.java — Added `DELETE_TOPIC_IN_BROKER_LIST` and `DELETE_SUBSCRIPTIONGROUP_LIST` request codes.
- **[Good]** TopicConfigManager.java / SubscriptionGroupManager.java — Added `deleteTopicList()` and `deleteSubscriptionGroupList()` batch methods.
- **[Good]** AdminBrokerProcessorTest.java — Comprehensive test coverage including:
  - Empty list validation (returns INVALID_PARAMETER)
  - System topic rejection in batch
  - Happy path with multiple topics
  - Duplicate handling in batch
  - Batch persist verification

### Analysis
The implementation is well-structured:
1. **Backward compatibility**: Existing single-item deletion APIs remain unchanged
2. **Consistent patterns**: Follows the `*_LIST` convention established by `CreateTopicListRequestBody`
3. **Proper validation**: Empty lists and system topics are rejected
4. **Test coverage**: Edge cases are well covered

### Suggestions
1. Consider adding a maximum batch size limit to prevent excessive memory usage or long-running operations.
2. The batch persist test verifies deduplication — good attention to edge cases.

### Verdict
✅ **Approved** — Well-designed performance optimization with proper validation and comprehensive test coverage.

---

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -758,7 +766,7 @@ private synchronized RemotingCommand updateAndCreateStaticTopic(ChannelHandlerCo
         return response;
     }
 
-    private synchronized RemotingCommand deleteTopic(ChannelHandlerContext ctx,
+    private RemotingCommand deleteTopic(ChannelHandlerContext ctx,
         RemotingCommand request) throws RemotingCommandException {
```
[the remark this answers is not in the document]
wang-jiahua: Done in b3e851de2. Adopted the second alternative you suggested: introduced a `private final Object adminTopicLock = new Object();` shared by both `deleteTopic(...)` and `deleteTopicList(...)`. The processor-level `synchronized` is gone, so unrelated admin operations (e.g. `updateAndCreateSubscriptionGroup`, getters) are no longer blocked, but topic delete paths remain mutually atomic. Concurrency between `deleteTopic` and `updateAndCreateTopic*` is a pre-existing trade-off and is intentionally left out of this PR's scope.

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -2197,6 +2199,25 @@ public void deleteTopicInBroker(final String addr, final String topic, final lon
...
     }
 
+    public void deleteTopicInBrokerList(final String addr, final List<String> topicList, final long timeoutMillis)
+        throws RemotingException, InterruptedException, MQClientException {
+        DeleteTopicListRequestBody requestBody = new DeleteTopicListRequestBody(topicList);
+        RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.DELETE_TOPIC_IN_BROKER_LIST, null);
+        request.setBody(requestBody.encode());
+        RemotingCommand response = this.remotingClient.invokeSync(MixAll.brokerVIPChannel(this.clientConfig.isVipChannelEnabled(), addr),
+            request, timeoutMillis);
+        assert response != null;
```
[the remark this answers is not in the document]
wang-jiahua: Done in b3e851de2. Replaced `assert response != null;` in `deleteTopicInBrokerList` with an explicit null check that throws `MQClientException(SYSTEM_ERROR, ...)`, so callers get a deterministic failure even with assertions disabled.

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -2257,6 +2278,27 @@ public void deleteSubscriptionGroup(final String addr, final String groupName, f
...
+    public void deleteSubscriptionGroupList(final String addr, final List<String> groupNameList, final boolean removeOffset,
+        final long timeoutMillis)
+        throws RemotingException, InterruptedException, MQClientException {
+        DeleteSubscriptionGroupListRequestBody requestBody = new DeleteSubscriptionGroupListRequestBody(groupNameList, removeOffset);
+        RemotingCommand request = RemotingCommand.createRequestCommand(RequestCode.DELETE_SUBSCRIPTIONGROUP_LIST, null);
+        request.setBody(requestBody.encode());
+
+        RemotingCommand response = this.remotingClient.invokeSync(MixAll.brokerVIPChannel(this.clientConfig.isVipChannelEnabled(), addr),
+            request, timeoutMillis);
+        assert response != null;
```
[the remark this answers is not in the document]
wang-jiahua: Done in b3e851de2. Same fix applied to `deleteSubscriptionGroupList`: explicit null check + `MQClientException(SYSTEM_ERROR, ...)` instead of `assert`.

client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
```diff
@@ -2257,6 +2278,27 @@ public void deleteSubscriptionGroup(final String addr, final String groupName, f
         throw new MQClientException(response.getCode(), response.getRemark());
     }
 
+    public void deleteSubscriptionGroupList(final String addr, final List<String> groupNameList, final boolean removeOffset,
```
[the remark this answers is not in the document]
wang-jiahua: Done in b3e851de2. Renamed the new client API parameter from `removeOffset` to `cleanOffset` so it now matches `DeleteSubscriptionGroupListRequestBody#cleanOffset` and the broker-side `requestBody.isCleanOffset()`. The legacy single-item `deleteSubscriptionGroup(..., boolean removeOffset, ...)` is left untouched to avoid breaking existing callers.

broker/src/main/java/org/apache/rocketmq/broker/topic/TopicConfigManager.java
```diff
@@ -626,6 +626,26 @@ public void deleteTopicConfig(final String topic) {
...
 
+    public void deleteTopicConfigList(final List<String> topics) {
+        if (topics == null || topics.isEmpty()) {
+            return;
+        }
+        boolean changed = false;
+        for (String topic : topics) {
+            TopicConfig old = removeTopicConfig(topic);
+            if (old != null) {
+                log.info("delete topic config OK, topic: {}", old);
```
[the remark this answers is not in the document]
wang-jiahua: Done in b3e851de2. Updated the log message in the new `deleteTopicConfigList` (and the analogous one added in `SubscriptionGroupManager#deleteSubscriptionGroupConfigList`) to log the topic / group name instead of the full config object, so the placeholder name now matches the value. The original `deleteTopicConfig` / `deleteSubscriptionGroupConfig` log lines are pre-existing and outside this PR's scope.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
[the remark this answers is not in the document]
wang-jiahua: Update: scope of this PR has been narrowed back to **only the new batch APIs**. The `synchronized` keyword on `deleteTopic` is now restored to its original behavior, and the new `deleteTopicList` / `deleteSubscriptionGroupList` are also marked `synchronized` to stay consistent with the existing `deleteTopic` / `updateAndCreate*` family. The race-condition concern you raised therefore no longer applies in this PR. Removing `synchronized` from `deleteTopic` is being discussed separately in #9997 / #9998 and is intentionally left out of this scope. Code at 75ae0ef8d.

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RequestCode.java
```diff
@@ -311,4 +311,7 @@ public class RequestCode {
     public static final int AUTH_LIST_ACL = 3010;
 
     public static final int SWITCH_TIMER_ENGINE = 5001;
+
+    public static final int DELETE_TOPIC_IN_BROKER_LIST = 5002;
+    public static final int DELETE_SUBSCRIPTIONGROUP_LIST = 5003;
```
RongtongJin: These new destructive request codes are not associated with any `CommandCustomHeader` annotated with `@RocketMQAction` / `@RocketMQResource`. The authorization builder resolves actions through `RequestHeaderRegistry`, which only scans annotated request headers; these batch APIs send `null` custom headers and put the topic/group list in the body, so authorization can produce an empty context list and allow the request without checking DELETE permission. Please add annotated headers for these request codes, or explicitly teach the authorization builder to parse these bodies and create DELETE contexts for each topic/group.

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -816,6 +814,84 @@ private synchronized RemotingCommand deleteTopic(ChannelHandlerContext ctx,
...
+        }
+
+        try {
+            for (String topic : topicsToClean) {
+                if (LiteMetadataUtil.isLiteMessageType(topic, brokerController)) {
+                    brokerController.getLiteLifecycleManager().cleanByParentTopic(topic);
+                }
+            }
+            // batch delete topic config in one persist
+            this.brokerController.getTopicConfigManager().deleteTopicConfigList(new ArrayList<>(topicsToClean));
```
RongtongJin: Consider the partial-failure retry case here. After this call removes the topic configs, a retry of the same batch rebuilds `topicsToClean` from the original request and `collectPopRetryTopics()`, but that helper only adds POP retry topics whose configs still exist. If the first attempt fails after this config deletion but before queue mapping / offset / timer metrics / message store cleanup finishes, those retry topics may not be rediscovered on retry, so the remaining cleanup for them can be skipped. Please make the delete plan retry-stable, or explicitly handle/document this partial-failure behavior.

RongtongJin: If making this fully retry-safe is too invasive, please at least add an explicit code comment near this call documenting the risk: once the topic configs are removed, a later retry may no longer rediscover derived POP retry topics, so callers should not assume this batch delete is fully idempotent after a partial failure.

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RequestCode.java
wang-jiahua: Good catch — fixed in 53cda760b.

Took the second alternative you suggested (teach the authorization builder to parse the body): added explicit cases for `DELETE_TOPIC_IN_BROKER_LIST` and `DELETE_SUBSCRIPTIONGROUP_LIST` in `DefaultAuthorizationContextBuilder.build(ChannelHandlerContext, RemotingCommand)`, mirroring the existing `LOCK_BATCH_MQ` / `UNLOCK_BATCH_MQ` pattern. Each topic / group in the body now produces its own `Action.DELETE` authorization context, blank entries are skipped, and the `rpcCode` is stamped accordingly. Annotated headers were not added because `CommandCustomHeader` cannot natively carry a `List<String>`, and the new bodies (`DeleteTopicListRequestBody` / `DeleteSubscriptionGroupListRequestBody`) already carry the full target list.

Covered by new assertions in `DefaultAuthorizationContextBuilderTest.buildRemoting` (validates one DELETE context per non-blank topic/group, correct `ResourceType`, and stamped `rpcCode`).

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
wang-jiahua: Fixed in 6bebee2c9. Moved `deleteTopicConfigList` after all other cleanup steps (queue mapping, offsets, inflight counter, timer metrics, message store). This way, if the batch fails partway through and the caller retries, `collectPopRetryTopics` → `selectTopicConfig` can still find the configs and rediscover derived POP retry topics.

RockteMQ-AI (commented):

### Summary
Adds batch deletion APIs for topics (`DELETE_TOPIC_IN_BROKER_LIST`) and subscription groups (`DELETE_SUBSCRIPTIONGROUP_LIST`), with proper per-resource authorization, input validation, and atomic-all-or-nothing semantics.

### Findings
- **[Info]** Authorization — Correctly emits one `DELETE` context per resource in the batch, preventing bypass when the body carries the target list instead of the header. This is a security-critical detail handled well.
- **[Info]** Error handling — Returns a structured error response with the list of failed items and the first exception. Good for debugging partial failures.
- **[Info]** Dedup — Uses `LinkedHashSet` to preserve insertion order while deduplicating. Reasonable choice.
- **[Info]** Refactoring — Extracts `collectPopRetryTopics()` from inline code in `deleteTopic()`. Good cleanup that avoids duplication.
- **[Info]** Cleanup order — Deletes client config and subscription first, then topic config last. Allows retry after partial failure.
- **[Warning]** `AdminBrokerProcessor.deleteTopicInBrokerList()` — The `RemotingCommand response` is created once at the top and reused. If `deleteTopicInBroker` internally sets response fields, ensure there's no state leakage between the success and error paths.
- **[Warning]** `MQClientAPIImpl.deleteTopicInBrokerList()` — The method creates the request body and sends it. Consider adding a timeout or batch size limit to prevent excessively large requests.

### Suggestions
- Consider adding a max batch size constant (e.g., 100) to prevent abuse or accidental large requests.
- The `DeleteTopicListRequestBody` and `DeleteSubscriptionGroupListRequestBody` classes are clean. Consider adding `@AllArgsConstructor` for convenience if not already present.
- Cross-repo: If batch deletion is exposed in the dashboard or CLI, ensure those clients are updated to use the new APIs.

Overall: Well-structured feature with proper authorization and error handling. A few minor suggestions but nothing blocking.

---

RockteMQ-AI (commented):

### Summary
Adds two new batch admin APIs — `DELETE_TOPIC_IN_BROKER_LIST` and `DELETE_SUBSCRIPTIONGROUP_LIST` — to `AdminBrokerProcessor`, following the existing `*_LIST` convention (e.g. `UPDATE_AND_CREATE_TOPIC_LIST`). Includes proper authorization context building for body-driven batch requests, new request body DTOs, and test coverage.

### Findings
- **[Info]** Authorization handling for body-driven batch APIs is correctly implemented — the `DefaultAuthorizationContextBuilder` decodes the request body and emits one `DELETE` context per item, preventing the annotation-based path from producing an empty context list (which would bypass permission checks).
- **[Info]** Good test coverage in `DefaultAuthorizationContextBuilderTest` — verifies blank entries are filtered, correct resource types, and per-item DELETE action.
- **[Warning]** The batch deletion is not atomic — if one topic/group fails to delete (e.g. still in use), the remaining items in the list may still be deleted. Consider documenting this partial-failure behavior in the API response, or adding a `failFast` flag.
- **[Info]** `LinkedHashSet` for deduplication while preserving insertion order is a good choice for predictable behavior.
- **[Info]** New `DeleteTopicListRequestBody` and `DeleteSubscriptionGroupListRequestBody` follow the existing pattern of `CreateTopicListRequestBody`.

### Suggestions
- Consider adding a response summary that reports per-item success/failure status (similar to how batch operations in other systems report partial results). This would help callers understand which items succeeded when partial failure occurs.
- Add a brief comment in `AdminBrokerProcessor` documenting the non-atomic semantics of batch deletion.

No blocking issues. Well-structured addition that follows existing conventions.

---

wang-jiahua:

| #10448 | 216k | 1.19ms | 296 | **Parity (feature PR, no perf change)** ✅ |

This is a feature PR (batch delete Topic/SubscriptionGroup), not a performance optimization. Benchmark confirms no regression.

fuyou001:
For `deleteTopicList`, I suggest reusing the existing single-topic delete path instead of maintaining a separate cleanup sequence. The batch handler can iterate over the requested topics, call `deleteTopic(...)` for each one, and guard the loop with a `RateLimiter` such as `RateLimiter.create(10.0)` so it deletes at most 10 topics per second.

That keeps the batch API aligned with the current single-delete semantics and avoids a large batch request creating a sudden cleanup burst on the broker. It would also be good to add a regression test covering the throttled batch path and invalid-entry handling before the loop starts.

RockteMQ-AI:
## Review by github-manager-bot (Re-review after new commit `4f0ab43`)

### Summary
Adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST` / `DELETE_SUBSCRIPTIONGROUP_LIST`) to reduce N RPCs + N `persist()` calls to 1 RPC + 1 persist. The new commit restructures the diff but retains the separate batch cleanup path.

### Findings

- **[Warning]** `AdminBrokerProcessor.java` — **Separate cleanup path vs. @fuyou001 suggestion.** The batch `deleteTopicList` maintains its own cleanup sequence rather than iterating over the existing `deleteTopic()`. This is intentional per the PR description (1 persist + 1 `deleteTopics()` call), and it does achieve the stated performance goal. However, any future fix to `deleteTopic()` (e.g., additional cleanup steps) must be manually mirrored here. Consider adding a comment at both methods cross-referencing each other, or extracting the common cleanup steps into a shared helper that both paths can call.

- **[Warning]** `AdminBrokerProcessor.java:deleteTopicList` — **Partial failure semantics.** If `messageStore.deleteTopics()` succeeds but `deleteTopicConfigList()` throws, the physical data is gone but config metadata remains. On retry, `collectPopRetryTopics` can still discover derived topics via `selectTopicConfig`, which is good. But the caller gets `SYSTEM_ERROR` without knowing what was partially cleaned. Consider logging the successfully cleaned topics before the exception, or returning a partial-success response.

- **[Info]** `MQClientAPIImpl.java:deleteTopicInBrokerList` — The `switch` with empty `case SUCCESS: return;` followed by `default: break;` then `throw` is functionally correct and consistent with the existing `deleteTopicInBroker` pattern. No action needed.

- **[Info]** `DefaultAuthorizationContextBuilder.java` — Good handling of body-driven auth contexts. The comment explaining why the annotation-based path doesn't work for batch APIs is helpful for future maintainers.

- **[Info]** `RequestCode.java` — New codes 5002/5003 are in the 5xxx range (timer/engine area). Not a bug, but worth confirming these don't collide with any in-flight PRs using the 5xxx range.

- **[Info]** Tests — Good coverage: empty input, system topic rejection, dedup verification (single persist call), and batch subscription group deletion. The `testDeleteTopicListBatchPersist` test correctly verifies that `deleteTopicConfig` (singular) is NOT called on the batch path.

### Regarding @fuyou001 RateLimiter Suggestion
The current approach prioritizes batch efficiency (1 persist + 1 store call). A `RateLimiter` wrapper around individual `deleteTopic()` calls would trade some throughput for safer burst control. This is a design decision for the maintainers — both approaches are valid. If the RateLimiter approach is adopted, the batch API can internally iterate with rate limiting while still saving the network round-trips.

### Assessment
Well-structured addition that follows the existing `*_LIST` convention. Authorization, dedup, validation, and test coverage are all solid. The two warnings above (cleanup path divergence and partial failure) are worth addressing but not blockers.

---

RockteMQ-AI (commented):

### Summary
Updated review for the latest push (4f0ab43). This PR adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST`, `DELETE_SUBSCRIPTIONGROUP_LIST`) to `AdminBrokerProcessor`, following the existing `*_LIST` convention. Includes per-resource authorization context building, new request body DTOs, client helpers, and test coverage.

### What Changed Since Last Review
- Single commit (4f0ab43) rebased/updated on 2026-06-24.
- A new reviewer suggestion from @fuyou001 proposes reusing the single-topic delete path with a `RateLimiter` instead of a separate batch cleanup sequence.

### Findings

- **[Info]** Authorization — `DefaultAuthorizationContextBuilder` correctly decodes the request body and emits one `DELETE` context per resource. This prevents the annotation-based path from producing an empty context list (which would bypass permission checks). Well handled.
- **[Info]** Dedup + ordering — `LinkedHashSet` preserves insertion order while deduplicating inputs. Good choice for predictable behavior.
- **[Info]** Refactoring — `collectPopRetryTopics(...)` is extracted from inline code in `deleteTopic()`, now shared by both single and batch paths. Reduces duplication.
- **[Info]** Batch persist — `TopicConfigManager.deleteTopicConfigList()` and `SubscriptionGroupManager.deleteSubscriptionGroupConfigList()` aggregate `persist()` once per batch instead of N times. Significant efficiency gain for large batches.
- **[Info]** Error reporting — Returns structured error with the list of failed items and the first exception. Good for debugging partial failures.
- **[Warning]** No batch size limit — Neither `deleteTopicList` nor `deleteSubscriptionGroupList` caps the input list size. A caller could send thousands of items in a single RPC, causing a long-running synchronized block and a large `messageStore.deleteTopics(...)` call. Consider adding a max batch size constant (e.g., 100 or 200) and returning `REQUEST_CODE_NOT_SUPPORTED` or a validation error when exceeded.
- **[Warning]** Rate limiting concern (raised by @fuyou001) — The current batch path does a burst of all deletes in one shot. If the reviewer suggestion is adopted (iterate + `RateLimiter`), it would smooth out the cleanup burst. If not adopted, at minimum document why a burst is acceptable for the expected use cases (tenant decommissioning, retention purge).
- **[Warning]** `synchronized` scope — Both `deleteTopicList` and `deleteSubscriptionGroupList` are `synchronized`, consistent with existing single-item methods. However, the batch path holds the lock for the entire batch duration. For large batches this could block other admin operations. This is acknowledged as out-of-scope (#9997/#9998) but worth noting for reviewers.
- **[Info]** Client null-check — `MQClientAPIImpl.deleteTopicInBrokerList()` and `deleteSubscriptionGroupList()` use explicit null checks on the response (instead of `assert`), producing a deterministic `MQClientException(SYSTEM_ERROR, ...)`. Good defensive coding.
- **[Info]** Backward compatibility — New `*_LIST` request codes are additive; older brokers will return `REQUEST_CODE_NOT_SUPPORTED` via `getUnknownCmdResponse`. No breaking changes.

### Suggestions
1. **Add a max batch size guard** — e.g., `if (topicList.size() > MAX_BATCH_SIZE) { return error response; }`. This is a low-cost safety measure.
2. **Address @fuyou001 feedback** — Either adopt the RateLimiter approach or explain why the current burst behavior is acceptable for the target use cases.
3. **Consider adding a timeout parameter** — The client methods accept `timeoutMillis` but the broker-side processing has no internal timeout. For very large batches, the broker could log a warning if processing exceeds a threshold.

Overall: Solid implementation with proper authorization, error handling, and test coverage. The main concern is the lack of batch size limits and the reviewer suggestion about rate limiting. Nothing blocking, but worth addressing before merge.

---

broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
```diff
@@ -1714,6 +1791,58 @@ private RemotingCommand deleteSubscriptionGroup(ChannelHandlerContext ctx,
...
+
+        try {
+            // batch delete subscription group config in one persist
+            this.brokerController.getSubscriptionGroupManager()
+                .deleteSubscriptionGroupConfigList(new ArrayList<>(groupNames));
+
+            boolean cleanOffset = requestBody.isCleanOffset();
+            boolean autoDeleteUnusedStats = this.brokerController.getBrokerConfig().isAutoDeleteUnusedStats();
+            for (String groupName : groupNames) {
+                if (cleanOffset || LiteMetadataUtil.isLiteGroupType(groupName, this.brokerController)) {
```
RongtongJin: Please snapshot this lite-group state before calling `deleteSubscriptionGroupConfigList(...)`. At this point the group config has already been removed, and `LiteMetadataUtil.isLiteGroupType(...)` calls `findSubscriptionGroupConfig(...)`. With the default `autoCreateSubscriptionGroup=true`, a missing group can be auto-created again when `cleanOffset` is false, so the batch delete can return success while the group config is recreated. A regression test for the `cleanOffset=false` path would catch this.

wang-jiahua: Good catch — fixed in 930d6e6d8.

Snapshotted lite-group state **before** calling `deleteSubscriptionGroupConfigList(...)`:

```java
boolean cleanOffset = requestBody.isCleanOffset();
boolean autoDeleteUnusedStats = ...;
// Snapshot lite-group state before deletion, because once the group config is removed,
// LiteMetadataUtil.isLiteGroupType(...) calls findSubscriptionGroupConfig(...) which may
// auto-recreate the group when autoCreateSubscriptionGroup=true.
Set<String> liteGroups = new HashSet<>();
if (!cleanOffset) {
    for (String groupName : groupNames) {
        if (LiteMetadataUtil.isLiteGroupType(groupName, this.brokerController)) {
            liteGroups.add(groupName);
        }
    }
}

this.brokerController.getSubscriptionGroupManager()
    .deleteSubscriptionGroupConfigList(new ArrayList<>(groupNames));

for (String groupName : groupNames) {
    if (cleanOffset || liteGroups.contains(groupName)) {
        ...
    }
}
```

This avoids the auto-recreation race. Will add a regression test for the `cleanOffset=false` path.

wang-jiahua:
@fuyou001 Thanks for the suggestion!

> For `deleteTopicList`, I suggest reusing the existing single-topic delete path instead of maintaining a separate cleanup sequence.

We ran a **broker-side A/B experiment** comparing both approaches with the same batch RPC (only the broker JAR differs):

- **Variant A** (your suggestion): `deleteTopicConfigList()` loops `deleteTopicConfig()` → each call triggers `persist()` (3 × fsync per item)
- **Variant B** (current): `deleteTopicConfigList()` removes all from memory, then calls `persist()` once (3 fsync total)

**Server benchmark results (50 topics/groups, Producer+Consumer running at ~180k TPS):**

| Metric | Variant A (loop persist) | Variant B (batch persist) | Speedup |
|--------|-------------------------|--------------------------|---------|
| **Topic batch delete** | 730ms | 133ms | **5.5x** |
| **Group batch delete** | 397ms | 18ms | **22x** |
| Producer TPS impact | 0% | 0% | — |
| Producer P99 RT | 5ms | 5ms | — |
| GC delta | 16 | 16 | — |

The key bottleneck is `persist()` → `fsync`. Each `deleteTopicConfig()` call serializes the full config to JSON, writes to disk, then fsyncs file + directory. With 50 items, that's 150 fsync calls in Variant A vs 3 in Variant B.

Both variants have zero impact on concurrent Producer/Consumer TPS and P99 RT because the delete operation completes in under 1 second either way. The difference is in the **delete operation latency itself** — important for admin tools that wait for the response.

**Regarding code reuse concern:** The batch path only diverges at the `removeTopicConfig` + single `persist()` level. All other cleanup logic (queue mapping, consumer offset, pop inflight counter, timer metrics, message store) is shared and unchanged.

Would appreciate your feedback!

RockteMQ-AI (commented):

### Summary
Re-review after rebase. PR adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST`, `DELETE_SUBSCRIPTIONGROUP_LIST`) following the existing `*_LIST` convention. The rebase preserves the previous design: body-driven batch requests, per-item authorization contexts, single-persist batching, and structured partial-failure responses.

### Findings

- **[Info]** Authorization — Body-driven batch APIs correctly emit one `DELETE` context per resource in `DefaultAuthorizationContextBuilder`, preventing the annotation-based path from producing an empty context list. Verified for both topic and subscription group batch codes.
- **[Info]** Batch persist — `TopicConfigManager#deleteTopicConfigList` and `SubscriptionGroupManager#deleteSubscriptionGroupConfigList` aggregate into a single `persist()` call per batch, avoiding N writes. Good.
- **[Info]** Dedup — `LinkedHashSet` preserves insertion order while deduplicating input. Consistent with existing `*_LIST` patterns.
- **[Info]** Refactoring — `collectPopRetryTopics()` extraction is clean and shared between single and batch delete paths, reducing duplication.
- **[Info]** Client null-check — `MQClientAPIImpl` methods use explicit `response == null` check (instead of `assert`) and throw `MQClientException(SYSTEM_ERROR, ...)`. Deterministic behavior.
- **[Info]** Request codes — `5002` / `5003` are additive; older brokers return `getUnknownCmdResponse` for unknown codes. No backward compatibility risk.
- **[Warning]** `AdminBrokerProcessor#deleteTopicList` — The `RemotingCommand response` is created once at the top. On the success path, `response.setCode(ResponseCode.SUCCESS)` and `response.setRemark(null)` are set explicitly. On the error path, a `DeleteTopicListResponseBody` is serialized into the response body. This is correct, but verify that no intermediate code path mutates `response` fields between the two branches (e.g., a logging interceptor that calls `response.setRemark()`).
- **[Warning]** No max batch size — Neither the server-side handler nor the client method enforces a maximum batch size. A caller could send thousands of topics in a single RPC, causing a long `synchronized` hold on the broker. Consider adding a server-side guard (e.g., reject if `topicList.size() > 1000`) to prevent accidental or malicious abuse.

### Suggestions
- Consider adding a `MAX_BATCH_SIZE` constant (e.g., 1000) and rejecting oversized requests early in `deleteTopicList()` / `deleteSubscriptionGroupList()` with a clear error message.
- The `DeleteSubscriptionGroupListRequestBody` has a 2-arg constructor `(groupNameList, cleanOffset)` and a 1-arg constructor `(groupNameList)`. The 1-arg constructor could delegate to the 2-arg one (`this(groupNameList, false)`) to reduce duplication.

### Verdict
Rebase looks clean — previous findings still hold. No new issues introduced. The two warnings above are non-blocking suggestions for future hardening.

---

RockteMQ-AI (commented):

### Summary
Re-review after new commit (2026-06-26). The new commit adds rate limiting to `deleteTopicList` via a configurable `batchDeleteTopicMaxRate` in `BrokerConfig`, addressing the previous "no max batch size" warning.

### Findings

- **[Info]** Rate limiting added — `BrokerConfig#batchDeleteTopicMaxRate` (default 10.0) + Guava `RateLimiter` in `deleteTopicList()`. This throttles per-topic deletion speed, partially addressing the previous batch-size concern. Good improvement.
- **[Warning]** Rate limiter only in `deleteTopicList`, not in `deleteSubscriptionGroupList` — The subscription group batch delete has no equivalent rate limiting. If the goal is to prevent long `synchronized` holds, both methods should be consistent. Consider adding the same pattern to `deleteSubscriptionGroupList`.
- **[Warning]** Rate limiter inside `synchronized` method — `deleteTopicList()` is `synchronized`. With rate limiting at 10/s, a batch of 1000 topics would hold the lock for ~100 seconds. This could block other admin operations. Consider:
  1. Moving rate-limited work outside the `synchronized` block (e.g., validate + prepare under lock, then rate-limited delete outside), or
  2. Adding a max batch size guard (e.g., reject if `topicList.size() > 1000`) in addition to rate limiting.
- **[Info]** `LiteMetadataUtil.isLiteMessageType()` and `getLiteLifecycleManager().cleanByParentTopic()` are called inside the rate-limited loop. Verify these are lightweight operations; if they involve RPC or heavy I/O, the rate limiter may not be sufficient to prevent long holds.
- **[Info]** Previous findings (authorization, batch persist, dedup, refactoring, client null-check, request codes) remain valid and unchanged.

### Suggestions
1. Add the same rate limiting pattern to `deleteSubscriptionGroupList` for consistency.
2. Consider adding a `MAX_BATCH_SIZE` constant as a hard upper bound, with rate limiting as a secondary throttle.
3. Document `batchDeleteTopicMaxRate` in the config description (e.g., "Max topics deleted per second in batch delete API. 0 means no limit.").

### Verdict
New rate limiting is a positive step. The two warnings above (inconsistency with subscription group path, synchronized + rate limiter interaction) are non-blocking but worth addressing before merge.

---

RockteMQ-AI (commented):
## Review by github-manager-bot (Re-review #2)

### Summary
Re-review after new commit `a258b0e` (2026-06-26). This PR adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST`, `DELETE_SUBSCRIPTIONGROUP_LIST`) to `AdminBrokerProcessor`, along with auth context integration, rate limiting, and client helpers.

### Findings

- **[Critical]** `AdminBrokerProcessor.java` — `deleteTopicList()` does **not** call `messageStore.deleteTopics(topicsToClean)`. The single-item `deleteTopic()` calls `this.brokerController.getMessageStore().deleteTopics(topicsToClean)` to clean CommitLog/ConsumeQueue data, but the batch path only calls `deleteTopicInBroker(topic)` per topic (which deletes config + queue mapping, **not** message store data). This will orphan message store data on disk. The batch path must include the `messageStore.deleteTopics()` call before or after the per-topic loop.

- **[Warning]** `TopicConfigManager.java:629` / `SubscriptionGroupManager.java:399` — `deleteTopicConfigList()` and `deleteSubscriptionGroupConfigList()` are defined but **never called** anywhere in this PR. `deleteTopicList()` calls `deleteTopicInBroker()` → `deleteTopicConfig()` one-by-one. Either these batch methods should be used (to aggregate `persist()` as the PR description claims), or they should be removed to avoid dead code.

- **[Warning]** `AdminBrokerProcessor.java` — The PR description states *"aggregate `persist()` once per batch"* and *"one-shot persist"*, but the current implementation calls `deleteTopicInBroker(topic)` per topic, which internally calls `deleteTopicConfig(topic)` — each of which likely triggers its own `persist()`. If batch persist is a goal, the `deleteTopicConfigList()` method should override the loop to persist once at the end, and `deleteTopicList()` should call it.

- **[Info]** `DefaultAuthorizationContextBuilder.java:324-362` — Auth context integration looks correct. Body-driven batch requests properly emit one `DELETE` context per topic/group, with blank entries filtered. Good that the annotation-based path gap was identified and handled explicitly.

- **[Info]** `AdminBrokerProcessor.java:815-820` — `collectPopRetryTopics()` refactoring is clean and correctly shared between single and batch paths. The `new ArrayList<>(topicsToClean)` snapshot before mutating the set in the batch path is a good defensive copy.

- **[Info]** `BrokerConfig.java:382` — `batchDeleteTopicMaxRate` default of `10.0` permits/sec is conservative and sensible. The `maxRate > 0` guard to disable the rate limiter is clean.

- **[Info]** `RequestCode.java:314-315` — Codes `5002`/`5003` are sequential after `SWITCH_TIMER_ENGINE = 5001`. No conflicts detected.

### Suggestions

1. **Must fix**: Add `messageStore.deleteTopics()` call in `deleteTopicList()`. Suggested placement — before the per-topic `deleteTopicInBroker` loop, matching the single-item flow:
```java
// After collecting topicsToClean (including pop retry topics):
try {
    if (this.brokerController.getMessageStore() != null) {
        this.brokerController.getMessageStore().deleteTopics(topicsToClean);
    }
} catch (Throwable t) {
    return buildErrorResponse(ResponseCode.SYSTEM_ERROR, t.getMessage());
}
// Then the existing rate-limited deleteTopicInBroker loop
```

2. **Should fix**: Either wire `deleteTopicConfigList()` / `deleteSubscriptionGroupConfigList()` into the batch path (with single-persist optimization), or remove them.

3. **Nit**: `deleteSubscriptionGroupList()` reuses `batchDeleteTopicMaxRate` for group deletion rate limiting. Consider renaming to `batchDeleteMaxRate` for clarity, or adding a separate config for group deletion.

---

RockteMQ-AI (commented):
## Review by github-manager-bot (Re-review #3)

### Summary
Re-review after new commit `1038b335` (2026-06-26). This PR adds batch deletion APIs (`DELETE_TOPIC_IN_BROKER_LIST`, `DELETE_SUBSCRIPTIONGROUP_LIST`) to `AdminBrokerProcessor`, with auth context integration, rate limiting, and client helpers.

### Changes Since Last Review
The latest commit appears to be a rebase/squash. The overall structure remains the same as the previous review.

### Findings

- **[Critical]** `AdminBrokerProcessor.java` — `deleteTopicList()` does **not** call `this.messageStore.deleteTopics(topicsToClean)`. The single-item `deleteTopic()` calls it after the per-topic loop to clean up actual message data in the store. Without this call, batch topic deletion removes broker-side metadata (topic config, queue mapping, etc.) but leaves orphaned message data on disk. This is the same issue flagged in the previous two reviews and remains unresolved.

  **Suggested fix** — add the store cleanup after the loop in `deleteTopicList()`:
  ```java
  for (String topic : topicsToClean) {
      if (rateLimiter != null) { rateLimiter.acquire(); }
      if (LiteMetadataUtil.isLiteMessageType(topic, brokerController)) {
          brokerController.getLiteLifecycleManager().cleanByParentTopic(topic);
      }
      deleteTopicInBroker(topic);
  }
  // ── add this ──
  this.messageStore.deleteTopics(topicsToClean);
  ```

- **[Warning]** `AdminBrokerProcessor.java:deleteTopicList()` — The method is `synchronized`, same as `deleteTopic()`. For a batch operation that may delete many topics with rate-limiting delays, holding the broker-wide lock for the entire duration could block other admin operations. Consider whether the critical section can be narrowed (e.g., validate first under the lock, then delete outside it).

- **[Info]** `SubscriptionGroupManager.deleteSubscriptionGroupConfigList()` and `TopicConfigManager.deleteTopicConfigList()` — These new batch methods simply loop over the single-item methods. They are not currently called by `deleteTopicList()` / `deleteSubscriptionGroupList()` (which call the single-item methods directly via `deleteTopicInBroker()` / `deleteSubscriptionGroupInBroker()`). If these batch methods are not used elsewhere, consider removing them to avoid dead code. If they are intended for future use, a brief Javadoc explaining the intended caller would help.

- **[Info]** `RequestCode.java` — New codes `5002` / `5003` are registered. No conflicts detected with existing codes. Good.

### What Looks Good
- Clean extraction of `collectPopRetryTopics()` from the single-item path — reduces duplication.
- `deleteSubscriptionGroupInBroker()` correctly checks `isLiteGroupType` **before** deleting the group config (avoids auto-recreation when `autoCreateSubscriptionGroup=true`).
- Auth context builder correctly emits per-resource DELETE actions for the batch request codes.
- Rate limiting via Guava `RateLimiter` with configurable max rate is a good safety measure.
- Comprehensive test coverage: empty/blank/system-topic validation, dedup, rate limiting, pop-retry-topic collection.

---

fuyou001 (approved):

PR #10461, opened 2026-06-10
[Fix #10460]: Pre-size ProxyContext map to avoid HashMap resize on request path

### LINKED ISSUE #10460: [Enhancement] Set initial capacity for ProxyContext internal HashMap to avoid resize

### Summary

`ProxyContext` (in `org.apache.rocketmq.proxy.common`) uses a `HashMap` with the default initial capacity (16) and default load factor (0.75) to hold context entries. The resize threshold is therefore `12`.

In typical request paths, the context is populated with the 10 standard keys defined in `ContextVariable` (`remote-address`, `local-address`, `client-id`, `channel`, `language`, `client-version`, `remaining-ms`, `action`, `protocol-type`, `namespace`) plus additional entries injected via `withVal(...)` from various processors / interceptors. As soon as the entry count exceeds 12, the underlying `HashMap` triggers a rehash/resize on the hot request path, which is unnecessary overhead given that the upper bound of entries is well-known.

### Motivation

`ProxyContext` is created on every inbound request (gRPC / Remoting), so this resize cost is paid per RPC. Pre-sizing the map eliminates a guaranteed reallocation + rehash on a hot path with effectively zero downside (the extra memory footprint is negligible, ~256 bytes per context).

Initialize the internal map with an explicit initial capacity (suggest `64`) so that under the default load factor the threshold (`48`) comfortably exceeds the realistic upper bound of context entries, and no resize will happen on the request path.

```java
// proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
private static final int DEFAULT_INITIAL_CAPACITY = 64;
private final Map<String, Object> value = new HashMap<>(DEFAULT_INITIAL_CAPACITY);
```

- Keep using the default capacity — rejected, incurs a deterministic resize on every request once entries cross 12.
- Use a smaller capacity such as 32 — works for most cases but still risks resize when interceptors/plugins push more values; 64 leaves comfortable headroom while staying small.
- Switch to a different map implementation (e.g. `IdentityHashMap`, array-backed) — out of scope; preserving `HashMap` semantics with a tuned initial capacity is the minimal change.

### Additional Context

`proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java`

Related code:

- `ProxyContext#value` field — the map being resized.
- `ContextVariable` — defines 10 well-known keys, all of which are usually set during a single request lifecycle.

Happy to send a PR.

HarshMehta112:

  - Fixes #10460 

  ### Brief Description

  `ProxyContext` is created on every gRPC/Remoting request. Its `value` map used the default `HashMap` capacity (16, threshold 12), so adding the 10 `ContextVariable` keys plus interceptor entries triggers a 
  resize on the hot path. Pre-size the map to avoid it:

  ```java
  private static final int DEFAULT_INITIAL_CAPACITY = 64;
  private final Map<String, Object> value = new HashMap<>(DEFAULT_INITIAL_CAPACITY);
  ```
  No behavior or API change.

  How Did You Test This Change?

  Pure capacity refactor, no logic change. Existing proxy module tests pass.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
@@ -23,7 +23,7 @@ import java.util.Map;
 
 public class ProxyContext {
     public static final String INNER_ACTION_PREFIX = "Inner";
-    private final Map<String, Object> value = new HashMap<>();
+    private final Map<String, Object> value = new HashMap<>(32);
 
     public static ProxyContext create() {
         return new ProxyContext();
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
@@ -23,7 +23,7 @@ import java.util.Map;
 
 public class ProxyContext {
     public static final String INNER_ACTION_PREFIX = "Inner";
-    private static final int DEFAULT_INITIAL_CAPACITY = 32;
+    private static final int DEFAULT_INITIAL_CAPACITY = 64;
     private final Map<String, Object> value = new HashMap<>(DEFAULT_INITIAL_CAPACITY);
 
     public static ProxyContext create() {
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
@@ -23,7 +23,8 @@ import java.util.Map;
 
 public class ProxyContext {
     public static final String INNER_ACTION_PREFIX = "Inner";
-    private final Map<String, Object> value = new HashMap<>(32);
+    private static final int DEFAULT_INITIAL_CAPACITY = 32;
+    private final Map<String, Object> value = new HashMap<>(DEFAULT_INITIAL_CAPACITY);
 
     public static ProxyContext create() {
         return new ProxyContext();


oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #10461 — [Fix #10460]: Pre-size ProxyContext map to avoid HashMap resize on request path
**Type:** Performance optimization (1 file, +2/-1)

### Assessment
Valid performance fix. `ProxyContext` is created per gRPC/Remoting request, and its `value` map previously used the default `HashMap` capacity (16, threshold 12). Adding ~10 `ContextVariable` keys plus interceptor entries would trigger a resize on the hot path.

### Changes
- Pre-sized `HashMap` to 64 (threshold 48), well above the expected key count
- Minimal, focused change with no API impact

### Verdict
✅ Clean, well-targeted optimization. Fixes #10460.

---

proxy/src/main/java/org/apache/rocketmq/proxy/common/ProxyContext.java
```diff
@@ -23,7 +23,8 @@
 
 public class ProxyContext {
     public static final String INNER_ACTION_PREFIX = "Inner";
-    private final Map<String, Object> value = new HashMap<>();
+    private static final int DEFAULT_INITIAL_CAPACITY = 64;
+    private final Map<String, Object> value = new HashMap<>(DEFAULT_INITIAL_CAPACITY);
```
RongtongJin, line 27: The change looks behaviorally safe, but the rationale should be tightened. Current built-in initialization does not always cross the default HashMap resize threshold: gRPC usually sets about 8 entries, or 9 with a deadline, and Remoting usually sets about 5-8 depending on available channel attributes. Resize mainly happens if later pipeline/interceptor/plugin entries add more values. Consider either using a smaller initial capacity such as 32, or keeping 64 while updating the PR description to frame this as headroom for extended contexts rather than a deterministic resize on every request.

RongtongJin (commented):
One inline note on the capacity choice and rationale.

RongtongJin (approved):

PR #10469, opened 2026-06-10
[ISSUE #10467] Reduce per-message allocation in createUniqID and valueOfMagicCode

### LINKED ISSUE #10467: [Enhancement] Reduce per-message allocation in createUniqID and valueOfMagicCode

### Summary

Reduce per-message allocation in `createUniqID` and `valueOfMagicCode`.

### Motivation

JFR profiling on the producer send path and broker dispatch path reveals two unnecessary per-message allocations:

1. **`MessageClientIDSetter.createUniqID()`** — allocates a `new char[LEN * 2]` on every send. Since the char array is only used to build a msgId string and is immediately discarded, it can be reused via `ThreadLocal`. JFR shows ~405 events/60s from this site.

2. **`MessageVersion.valueOfMagicCode(int)`** — calls `MessageVersion.values()` which copies the enum values array on every invocation (JDK spec), then performs an O(n) linear scan. Since there are only two known magic codes, a direct if-else is both faster and avoids the array copy.

- `createUniqID()`: replace `new char[LEN * 2]` with `ThreadLocal<char[]>` initialized once per thread.
- `valueOfMagicCode()`: replace `values()` loop with direct if-else matching the two known magic code constants.

- Use `StringBuilder` pool instead of `ThreadLocal<char[]>` — more overhead for a simple char array.
- Cache the full msgId string — not possible since it includes timestamp and sequence number.

### Additional Context

JFR shows ~405 char[] allocation events per 60s from `createUniqID`.

wang-jiahua:

Fixes #10467

### Brief Description

Two independent per-message allocation reductions:

1. **`MessageClientIDSetter.createUniqID()`**: replaces `new char[LEN * 2]` per call with a `ThreadLocal<char[]>` that is allocated once per thread and reused. The char array is only used to build the msgId string within the method and does not escape. JFR shows ~405 `char[]` allocation events/60s from this site eliminated.

2. **`MessageVersion.valueOfMagicCode(int)`**: replaces `Enum.values()` (which copies the backing array per JDK spec) + O(n) loop with direct `if-else` matching on the two known magic code constants (`MESSAGE_MAGIC_CODE` and `MESSAGE_MAGIC_CODE_V2`). Eliminates the per-call array copy on the broker dispatch path.

#### Compatibility

- `createUniqID()` produces the same output format — only the internal char buffer allocation strategy changes.
- `valueOfMagicCode()` behavior is identical for all valid inputs. Invalid magic codes still throw `IllegalArgumentException`.

```bash
mvn -pl common -Dcheckstyle.skip=false -Dspotbugs.skip=true validate
# 0 Checkstyle violations

mvn -pl common -Dcheckstyle.skip -Dspotbugs.skip=true -Djacoco.skip=true test
# All tests pass
```

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/MessageClientIDSetter.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/MessageClientIDSetter.java
@@ -111,9 +111,11 @@ public class MessageClientIDSetter {
         return value & 0x0000FFFF;
     }
 
+    private static final ThreadLocal<char[]> UNIQ_ID_BUF = ThreadLocal.withInitial(() -> new char[LEN * 2]);
+
     public static String createUniqID() {
-        char[] sb = new char[LEN * 2];
-        System.arraycopy(FIX_STRING, 0, sb, 0, FIX_STRING.length);
+        char[] buf = UNIQ_ID_BUF.get();
+        System.arraycopy(FIX_STRING, 0, buf, 0, FIX_STRING.length);
         long current = System.currentTimeMillis();
         if (current >= nextStartTime) {
             setStartTime(current);
@@ -124,10 +126,10 @@ public class MessageClientIDSetter {
             diff = 0;
         }
         int pos = FIX_STRING.length;
-        UtilAll.writeInt(sb, pos, diff);
+        UtilAll.writeInt(buf, pos, diff);
         pos += 8;
-        UtilAll.writeShort(sb, pos, COUNTER.getAndIncrement());
-        return new String(sb);
+        UtilAll.writeShort(buf, pos, COUNTER.getAndIncrement());
+        return new String(buf);
     }
 
     public static void setUniqID(final Message msg) {
--- a/common/src/main/java/org/apache/rocketmq/common/message/MessageVersion.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/MessageVersion.java
@@ -71,13 +71,14 @@ public enum MessageVersion {
     }
 
     public static MessageVersion valueOfMagicCode(int magicCode) {
-        for (MessageVersion version : MessageVersion.values()) {
-            if (version.getMagicCode() == magicCode) {
-                return version;
-            }
+        switch (magicCode) {
+            case MessageDecoder.MESSAGE_MAGIC_CODE:
+                return MESSAGE_VERSION_V1;
+            case MessageDecoder.MESSAGE_MAGIC_CODE_V2:
+                return MESSAGE_VERSION_V2;
+            default:
+                throw new IllegalArgumentException("Invalid magicCode " + magicCode);
         }
-
-        throw new IllegalArgumentException("Invalid magicCode " + magicCode);
     }
 
     public int getMagicCode() {
--- a/common/src/test/java/org/apache/rocketmq/common/message/MessageVersionTest.java
+++ b/common/src/test/java/org/apache/rocketmq/common/message/MessageVersionTest.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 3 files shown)


common/src/main/java/org/apache/rocketmq/common/message/MessageVersion.java
```diff
@@ -71,12 +71,12 @@ public void putTopicLength(ByteBuffer buffer, int topicLength) {
...
-                return version;
-            }
+        if (magicCode == MessageDecoder.MESSAGE_MAGIC_CODE) {
+            return MESSAGE_VERSION_V1;
+        }
+        if (magicCode == MessageDecoder.MESSAGE_MAGIC_CODE_V2) {
+            return MESSAGE_VERSION_V2;
         }
-
         throw new IllegalArgumentException("Invalid magicCode " + magicCode);
```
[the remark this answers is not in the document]
wang-jiahua: Intentional trade-off. A static lookup map built from `values()` would achieve the same result but adds complexity for exactly two known constants. The if-else is trivially verifiable and a new version would just add one more line. The `values()` call this replaces was the problem — it copies the enum backing array on every invocation (JDK spec), which this fix eliminates.

common/src/main/java/org/apache/rocketmq/common/message/MessageClientIDSetter.java
```diff
@@ -111,8 +111,10 @@ public static int getPidFromID(String msgID) {
         return value & 0x0000FFFF;
     }
 
+    private static final ThreadLocal<char[]> UNIQ_ID_BUF = ThreadLocal.withInitial(() -> new char[LEN * 2]);
+
     public static String createUniqID() {
-        char[] sb = new char[LEN * 2];
+        char[] sb = UNIQ_ID_BUF.get();
```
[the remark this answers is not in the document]
wang-jiahua: The retained buffer is `LEN * 2` bytes (~32–64 bytes depending on IP version) per thread — negligible even with hundreds of threads. RocketMQ producer threads are long-lived pool threads, so the ThreadLocal lifecycle aligns with thread lifecycle. Added a brief inline comment noting the per-thread retention.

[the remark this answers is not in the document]
wang-jiahua: Good point — renamed `sb` to `buf` for clarity.

oss-sentinel-ai (approved):
## Review: Approved ✅

**PR:** #10469 — Reduce per-message allocation in createUniqID and valueOfMagicCode
**Type:** Performance (2 files, +12/-10)

### Assessment
Optimizes hot path allocations:
- ThreadLocal<char[]> for createUniqID() (eliminates ~405 char[] allocations/60s)
- Direct if-else for valueOfMagicCode() (eliminates Enum.values() array copy)

### Verdict
✅ Solid performance optimization based on JFR profiling. Fixes #10467.

---

common/src/main/java/org/apache/rocketmq/common/message/MessageVersion.java
```diff
@@ -71,12 +71,12 @@ public void putTopicLength(ByteBuffer buffer, int topicLength) {
     }
 
     public static MessageVersion valueOfMagicCode(int magicCode) {
-        for (MessageVersion version : MessageVersion.values()) {
-            if (version.getMagicCode() == magicCode) {
-                return version;
-            }
+        if (magicCode == MessageDecoder.MESSAGE_MAGIC_CODE) {
```
fuyou001: add unit test  magicCode 
 @Test
  public void testValueOfMagicCode() {
      for (MessageVersion version : MessageVersion.values()) {
          assertThat(MessageVersion.valueOfMagicCode(version.getMagicCode()))
              .isEqualTo(version);
      }
  }

fuyou001: replace if with switch

wang-jiahua: Done — switched to `switch` statement. Reads slightly cleaner with the `default: throw` pattern.

wang-jiahua: Good suggestion! Added `MessageVersionTest` with `testValueOfMagicCode()` that iterates all versions and asserts round-trip through `valueOfMagicCode(getMagicCode())`. Also added a negative test for invalid magic codes.

fuyou001 (approved):

RongtongJin (approved):

RongtongJin (approved):

PR #10516, opened 2026-06-15
[ISSUE #10510] Add gRPC server permit keepalive configuration

### LINKED ISSUE #10510: [Enhancement] Make gRPC server keepalive parameters configurable in ProxyConfig
## Motivation

When a TCP connection is broken without a RST (e.g., intermediate network device silently drops packets, process killed with `kill -9`), the gRPC client needs to rely on HTTP/2 PING (keepalive) to detect the dead connection. Until the keepalive mechanism detects the failure, **all requests on that connection will fail** and have to wait until their deadline expires.

Currently, the gRPC server in the Proxy does **not** configure `permitKeepAliveTime` or `permitKeepAliveWithoutCalls`, which means gRPC Netty server defaults apply:

- `permitKeepAliveTime` = **5 minutes** — clients cannot send keepalive pings more frequently than every 5 minutes, or the server will send a GOAWAY
- `permitKeepAliveWithoutCalls` = **false** — keepalive pings on idle connections (no active RPCs) are rejected by the server

This causes two problems:

1. **Slow dead-connection detection**: The `rocketmq-clients` Java SDK sets `keepAliveTime = 300s` (5 min), so worst-case detection time is **5 min + 30s = 5.5 minutes**. During this window, all sends to the affected endpoint fail.
2. **Idle connection keepalive ineffective**: Although `rocketmq-clients` sets `keepAliveWithoutCalls(true)`, the server's default `permitKeepAliveWithoutCalls = false` silently rejects these pings, making idle connection health detection impossible.

## Proposed Changes

Add the following configurable parameters to `ProxyConfig`:

| Parameter | Default Value | Description |
|-----------|---------------|-------------|
| `grpcServerPermitKeepAliveTimeMillis` | `10000` (10s) | Minimum time a client should wait before sending each keepalive ping |
| `grpcServerPermitKeepAliveWithoutCalls` | `true` | Whether to allow keepalive pings when there are no outstanding RPCs |

Apply these in `GrpcServerBuilder`:

```java
serverBuilder
    .permitKeepAliveTime(config.getGrpcServerPermitKeepAliveTimeMillis(), TimeUnit.MILLISECONDS)
    .permitKeepAliveWithoutCalls(config.isGrpcServerPermitKeepAliveWithoutCalls());
```

### Impact

- **Overhead is minimal**: Each keepalive PING frame is only ~70 bytes (including TCP/IP headers). At 30s interval per connection, this adds ~8KB/hour per connection.
- **Backward compatible**: Default values are more permissive than the current implicit defaults, but won't break existing clients. Clients with longer keepalive intervals (e.g., current 300s) will continue to work fine.
- Once the server permits shorter keepalive intervals, client-side improvements (reducing `keepAliveTime` from 300s to 30s) can bring dead-connection detection time from **5.5 minutes down to ~40 seconds**.

### Related Code

- `GrpcServerBuilder`: `proxy/src/main/java/org/apache/rocketmq/proxy/grpc/GrpcServerBuilder.java`
- `ProxyConfig`: `proxy/src/main/java/org/apache/rocketmq/proxy/config/ProxyConfig.java`
- Client keepalive settings: `keepAliveTime=300s, keepAliveTimeout=30s, keepAliveWithoutCalls=true` in rocketmq-clients Java SDK (`RpcClientImpl.java`)

RockteMQ-AI:
## Motivation

When a TCP connection is broken without a RST, the gRPC client needs to rely on HTTP/2 PING (keepalive) to detect the dead connection. Currently, the gRPC server in the Proxy does **not** configure `permitKeepAliveTime` or `permitKeepAliveWithoutCalls`, which means gRPC Netty server defaults apply:

- `permitKeepAliveTime` = **5 minutes** — clients cannot send keepalive pings more frequently than every 5 minutes
- `permitKeepAliveWithoutCalls` = **false** — keepalive pings on idle connections are rejected

This causes slow dead-connection detection (up to 5.5 minutes) and ineffective idle connection health checks.

Fixes #10510

## Changes

Add two configurable parameters to `ProxyConfig`:

| Parameter | Default | Description |
|-----------|---------|-------------|
| `grpcServerPermitKeepAliveTimeMillis` | 10000 (10s) | Minimum time between client keepalive pings |
| `grpcServerPermitKeepAliveWithoutCalls` | true | Allow keepalive pings when no active RPCs |

Apply these in `GrpcServerBuilder`:
```java
serverBuilder
    .permitKeepAliveTime(config.getGrpcServerPermitKeepAliveTimeMillis(), TimeUnit.MILLISECONDS)
    .permitKeepAliveWithoutCalls(config.isGrpcServerPermitKeepAliveWithoutCalls());
```

## Impact

- **Backward compatible**: Default values are more permissive than current implicit defaults
- **Minimal overhead**: ~8KB/hour per connection at 30s interval
- **Enables faster detection**: Once clients reduce keepAliveTime, dead-connection detection can drop from 5.5 minutes to ~40 seconds

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/config/ProxyConfig.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/config/ProxyConfig.java
@@ -156,6 +156,8 @@ public class ProxyConfig implements ConfigFile {
     private long grpcClientConsumerMaxLongPollingTimeoutMillis = Duration.ofSeconds(20).toMillis();
     private int grpcClientConsumerLongPollingBatchSize = 32;
     private long grpcClientIdleTimeMills = Duration.ofSeconds(120).toMillis();
+    private long grpcServerPermitKeepAliveTimeMillis = 10000;
+    private boolean grpcServerPermitKeepAliveWithoutCalls = true;
 
     private int channelExpiredInSeconds = 60;
     private int contextExpiredInSeconds = 30;
@@ -1219,6 +1221,22 @@ public class ProxyConfig implements ConfigFile {
 
     public void setGrpcClientIdleTimeMills(final long grpcClientIdleTimeMills) {
         this.grpcClientIdleTimeMills = grpcClientIdleTimeMills;
+    }
+
+    public long getGrpcServerPermitKeepAliveTimeMillis() {
+        return grpcServerPermitKeepAliveTimeMillis;
+    }
+
+    public void setGrpcServerPermitKeepAliveTimeMillis(long grpcServerPermitKeepAliveTimeMillis) {
+        this.grpcServerPermitKeepAliveTimeMillis = grpcServerPermitKeepAliveTimeMillis;
+    }
+
+    public boolean isGrpcServerPermitKeepAliveWithoutCalls() {
+        return grpcServerPermitKeepAliveWithoutCalls;
+    }
+
+    public void setGrpcServerPermitKeepAliveWithoutCalls(boolean grpcServerPermitKeepAliveWithoutCalls) {
+        this.grpcServerPermitKeepAliveWithoutCalls = grpcServerPermitKeepAliveWithoutCalls;
     }
 
     public String getRegionId() {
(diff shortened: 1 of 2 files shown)


qianye1001:
grpcServerPermitKeepAliveTimeMillis default value set 30s

qianye1001:
@RockteMQ-AI grpcServerPermitKeepAliveTimeMillis default value set 30s

RockteMQ-AI:
Done. The default value of `grpcServerPermitKeepAliveTimeMillis` has been updated to `30000` (30s) in commit cbf8018.

RockteMQ-AI:
Thanks for the feedback! The default value of `grpcServerPermitKeepAliveTimeMillis` has been updated to `30000` (30 seconds) in commit cbf8018.

---
*Automated reply by github-manager-bot*

RockteMQ-AI:
Hi @qianye1001, thanks for the suggestion! The default value of `grpcServerPermitKeepAliveTimeMillis` has been updated to `30000` (30 seconds) in commit cbf80183d. Please let us know if any further adjustments are needed.

lizhimins (approved):

PR #10519, opened 2026-06-16
[ISSUE #10489] Support BatchChangeInvisibleTime to reduce network round-trips and RocksDB flushes

### LINKED ISSUE #10489: [Enhancement] Support BatchChangeInvisibleTime to reduce network round-trips and RocksDB flushes

### Summary

Add `BatchChangeInvisibleTime` support in the Proxy and Broker, following the existing `BatchAck` pattern. Currently, when the Proxy needs to change invisible time for multiple messages (client disconnect, write-back failure, periodic renewal, message filtering), it sends individual `CHANGE_MESSAGE_INVISIBLETIME` requests to the Broker one by one. This causes excessive network round-trips and, on the PopConsumerKVService (RocksDB) path, excessive RocksDB flushes (2N `db.write()` calls for N messages).

### Motivation

Multiple scenarios in the Proxy require batch `changeInvisibleTime`:

1. **Client disconnect** (`DefaultReceiptHandleManager.returnHandleGroup`): When a client goes offline, the Proxy scans all unacked receipt handles for that client and calls `changeInvisibleTime` for each one individually. A consumer with many in-flight messages triggers N separate network round-trips.

2. **ReceiveMessage write-back failure** (`ReceiveMessageResponseStreamWriter`): When writing messages back to the gRPC client fails mid-stream, the Proxy nacks all remaining messages one by one.

3. **Periodic renewal** (`DefaultReceiptHandleManager.scheduleRenewTask`): The Proxy periodically renews receipt handles that are about to expire. Each handle is renewed individually via a separate `changeInvisibleTime` call.

4. **Pop message filtering** (`ConsumerProcessor.filterPopResult`): Messages filtered as `TO_RETURN` are changed invisible time individually.

On the Broker side with `PopConsumerKVServiceEnable=true`, each individual `changeInvisibleTime` call in `PopConsumerService.changeInvisibilityDuration()` does:
- `popConsumerStore.writeRecords(Collections.singletonList(ckRecord))` — 1 RocksDB WriteBatch with 1 entry
- `popConsumerStore.deleteRecords(Collections.singletonList(ackRecord))` — 1 RocksDB WriteBatch with 1 entry

For N messages, this results in 2N `db.write()` calls. Since `PopConsumerRocksdbStore` uses `sync=true` on WriteOptions, each write forces a WAL flush, making this a significant bottleneck.

Add a new `BATCH_CHANGE_MESSAGE_INVISIBLETIME` request code, following the existing `BatchAck` (`BATCH_ACK_MESSAGE`) pattern:

**Request format** — Each entry carries the same parameters as the single `ChangeInvisibleTimeRequestHeader`:
- `consumerGroup`, `topic`, `queueId`, `extraInfo`, `offset`, `invisibleTime`, `liteTopic`, `suspend`

**Response format** — Per-entry results with new receipt handle info (same as `ChangeInvisibleTimeResponseHeader`):
- `code` (SUCCESS or error), `popTime`, `invisibleTime`, `reviveQid`

**Layer-by-layer changes:**

1. **Remoting**: Add `RequestCode.BATCH_CHANGE_MESSAGE_INVISIBLETIME`, `BatchChangeInvisibleTimeRequestBody` (with `List<ChangeInvisibleTimeRequestEntry>`), `BatchChangeInvisibleTimeResponseBody` (with per-entry results)

2. **Broker** — True batch processing:
   - `PopConsumerService.batchChangeInvisibilityDuration()`: Collect all CK records and ack records, then call `writeRecords(allCkRecords)` once and `deleteRecords(allAckRecords)` once. This reduces N messages from 2N `db.write()` to just 2 `db.write()` calls.
   - `ChangeInvisibleTimeProcessor`: Handle `BATCH_CHANGE_MESSAGE_INVISIBLETIME` — on KV path, delegate to the batch method above; on traditional revive path, iterate and call `appendCheckPointThenAckOrigin()` per entry.
   - `BrokerController`: Register the new request code to `changeInvisibleTimeProcessor`.

3. **Client**: Add `batchChangeInvisibleTimeAsync()` to `MQClientAPIImpl` and `MQClientAPIExt`.

4. **Proxy Service**: Add `batchChangeInvisibleTime()` to `MessageService`, `ClusterMessageService`, and `LocalMessageService`.

5. **Proxy Processor**: Add `batchChangeInvisibleTime()` to `ConsumerProcessor` (group by broker, filter expired handles), `MessagingProcessor` interface, and `DefaultMessagingProcessor`.

6. **Caller integration**:
   - `DefaultReceiptHandleManager.returnHandleGroup()`: Collect all handles, batch send
   - `DefaultReceiptHandleManager.scheduleRenewTask()`: Collect handles needing renewal per group key, batch send, update each handle with per-entry result
   - `ReceiveMessageResponseStreamWriter`: Collect all nack messages, batch send
   - `ConsumerProcessor.filterPopResult()`: Collect all TO_RETURN messages, batch send

7. **Config**: Add `enableBatchChangeInvisibleTime` flag (default false) for gradual rollout.

1. **Proxy-level batching only (no Broker changes)**: Group entries by broker at the Proxy, but still send individual `CHANGE_MESSAGE_INVISIBLETIME` requests. This reduces Proxy-side processing overhead but not the number of network round-trips or RocksDB flushes. Rejected because the Broker-side RocksDB flush reduction is the main performance win.

2. **Reuse `BATCH_ACK_MESSAGE` with BitSet merging**: `BatchAck` merges multiple offsets sharing the same CK into a BitSet. This doesn't work for `ChangeInvisibleTime` because each entry requires its own new checkpoint write + old checkpoint ack, and each entry may have a different `invisibleTime`. A list-based request body is necessary.

3. **Async fire-and-forget (no per-entry response)**: For the `returnHandleGroup` scenario, per-entry results aren't strictly needed. However, the renewal scenario (`startRenewMessage`) requires per-entry new receipt handles to update the cached handle via `messageReceiptHandle.updateReceiptHandle()`. Supporting per-entry responses makes the batch API general-purpose across all scenarios.

### Additional Context

**RocksDB batch optimization detail:**

Current (N messages = 2N `db.write()`):
```java
// PopConsumerService.changeInvisibilityDuration() — called N times
this.popConsumerStore.writeRecords(Collections.singletonList(ckRecord));   // 1 WriteBatch, 1 entry
this.popConsumerStore.deleteRecords(Collections.singletonList(ackRecord)); // 1 WriteBatch, 1 entry
```

Proposed (N messages = 2 `db.write()`):
```java
// PopConsumerService.batchChangeInvisibilityDuration() — called once
this.popConsumerStore.writeRecords(allCkRecords);     // 1 WriteBatch, N entries
this.popConsumerStore.deleteRecords(allAckRecords);    // 1 WriteBatch, N entries
```

`PopConsumerRocksdbStore.writeRecords/deleteRecords` already accept `List<PopConsumerRecord>` and use `WriteBatch` internally, so the infrastructure for batch RocksDB writes is already in place — we just need to stop passing `Collections.singletonList()`.

qianye1001:
Fixes #10489

## Summary
- Add remoting/client/proxy support for BatchChangeInvisibleTime with a batch header scoped to one topic and consumer group for broker-side auth.
- Add broker batch handling for PopConsumer KV change invisible time, including buffer-aware cache handling and delete-before-put RocksDB batch semantics.
- Wire proxy batch call sites for receipt handle renew/cleanup, receive write-back failure nack, and filtered TO_RETURN messages, gated by `enableBatchChangeInvisibleTime`.

## Tests
- `mvn -pl common,remoting,client,broker,proxy -DskipTests compile test-compile`
- `mvn -pl common,remoting,client,broker,proxy -DfailIfNoTests=false -Dtest=BatchChangeInvisibleTimeTest,MQClientAPIImplTest#testBatchChangeInvisibleTimeAsyncSendsRequestHeader+testProcessBatchChangeInvisibleTimeResponse,ChangeInvisibleTimeProcessorTest#testProcessBatchRequestRejectsMismatchedTopicOrGroup+testProcessBatchKvRequestDoesNotBuildSingleHeader+testProcessBatchRequestConvertsBadEntryToPerEntryFailure+testProcessBatchRequestRejectsOversizedBody,PopConsumerCacheTest,PopConsumerRocksdbStoreTest,ConsumerProcessorTest#testBatchChangeInvisibleTimeSplitByRealTopic+testBatchChangeInvisibleTime+testBatchChangeInvisibleTimeSplitOversizedBrokerGroup,ReceiveMessageResponseStreamWriterTest#testWriteBackFailureShouldBatchChangeInvisibleTime,ReceiptHandleProcessorTest#testBatchRenewMessage+testBatchClearGroup test`
- `mvn -pl common,remoting,client,broker,proxy -DfailIfNoTests=false -Dtest=ReceiveMessageResponseStreamWriterTest,ReceiptHandleProcessorTest test`

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -497,18 +497,35 @@ public class PopConsumerService extends ServiceThread {
     public void changeInvisibilityDuration(long popTime, long invisibleTime, long changedPopTime,
                                            long changedInvisibleTime, String groupId, String topicId,
                                            int queueId, long offset, boolean suspend) {
-        ChangeInvisibleTimeRequestEntry changeRecord = new ChangeInvisibleTimeRequestEntry();
-        changeRecord.setConsumerGroup(groupId);
-        changeRecord.setTopic(topicId);
-        changeRecord.setQueueId(queueId);
-        changeRecord.setOffset(offset);
-        changeRecord.setInvisibleTime(changedInvisibleTime);
-        changeRecord.setPopTime(popTime);
-        changeRecord.setOldInvisibleTime(invisibleTime);
-        changeRecord.setChangedPopTime(changedPopTime);
-        changeRecord.setChangedInvisibleTime(changedInvisibleTime);
-        changeRecord.setSuspend(suspend);
-        batchChangeInvisibilityDuration(Collections.singletonList(changeRecord));
+        if (brokerConfig.isPopConsumerKVServiceLog()) {
+            log.info("PopConsumerService change, time={}, invisible={}, " +
+                    "groupId={}, topic={}, queueId={}, offset={}, new time={}, new invisible={}",
+                popTime, invisibleTime, groupId, topicId, queueId, offset, changedPopTime, changedInvisibleTime);
+        }
+
+        PopConsumerRecord ckRecord = new PopConsumerRecord(
+            changedPopTime, groupId, topicId, queueId, 0, changedInvisibleTime, offset, null, suspend);
+
+        PopConsumerRecord ackRecord = new PopConsumerRecord(
+            popTime, groupId, topicId, queueId, 0, invisibleTime, offset, null, suspend);
+
+        // No need to generate new records when the group does not exist,
+        // because these retry messages will not be consumed by anyone.
+        boolean skipWrite = brokerConfig.isPopReviveSkipIfGroupAbsent() &&
+            !brokerController.getSubscriptionGroupManager().containsSubscriptionGroup(groupId);
+
+        if (skipWrite) {
+            log.info("PopConsumerService change invisibility skip, time={}, " +
+                "groupId={}, topicId={}, queueId={}, offset={}", popTime, groupId, topicId, queueId, offset);
+        }
+
+        List<PopConsumerRecord> ckRecords = skipWrite ? Collections.emptyList() : Collections.singletonList(ckRecord);
+        List<PopConsumerRecord> ackRecords = Collections.singletonList(ackRecord);
+        if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
+            popConsumerCache.writeAndDeleteRecords(ckRecords, ackRecords);
+        } else {
+            this.popConsumerStore.writeAndDeleteRecords(ckRecords, ackRecords);
+        }
     }
 
     public void batchChangeInvisibilityDuration(List<ChangeInvisibleTimeRequestEntry> changeRecords) {
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -517,21 +517,14 @@ public class PopConsumerService extends ServiceThread {
         if (skipWrite) {
             log.info("PopConsumerService change invisibility skip, time={}, " +
                 "groupId={}, topicId={}, queueId={}, offset={}", popTime, groupId, topicId, queueId, offset);
-        } else {
-            this.popConsumerStore.writeRecords(Collections.singletonList(ckRecord));
         }
 
+        List<PopConsumerRecord> ckRecords = skipWrite ? Collections.emptyList() : Collections.singletonList(ckRecord);
+        List<PopConsumerRecord> ackRecords = Collections.singletonList(ackRecord);
         if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
-            if (popConsumerCache.deleteRecords(Collections.singletonList(ackRecord)).isEmpty()) {
-                return;
-            }
-        }
-
-        // If the new CK has the same key as the old CK (same visibilityTimeout),
-        // the write already overwrites the old record in RocksDB, skip delete
-        // to avoid removing the newly written record.
-        if (skipWrite || ckRecord.getVisibilityTimeout() != ackRecord.getVisibilityTimeout()) {
-            this.popConsumerStore.deleteRecords(Collections.singletonList(ackRecord));
+            popConsumerCache.writeAndDeleteRecords(ckRecords, ackRecords);
+        } else {
+            this.popConsumerStore.writeAndDeleteRecords(ckRecords, ackRecords);
         }
     }
 
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -535,6 +535,120 @@ public class PopConsumerService extends ServiceThread {
         }
     }
 
+    public void batchChangeInvisibilityDuration(List<ChangeInvisibilityRecord> changeRecords) {
+        if (changeRecords == null || changeRecords.isEmpty()) {
+            return;
+        }
+
+        List<PopConsumerRecord> ckRecords = new ArrayList<>(changeRecords.size());
+        List<PopConsumerRecord> ackRecords = new ArrayList<>(changeRecords.size());
+        List<PopConsumerRecord> storeAckRecords = new ArrayList<>(changeRecords.size());
+
+        for (ChangeInvisibilityRecord changeRecord : changeRecords) {
+            if (brokerConfig.isPopConsumerKVServiceLog()) {
+                log.info("PopConsumerService batch change, time={}, invisible={}, " +
+                        "groupId={}, topic={}, queueId={}, offset={}, new time={}, new invisible={}",
+                    changeRecord.getPopTime(), changeRecord.getInvisibleTime(), changeRecord.getGroupId(),
+                    changeRecord.getTopicId(), changeRecord.getQueueId(), changeRecord.getOffset(),
+                    changeRecord.getChangedPopTime(), changeRecord.getChangedInvisibleTime());
+            }
+
+            PopConsumerRecord ckRecord = new PopConsumerRecord(
+                changeRecord.getChangedPopTime(), changeRecord.getGroupId(), changeRecord.getTopicId(),
+                changeRecord.getQueueId(), 0, changeRecord.getChangedInvisibleTime(), changeRecord.getOffset(),
+                null, changeRecord.isSuspend());
+
+            PopConsumerRecord ackRecord = new PopConsumerRecord(
+                changeRecord.getPopTime(), changeRecord.getGroupId(), changeRecord.getTopicId(),
+                changeRecord.getQueueId(), 0, changeRecord.getInvisibleTime(), changeRecord.getOffset(),
+                null, changeRecord.isSuspend());
+
+            boolean skipWrite = brokerConfig.isPopReviveSkipIfGroupAbsent() &&
+                !brokerController.getSubscriptionGroupManager().containsSubscriptionGroup(changeRecord.getGroupId());
+
+            if (skipWrite) {
+                log.info("PopConsumerService batch change invisibility skip, time={}, " +
+                    "groupId={}, topicId={}, queueId={}, offset={}", changeRecord.getPopTime(),
+                    changeRecord.getGroupId(), changeRecord.getTopicId(), changeRecord.getQueueId(),
+                    changeRecord.getOffset());
+            } else {
+                ckRecords.add(ckRecord);
+            }
+
+            ackRecords.add(ackRecord);
+            if (skipWrite || ckRecord.getVisibilityTimeout() != ackRecord.getVisibilityTimeout()) {
+                storeAckRecords.add(ackRecord);
+            }
+        }
+
+        if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
+            popConsumerCache.writeAndDeleteRecords(ckRecords, ackRecords);
+        } else {
+            this.popConsumerStore.writeAndDeleteRecords(ckRecords, storeAckRecords);
+        }
+    }
+
+    public static class ChangeInvisibilityRecord {
+        private final long popTime;
+        private final long invisibleTime;
+        private final long changedPopTime;
+        private final long changedInvisibleTime;
+        private final String groupId;
+        private final String topicId;
+        private final int queueId;
+        private final long offset;
+        private final boolean suspend;
+
+        public ChangeInvisibilityRecord(long popTime, long invisibleTime, long changedPopTime,
+            long changedInvisibleTime, String groupId, String topicId, int queueId, long offset, boolean suspend) {
+            this.popTime = popTime;
+            this.invisibleTime = invisibleTime;
+            this.changedPopTime = changedPopTime;
+            this.changedInvisibleTime = changedInvisibleTime;
+            this.groupId = groupId;
+            this.topicId = topicId;
+            this.queueId = queueId;
+            this.offset = offset;
+            this.suspend = suspend;
+        }
+
+        public long getPopTime() {
+            return popTime;
+        }
+
+        public long getInvisibleTime() {
+            return invisibleTime;
+        }
+
+        public long getChangedPopTime() {
+            return changedPopTime;
+        }
+
+        public long getChangedInvisibleTime() {
+            return changedInvisibleTime;
+        }
+
+        public String getGroupId() {
+            return groupId;
+        }
+
+        public String getTopicId() {
+            return topicId;
+        }
+
+        public int getQueueId() {
+            return queueId;
+        }
+
+        public long getOffset() {
+            return offset;
+        }
+
+        public boolean isSuspend() {
+            return suspend;
+        }
+    }
+
     // Use broker escape bridge to support remote read
     public CompletableFuture<Triple<MessageExt, String, Boolean>> getMessageAsync(PopConsumerRecord consumerRecord) {
         return this.brokerController.getEscapeBridge().getMessageAsync(consumerRecord.getTopicId(),
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -497,36 +497,18 @@ public class PopConsumerService extends ServiceThread {
     public void changeInvisibilityDuration(long popTime, long invisibleTime, long changedPopTime,
                                            long changedInvisibleTime, String groupId, String topicId,
                                            int queueId, long offset, boolean suspend) {
-
-        if (brokerConfig.isPopConsumerKVServiceLog()) {
-            log.info("PopConsumerService change, time={}, invisible={}, " +
-                    "groupId={}, topic={}, queueId={}, offset={}, new time={}, new invisible={}",
-                popTime, invisibleTime, groupId, topicId, queueId, offset, changedPopTime, changedInvisibleTime);
-        }
-
-        PopConsumerRecord ckRecord = new PopConsumerRecord(
-            changedPopTime, groupId, topicId, queueId, 0, changedInvisibleTime, offset, null, suspend);
-
-        PopConsumerRecord ackRecord = new PopConsumerRecord(
-            popTime, groupId, topicId, queueId, 0, invisibleTime, offset, null, suspend);
-
-        // No need to generate new records when the group does not exist,
-        // because these retry messages will not be consumed by anyone.
-        boolean skipWrite = brokerConfig.isPopReviveSkipIfGroupAbsent() &&
-            !brokerController.getSubscriptionGroupManager().containsSubscriptionGroup(groupId);
-
-        if (skipWrite) {
-            log.info("PopConsumerService change invisibility skip, time={}, " +
-                "groupId={}, topicId={}, queueId={}, offset={}", popTime, groupId, topicId, queueId, offset);
-        }
-
-        List<PopConsumerRecord> ckRecords = skipWrite ? Collections.emptyList() : Collections.singletonList(ckRecord);
-        List<PopConsumerRecord> ackRecords = Collections.singletonList(ackRecord);
-        if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
-            popConsumerCache.writeAndDeleteRecords(ckRecords, ackRecords);
-        } else {
-            this.popConsumerStore.writeAndDeleteRecords(ckRecords, ackRecords);
-        }
+        ChangeInvisibleTimeRequestEntry changeRecord = new ChangeInvisibleTimeRequestEntry();
+        changeRecord.setConsumerGroup(groupId);
+        changeRecord.setTopic(topicId);
+        changeRecord.setQueueId(queueId);
+        changeRecord.setOffset(offset);
+        changeRecord.setInvisibleTime(changedInvisibleTime);
+        changeRecord.setPopTime(popTime);
+        changeRecord.setOldInvisibleTime(invisibleTime);
+        changeRecord.setChangedPopTime(changedPopTime);
+        changeRecord.setChangedInvisibleTime(changedInvisibleTime);
+        changeRecord.setSuspend(suspend);
+        batchChangeInvisibilityDuration(Collections.singletonList(changeRecord));
     }
 
     public void batchChangeInvisibilityDuration(List<ChangeInvisibleTimeRequestEntry> changeRecords) {
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -535,7 +535,6 @@ public class PopConsumerService extends ServiceThread {
 
         List<PopConsumerRecord> ckRecords = new ArrayList<>(changeRecords.size());
         List<PopConsumerRecord> ackRecords = new ArrayList<>(changeRecords.size());
-        List<PopConsumerRecord> storeAckRecords = new ArrayList<>(changeRecords.size());
 
         for (ChangeInvisibleTimeRequestEntry changeRecord : changeRecords) {
             if (brokerConfig.isPopConsumerKVServiceLog()) {
@@ -546,15 +545,11 @@ public class PopConsumerService extends ServiceThread {
                     changeRecord.getChangedPopTime(), changeRecord.getChangedInvisibleTime());
             }
 
-            PopConsumerRecord ckRecord = new PopConsumerRecord(
-                changeRecord.getChangedPopTime(), changeRecord.getConsumerGroup(), changeRecord.getTopic(),
-                changeRecord.getQueueId(), 0, changeRecord.getChangedInvisibleTime(), changeRecord.getOffset(),
-                null, changeRecord.isSuspend());
-
             PopConsumerRecord ackRecord = new PopConsumerRecord(
                 changeRecord.getPopTime(), changeRecord.getConsumerGroup(), changeRecord.getTopic(),
                 changeRecord.getQueueId(), 0, changeRecord.getOldInvisibleTime(), changeRecord.getOffset(),
                 null, changeRecord.isSuspend());
+            ackRecords.add(ackRecord);
 
             boolean skipWrite = brokerConfig.isPopReviveSkipIfGroupAbsent() &&
                 !brokerController.getSubscriptionGroupManager().containsSubscriptionGroup(changeRecord.getConsumerGroup());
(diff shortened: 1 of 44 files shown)


RockteMQ-AI (commented):

### Summary
This PR adds batch support for `ChangeInvisibleTime` across the full stack (remoting protocol → client → broker → proxy), reducing network round-trips and RocksDB flushes when multiple pop consumer records need invisible-time updates in the same batch. The feature is gated by `enableBatchChangeInvisibleTime` and falls back to per-entry processing when disabled.

### Findings

- **[Critical]** `PopConsumerCache.writeAndDeleteRecords` — **Busy-wait spin lock**: `while (!consumerLockService.tryLock(...)) { Thread.yield(); }` is a CPU-consuming spin loop. Under high contention (many concurrent batch operations on the same group/topic), this can waste significant CPU. Consider using a timed lock (e.g., `tryLock` with a timeout + retry with backoff) or a blocking lock abstraction.

- **[Critical]** `PopConsumerCache.writeAndDeleteRecords` — **Lock topic derived from first record only**: The lock topic is parsed from `lockRecord` which comes from `validateAndGetLockRecord(writeRecordList, deleteRecordList)`. If the batch somehow contains records spanning different topics (despite the header-level validation), this could lead to incorrect locking. Please add an assertion or defensive check that all records in the batch share the same topic+group.

- **[Warning]** `ChangeInvisibleTimeProcessor.processBatchRequest` — **Batch size splitting**: The proxy splits oversized batches (`splitOversizedBrokerGroup`), but the broker-side `processBatchRequest` also checks `popBatchChangeInvisibleTimeMaxSize`. If the proxy split threshold and broker max size are misconfigured, the broker will reject the batch and all entries fail. Consider documenting the expected relationship between these two config values, or having the broker auto-split instead of rejecting.

- **[Warning]** `DefaultReceiptHandleManager.scheduleRenewTask` — **No max batch size for renew**: The renew path collects all expiring handles in a group into a single batch call. If a group has thousands of handles expiring simultaneously (e.g., after a long consumer pause), this could create a very large batch. Consider capping the batch size and splitting into multiple calls.

- **[Warning]** `ConsumerProcessor.filterPopResult` — **TO_RETURN batch not flushed on error path**: The `toReturnMessageList` is accumulated during the message loop and sent after the loop via `batchChangeInvisibleTime`. If an exception occurs during the loop (e.g., in the `REJECT` case), the accumulated TO_RETURN handles may not be processed. Consider wrapping the post-loop batch send in a `finally` block or ensuring it runs regardless of earlier exceptions.

- **[Warning]** `PopConsumerCache.writeAndDeleteRecords` — **Buffer eviction race**: When records are in the in-memory buffer and a `writeAndDeleteRecords` is called, the method splits records into buffer vs. store paths. The buffer deletion path calls `buffer.remove(record)` while the store path adds to `storeDeleteRecords`. If a concurrent `scanExpiredRecords` promotes a record from buffer to store between the split and the actual store delete, the record could be missed. The lock should prevent this, but please verify the lock scope covers the full split+write+delete sequence.

- **[Info]** `PopConsumerRocksdbStore.writeAndDeleteRecords` — **Delete-before-put ordering**: The `WriteBatch` processes all deletes before puts. This is correct for the case where the same key appears in both lists (old record deleted, new record written with different key due to changed popTime). The ordering is sound.

- **[Info]** Good defensive handling of null entries in `prepareBatchRequestEntries` and `processBatchChangeInvisibleTimeResponse`. Each null entry gets a failure response without aborting the entire batch.

- **[Info]** Test coverage is comprehensive — covers cache, RocksDB store, service layer, processor, client API, and proxy components. The split-by-broker and oversized-batch tests in the proxy are particularly valuable.

### Suggestions

1. Replace the `Thread.yield()` spin loop with a timed wait or `CompletableFuture`-based lock to avoid CPU waste under contention.
2. Add a defensive assertion that all records in a batch share the same topic+group before acquiring the lock.
3. Consider aligning or documenting the relationship between proxy `batchChangeInvisibleTimeMaxSize` and broker `popBatchChangeInvisibleTimeMaxSize`.
4. Add a max batch size cap in the renew path to handle edge cases with large numbers of expiring handles.

### Cross-repo Note
This PR adds new request/response protocol types (`BatchChangeInvisibleTimeRequestHeader`, `BatchChangeInvisibleTimeRequestBody`, etc.) in `rocketmq-remoting`. The `rocketmq-clients` repos do not need changes since the batch protocol is internal (broker ↔ proxy), not client-facing.

---

RockteMQ-AI (commented):

### Summary
Re-review after 2 new commits:
1. `5df26c4` Fix single change invisible time KV write batch
2. `76cc6b4` Serialize oversized batch change invisible time chunks

### Changes Since Last Review

**Commit 1 — Single-entry path fix:** The `changeInvisibilityDuration` method now uses `writeAndDeleteRecords` for both the single-entry and batch paths, ensuring a single batched KV operation instead of separate write + delete calls. This is a clean improvement.

**Commit 2 — Oversized batch chunking:** The proxy now splits large `handleMessageList` into chunks of `batchChangeInvisibleTimeMaxNum` and processes them **sequentially** via `thenCompose`. This avoids overwhelming a single broker with many concurrent write-heavy batches. The fallback to per-item requests on failure is preserved.

### Updated Findings

- **[Resolved]** The single-entry path now correctly uses the batch API — good fix.
- **[Resolved]** Oversized batches are now chunked and serialized — addresses the concern about large batch pressure.
- **[Info]** `PopConsumerCache.writeAndDeleteRecords` — The spin-lock pattern (`while (!tryLock) { Thread.yield(); }`) is consistent with the existing `deleteRecords` method in the same class. While not ideal under high contention, this follows the established codebase convention. Acceptable for now; consider a timed backoff in a future cleanup.
- **[Positive]** The `validateAndGetLockRecord` method correctly ensures all records share the same lock key, preventing cross-topic/group lock conflicts.
- **[Positive]** Good test coverage for the new `writeAndDeleteRecords` method — tests cover empty deletes, buffered records, topic lock routing, and mixed lock key rejection.

### Verdict
The two new commits meaningfully address the previous review findings. The implementation looks solid.

---

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RequestCode.java
```diff
@@ -82,6 +82,7 @@ public class RequestCode {
     public static final int BATCH_ACK_MESSAGE = 200151;
     public static final int PEEK_MESSAGE = 200052;
     public static final int CHANGE_MESSAGE_INVISIBLETIME = 200053;
+    public static final int BATCH_CHANGE_MESSAGE_INVISIBLETIME = 200153;
```
lizhimins, line 85: suggest change

RockteMQ-AI (commented):
## Review by github-manager-bot (Re-review after 4 new commits)

### Summary
Four new commits since the previous review:
1. **Refine batch entries** — Introduces `ChangeInvisibleTimeRequestEntry` as a first-class remoting type with proper serialization
2. **Remove redundant validation** — Removes the batch-level topic/group validation in `PopConsumerCache` (now enforced at the header level)
3. **Simplify write/delete** — Replaces the busy-wait spin lock with a clean split-buffer/store approach
4. **Fix serializable compat** — Skips `transient` fields in the reflection-based compatibility test

### Previous Findings — Status

- ~~**[Critical]** Busy-wait spin lock in `writeAndDeleteRecords`~~ — **Resolved.** The spin lock is removed. The new implementation splits records into buffer vs. store paths and processes them independently. Much cleaner.

- ~~**[Critical]** Lock topic derived from first record only~~ — **Resolved.** The lock mechanism is removed entirely. The batch header enforces topic+group consistency, and the split logic processes each record independently.

- **[Warning]** Batch size splitting — **Partially addressed.** The broker now checks `batchMaxNum` and rejects oversized batches. The proxy splits oversized groups via `splitOversizedBrokerGroup`. However, see remaining concern below.

- **[Warning]** TO_RETURN batch not flushed on error — **Likely resolved.** The loop catches per-message exceptions and continues, so the post-loop batch send should execute. But consider wrapping the post-loop `batchChangeInvisibleTime` call in a `finally` block for extra safety.

- **[Warning]** Buffer eviction race — **Resolved.** The new `writeAndDeleteRecords` cleanly separates buffer and store paths, avoiding the race between buffer eviction and batch write.

### Remaining Findings

- **[Warning]** `ConsumerProcessor.scheduleRenewTask` — **No max batch size for renew.** The renew path still collects all expiring handles in a group into a single `batchChangeInvisibleTime` call. If a group has thousands of handles expiring simultaneously (e.g., after a long consumer pause), this creates a very large batch. Consider capping the batch size (e.g., `proxyConfig.getBatchChangeInvisibleTimeMaxNum()`) and splitting into multiple calls.

- **[Warning]** `ChangeInvisibleTimeProcessor.processBatchRequestAsync` — **Batch rejection vs. auto-split.** When `requestEntries.size() > batchMaxNum`, the broker rejects the entire batch. If the proxy split threshold and broker max size are misconfigured, all entries fail. Consider auto-splitting at the broker level instead of rejecting, or at minimum documenting the expected relationship between `popBatchChangeInvisibleTimeMaxSize` (proxy) and `batchChangeInvisibleTimeMaxNum` (broker).

- **[Info]** `ChangeInvisibleTimeRequestEntry` — Good addition of a first-class remoting type. The `encode()`/`decode()` methods handle the serialization correctly. The `@JSONField` annotations ensure proper JSON mapping.

- **[Info]** `RemotingSerializableCompatTest` — Skipping `transient` fields in the compatibility check is correct. Transient fields are not part of the wire format.

### Cross-repo Note
This PR adds `BATCH_CHANGE_MESSAGE_INVISIBLETIME` as a new `RequestCode`. Ensure the `RequestCode` value does not conflict with any in-flight PRs in `apache/rocketmq`. The corresponding client-side support is included in this PR, so no cross-repo coordination is needed.

---

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/RequestCode.java
qianye1001: 和 batch ack 一样的逻辑，就用这个吧

fuyou001:
**[P1] Preserve input ordering when returning batch change-invisible-time results**

`batchChangeInvisibleTime` does not preserve the order of `handleMessageList`: expired-handle results are appended first, while the remaining results are appended later in `HashMap` group iteration order. However, `ReceiptHandleProcessor.batchChangeInvisibleTime` consumes the returned list by index and completes the future for `indexes.get(i)` with `results.get(i)`.

When a renewal batch contains handles from different brokers/real topics, or contains an expired handle, a result can therefore be assigned to a different message. This may update message A with message B receipt handle and cause subsequent renew/ack operations to target the wrong offset.

Please preserve the original input order in the returned result list, or correlate each result with its `ReceiptHandleMessage` instead of consuming results positionally. A regression test should cover interleaved broker/topic groups and an expired handle.

broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
```diff
@@ -517,21 +518,67 @@ public void changeInvisibilityDuration(long popTime, long invisibleTime, long ch
...
         if (brokerConfig.isEnablePopBufferMerge() && popConsumerCache != null) {
-            if (popConsumerCache.deleteRecords(Collections.singletonList(ackRecord)).isEmpty()) {
-                return;
+            popConsumerCache.writeAndDeleteRecords(ckRecords, ackRecords);
+        } else {
+            this.popConsumerStore.writeAndDeleteRecords(ckRecords, ackRecords);
+        }
+    }
+
+    public void batchChangeInvisibilityDuration(List<ChangeInvisibleTimeRequestEntry> changeRecords) {
```
fuyou001, line 531: changeInvisibilityDuration can call batchChangeInvisibilityDuration

proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
```diff
@@ -404,6 +434,150 @@ protected CompletableFuture<List<BatchAckResult>> processBrokerHandle(ProxyConte
...
+                futures[futureIndex++] = processBrokerChangeInvisibleTime(
+                    ctx, consumerGroup, topic, brokerHandleList, invisibleTime, timeoutMillis, suspend);
+            }
+            CompletableFuture.allOf(futures).whenComplete((val, throwable) -> {
+                if (throwable != null) {
+                    future.completeExceptionally(throwable);
+                    return;
+                }
+                for (CompletableFuture<List<BatchChangeInvisibleTimeResult>> resultFuture : futures) {
+                    batchResultList.addAll(resultFuture.join());
```
RongtongJin: This changes the result ordering contract. `batchResultList` is built from expired handles first, then appends each broker/topic group using `HashMap.values()` iteration order. However, `ReceiptHandleProcessor.batchChangeInvisibleTime` consumes the returned list by index and completes `futureList.get(indexes.get(i))` with `results.get(i)`. If one renewal batch contains handles for multiple brokers, the result list can differ from the original `handleMessageList` order, causing a message to receive another message's renewed receipt handle. Please either preserve the original input order when building the result list, or have the caller map results back by `BatchChangeInvisibleTimeResult.getReceiptHandleMessage()` instead of by index.

RockteMQ-AI (commented):

### Summary
One new commit since the previous review:
- **Fix test flake** — Reorders mock setup before  in , properly mocks  to return a completed , and adds try/finally cleanup.

### New Commit Assessment

- **[Info]**  — Good fix. The previous ordering ( → ) had a race where the background thread could attempt processing before the mock was configured. The new ordering (mock setup →  → ) eliminates the race. The try/finally cleanup ( + ) is also a good practice to avoid test interference.

### Previous Findings — No Change

All previous findings remain as stated in the last review. The remaining warnings about batch size capping in  and broker-level batch rejection vs. auto-split are still applicable.

No new issues found in this commit.

---

RockteMQ-AI (commented):

### Summary
One new commit since the previous review:
- **Fix test flake** — Reorders mock setup before `start()` in `ReceiptHandleProcessorTest.testStart()`, properly mocks `changeInvisibleTime` to return a completed `CompletableFuture<AckResult>`, and adds try/finally cleanup.

### New Commit Assessment

- **[Info]** `ReceiptHandleProcessorTest.testStart` — Good fix. The previous ordering (`start()` then `addReceiptHandle()`) had a race where the background thread could attempt processing before the mock was configured. The new ordering (mock setup, `addReceiptHandle()`, then `start()`) eliminates the race. The try/finally cleanup (`removeReceiptHandle` + `shutdown`) is also a good practice to avoid test interference.

### Previous Findings — No Change

All previous findings remain as stated in the last review. The remaining warnings about batch size capping in `ConsumerProcessor.scheduleRenewTask` and broker-level batch rejection vs. auto-split are still applicable.

No new issues found in this commit.

---

fuyou001:
**[P1] Do not fall back to single requests after an ambiguous batch failure**

`processBrokerChangeInvisibleTime` falls back to individual `changeInvisibleTime` calls for every exception from `batchChangeInvisibleTime`, including timeouts, connection resets, and lost responses. Those failures do not prove that the broker did not apply the batch. The broker may have already persisted the new CK records and only the response failed to reach the proxy.

Retrying each entry with the original receipt handle is not idempotent: each execution generates a new `changedPopTime`, so the fallback can create a second CK while the CK created by the successful batch remains. Both CKs may later revive the same message, causing duplicate delivery.

Please restrict fallback to an explicit response proving that the broker does not support the batch request code. Transport failures and timeouts should not be retried this way unless the operation is made idempotent, for example with a request identifier deduplicated by the broker. A regression test should cover the case where the broker applies the batch but the client future completes with a timeout.

fuyou001:
**[P1] Make cache ownership transfer atomic with concurrent cleanup**

`PopConsumerCache.writeAndDeleteRecords` first classifies an old CK with `consumerRecords.contains(record)`, then performs the RocksDB write/delete batch, and only afterward calls `deleteRecords(bufferDeleteRecords)`. This check and removal are not atomic with `cleanupRecords`.

A cleanup thread can move the old CK from `recordTreeMap` to `removeTreeMap`, flush it to RocksDB, and clear it from the cache after `contains` returned true but before the final cache deletion. The renewal path then writes the new CK without including the old CK in `storeDeleteRecords`; its final cache deletion fails, and that failure result is ignored. RocksDB can therefore retain both the old and new CK, potentially reviving the same message more than once. If the keys coincide, a late cleanup write can also overwrite the new value with the old one.

Please coordinate cleanup and renewal through an atomic claim/removal or another explicit ownership state transition before deciding whether the old CK must be deleted from RocksDB. A concurrent regression test should pause between classification and deletion while cleanup stages and flushes the same record.

RockteMQ-AI (commented):
## Review by github-manager-bot (Re-review after 2 new commits)

### Summary
Two new commits since the previous review:
1. **`7a2e4bc` Fix batch change invisible time result ordering** — Refactors `ConsumerProcessor.batchChangeInvisibleTime` to preserve input order in results by using index-based arrays instead of appending to a list.
2. **`3c5cc1b` Keep single invisible time change direct** — Reverts the delegation of `changeInvisibilityDuration` to the batch path, restoring the original direct implementation.

### New Commit Assessment

- **[Info]** `ConsumerProcessor.batchChangeInvisibleTime` ordering fix — Good change. The previous implementation grouped messages by broker and appended results in broker-group order, losing the original input ordering. The new approach:
  - Uses `BatchChangeInvisibleTimeResult[]` array indexed by original position
  - Tracks indices via `brokerHandleIndexMap` instead of message lists
  - Converts array to list via `buildBatchChangeInvisibleTimeResultList` at the end
  - This ensures callers receive results in the same order as the input `handleMessageList`, which is important for correlating results with requests.

- **[Info]** `ConsumerProcessorTest.testBatchChangeInvisibleTimePreserveInputOrderWithExpiredAndInterleavedGroups` — Comprehensive test that verifies ordering is preserved with expired handles, interleaved broker assignments, and multiple groups. Good coverage.

- **[Info]** `PopConsumerService.changeInvisibilityDuration` revert — Correct decision. Delegating a single-item call through the batch path (`batchChangeInvisibilityDuration(Collections.singletonList(...))`) added unnecessary overhead (record conversion, batch assembly) for the common single-item case. The direct path with `PopConsumerRecord` is more efficient and avoids the extra allocation. The net diff for this file is zero.

### Previous Findings — No Change

All previous findings remain as stated in the last review. The remaining warnings about batch size capping in `ConsumerProcessor.scheduleRenewTask` and broker-level batch rejection vs. auto-split are still applicable.

No new issues found in these commits.

---

RockteMQ-AI (approved):
## Review by github-manager-bot (Re-review after 2 new commits)

### Summary
Two new commits since the previous review:
1. **`7ed9d80c` Fix receipt handle group async cleanup test** — Minor test adjustment in `ReceiptHandleGroupTest.java` (+3/-3).
2. **`7b533ec6` Mark broker shutdown test medium in Bazel** — Adds timeout/size annotation to a Bazel test target in `broker/BUILD.bazel` (+3/-0).

### Diff Verification
✅ All previously reviewed changes remain intact:
- Batch `ChangeInvisibleTime` remoting protocol
- `ConsumerProcessor.batchChangeInvisibleTime()` with index-based result ordering
- Single invisible time change kept as direct path (not delegated to batch)
- Chunk serialization for oversized batches
- Pop cache batch write/delete simplification
- Remoting serializable compat for transient fields

✅ New commits are test-only changes:
- Test fix is a minor assertion/ordering adjustment
- Bazel annotation is a build configuration change

### Verdict
Test fixes are clean and don't affect production code. The overall PR design (batch invisible time with chunking, backward-compatible single-item path) remains sound. LGTM.

---

broker/src/main/java/org/apache/rocketmq/broker/processor/ChangeInvisibleTimeProcessor.java
```diff
@@ -186,6 +206,248 @@ public CompletableFuture<RemotingCommand> processRequestAsync(final Channel chan
...
+
+        return CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).thenApply(ignored -> {
+            BatchChangeInvisibleTimeResponseBody responseBody = new BatchChangeInvisibleTimeResponseBody();
+            responseBody.setEntries(Arrays.asList(responseEntries));
+            response.setBody(responseBody.encode());
+            return response;
+        });
+    }
+
+    protected boolean validateBatchRequestEntries(BatchChangeInvisibleTimeRequestHeader batchRequestHeader,
```
lollipopjin: How about renaming it to normalizeAndValidateBatchRequestEntries to reflect that it mutates the input requestEntries (back-filling consumerGroup / topic from the header)?

qianye1001:
@fuyou001 Thanks, agreed. Fixed in d1d6e3e8ddbc13224a216fa1473e55225eb27ec1.

The proxy fallback is now restricted to the explicit unsupported case only: the real exception must be an `MQBrokerException` with `ResponseCode.REQUEST_CODE_NOT_SUPPORTED`. For ambiguous failures such as timeout/transport failure/lost response, `processBrokerChangeInvisibleTime` now completes the batch entries with `INTERNAL_SERVER_ERROR` and does not retry single `changeInvisibleTime`, so it will not create a second CK with a new `changedPopTime` after the broker may already have applied the batch.

Added regression coverage in `ConsumerProcessorTest` for both paths:
- `testBatchChangeInvisibleTimeFallbackOnlyWhenRequestCodeNotSupported`
- `testBatchChangeInvisibleTimeDoesNotFallbackOnAmbiguousFailure`

Verified locally:
- `mvn -pl proxy -am -DfailIfNoTests=false -Dtest=ConsumerProcessorTest test`
- `mvn -pl broker -am -DfailIfNoTests=false -Dtest=ChangeInvisibleTimeProcessorTest test`

Comment addressed: https://github.com/apache/rocketmq/pull/10519#issuecomment-4741426672

qianye1001:
@fuyou001 Thanks for the detailed analysis. I considered this one and I do not plan to address it in this batch-change PR.

The reason is that this is not a new correctness model introduced by the batch path. The existing single `changeInvisibilityDuration` path already follows the same pattern: write the new CK, try to remove the old CK from the pop buffer, and then delete the old CK from RocksDB when needed. A concurrent cleanup can theoretically race with that path as well, so fixing this properly would be a broader PopConsumerCache ownership/cleanup change, not a batch-change-specific fix.

Making the ownership transfer atomic would likely require coordinating renewal/change and cleanup with an additional lock or explicit ownership state transition on the hot path. For the batch path that cost is relatively high, especially because the observed impact is a low-probability duplicate delivery scenario and POP consumption already has at-least-once semantics. I would prefer not to add that synchronization cost in this PR.

So for this PR I am keeping the batch behavior aligned with the existing single-change semantics and limiting the changes to reducing broker round trips and store writes/deletes. If we want to harden this race, I think it should be handled separately for both single and batch paths with a dedicated design and benchmark.

Comment addressed: https://github.com/apache/rocketmq/pull/10519#issuecomment-4741428792

RockteMQ-AI (commented):
## Review by github-manager-bot (Re-review after 3 new commits)

### Summary
Three new commits since the previous review:
1. **`d1d6e3e` Fix batch change invisible time fallback** — Critical correctness fix for the fallback logic
2. **`3e563b2` ci: re-trigger CI** — CI housekeeping
3. **`dae2063` Stabilize producer backpressure test** — Removes flaky timing-dependent test

### Changes Since Last Review

**1. Fallback Logic Fix (`ConsumerProcessor.java`)**
- **[Info]** Previously, any failure from `batchChangeInvisibleTime` triggered a fallback to single `changeInvisibleTime` requests. This was unsafe: if the batch partially succeeded on the broker (e.g., timeout after applying some changes), falling back would create duplicate checkpoints.
- **[Info]** Now only falls back when the error is `REQUEST_CODE_NOT_SUPPORTED` (broker genuinely does not support the batch API). For all other failures (timeout, internal error), returns errors directly via `buildBatchChangeInvisibleTimeExceptionResults`.
- **[Info]** The `isBatchChangeInvisibleTimeNotSupported()` helper correctly unwraps the real exception via `ExceptionUtils.getRealException()` before checking the type and response code.

**2. Method Rename (`ChangeInvisibleTimeProcessor.java`)**
- **[Info]** `validateBatchRequestEntries` → `normalizeAndValidateBatchRequestEntries`. The new name accurately reflects that the method both normalizes and validates entries.

**3. Test Stabilization (`DefaultMQProducerTest.java`)**
- **[Info]** `testRunningSetBackCompress` simplified: removed multi-threaded `CountDownLatch` approach, replaced with direct semaphore manipulation. Eliminates timing-dependent flakiness.

**4. New Tests (`ConsumerProcessorTest.java`)**
- **[Info]** `testBatchChangeInvisibleTimeFallbackOnlyWhenRequestCodeNotSupported` — Verifies fallback occurs only for `REQUEST_CODE_NOT_SUPPORTED`.
- **[Info]** `testBatchChangeInvisibleTimeDoesNotFallbackOnAmbiguousFailure` — Verifies no fallback for ambiguous failures (e.g., `TimeoutException`), preventing duplicate checkpoints.

### Findings
- **[Info]** The fallback fix is a significant correctness improvement. The distinction between "broker does not support batch" (safe to fall back) and "batch may have partially succeeded" (unsafe to fall back) is well-reasoned.
- **[Info]** Test coverage for the two fallback paths (supported vs. unsupported) is thorough.
- **[Info]** The `buildReceiptHandleMessages` helper reduces test boilerplate.

### Suggestions
- **[Info]** Consider logging the actual exception type in the non-fallback path at `INFO` level (currently `WARN`) to help operators distinguish between "broker too old" and "transient failure" in production logs. Minor nit.

Overall the incremental changes look solid. The fallback correctness fix is important.

---

remoting/src/main/java/org/apache/rocketmq/remoting/protocol/body/BatchChangeInvisibleTimeRequestBody.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.remoting.protocol.body;
+
+import java.util.List;
+import org.apache.rocketmq.remoting.protocol.RemotingSerializable;
+
+public class BatchChangeInvisibleTimeRequestBody extends RemotingSerializable {
```
lizhimins, line 23: need to check if the batch size exceeds the length limit of the frequency.

fuyou001 (dismissed):

RongtongJin (dismissed):

qianye1001:
The remaining 2 CI failures are unrelated flaky tests:

- **bazel-compile**: `OffsetResetIT` — not modified by this PR, known flaky integration test
- **maven-compile (macos)**: `ServiceThreadTest.serviceThreadShouldNotLoseWakeupUnderStress` — not modified by this PR, timing jitter on macOS runner

Re-running both jobs.

PR #10528, opened 2026-06-17
[ISSUE #10527] Reduce auxiliary component allocation via AtomicLong, string caches, StringBuilder reuse, and dirty flag

### LINKED ISSUE #10527: [Enhancement] Reduce auxiliary component allocation via AtomicLong, string caches, StringBuilder reuse, and dirty flag

### Summary

Reduce allocation in auxiliary store components by replacing boxed maps with primitive atomics, caching string keys, reusing StringBuilders, and optimizing timer wheel flush.

### Motivation

JFR profiling reveals several allocation hotspots in auxiliary store components that run on every message:

1. **`QueueOffsetOperator`** — uses `ConcurrentMap<String, Long>` for queue offsets, boxing every `long` update into a `Long` object and requiring map lookups.
2. **`BrokerStatsManager`** — `buildStatsKey`/`topicQueueKey`/`consumerOffset` methods create new strings on every call. `incQueue*` methods use `Integer` parameters causing autoboxing.
3. **`IndexService`** — builds index keys with `StringBuilder` allocated per-call.
4. **`TimerWheel`** — flushes all slots unconditionally, even when no changes occurred.

1. `QueueOffsetOperator`: Replace `ConcurrentMap<String, Long>` with `ConcurrentMap<String, AtomicLong>` — eliminates boxing on every update.
2. `BrokerStatsManager`: Cache `buildStatsKey`/`topicQueueKey`/`consumerOffset` string results. Change `Integer` parameters to `int` to eliminate autoboxing.
3. `IndexService`: Reuse `StringBuilder` via ThreadLocal.
4. `TimerWheel`: Add volatile `dirty` flag — skip flush when no changes since last flush.
5. `AppendMessageResult`: Add constructor with pre-computed fields to avoid redundant allocation.

- Use `LongAdder` instead of `AtomicLong` — `AtomicLong` is sufficient for moderate contention and provides `get()` for reads.
- Use `String.format` cache — `String.concat` is faster for small fixed key patterns.
- Use object pool for `StringBuilder` — ThreadLocal is simpler and thread-safe by design.

### Additional Context

Part of a larger JFR-driven optimization effort. Related PRs: #10443, #10444, #10514, #10524, #10526.

wang-jiahua:

Fixes #10527

### Brief Description

Reduce allocation in auxiliary store components through independent optimizations:

1. **`BrokerStatsManager`** — Change `Integer` parameters to `int` in `incQueue*` methods to eliminate autoboxing. (String caching was removed after microbenchmark showed ConcurrentHashMap lookup is slower than JIT-optimized String concat.)
2. **`IndexService`** — Reuse a `ThreadLocal<StringBuilder>` for key building instead of allocating per call (`buildKey` is reachable from query thread pools and concurrent dispatch, so the builder must be per-thread).
3. **`TimerWheel`** — Add a volatile `dirty` flag to skip flush when nothing changed; the flag is cleared before diffing so a concurrent `putSlot` can only cause an extra flush, never a missed one.
4. **`AppendMessageResult`** — Add constructor with pre-computed fields to avoid redundant allocation.

1. **Unit tests**: All 15 `BrokerStatsManagerTest` tests pass.
2. **Compilation**: `store` module compiles cleanly on JDK 21.
3. **Commercial compatibility**: No commercial classes extend any modified class.
4. **Microbenchmark**: String caching was removed because a microbenchmark showed it was 120% slower than JIT-optimized String concatenation (ConcurrentHashMap lookup overhead exceeds String allocation cost for short strings). A `QueueOffsetOperator` AtomicLong conversion originally in this PR was dropped during review: `getTopicQueueTable()` exposes the live map, and callers (including reflection-based extensions) may rely on mutating it, so a snapshot-returning accessor could break them silently.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
@@ -49,6 +49,7 @@ public class IndexService implements CommitLogDispatchStore {
     private final String storePath;
     private final ArrayList<IndexFile> indexFileList = new ArrayList<>();
     private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+    private static final ThreadLocal<StringBuilder> KEY_BUILDER = ThreadLocal.withInitial(() -> new StringBuilder(128));
 
     public IndexService(final DefaultMessageStore store) {
         this.defaultMessageStore = store;
@@ -215,10 +216,15 @@ public class IndexService implements CommitLogDispatchStore {
     }
 
     private String buildKey(final String topic, final String key) {
-        return topic + "#" + key;
+        StringBuilder keyBuilder = KEY_BUILDER.get();
+        keyBuilder.setLength(0);
+        return keyBuilder.append(topic).append('#').append(key).toString();
     }
+
     private String buildKey(final String topic, final String key, final String indexType) {
-        return topic + "#" + indexType + "#" + key;
+        StringBuilder keyBuilder = KEY_BUILDER.get();
+        keyBuilder.setLength(0);
+        return keyBuilder.append(topic).append('#').append(indexType).append('#').append(key).toString();
     }
 
     public void buildIndex(DispatchRequest req) {
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerWheel.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerWheel.java
@@ -55,6 +55,8 @@ public class TimerWheel {
     };
     private final int wheelLength;
 
+    private volatile boolean dirty;
+
     private long snapOffset;
 
     public TimerWheel(String fileName, int slotsTotal, int precisionMs) throws IOException {
@@ -128,14 +130,26 @@ public class TimerWheel {
         if (mappedByteBuffer == null) {
             return;
         }
+        if (!dirty) {
+            return;
+        }
+        // Clear the flag before diffing: a concurrent putSlot during the diff re-marks it and the
+        // next flush picks the change up, so an interleaving can only cause an extra flush, never a
+        // missed one. All threads share the same underlying direct buffer (localBuffer duplicates
+        // only isolate position/limit), so the data itself is always visible here.
+        dirty = false;
         ByteBuffer bf = localBuffer.get();
-        bf.position(0);
-        bf.limit(wheelLength);
-        mappedByteBuffer.position(0);
-        mappedByteBuffer.limit(wheelLength);
-        for (int i = 0; i < wheelLength; i++) {
-            if (bf.get(i) != mappedByteBuffer.get(i)) {
-                mappedByteBuffer.put(i, bf.get(i));
+        int longAligned = wheelLength & ~7;
+        for (int i = 0; i < longAligned; i += 8) {
+            long local = bf.getLong(i);
+            if (local != mappedByteBuffer.getLong(i)) {
+                mappedByteBuffer.putLong(i, local);
+            }
+        }
+        for (int i = longAligned; i < wheelLength; i++) {
+            byte b = bf.get(i);
+            if (b != mappedByteBuffer.get(i)) {
+                mappedByteBuffer.put(i, b);
             }
         }
         this.mappedByteBuffer.force();
@@ -287,11 +301,10 @@ public class TimerWheel {
 
     public void putSlot(long timeMs, long firstPos, long lastPos) {
         localBuffer.get().position(getSlotIndex(timeMs) * Slot.SIZE);
-        // To be compatible with previous version.
-        // The previous version's precision is fixed at 1000ms and it store timeMs / 1000 in slot.
         localBuffer.get().putLong(timeMs / precisionMs);
         localBuffer.get().putLong(firstPos);
         localBuffer.get().putLong(lastPos);
+        dirty = true;
     }
 
     public void putSlot(long timeMs, long firstPos, long lastPos, int num, int magic) {
@@ -301,6 +314,7 @@ public class TimerWheel {
         localBuffer.get().putLong(lastPos);
         localBuffer.get().putInt(num);
         localBuffer.get().putInt(magic);
+        dirty = true;
     }
 
     public void reviseSlot(long timeMs, long firstPos, long lastPos, boolean force) {
@@ -313,11 +327,13 @@ public class TimerWheel {
         } else {
             if (IGNORE != firstPos) {
                 localBuffer.get().putLong(firstPos);
+                dirty = true;
             } else {
                 localBuffer.get().getLong();
             }
             if (IGNORE != lastPos) {
                 localBuffer.get().putLong(lastPos);
+                dirty = true;
             }
         }
     }
--- a/store/src/main/java/org/apache/rocketmq/store/AppendMessageResult.java
+++ b/store/src/main/java/org/apache/rocketmq/store/AppendMessageResult.java
@@ -84,6 +84,18 @@ public class AppendMessageResult {
         this.msgNum = msgNum;
     }
 
+    public AppendMessageResult(AppendMessageStatus status, long wroteOffset, int wroteBytes, String msgId,
+            long storeTimestamp, long logicsOffset, long pagecacheRT, int msgNum) {
+        this.status = status;
+        this.wroteOffset = wroteOffset;
+        this.wroteBytes = wroteBytes;
+        this.msgId = msgId;
+        this.storeTimestamp = storeTimestamp;
+        this.logicsOffset = logicsOffset;
+        this.pagecacheRT = pagecacheRT;
+        this.msgNum = msgNum;
+    }
+
     public long getPagecacheRT() {
         return pagecacheRT;
     }
--- a/store/src/test/java/org/apache/rocketmq/store/stats/BrokerStatsManagerTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/stats/BrokerStatsManagerTest.java
@@ -45,7 +45,7 @@ public class BrokerStatsManagerTest {
     private BrokerStatsManager brokerStatsManager;
 
     private static final String TOPIC = "TOPIC_TEST";
-    private static final Integer QUEUE_ID = 0;
+    private static final int QUEUE_ID = 0;
     private static final String GROUP_NAME = "GROUP_TEST";
     private static final String CLUSTER_NAME = "DefaultCluster";
 


RockteMQ-AI (commented):

### Summary
This PR reduces auxiliary component allocation in the store layer through five optimization strategies: reusable `StringBuilder` in `IndexService`, `AtomicLong`-based offset tracking in `QueueOffsetOperator`, string caching in `BrokerStatsManager`, dirty-flag flush gating in `TimerWheel`, and a new `AppendMessageResult` constructor overload. Overall direction is sound — reducing per-message allocation pressure in hot paths.

### Findings

- **[Critical]** `IndexService.java:52` — **Thread-safety violation with `reusableKeyBuilder`**. `IndexService` is shared across threads (commitlog dispatch, query, etc.), and `buildKey()` is called from both `queryOffset()` (read lock) and `buildIndex()` (write lock). A shared mutable `StringBuilder` without synchronization will produce corrupted keys under concurrent access. The JVM may even reorder `setLength(0)` and `append()` calls across threads. This is a **data corruption risk** — keys will be silently wrong, causing index lookups to miss or return incorrect offsets.
  - **Fix**: Either use a `ThreadLocal<StringBuilder>`, pass the builder as a parameter from callers that already hold appropriate locks, or revert to the original string concatenation (the JIT will optimize `topic + "#" + key` into a `StringBuilder` anyway, and the allocation is short-lived).

- **[Critical]** `TimerWheel.java:58,137,151,303,313,326,332` — **`dirty` flag race condition**. The pattern `if (!dirty) return; ... dirty = false;` is a non-atomic check-then-act. Between the `if (!dirty)` guard and the final `dirty = false`, another thread calling `setDirty(true)` can have its update lost — the flush proceeds but then resets the flag, losing the knowledge that new data was written. Over time this can cause the timer wheel to skip flushes, leading to data loss on crash.
  - **Fix**: Use `AtomicBoolean` with `compareAndSet(true, false)` to atomically claim the dirty state, or synchronize the flush method. Alternatively, accept that some flushes may be redundant (safe) by checking `dirty` again after the flush.

- **[Warning]** `BrokerStatsManager.java:496` — **Unbounded cache growth**. The `statsKeyByGroupCache` and similar `ConcurrentHashMap<String, String[]>` caches grow without eviction. In deployments with thousands of topics × groups, this can accumulate significant heap. Consider adding a size cap or using a bounded cache (e.g., Caffeine with `maximumSize`).

- **[Info]** `QueueOffsetOperator.java` — The change from `Long` to `AtomicLong` for offset values is a good correctness improvement (atomic increment vs. non-atomic read-modify-write). The `getTopicQueueTable()` snapshot copy is appropriate for iteration safety.

### Suggestions
1. The `IndexService` thread-safety issue is a blocker — please fix before merge.
2. The `TimerWheel` dirty flag issue should also be addressed; consider `AtomicBoolean.getAndSet(false)` as a minimal fix.
3. Consider adding a JMH benchmark for the hot paths to quantify the allocation reduction.

---

store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
```diff
@@ -49,6 +49,7 @@ public class IndexService implements CommitLogDispatchStore {
     private final String storePath;
     private final ArrayList<IndexFile> indexFileList = new ArrayList<>();
     private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+    private final StringBuilder reusableKeyBuilder = new StringBuilder(128);
```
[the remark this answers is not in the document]
wang-jiahua: @/tmp/reply-s4-idx.txt

store/src/main/java/org/apache/rocketmq/store/timer/TimerWheel.java
```diff
@@ -55,6 +55,8 @@ protected ByteBuffer initialValue() {
     };
     private final int wheelLength;
 
+    private volatile boolean dirty;
```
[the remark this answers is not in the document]
wang-jiahua: @/tmp/reply-s4-timer.txt

store/src/main/java/org/apache/rocketmq/store/stats/BrokerStatsManager.java
```diff
@@ -464,47 +470,92 @@ public void incGroupAckNums(final String group, final String topic, final int in
...
+        String[] keys = statsKeyByQueueCache.get(topic);
+        if (keys != null && queueId >= 0 && queueId < keys.length) {
+            String cached = keys[queueId];
+            if (cached != null) {
+                return cached;
+            }
         }
-        strBuilder.append(topic).append("@").append(queueId);
-        return strBuilder.toString();
+        String key = topic + "@" + queueId;
```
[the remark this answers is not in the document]
wang-jiahua: @/tmp/reply-s4-null.txt

store/src/main/java/org/apache/rocketmq/store/stats/BrokerStatsManager.java
```diff
@@ -380,32 +386,32 @@ public void onGroupDeleted(final String group) {
         this.momentStatsItemSetFallTime.delValueBySuffixKey(group, "@");
     }
 
-    public void incQueuePutNums(final String topic, final Integer queueId) {
+    public void incQueuePutNums(final String topic, final int queueId) {
```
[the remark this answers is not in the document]
wang-jiahua: @/tmp/reply-s4-int.txt

store/src/main/java/org/apache/rocketmq/store/stats/BrokerStatsManager.java
```diff
@@ -141,6 +142,11 @@ public class BrokerStatsManager {
     private ScheduledExecutorService cleanResourceExecutor;
 
     private final HashMap<String, StatsItemSet> statsTable = new HashMap<>();
+    private final ConcurrentHashMap<String, String[]> latencyKeyCache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, String[]> statsKeyByQueueCache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, ConcurrentHashMap<String, String>> statsKeyByGroupCache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, ConcurrentHashMap<String, String[]>> statsKeyByTopicQueueGroupCache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, ConcurrentHashMap<String, String[]>> statsKeyByQueueTopicGroupCache = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
wang-jiahua: @/tmp/reply-s4-cache.txt

RockteMQ-AI (commented):

### Summary
Re-review after new commit `f2503ca0` (2026-06-19). The update adds a new `AppendMessageResult` constructor overload and refines `QueueOffsetOperator` to use `AtomicLong` for offset tracking. However, the two **critical** issues from the previous review remain unaddressed.

### Unresolved Critical Issues

- **[Critical]** `IndexService.java` — **`reusableKeyBuilder` thread-safety violation persists**. The shared `StringBuilder` instance field is still used in `buildKey()` without synchronization. `IndexService` is accessed from multiple threads (dispatch, query). Under concurrent access, `setLength(0)` + `append()` interleaving will produce corrupted keys, causing silent index lookup failures.
  - **Fix**: Use `ThreadLocal<StringBuilder>`, or revert to `topic + "#" + key` (JIT optimizes this).

- **[Critical]** `TimerWheel.java` — **`dirty` flag race condition persists**. The `if (!dirty) return; ... dirty = false;` pattern is a non-atomic check-then-act. Between the guard and the reset, another thread can set `dirty = true` which then gets silently cleared, causing missed flushes and potential data loss on crash.
  - **Fix**: Use `AtomicBoolean.compareAndSet(true, false)` to atomically claim the dirty state.

### New Changes

- **[Info]** `AppendMessageResult.java` — New 8-arg constructor (drops `queueOffset` and `logicsOffset`). Callers that don't need these fields can use this lighter constructor. No issues.

- **[Info]** `QueueOffsetOperator.java` — `AtomicLong`-based offset tracking is a good correctness improvement. The `getTopicQueueTable()` snapshot copy is appropriate for iteration safety. The `setTopicQueueTable` conversion from `Map<String, Long>` to `Map<String, AtomicLong>` is a one-time cost during load, acceptable.

- **[Info]** `BrokerStatsManagerTest.java` — `Integer QUEUE_ID` → `int QUEUE_ID` avoids autoboxing. Minor but correct.

### Action Required
The two critical thread-safety issues **must be resolved before merge**. Please fix `IndexService` and `TimerWheel` as suggested.

---

wang-jiahua:

| #10528 | 224k | 1.14ms | 332 | **+3.2% TPS, -3.4% RT** ✅ |

Auxiliary component optimizations (QueueOffsetOperator AtomicLong, IndexService StringBuilder reuse, TimerWheel dirty flag) reduce boxing and string allocation on the put message path.

RockteMQ-AI (changes requested):

### Summary
Reduces auxiliary component allocation via AtomicLong counters, cached string keys, StringBuilder reuse, and dirty flag patterns.

### Issues Found

**[Critical] `IndexService.reusableKeyBuilder` is not thread-safe**

```java
private final StringBuilder reusableKeyBuilder = new StringBuilder(128);

private String buildKey(final String topic, final String key) {
    reusableKeyBuilder.setLength(0);
    return reusableKeyBuilder.append(topic).append(\"#\").append(key).toString();
}
```

`IndexService` is a shared service used by multiple threads (dispatch threads, query threads). `buildKey()` is called from `queryOffset()` and `buildIndex()` which can execute concurrently. `StringBuilder` is not thread-safe — concurrent `setLength(0)` + `append()` calls will produce corrupted keys or throw `StringIndexOutOfBoundsException`.

**Suggested fix:** Use `ThreadLocal<StringBuilder>` instead:
```java
private static final ThreadLocal<StringBuilder> KEY_BUILDER = ThreadLocal.withInitial(() -> new StringBuilder(128));

private String buildKey(final String topic, final String key) {
    StringBuilder sb = KEY_BUILDER.get();
    sb.setLength(0);
    return sb.append(topic).append(\"#\").append(key).toString();
}
```

### Other Findings
- **[Info]** `QueueOffsetOperator` AtomicLong migration — Clean. Eliminates boxing/unboxing on every offset update. The `ConcurrentHashMap<String, AtomicLong>` pattern is correct.
- **[Info]** `AppendMessageResult` new constructor — Good, avoids unused StringBuilder parameter.
- **[Info]** `ConsumeQueue.topicQueueKey` caching — Good optimization, avoids repeated string concatenation.

---

store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
```diff
@@ -215,10 +216,12 @@ public QueryOffsetResult queryOffset(String topic, String key, int maxNum, long
     }
 
     private String buildKey(final String topic, final String key) {
-        return topic + "#" + key;
+        reusableKeyBuilder.setLength(0);
```
RockteMQ-AI: **[Critical] Thread-safety bug**: `reusableKeyBuilder` is an instance field shared across threads, but `buildKey()` is called from `queryOffset()` which can be invoked concurrently by multiple admin/consumer query threads. StringBuilder is NOT thread-safe — concurrent calls to `setLength(0)` and `append()` will interleave, producing corrupted keys like `topic1#key2` instead of `topic1#key1`.

This will cause index lookups to return wrong results or miss entries entirely.

**Fix**: Use `ThreadLocal<StringBuilder>` instead of a shared instance field, or revert to string concatenation (the JIT compiler optimizes it well).

store/src/main/java/org/apache/rocketmq/store/timer/TimerWheel.java
```diff
@@ -128,17 +130,25 @@ public void flush() {
         if (mappedByteBuffer == null) {
             return;
         }
+        if (!dirty) {
```
RockteMQ-AI, line 133: **[Critical] Thread-safety bug**: The `dirty` flag is instance-level (shared across threads), but `localBuffer` is `ThreadLocal<ByteBuffer>` (per-thread). This creates a race condition:

1. Thread A calls `putSlot()` → sets `dirty = true`
2. Thread B calls `flush()` → sees `dirty = true` → flushes Thread B's local buffer
3. Thread B's buffer may be uninitialized (all zeros) or contain stale data
4. Flushing zeros over the mapped file corrupts the timer wheel

**Fix**: Make `dirty` a `ThreadLocal<Boolean>` to match the per-thread buffer, or use a synchronized block around flush operations.

store/src/main/java/org/apache/rocketmq/store/queue/QueueOffsetOperator.java
```diff
@@ -123,10 +132,18 @@ public void setLmqTopicQueueTable(ConcurrentMap<String, Long> lmqTopicQueueTable
     }
 
     public ConcurrentMap<String, Long> getTopicQueueTable() {
-        return topicQueueTable;
+        ConcurrentMap<String, Long> snapshot = new ConcurrentHashMap<>(this.topicQueueTable.size() * 2);
```
RockteMQ-AI: **[Warning] Behavioral change**: `getTopicQueueTable()` now returns a snapshot copy instead of the live map. While this is safer for thread safety, it's a behavioral change that could affect callers who expected to modify the map directly.

If any code relies on modifying the returned map and seeing those changes reflected in the internal state, it will break silently.

**Recommendation**: Document this change in the method's Javadoc, or consider whether callers actually need live access.

RockteMQ-AI (changes requested):
### Summary
This PR optimizes auxiliary component allocation in the store layer — reducing StringBuilder allocations in IndexService, switching QueueOffsetOperator to AtomicLong for better concurrency, and adding a dirty-flag optimization to TimerWheel flush. The intent is sound, but two of the optimizations introduce **critical thread-safety bugs** that must be fixed before merge.

### Findings
- **[Critical]** IndexService.java:219 — `reusableKeyBuilder` is a shared instance field used by `buildKey()` which can be called concurrently. StringBuilder is not thread-safe; concurrent `setLength(0)` + `append()` will corrupt keys.
- **[Critical]** TimerWheel.java:133 — `dirty` flag is instance-level but `localBuffer` is ThreadLocal. Thread A sets dirty=true, Thread B flushes its own (possibly zeroed) buffer, corrupting the timer wheel.
- **[Warning]** QueueOffsetOperator.java:135 — `getTopicQueueTable()` now returns a snapshot copy instead of the live map. Behavioral change should be documented.

### Suggestions
1. For IndexService: use `ThreadLocal<StringBuilder>` or revert to string concatenation (JIT optimizes it well)
2. For TimerWheel: make `dirty` a `ThreadLocal<Boolean>` to match the per-thread buffer, or synchronize flush
3. For QueueOffsetOperator: add Javadoc noting the snapshot semantics of `getTopicQueueTable()`

---

store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
wang-jiahua: Confirmed — `buildKey` is reachable from `queryOffset` (query thread pools) and from concurrent dispatch when `enableBuildConsumeQueueConcurrently` is on, so a shared instance StringBuilder can interleave. Switched to `ThreadLocal<StringBuilder>` in 43be17b98, which keeps the allocation win on the hot build path and is safe on every thread.

store/src/main/java/org/apache/rocketmq/store/timer/TimerWheel.java
wang-jiahua: The per-thread-buffer scenario doesn't apply here: `localBuffer` holds `byteBuffer.duplicate()`, and duplicates share the single underlying direct buffer — the ThreadLocal only isolates position/limit cursors, so there is exactly one copy of the wheel data and no thread can flush stale/uninitialized content. That said, this comment made me re-examine the flag ordering and there was a real lost-update window: clearing `dirty` after the diff could swallow a concurrent `putSlot`'s mark, leaving the last write unflushed until the next putSlot. Fixed in 43be17b98 by clearing the flag before diffing — any interleaving now causes at most one extra flush, never a missed one.

store/src/main/java/org/apache/rocketmq/store/queue/QueueOffsetOperator.java
wang-jiahua: Agreed, and on reflection the risk is worse than a doc issue: callers (including extensions that reach this via reflection, e.g. recovery flows pre-populating offsets) may rely on mutating the returned map and would break silently with a snapshot. I reverted the QueueOffsetOperator changes entirely in 43be17b98 — `getTopicQueueTable()` returns the live map again and the AtomicLong conversion is dropped from this PR.

RockteMQ-AI (approved):
### Summary
Performance optimization PR addressing auxiliary component allocation. New commits since the previous review address the requested changes with clean implementations.

### Changes Reviewed
- **ThreadLocal StringBuilder reuse** in  — avoids per-call allocation on hot path
- **Dirty flag optimization** in  — skips flush when no writes occurred since last flush, with proper race condition handling
- **Long-aligned comparison** for flush — efficient 8-byte comparison instead of byte-by-byte
- **Integer → int** in test — avoids unnecessary autoboxing

All changes look correct and well-targeted. The dirty flag pattern in TimerWheel properly handles the race between / and .

LGTM.

---
*Automated review by RocketMQ-AI*

PR #10531, opened 2026-06-19
[ISSUE #10529] Fix incorrect queue permissions in gRPC QueryRouteResponse

itxaiohanglover:
## What Changed

Fixed the permission assignment logic in `RouteActivity#genMessageQueueFromQueueData` so that each `MessageQueue`'s permission is derived from its queue id rather than being assigned by grouped permission counts.

## Why

When the gRPC proxy builds `QueryRouteResponse`, the old implementation calculated the number of read-only, write-only, and read-write queues first, then appended them in groups (READ, WRITE, READ_WRITE) while increasing the queue id. This preserved the count of each permission type but assigned the permission to the wrong queue id.

For example, with `readQueueNums = 4`, `writeQueueNums = 8`, and `PERM_READ | PERM_WRITE`:

- **Expected:** queue ids `0..3` -> `READ_WRITE`, queue ids `4..7` -> `WRITE`
- **Before fix:** queue ids `0..3` -> `WRITE`, queue ids `4..7` -> `READ_WRITE`

Since `MessageQueue.id` is part of the queue identity, the permission must match the actual queue id. The internal route selection logic also builds read/write queues by iterating queue ids from `0` to `readQueueNums - 1` or `writeQueueNums - 1`, confirming that the permission should be derived per queue id.

## How

Instead of grouping by permission type, iterate queue ids from `0` to `max(readQueueNums, writeQueueNums) - 1` and derive the permission per id:
- `id < readQueueNums && id < writeQueueNums` -> `READ_WRITE`
- `id < writeQueueNums` only -> `WRITE`
- `id < readQueueNums` only -> `READ`

The write-only, read-only, and no-access branches remain unchanged.

## Test Plan

- [x] Updated `RouteActivityTest#testGenPartitionFromQueueData` to verify the per-queue-id permission mapping for `readQueueNums = 4, writeQueueNums = 8` (ids `0..3` -> `READ_WRITE`, ids `4..7` -> `WRITE`)
- [x] Added test case for `readQueueNums = 8, writeQueueNums = 4` (ids `0..3` -> `READ_WRITE`, ids `4..7` -> `READ`)
- [x] All existing test cases (equal read/write, read-only, write-only, no-access) continue to pass
- [ ] CI passes (`maven-compile`, `license-checker`)

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
@@ -242,58 +242,59 @@ public class RouteActivity extends AbstractMessagingActivity {
         TopicMessageType topicMessageType, Broker broker) {
         List<MessageQueue> messageQueueList = new ArrayList<>();
 
-        int r = 0;
-        int w = 0;
-        int rw = 0;
-        int n = 0;
-        if (PermName.isWriteable(queueData.getPerm()) && PermName.isReadable(queueData.getPerm())) {
-            rw = Math.min(queueData.getWriteQueueNums(), queueData.getReadQueueNums());
-            r = queueData.getReadQueueNums() - rw;
-            w = queueData.getWriteQueueNums() - rw;
-        } else if (PermName.isWriteable(queueData.getPerm())) {
-            w = queueData.getWriteQueueNums();
-        } else if (PermName.isReadable(queueData.getPerm())) {
-            r = queueData.getReadQueueNums();
+        int readQueueNums = queueData.getReadQueueNums();
+        int writeQueueNums = queueData.getWriteQueueNums();
+        boolean isReadable = PermName.isReadable(queueData.getPerm());
+        boolean isWriteable = PermName.isWriteable(queueData.getPerm());
+
+        if (isReadable && isWriteable) {
+            // Derive the permission per queue id so that it matches the actual queue id.
+            // Queue id < readQueueNums means readable, id < writeQueueNums means writable.
+            int totalQueueNums = Math.max(readQueueNums, writeQueueNums);
+            for (int queueId = 0; queueId < totalQueueNums; queueId++) {
+                Permission permission;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
This PR fixes a real bug in `RouteActivity#genMessageQueueFromQueueData` where queue permissions were assigned to wrong queue IDs in the gRPC `QueryRouteResponse`. The old code grouped queues by permission type (READ → WRITE → READ_WRITE) and assigned sequential IDs within each group, causing a mismatch between the actual queue ID and its permission.

### Findings

- **[Info]** `RouteActivity.java:245-297` — The refactored logic correctly derives permission per queue ID by iterating from `0` to `max(readQueueNums, writeQueueNums) - 1`. This matches the internal routing logic that builds read/write queues by iterating queue IDs from `0`.

- **[Info]** `RouteActivity.java:257-264` — The three-way branching (`READ_WRITE` / `WRITE` / `READ`) within the combined readable+writable case is clean and correct. The queue ID directly determines the permission, which is the right approach.

- **[Info]** `RouteActivity.java:284-286` — The no-access branch preserves the original `max(1, max(writeQueueNums, readQueueNums))` behavior. This is fine for backward compatibility, though one could argue that `max(1, ...)` for a no-permission case is already questionable — but that is out of scope for this fix.

- **[Info]** `RouteActivityTest.java` — Good test coverage. The new assertions verify both queue count and the specific permission-to-ID mapping for asymmetric cases (4R/8W and 8R/4W), which would have caught the original bug.

### Duplicate PR Notice
Note that PR #10530 by @Kris20030907 targets the same issue (#10529) with a similar fix scope. The maintainers may want to coordinate between the two contributions.

### Suggestions
- Minor: The `MessageQueue.newBuilder()...` pattern is repeated across all branches. A small helper method like `buildQueue(broker, topic, queueId, permission, topicMessageType)` could reduce duplication, but this is a style preference and not blocking.

### Verdict
The fix is correct, well-explained, and properly tested. **LGTM.** 👍

---

itxaiohanglover:
Closing this as a duplicate of #10530, which was submitted first and addresses the same issue with the same fix.

PR #10539, opened 2026-06-22
[ISSUE #10537] Cache RPC Attributes and use processTimerElapsedMs in NettyRemotingAbstract

### LINKED ISSUE #10537: [ISSUE #10537] Cache RPC Attributes and use processTimerElapsedMs in NettyRemotingAbstract
NettyRemotingAbstract creates a new AttributesBuilder per RPC. This PR adds a getOrBuildAttributes() cache.

wang-jiahua:

Fixes #10537

## Brief Description

Cache RPC `Attributes` objects in `RemotingMetricsManager` to avoid per-RPC `Attributes.builder().build()` allocation.

- **RemotingMetricsManager**: add `getOrBuildAttributes(reqCode, respCode, isLongPolling, result)` with a long-key cache (35-bit packed key, no String allocation). Falls back to direct build for unknown `result` values.
- **NettyRemotingAbstract**: use cached Attributes from `getOrBuildAttributes()` instead of inline `attributesBuilder.build()` at each metrics recording point.
- **NettyRemotingAbstract**: add `log.isDebugEnabled()` guard for debug log.

## How Did You Test This Change

Server benchmark: 8C30G broker, 1KB msg, 128 queues, sync mode, 256 threads. CI checkstyle + compilation passed.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/metrics/RemotingMetricsManager.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/metrics/RemotingMetricsManager.java
@@ -30,6 +30,7 @@ import io.opentelemetry.sdk.metrics.ViewBuilder;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 import org.apache.rocketmq.common.Pair;
 import org.apache.rocketmq.common.metrics.NopLongHistogram;
@@ -37,14 +38,21 @@ import org.apache.rocketmq.common.metrics.NopLongHistogram;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.HISTOGRAM_RPC_LATENCY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_PROTOCOL_TYPE_KEY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.PROTOCOL_TYPE_REMOTING;
+import org.apache.rocketmq.remoting.common.RemotingHelper;
 
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_IS_LONG_POLLING_KEY;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_REQUEST_CODE_KEY;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_RESPONSE_CODE_KEY;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_RESULT_KEY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_CANCELED;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_ONEWAY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_SUCCESS;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_WRITE_CHANNEL_FAILED;
 
 public class RemotingMetricsManager {
     private LongHistogram rpcLatency = new NopLongHistogram();
     private Supplier<AttributesBuilder> attributesBuilderSupplier;
+    private final ConcurrentHashMap<Long, Attributes> attributesCache = new ConcurrentHashMap<>();
 
     public RemotingMetricsManager() {
     }
@@ -85,6 +93,39 @@ public class RemotingMetricsManager {
         ViewBuilder viewBuilder = View.builder()
             .setAggregation(Aggregation.explicitBucketHistogram(rpcCostTimeBuckets));
         return Lists.newArrayList(new Pair<>(selector, viewBuilder));
+    }
+
+    public Attributes getOrBuildAttributes(int requestCode, int responseCode,
+        boolean isLongPolling, String result) {
+        int resultIdx;
+        if (RESULT_SUCCESS.equals(result)) resultIdx = 0;
+        else if (RESULT_ONEWAY.equals(result)) resultIdx = 1;
+        else if (RESULT_WRITE_CHANNEL_FAILED.equals(result)) resultIdx = 2;
+        else if (RESULT_CANCELED.equals(result)) resultIdx = 3;
+        else resultIdx = -1;
+
+        if (resultIdx < 0 || requestCode < 0 || requestCode > 0xFFFF || responseCode < 0 || responseCode > 0xFFFF) {
+            // Rare shapes (unknown result, or codes outside 16 bits such as negative response
+            // codes) fall back to direct building so distinct inputs can never collide in the cache.
+            return buildAttributes(requestCode, responseCode, isLongPolling, result);
+        }
+
+        long key = ((long) requestCode << 19)
+            | ((long) responseCode << 3)
+            | (isLongPolling ? 4L : 0L)
+            | resultIdx;
+        return attributesCache.computeIfAbsent(key,
+            k -> buildAttributes(requestCode, responseCode, isLongPolling, result));
+    }
+
+    private Attributes buildAttributes(int requestCode, int responseCode,
+        boolean isLongPolling, String result) {
+        return newAttributesBuilder()
+            .put(LABEL_IS_LONG_POLLING_KEY, isLongPolling)
+            .put(LABEL_REQUEST_CODE_KEY, RemotingHelper.getRequestCodeDesc(requestCode))
+            .put(LABEL_RESPONSE_CODE_KEY, RemotingHelper.getResponseCodeDesc(responseCode))
+            .put(LABEL_RESULT_KEY, result)
+            .build();
     }
 
     public String getWriteAndFlushResult(Future<?> future) {
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -66,10 +66,9 @@ import org.apache.rocketmq.remoting.protocol.RemotingCommand;
 import org.apache.rocketmq.remoting.protocol.RemotingSysResponseCode;
 import org.apache.rocketmq.remoting.protocol.ResponseCode;
 
-import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_IS_LONG_POLLING;
-import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_REQUEST_CODE;
-import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_RESPONSE_CODE;
-import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_RESULT;
+import io.opentelemetry.api.common.Attributes;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_REQUEST_CODE_KEY;
+import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.LABEL_RESULT_KEY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_ONEWAY;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_PROCESS_REQUEST_FAILED;
 import static org.apache.rocketmq.remoting.metrics.RemotingMetricsConstant.RESULT_WRITE_CHANNEL_FAILED;
@@ -235,19 +234,14 @@ public abstract class NettyRemotingAbstract {
         if (response == null) {
             return;
         }
-        final AttributesBuilder attributesBuilder;
-        if (remotingMetricsManager != null) {
-            attributesBuilder = remotingMetricsManager.newAttributesBuilder();
-            attributesBuilder.put(LABEL_IS_LONG_POLLING, request.isSuspended())
-                .put(LABEL_REQUEST_CODE, RemotingHelper.getRequestCodeDesc(request.getCode()))
-                .put(LABEL_RESPONSE_CODE, RemotingHelper.getResponseCodeDesc(response.getCode()));
-        } else {
-            attributesBuilder = null;
-        }
+        final int requestCode = request.getCode();
+        final int responseCode = response.getCode();
+        final boolean isLongPolling = request.isSuspended();
         if (request.isOnewayRPC()) {
-            if (attributesBuilder != null) {
-                attributesBuilder.put(LABEL_RESULT, RESULT_ONEWAY);
-                remotingMetricsManager.getRpcLatency().record(request.getProcessTimer().elapsed(TimeUnit.MILLISECONDS), attributesBuilder.build());
+            if (remotingMetricsManager != null) {
+                Attributes attrs = remotingMetricsManager.getOrBuildAttributes(
+                    requestCode, responseCode, isLongPolling, RESULT_ONEWAY);
+                remotingMetricsManager.getRpcLatency().record(request.getProcessTimer().elapsed(TimeUnit.MILLISECONDS), attrs);
             }
             return;
         }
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (commented):

### Summary
Caches per-request-code OpenTelemetry `Attributes` objects and replaces `Stopwatch.elapsed(TimeUnit.MILLISECONDS)` with `processTimerElapsedMs()` (nanoTime-based) across the remoting layer. Also adds cached `Constructor` lookup for custom header classes and reduces default HashMap capacity. Multiple coordinated micro-optimizations targeting the broker hot path.

### Findings

- **[Warning]** `NettyRemotingServer.java:432-433` — Changed `log.info("channel writable false. host:{}", ...)` to `log.debug(...)`. This is a behavioral change: production users who relied on INFO-level channel writability warnings for backpressure monitoring will lose this visibility. If this was intentional (to reduce log noise), it is fine, but worth confirming.

- **[Info]** `RemotingMetricsManager.java` — The attributes cache uses a `ConcurrentHashMap` with a volatile fast-path (`lastAttrsCacheKey` / `lastCachedAttrs`). The non-atomic read-then-write pattern means two threads with the same cache key could both rebuild `Attributes`, but `computeIfAbsent` ensures correctness. The volatile fast-path is a reasonable optimization for the common case where the same request type dominates.

- **[Info]** `RemotingMetricsManager.java:214-216` — Cache key encoding: `(requestCode << 19) | (responseCode << 3) | (isLongPolling ? 4 : 0) | resultIndex`. This assumes requestCode fits in 16 bits (max 65535) and responseCode fits in 16 bits. Both are safe given current `ResponseCode` and `RequestCode` values. Consider adding a comment documenting these bit-width assumptions for future maintainers.

- **[Info]** `RemotingCommand.java` — Replacing `Stopwatch processTimer` with `long processTimeStartNanos` eliminates one object allocation per request. The new `processTimerElapsedMs()` method correctly computes elapsed milliseconds from `System.nanoTime()`. The deprecated `getProcessTimer()` now returns a fresh `Stopwatch` which would give incorrect elapsed time if called — the `@Deprecated` annotation is appropriate.

- **[Info]** `RemotingCommand.java:231-256` — Cached `Constructor` lookup via `CONSTRUCTOR_CACHE` (ConcurrentHashMap with `computeIfAbsent`). Thread-safe. The `setAccessible(true)` call is necessary for non-public constructors. Clean.

- **[Info]** `RemotingCommand.java:341` — HashMap initial capacity reduced from 256 to 16. With default load factor 0.75, this supports 12 entries before resizing — sufficient for typical header field counts. Good optimization.

- **[Info]** `TopicQueueMappingContext.java` — Added `EMPTY` static singleton to avoid allocating empty context objects. Clean.

### Suggestions

1. The log level change in `NettyRemotingServer.java` (channel writability) is the most notable behavioral change. If intentional, no action needed.

2. Consider adding a brief comment on the cache key encoding in `RemotingMetricsManager` documenting the bit-width assumptions (requestCode ≤ 16 bits, responseCode ≤ 16 bits).

Overall: Well-structured set of micro-optimizations with consistent patterns. The Stopwatch → nanoTime replacement is a clean improvement.

---

RockteMQ-AI (commented):

### Summary
PR squashed into 2 commits. Core approach unchanged: cache OpenTelemetry `Attributes` objects and replace `Stopwatch` with `long processTimerNanos` to eliminate per-RPC allocations. New `log.isDebugEnabled()` guard added for debug logging.

### CI Status
**CI is failing** — SpotBugs reports 4 `ES_COMPARING_PARAMETER_STRING_WITH_EQ` violations in `RemotingMetricsManager.getOrBuildAttributes()` (lines 102-105). The `result == RESULT_SUCCESS` style comparisons on a method parameter trigger SpotBugs even though the values are interned constants.

### Findings

- **[Critical]** `RemotingMetricsManager.java:102-105` — String `==` comparison on method parameter `result`. Fix: use `.equals()` or add `@SuppressFBWarnings("ES_COMPARING_PARAMETER_STRING_WITH_EQ")` with justification that all callers pass interned constant strings. The `.equals()` approach is safer and preferred.

- **[Info]** `RemotingMetricsManager.java` — Cache key encoding `(requestCode << 19) | (responseCode << 3) | (isLongPolling ? 4 : 0) | resultIdx` is correct for current value ranges. The `lastAttrsCacheKey`/`lastCachedAttrs` volatile fast-path is a reasonable optimization for the dominant-case pattern.

- **[Info]** `RemotingCommand.java` — `processTimerElapsedMs()` correctly computes elapsed ms from `System.nanoTime()`. The deprecated `getProcessTimer()` returning a fresh `Stopwatch` is a safe backward-compat shim.

- **[Info]** `NettyRemotingAbstract.java` — `log.isDebugEnabled()` guard avoids string concatenation overhead when debug is disabled. Good.

### Suggestions
1. Fix the SpotBugs violations to unblock CI — this is the only blocker.
2. Consider adding a brief comment on the cache key bit-width assumptions for future maintainers.

---

RockteMQ-AI (commented):

### Summary
New squashed commit. Core approach unchanged: cache OpenTelemetry `Attributes` via `ConcurrentHashMap` + volatile fast-path, and replace `Stopwatch` with `long processTimerNanos`.

### CI Status
**CI is still failing** — SpotBugs reports 4 `ES_COMPARING_PARAMETER_STRING_WITH_EQ` violations in `RemotingMetricsManager.getOrBuildAttributes()`:

```
RemotingMetricsManager.java:102 — result == RESULT_SUCCESS
RemotingMetricsManager.java:103 — result == RESULT_ONEWAY
RemotingMetricsManager.java:104 — result == RESULT_WRITE_CHANNEL_FAILED
RemotingMetricsManager.java:105 — result == RESULT_CANCELED
```

**This was flagged in the previous review but not yet fixed.** SpotBugs treats `==` comparison on a method parameter as a bug even when the constants are interned, because the parameter type is `String` and callers could pass non-interned strings.

### Fix Required

Replace `==` with `.equals()` in `getOrBuildAttributes()`:

```java
public Attributes getOrBuildAttributes(int requestCode, int responseCode,
    boolean isLongPolling, String result) {
    int resultIdx;
    if (RESULT_SUCCESS.equals(result)) resultIdx = 0;
    else if (RESULT_ONEWAY.equals(result)) resultIdx = 1;
    else if (RESULT_WRITE_CHANNEL_FAILED.equals(result)) resultIdx = 2;
    else if (RESULT_CANCELED.equals(result)) resultIdx = 3;
    else resultIdx = -1;
    // ... rest unchanged
```

### Findings

- **[Critical]** `RemotingMetricsManager.java:102-105` — **Not fixed.** Change `result == CONSTANT` to `CONSTANT.equals(result)` to resolve SpotBugs `ES_COMPARING_PARAMETER_STRING_WITH_EQ`. This is blocking CI.

- **[Info]** The `ConcurrentHashMap<Long, Attributes>` + volatile fast-path caching strategy is sound. The `lastAttrsCacheKey` / `lastCachedAttrs` volatile pair avoids HashMap lookup on the hottest path. Good optimization.

- **[Info]** The bit-packing key `((long) requestCode << 19) | ((responseCode & 0xFFFF) << 3) | ...` is correct given the value ranges (requestCode fits in 13 bits, responseCode in 16 bits).

---

RockteMQ-AI (approved):
## Review by github-manager-bot (Re-review after new commit `611d8ea`)

### Summary
New commit confirmed — OpenTelemetry `Attributes` are now cached per request-code/response-code combination using a `ConcurrentHashMap`. Debug log string concatenation is also guarded with `log.isDebugEnabled()`.

### Assessment
- **[Correctness]** The bit-packing key `(requestCode << 19) | (responseCode << 3) | flags` correctly encodes the attribute combination. The `resultIdx` mapping is consistent with the string constants. Cache invalidation is not needed since the set of request/response codes is finite and small.
- **[Performance]** Eliminates per-request `Attributes` object allocation and string formatting on the metrics hot path. The `isDebugEnabled()` guard avoids unnecessary string concatenation. Good optimization.
- **[Thread Safety]** `ConcurrentHashMap.computeIfAbsent` is used correctly for thread-safe lazy initialization.
- **[Info]** One minor note: the cache is unbounded. In practice the number of distinct (requestCode, responseCode, isLongPolling, result) combinations is small and bounded, so this is not a concern.

LGTM — all CI checks pass. Clean performance improvement. ✅

---

remoting/src/main/java/org/apache/rocketmq/remoting/metrics/RemotingMetricsManager.java
```diff
@@ -87,6 +95,37 @@ public List<Pair<InstrumentSelector, ViewBuilder>> getMetricsView() {
...
+        else if (RESULT_ONEWAY.equals(result)) resultIdx = 1;
+        else if (RESULT_WRITE_CHANNEL_FAILED.equals(result)) resultIdx = 2;
+        else if (RESULT_CANCELED.equals(result)) resultIdx = 3;
+        else resultIdx = -1;
+
+        if (resultIdx < 0) {
+            return buildAttributes(requestCode, responseCode, isLongPolling, result);
+        }
+
+        long key = ((long) requestCode << 19)
```
RongtongJin, line 113: The caching logic looks reasonable, but the packed key is now part of the correctness boundary for remoting metrics. Could you add focused tests to verify that `requestCode`, `responseCode`, `isLongPolling`, and `result` all produce distinct cached Attributes where expected, and that unknown result values still use the non-cached path? This would reduce the risk of future metric label mix-ups, especially since the current patch coverage is very low.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
```diff
@@ -235,19 +234,14 @@ public static void writeResponse(Channel channel, RemotingCommand request, @Null
...
+        final int responseCode = response.getCode();
+        final boolean isLongPolling = request.isSuspended();
         if (request.isOnewayRPC()) {
-            if (attributesBuilder != null) {
-                attributesBuilder.put(LABEL_RESULT, RESULT_ONEWAY);
-                remotingMetricsManager.getRpcLatency().record(request.getProcessTimer().elapsed(TimeUnit.MILLISECONDS), attributesBuilder.build());
+            if (remotingMetricsManager != null) {
+                Attributes attrs = remotingMetricsManager.getOrBuildAttributes(
+                    requestCode, responseCode, isLongPolling, RESULT_ONEWAY);
+                remotingMetricsManager.getRpcLatency().record(request.getProcessTimer().elapsed(TimeUnit.MILLISECONDS), attrs);
```
RongtongJin, line 244: The PR description mentions switching to `processTimerElapsedMs`, but this diff still records latency using `request.getProcessTimer().elapsed(TimeUnit.MILLISECONDS)` in these paths. If the timing change is no longer part of this PR, please update the description so reviewers can evaluate the actual scope.

RockteMQ-AI (approved):

### Summary
Re-review triggered by new commit `84e42d4a`. This PR caches `Attributes` objects in `RemotingMetricsManager` to avoid per-RPC allocation on the hot path, and guards debug log string concatenation with `isDebugEnabled()`.

### Findings

- **[Info] remoting/.../RemotingMetricsManager.java** — The `attributesCache` (`ConcurrentHashMap<Long, Attributes>`) has no eviction policy. In practice, RocketMQ has a bounded set of request/response codes so the cache should stabilize at a manageable size. However, consider documenting the expected upper bound or adding a size guard (e.g., `if (cache.size() > THRESHOLD) return buildAttributes(...)`) as a safety net against unexpected growth.

- **[Info] remoting/.../RemotingMetricsManager.java:112-115** — Bit-packing key construction: `requestCode` is shifted by 19 bits. This assumes `requestCode` fits in 19 bits (0–524287). RocketMQ request codes are well within this range, but a defensive comment documenting the assumption would help future maintainers.

- **[Info] remoting/.../NettyRemotingAbstract.java** — Good improvement: extracting `processTimerElapsedMs` and using it consistently across `processResponseCommand`, `processRequestCommand`, and `processResponseCommandAsync`. The `isDebugEnabled()` guard before string concatenation is a solid optimization for the hot path.

- **[Info] remoting/.../RemotingMetricsConstant.java** — Renaming labels to include `_KEY` suffix improves clarity. Clean migration.

### Positive Aspects
- Well-structured `getOrBuildAttributes` with clear result-index mapping
- Comprehensive test coverage including cache hit verification and unknown-result bypass
- Thread-safe via `ConcurrentHashMap.computeIfAbsent`
- Good deduplication of metrics recording across writeResponse methods

LGTM 👍

---

remoting/src/main/java/org/apache/rocketmq/remoting/metrics/RemotingMetricsManager.java
wang-jiahua: Added `RemotingMetricsManagerTest` with 7 test cases in the latest push:

- `testCacheHitSameArgs`: same args return cached instance
- `testDifferentRequestCode` / `testDifferentResponseCode` / `testDifferentIsLongPolling` / `testDifferentResult`: each parameter change produces distinct cached Attributes
- `testUnknownResultNotCached`: unknown result values use the non-cached path (new instance each time)
- `testKnownResultsCached`: all 4 known result types (SUCCESS, ONEWAY, WRITE_CHANNEL_FAILED, CANCELED) are properly cached

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
wang-jiahua: Updated the PR description — removed the `processTimerElapsedMs` mention. This PR only adds the Attributes cache and the `isDebugEnabled()` guard. The timing change is not part of this PR.

RockteMQ-AI (commented):

### Summary
Caches metrics `Attributes` objects in a `ConcurrentHashMap` keyed by a packed long value combining requestCode, responseCode, isLongPolling, and result index. Also introduces `processTimerElapsedMs()` on `RemotingCommand` to avoid `TimeUnit` imports and Stopwatch overhead.

### Findings
- **[Warning]** Key packing in `getOrBuildAttributes()` — The key packs requestCode (16 bits) + responseCode (15 bits) + isLongPolling (1 bit) + resultIdx (4 bits) into a long. This assumes response codes fit in 15 bits (max 32767). If RocketMQ ever defines response codes above 32767, there will be silent key collisions.
- **[Info]** `ConcurrentHashMap` for attributes cache — Appropriate for concurrent access. Consider adding a size bound or eviction policy to prevent unbounded growth if the cardinality of (requestCode, responseCode, isLongPolling, result) combinations is high.
- **[Info]** `processTimerElapsedMs()` — Clean abstraction. Avoids repeated `TimeUnit.MILLISECONDS` import and call pattern across multiple processors.

### Suggestions
- Add a comment documenting the key bit layout and the assumption that response codes fit in 15 bits.
- Consider using `Caffeine` or `Guava` cache with a max size if the cardinality concern is real.

---

remoting/src/main/java/org/apache/rocketmq/remoting/metrics/RemotingMetricsManager.java
```diff
@@ -87,6 +95,37 @@ public List<Pair<InstrumentSelector, ViewBuilder>> getMetricsView() {
...
+            | ((long) (responseCode & 0xFFFF) << 3)
+            | (isLongPolling ? 4L : 0L)
+            | resultIdx;
+        return attributesCache.computeIfAbsent(key,
+            k -> buildAttributes(requestCode, responseCode, isLongPolling, result));
```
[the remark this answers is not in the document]
wang-jiahua: Good catch — negative response codes would collide after masking, and a negative requestCode would sign-extend over the whole key. Fixed in 4832ff0c5: the cache is now only used when both codes fit in 0..0xFFFF (and the result is known); anything else falls back to building attributes directly, so distinct inputs can never collide.

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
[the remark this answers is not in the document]
wang-jiahua: The code is correct as-is: `getProcessTimer().elapsed(TimeUnit.MILLISECONDS)` is allocation-free and not deprecated, so a separate `processTimerElapsedMs()` accessor would add API surface for no gain. The issue text was stale (it described an earlier plan) — I've updated #10537 to drop that sentence.

RockteMQ-AI (approved):
### Summary
Performance optimization for RPC metrics: caches `Attributes` objects in `RemotingMetricsManager` using a ConcurrentHashMap, avoiding per-RPC allocation. Also switches to `processTimerElapsedMs` for more accurate timing.

New commits since the previous review properly implement the caching pattern with correct thread safety. The changes are well-scoped and the performance improvement is meaningful for high-throughput scenarios.

LGTM.

---
*Automated review by RocketMQ-AI*

PR #10542, opened 2026-06-22
[ISSUE #10541] Avoid duplicate subscription change logs

### LINKED ISSUE #10541: [Bug] Repeated subscription changed logs for unchanged gRPC subscriptions

- [x] I have confirmed this issue has not been reported before.
- [x] I have searched the existing issues.

RocketMQ proxy with gRPC consumers and proxy heartbeat sync enabled.

### RocketMQ version

develop branch

### Describe the Bug

For gRPC consumers, `ClientActivity.registerConsumer()` rebuilds the subscription data set from the client `Settings` on each registration/heartbeat path. `FilterAPI.build()` creates a new `SubscriptionData`, whose default `subVersion` is initialized with `System.currentTimeMillis()`.

When the subscription topic and filter expression are unchanged, the newly built `SubscriptionData` still has a newer `subVersion`. `ConsumerGroupInfo.updateSubscription()` currently logs `subscription changed` whenever the incoming `subVersion` is greater than the stored one, so the broker/proxy can repeatedly log subscription changes even though the subscription content is identical.

This is especially visible on the proxy internal heartbeat sync path, where the log thread can be `HeartbeatSyncer`.

### Steps to Reproduce

1. Start RocketMQ proxy in cluster mode.
2. Register a gRPC consumer with a stable subscription.
3. Let the client continue sending heartbeat/settings updates or let proxy heartbeat sync propagate the registration data.
4. Observe repeated `subscription changed` logs with identical topic/filter content but different `subVersion` values.

The broker/proxy should only log `subscription changed` when the actual subscription content changes.

### What Did You See Instead?

The log is printed repeatedly because only `subVersion` changes.

### Additional Context

The repeated log is mostly noise and internal sync overhead, but it can make operational logs misleading and noisy for stable gRPC consumers.

qianye1001:
ConsumerGroupInfo logs a subscription change whenever it receives a newer subVersion, even when the topic and filter are unchanged. This produces misleading, repeated logs during gRPC heartbeat/settings registration and version-only updates from other clients.

Compare subscription content before logging a change, while still accepting and storing newer versions. Keep the existing subscription equality and consumer notification behavior unchanged.

Closes #10541

### Validation

- Added coverage for silent version-only updates, rejection of stale updates, real TAG/SQL changes, and unchanged active-consumer logging behavior.
- JDK 11: 38 tests passed, with Checkstyle and SpotBugs checks passing.
- `mvn -B -pl proxy -am -Dtest=SubscriptionDataTest,ConsumerGroupInfoTest,ConsumerManagerTest,ClientActivityTest -DfailIfNoTests=false test`

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerGroupInfo.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/ConsumerGroupInfo.java
@@ -185,7 +185,7 @@ public class ConsumerGroupInfo {
                         sub.toString());
                 }
             } else if (sub.getSubVersion() > old.getSubVersion()) {
-                if (this.consumeType == ConsumeType.CONSUME_PASSIVELY) {
+                if (this.consumeType == ConsumeType.CONSUME_PASSIVELY && !old.equalsIgnoreSubVersion(sub)) {
                     log.info("subscription changed, group: {} OLD: {} NEW: {}",
                         this.groupName,
                         old.toString(),
--- a/broker/src/test/java/org/apache/rocketmq/broker/client/ConsumerGroupInfoTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/client/ConsumerGroupInfoTest.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.broker.client;
+
+import java.util.Collections;
+import org.apache.rocketmq.common.constant.LoggerName;
+import org.apache.rocketmq.common.consumer.ConsumeFromWhere;
+import org.apache.rocketmq.common.filter.ExpressionType;
+import org.apache.rocketmq.logging.ch.qos.logback.classic.Level;
+import org.apache.rocketmq.logging.ch.qos.logback.classic.Logger;
+import org.apache.rocketmq.logging.ch.qos.logback.classic.spi.ILoggingEvent;
+import org.apache.rocketmq.logging.ch.qos.logback.core.read.ListAppender;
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (commented):

### Summary
This PR fixes unnecessary `subVersion` refresh in the proxy gRPC path. When `FilterAPI.build()` creates `SubscriptionData`, it always initializes `subVersion` to the current timestamp. For unchanged subscriptions, this causes downstream heartbeat sync and registration logs to falsely appear as subscription changes. The fix looks up the existing `ConsumerGroupInfo` and reuses the old `subVersion` when the subscription content has not actually changed.

### Findings

- **[Info]** `ClientActivity.java:548-551` — `reuseSubscriptionVersion` temporarily sets `subVersion` to the old value for the equality comparison, then restores it if not equal. This is functionally correct but the temporary mutation pattern is slightly fragile. Consider computing equality on the relevant fields (topic, expression type, tags/set) directly instead of relying on `equals()` after mutating `subVersion`. This would make the intent clearer and avoid any risk if `SubscriptionData.equals()` semantics change in the future.

- **[Info]** `ClientActivity.java:537` — `getConsumerGroupInfo(ctx, consumerGroup)` is called once per `buildSubscriptionDataSet` invocation, which is fine. Just confirming this is a lightweight lookup (map-based) and not a heavy operation. If it ever becomes expensive, the call should be hoisted further up.

- **[Info]** `ClientActivity.java:533` — The method signature of `buildSubscriptionDataSet` changed from `(List<SubscriptionEntry>)` to `(ProxyContext, String, List<SubscriptionEntry>)`. Since this is a `protected` method, any subclass overriding it will need to update accordingly. A quick search suggests no overrides in the current codebase, but worth confirming.

### Positive Notes
- Clean, well-scoped fix that addresses the root cause at the proxy gRPC layer.
- Good test coverage with both "unchanged subscription" and "changed subscription" scenarios in `ClientActivityTest`.
- Proper null-safety in `reuseSubscriptionVersion` — handles null `consumerGroupInfo` and null `oldSubscriptionData` gracefully.
- The approach of comparing with the old version and only keeping the new one when content actually changed is the right strategy.

### Suggestions
- Consider adding a brief Javadoc on `reuseSubscriptionVersion` explaining the contract (mutates `subVersion` to old value, restores if subscription content differs).

---

fuyou001 (dismissed):
LGTM

RockteMQ-AI (approved):
LGTM — CI passed, ready for merge.

---

RongtongJin (dismissed):

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits were pushed (HEAD changed from `5f492fdf` to `5a314b2a`). The changes remain clean and focused:

- **`equalsIgnoreSubVersion()`** in `SubscriptionData` correctly compares all content fields (topic, subString, expressionType, tagsSet, codeSet, classFilterMode) while excluding `subVersion`.
- The guard condition in `ConsumerGroupInfo.updateSubscription()` now suppresses misleading "subscription changed" logs when only the version number advances without actual content changes.
- Pull consumer behavior is preserved (never logs content changes).
- Tests are comprehensive: covering silent version-only updates, real TAG/SQL changes, active consumer logging, null handling, and expression type differences.

No issues found. LGTM.

---
*Automated review by github-manager (RocketMQ-AI)*

PR #10544, opened 2026-06-23
[ISSUE #10543] resolve data race in ServiceThread wakeup mechanism

### LINKED ISSUE #10543: [Bug] ServiceThread#wakeup() can be lost when it races with waitForRunning()'s waitPoint.reset()

OS: any

### RocketMQ version

5.x

### JDK Version

JDK 8

### Describe the Bug

`ServiceThread#wakeup()` and `ServiceThread#waitForRunning()` coordinate through an `AtomicBoolean hasNotified` and a resettable latch `CountDownLatch2 waitPoint`:

```java
public void wakeup() {
    if (hasNotified.compareAndSet(false, true)) {
        waitPoint.countDown(); // notify
    }
}

protected void waitForRunning(long interval) {
    if (hasNotified.compareAndSet(true, false)) { // fast path
        this.onWaitEnd();
        return;
    }
    waitPoint.reset();                            // <-- race point
    try {
        waitPoint.await(interval, TimeUnit.MILLISECONDS);
        log.error("Interrupted", e);
        Thread.currentThread().interrupt();
    } finally {
        hasNotified.set(false);
        this.onWaitEnd();
    }
}
```

There is a race between `waitPoint.countDown()` in `wakeup()` and `waitPoint.reset()` in `waitForRunning()`. If a `wakeup()` lands after the fast-path CAS has already returned `false` but before `waitPoint.reset()` executes, the `countDown()` is overwritten by `reset()`, and the worker then blocks on `await(interval)` for the full interval (default `1000ms`) even though it was already notified. The wakeup is effectively swallowed.

### Steps to Reproduce

Interleaving of worker thread A (running `waitForRunning(1000)`) and caller thread B (calling `wakeup()`). Initial state: `hasNotified=false`, `waitPoint` count = 1.

| Step | Thread A (`waitForRunning`) | Thread B (`wakeup`) | State |
|------|------------------------------|----------------------|-------|
| 1 | `hasNotified.CAS(true,false)` → `false`, skip fast path | | `hasNotified=false`, count=1 |
| 2 | (about to call `reset()`) | `hasNotified.CAS(false,true)` → `true`; `waitPoint.countDown()` | `hasNotified=true`, count=0 |
| 3 | `waitPoint.reset()` → count restored to 1 (**countDown discarded**) | | `hasNotified=true`, count=1 |
| 4 | `waitPoint.await(1000)` blocks; no further `countDown` is coming | | blocks ~1000ms |
| 5 | after timeout: `finally` sets `hasNotified=false`, `onWaitEnd()` | | wakeup lost |

A deterministic reproduction is included in the repo (`ServiceThreadTest`): it injects a `CountDownLatch2` whose `reset()` blocks until a `wakeup()` has been delivered, forcing the exact interleaving above, then asserts the wait returned promptly.

A `wakeup()` that happens concurrently with entering `waitForRunning()` should cause the wait to return (almost) immediately — the thread should react to the notification, not sleep for the whole `interval`.

### What Did You See Instead?

The `countDown()` is wiped out by `reset()`, so `await(interval)` sleeps the full interval (≈1000ms by default). The wakeup is delayed by up to one interval, which adds latency to every service thread relying on `wakeup()` for prompt reaction (flush services, reput/dispatch, HA, etc.). Under bursty notification patterns this manifests as periodic ~1s stalls.

### Additional Context

_No response_

ShadowySpirits:
Fixes #10543

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/ServiceThread.java
+++ b/common/src/main/java/org/apache/rocketmq/common/ServiceThread.java
@@ -18,6 +18,7 @@ package org.apache.rocketmq.common;
 
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.LockSupport;
 
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.logging.org.slf4j.Logger;
@@ -28,9 +29,8 @@ public abstract class ServiceThread implements Runnable {
 
     private static final long JOIN_TIME = 90 * 1000;
 
-    protected Thread thread;
-    protected final CountDownLatch2 waitPoint = new CountDownLatch2(1);
-    protected volatile AtomicBoolean hasNotified = new AtomicBoolean(false);
+    protected volatile Thread thread;
+    protected final AtomicBoolean hasNotified = new AtomicBoolean(false);
     protected volatile boolean stopped = false;
     protected boolean isDaemon = false;
 
@@ -97,32 +97,49 @@ public abstract class ServiceThread implements Runnable {
             return;
         }
         this.stopped = true;
+        // wake up the parked worker so it observes the stop flag promptly
+        wakeup();
         log.info("makestop thread[{}] ", this.getServiceName());
     }
 
     public void wakeup() {
+        if (hasNotified.get()) {
+            return;
+        }
         if (hasNotified.compareAndSet(false, true)) {
-            waitPoint.countDown(); // notify
+            LockSupport.unpark(this.thread); // notify
         }
     }
 
     protected void waitForRunning(long interval) {
+        // Publish the parking thread so wakeup() can target it (also handles restart).
+        this.thread = Thread.currentThread();
+
         if (hasNotified.compareAndSet(true, false)) {
             this.onWaitEnd();
             return;
         }
 
-        //entry to wait
-        waitPoint.reset();
-
-        try {
-            waitPoint.await(interval, TimeUnit.MILLISECONDS);
-        } catch (InterruptedException e) {
-            log.error("Interrupted", e);
-        } finally {
-            hasNotified.set(false);
-            this.onWaitEnd();
+        // LockSupport permits are sticky: an unpark delivered before park makes the next park
+        // return at once, and the loop re-checks hasNotified, so no wakeup can be lost.
+        long deadline = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(interval);
+        while (!hasNotified.get()) {
+            if (stopped) {
+                break;
+            }
+            long remain = deadline - System.nanoTime();
+            if (remain <= 0) {
+                break;
+            }
+            LockSupport.parkNanos(this, remain);
+            if (Thread.interrupted()) {
+                Thread.currentThread().interrupt();
+                break;
+            }
         }
+
+        hasNotified.set(false);
+        this.onWaitEnd();
     }
 
     protected void onWaitEnd() {
--- a/common/src/test/java/org/apache/rocketmq/common/ServiceThreadTest.java
+++ b/common/src/test/java/org/apache/rocketmq/common/ServiceThreadTest.java
@@ -17,10 +17,31 @@
 
 package org.apache.rocketmq.common;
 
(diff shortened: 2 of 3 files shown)


lizhimins (approved):

RockteMQ-AI (commented):

### Summary
This PR fixes a data race in `ServiceThread` where `wakeup()` could be lost when it races with `waitForRunning()`'s `waitPoint.reset()`. The fix replaces the `CountDownLatch2`-based signaling with `LockSupport.park/unpark`, whose sticky permit semantics guarantee that no wakeup signal can be dropped. The `CommitLog` subclass is also cleaned up to use the parent `wakeup()` method instead of duplicating the CAS + countDown pattern inline.

### Findings

- **[Warning]** `ServiceThread.java:101-105` — The `wakeup()` method has a two-step check: an early `hasNotified.get()` guard followed by `compareAndSet(false, true)`. The early `get()` is a valid optimization (avoids an unnecessary CAS when the notification is already pending), but it introduces a subtle semantic: if a concurrent `wakeup()` caller sees `hasNotified == true` via the early `get()`, it skips the `unpark()` call entirely. This is correct **only if** the pending notification will eventually reach the worker. Consider the interleaving where `makeStop()` sets `stopped = true` then calls `wakeup()`, but the early `get()` sees a stale `true` from a **prior** (already consumed) notification that was immediately re-armed by another thread. In this narrow window the worker could be parked without a corresponding `unpark()`, relying solely on the `stopped` flag check inside the park loop to break out. This works correctly today because `stopped` is checked in the loop, but the intent would be clearer if the early `get()` were removed, letting the CAS be the sole arbiter:

  ```java
  public void wakeup() {
      if (hasNotified.compareAndSet(false, true)) {
          LockSupport.unpark(this.thread);
      }
  }
  ```

  This is a minor readability/robustness suggestion, not a correctness blocker.

- **[Info]** `ServiceThread.java:109` — `this.thread = Thread.currentThread()` is written on every entry to `waitForRunning()`. Since `thread` is `volatile`, this is a store barrier on every iteration. In practice this is negligible (the volatile write cost is dwarfed by the park/unpark overhead), but worth noting for hot-path awareness. An alternative would be to set `thread` once in `start()` and only update it on restart, but the current approach is simpler and handles thread restart correctly.

- **[Info]** `ServiceThread.java:31` — Making `thread` `volatile` is correct and necessary for cross-thread visibility between `wakeup()` (caller thread) and `waitForRunning()` (worker thread). Good fix.

- **[Info]** `ServiceThread.java:97-100` — `makeStop()` now calls `wakeup()` after setting `stopped = true`. This is an improvement over the old code, which did not explicitly wake the worker on stop. The worker's park loop checks `stopped` and breaks promptly. Good addition.

- **[Info]** `CommitLog.java:1795` — Replacing the inline `hasNotified.compareAndSet(false, true) + waitPoint.countDown()` with `this.wakeup()` is a clean refactoring. It ensures all subclasses benefit from the fixed wakeup logic and eliminates code duplication.

### Positive Notes
- The LockSupport park/unpark approach is the textbook solution for lost-wakeup problems. Sticky permits eliminate the entire class of race conditions that `CountDownLatch2.reset()` was vulnerable to.
- Excellent test coverage: basic functionality (`testWakeup`, `testWaitForRunning`), latency assertion (`testWakeupDeliveredPromptly`), single-threaded stress (`testNoWakeupLostUnderStress`), and multi-threaded stress (`serviceThreadShouldNotLoseWakeupUnderStress`). The stress tests directly target the race window that the original bug exploited.
- The park loop with deadline + `stopped` check is well-structured and handles interruption correctly (`Thread.interrupted()` restores the interrupt flag).
- The `CommitLog` cleanup is a nice side benefit — removing duplicated signaling logic from subclasses.

### Suggestions
- Consider removing the early `hasNotified.get()` in `wakeup()` to simplify the method and make the CAS the single synchronization point (see Warning above).
- The `waitForRunning()` park loop could benefit from a brief inline comment explaining why `this.thread` is assigned before the CAS (to ensure `unpark()` targets the correct thread even if called concurrently).

---

lizhimins:
image

RockteMQ-AI (approved):
LGTM — CI passed, ready for merge.

---

RockteMQ-AI (approved):
LGTM. All CI checks passed. ✅

PR #10548, opened 2026-06-25
[ISSUE #10545] Support queue selector for transactional sends

### LINKED ISSUE #10545: [Enhancement] Support `MessageQueueSelector` for transactional messages

### Summary

# Support `MessageQueueSelector` for transactional messages

## Background

RocketMQ currently supports selecting a target queue for normal messages through APIs such as:

```java
DefaultMQProducer#send(Message msg, MessageQueueSelector selector, Object arg)
```

This is useful for scenarios that require deterministic queue routing, for example ordered messages, sharding by business key, traffic isolation, canary release, or custom queue lane routing.

However, transactional messages currently do not expose an equivalent queue selection API.

The current transactional send API is:

```java
TransactionMQProducer#sendMessageInTransaction(Message msg, Object arg)
```

Internally, `TransactionMQProducer` delegates to `DefaultMQProducerImpl#sendMessageInTransaction(...)`, which marks the message as a prepared transaction message and then calls the default send path:

```java
sendResult = this.send(msg);
```

As a result, transactional messages go through `sendDefaultImpl(...)`, where the client chooses a queue by `selectOneMessageQueue(...)`. There is no way for users to provide a custom `MessageQueueSelector`.

Relevant source references:

- `TransactionMQProducer#sendMessageInTransaction(...)`
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/client/src/main/java/org/apache/rocketmq/client/producer/TransactionMQProducer.java#L705-L719>
- `DefaultMQProducerImpl#sendMessageInTransaction(...)`
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java#L1222-L1242>
- Normal message selector path:
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java#L513-L526>
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java#L1105-L1113>

## Broker-side behavior

From the broker side, the final queue of a transactional message is determined by the queue selected when the half message is sent.

When the broker receives a prepared transactional message, it stores the original topic and queue id into message properties:

```text
PROPERTY_REAL_TOPIC
PROPERTY_REAL_QUEUE_ID
```

Then the message is rewritten to the internal half topic.

During commit, the broker restores the final message topic and queue id from these properties.

Relevant source references:

- Store original topic and queue id before writing half message:
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageBridge.java#L203-L211>
- Restore original topic and queue id when committing:
  - <https://github.com/apache/rocketmq/blob/rocketmq-all-4.9.7/broker/src/main/java/org/apache/rocketmq/broker/processor/EndTransactionProcessor.java#L194-L215>

This means queue selection must happen before the half message is sent. After the half message is stored, the commit phase only restores the previously selected queue and does not perform routing again.

## Problem

For transactional messages, users cannot provide a `MessageQueueSelector`, so they cannot control which queue the half message is written to.

This makes it difficult to support scenarios such as:

- deterministic transaction message sharding by business key
- canary / gray release queue lanes
- traffic isolation by queue
- keeping normal messages and transactional messages under the same queue routing strategy

For example, an application may use a custom selector for normal messages to route gray traffic to a subset of queues. But when the same flow uses transactional messages, the routing strategy cannot be reused because transactional send always uses the default queue selection logic.

## Proposal

Add overloaded transactional send APIs that accept `MessageQueueSelector`, similar to normal message sending APIs.

For example:

```java
TransactionSendResult sendMessageInTransaction(
    Message msg,
    MessageQueueSelector selector,
    Object selectorArg,
    Object transactionArg
) throws MQClientException;
```

Internally, the transactional send flow could still keep the existing transaction lifecycle unchanged:

1. validate transaction listener
2. mark message with `PROPERTY_TRANSACTION_PREPARED`
3. set producer group
4. send the half message using selector-based queue selection
5. execute local transaction
6. end transaction with commit / rollback / unknown

The only expected difference is step 4: instead of always calling the default send path, allow the half message to be sent through the selector-based path.

## Compatibility

This can be added as an overload, so existing APIs and behavior remain unchanged.

Existing users of:

```java
sendMessageInTransaction(Message msg, Object arg)
```

would continue to use the current default queue selection logic.

Users who need deterministic transactional queue routing can opt in to the new overload.

## Expected behavior

When a transactional message is sent with a `MessageQueueSelector`, the selected queue should become the original queue of the half message. The broker should then store this queue id as `PROPERTY_REAL_QUEUE_ID` and restore it when the transaction is committed, consistent with the existing broker transaction message flow.

## Additional context

This request does not require broker-side transaction protocol changes. The broker already restores the final queue from the queue id selected before the half message is stored. The missing capability is mainly on the client API and transactional send path.

### Motivation

Transactional messages currently do not provide an API to select the target message queue.

For normal messages, RocketMQ supports `MessageQueueSelector`, which allows users to route messages deterministically by business key, shard key, canary/gray label, or other custom routing rules. However, transactional messages always go through the default queue selection path internally.

From the broker-side transaction flow, the final queue of a transactional message is determined when the half message is sent. The broker stores the original queue id as `PROPERTY_REAL_QUEUE_ID` and restores it when the transaction is committed. Therefore, if users cannot select the queue before the half message is written, they cannot control the final queue of the committed transactional message.

This makes it difficult to keep transactional messages and normal messages under the same routing strategy, especially in scenarios such as deterministic sharding, traffic isolation, ordered business lanes, and canary/gray release queue lanes.

Transactional messages currently do not provide an API to select the target message queue.

For normal messages, RocketMQ supports `MessageQueueSelector`, which allows users to route messages deterministically by business key, shard key, canary/gray label, or other custom routing rules. However, transactional messages always go through the default queue selection path internally.

From the broker-side transaction flow, the final queue of a transactional message is determined when the half message is sent. The broker stores the original queue id as `PROPERTY_REAL_QUEUE_ID` and restores it when the transaction is committed. Therefore, if users cannot select the queue before the half message is written, they cannot control the final queue of the committed transactional message.

This makes it difficult to keep transactional messages and normal messages under the same routing strategy, especially in scenarios such as deterministic sharding, traffic isolation, ordered business lanes, and canary/gray release queue lanes.

I considered the following alternatives:

1. Put the desired queue id into message properties or user headers.

This does not work because the broker does not read the target queue id from message properties. The queue id is set in `SendMessageRequestHeader` by the client, based on the selected `MessageQueue`.

2. Select or change the queue during transaction commit.

This does not fit the current broker transaction flow. For transactional messages, the broker stores the original queue id as `PROPERTY_REAL_QUEUE_ID` when the half message is written. During commit, the broker only restores the original queue id from the half message. It does not perform queue selection again.

3. Add custom broker-side routing logic based on message properties.

This is possible in theory, but it would push application-level routing rules into the broker. It also introduces compatibility and maintenance risks around half messages, commit/rollback, transaction check, operation messages, and future upgrades.

4. Use normal messages with an outbox pattern instead of RocketMQ transactional messages.

This can be used as an application-level workaround, but it changes the programming model and does not solve the missing queue-selection capability for RocketMQ transactional messages.

### Additional Context

_No response_

Alaske:

Fixes #10545

### Brief Description

This PR adds `MessageQueueSelector` support for transactional message sending in the Java client.

Currently normal message sends can use `MessageQueueSelector`, but transactional sends only support:

```java
sendMessageInTransaction(Message msg, Object arg)
```

This PR adds a selector-aware overload so users can control the real queue used by the transactional half message:

```java
sendMessageInTransaction(
    Message msg,
    MessageQueueSelector selector,
    Object selectorArg,
    Object transactionArg
)
```

### How Did You Implement It

- Added a Java 8 `default` method to `MQProducer` to preserve binary compatibility for third-party implementations.
- Kept `DefaultMQProducer` behavior unchanged by explicitly rejecting transactional sends and directing users to `TransactionMQProducer`.
- Implemented the new overload in `TransactionMQProducer`.
- Added a selector-aware transaction send path in `DefaultMQProducerImpl`.
- Kept `selectorArg` and `transactionArg` separate:
  - `selectorArg` is passed only to `MessageQueueSelector`.
  - `transactionArg` is passed only to `TransactionListener#executeLocalTransaction`.
- Reused the existing selector queue resolution and specified-queue send validation so selected queues follow the same validation semantics as normal selector sends.
- Did not change broker, store, or remoting protocol behavior. The existing broker transaction flow already stores `REAL_TOPIC` / `REAL_QID` for half messages and restores the real queue when committing.

### How to Verify It

Added unit tests covering:

- `TransactionMQProducer` sends transactional messages to the queue selected by `MessageQueueSelector`.
- `selectorArg` is passed to the selector and `transactionArg` is passed to the local transaction listener.
- `DefaultMQProducer` still rejects transactional sends.
- Null selector is rejected.
- A selector returning a queue with a different topic is rejected consistently with normal selector send behavior.
- End transaction routing uses the selected broker/queue.

Verified with JDK 8:

```bash
mvn -pl client \
  -Dtest=DefaultMQProducerTest#assertSendMessageInTransactionByQueueSelector+testSendMessageInTransactionByQueueSelector+testSendMessageInTransactionByNullQueueSelector+testSendMessageInTransactionBySelectorWithDifferentTopicQueue \
  -DskipITs test
```

Result:

```text
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
Checkstyle: 0 violations
SpotBugs: 0 issues
```

The public API schema for `DefaultMQProducer` was also updated.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
@@ -951,6 +951,27 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
     }
 
     /**
+     * This method is used to send transactional messages with a message queue selector.
+     *
+     * <p>The {@code selectorArg} is passed only to {@link MessageQueueSelector#select(List, Message, Object)}.
+     * The {@code transactionArg} is passed only to {@link TransactionListener#executeLocalTransaction(Message, Object)}.
+     * This method follows the existing selector send semantics and does not perform the broker or queue reselection
+     * retry used by the default send path.</p>
+     *
+     * @param msg Transactional message to send.
+     * @param selector Message queue selector.
+     * @param selectorArg Argument used by the selector.
+     * @param transactionArg Argument used along with local transaction executor.
+     * @return Transaction result.
+     * @throws MQClientException if there is any client error.
+     */
+    @Override
+    public TransactionSendResult sendMessageInTransaction(Message msg,
+        MessageQueueSelector selector, Object selectorArg, Object transactionArg) throws MQClientException {
+        throw new RuntimeException("sendMessageInTransaction not implement, please use TransactionMQProducer class");
+    }
+
+    /**
      * This method will be removed in a certain version after April 5, 2020, so please do not use this method.
      *
      * @param key        accessKey
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -1433,6 +1433,21 @@ public class DefaultMQProducerImpl implements MQProducerInner {
     public TransactionSendResult sendMessageInTransaction(final Message msg,
         final TransactionListener localTransactionListener, final Object arg)
         throws MQClientException {
+        return sendMessageInTransaction(msg, localTransactionListener, null, null, arg);
+    }
+
+    public TransactionSendResult sendMessageInTransaction(final Message msg,
+        final MessageQueueSelector selector, final Object selectorArg, final Object arg)
+        throws MQClientException {
+        if (selector == null) {
+            throw new MQClientException("MessageQueueSelector is null", null);
+        }
+        return sendMessageInTransaction(msg, null, selector, selectorArg, arg);
+    }
+
+    private TransactionSendResult sendMessageInTransaction(final Message msg,
+        final TransactionListener localTransactionListener, final MessageQueueSelector selector,
+        final Object selectorArg, final Object arg) throws MQClientException {
         TransactionListener transactionListener = getCheckListener();
         if (null == localTransactionListener && null == transactionListener) {
             throw new MQClientException("tranExecutor is null", null);
@@ -1445,7 +1460,13 @@ public class DefaultMQProducerImpl implements MQProducerInner {
         MessageAccessor.putProperty(msg, MessageConst.PROPERTY_TRANSACTION_PREPARED, "true");
         MessageAccessor.putProperty(msg, MessageConst.PROPERTY_PRODUCER_GROUP, this.defaultMQProducer.getProducerGroup());
         try {
-            sendResult = this.send(msg);
+            if (selector == null) {
+                sendResult = this.send(msg);
+            } else {
+                MessageQueue mq = this.invokeMessageQueueSelector(msg, selector, selectorArg,
+                    this.defaultMQProducer.getSendMsgTimeout());
+                sendResult = this.send(msg, mq);
+            }
         } catch (Exception e) {
             throw new MQClientException("send message Exception", e);
         }
--- a/client/src/main/java/org/apache/rocketmq/client/producer/MQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/MQProducer.java
@@ -99,10 +99,8 @@ public interface MQProducer extends MQAdmin {
      * @return transaction result.
      * @throws MQClientException if the message cannot be sent.
      */
-    default TransactionSendResult sendMessageInTransaction(final Message msg,
-        final MessageQueueSelector selector, final Object selectorArg, final Object transactionArg) throws MQClientException {
-        throw new RuntimeException("sendMessageInTransaction not implement, please use TransactionMQProducer class");
-    }
+    TransactionSendResult sendMessageInTransaction(final Message msg,
+                                                   final MessageQueueSelector selector, final Object selectorArg, final Object transactionArg) throws MQClientException;
 
     //for batch
     SendResult send(final Collection<Message> msgs) throws MQClientException, RemotingException, MQBrokerException,
--- a/client/src/main/java/org/apache/rocketmq/client/producer/MQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/MQProducer.java
@@ -84,6 +84,26 @@ public interface MQProducer extends MQAdmin {
     TransactionSendResult sendMessageInTransaction(final Message msg,
         final Object arg) throws MQClientException;
 
+    /**
+     * Send a transactional message with a message queue selector.
+     *
+     * <p>The {@code selectorArg} is passed only to {@link MessageQueueSelector#select(List, Message, Object)}.
+     * The {@code transactionArg} is passed only to {@link TransactionListener#executeLocalTransaction(Message, Object)}.
+     * This method follows the existing selector send semantics and does not perform the broker or queue reselection
+     * retry used by the default send path.</p>
+     *
+     * @param msg transactional message to send.
+     * @param selector message queue selector.
+     * @param selectorArg argument used by the selector.
+     * @param transactionArg argument used by the local transaction executor.
+     * @return transaction result.
+     * @throws MQClientException if the message cannot be sent.
+     */
+    default TransactionSendResult sendMessageInTransaction(final Message msg,
+        final MessageQueueSelector selector, final Object selectorArg, final Object transactionArg) throws MQClientException {
+        throw new RuntimeException("sendMessageInTransaction not implement, please use TransactionMQProducer class");
+    }
+
     //for batch
     SendResult send(final Collection<Message> msgs) throws MQClientException, RemotingException, MQBrokerException,
         InterruptedException;
--- a/client/src/main/java/org/apache/rocketmq/client/producer/TransactionMQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/TransactionMQProducer.java
@@ -89,6 +89,35 @@ public class TransactionMQProducer extends DefaultMQProducer {
         return this.defaultMQProducerImpl.sendMessageInTransaction(msg, null, arg);
     }
 
+    /**
+     * Send a transactional message with a message queue selector.
+     *
+     * <p>The {@code selectorArg} is passed only to {@link MessageQueueSelector#select(List, Message, Object)}.
+     * The {@code transactionArg} is passed only to {@link TransactionListener#executeLocalTransaction(Message, Object)}.
+     * This method follows the existing selector send semantics and does not perform the broker or queue reselection
+     * retry used by the default send path.</p>
+     *
+     * @param msg transactional message to send.
+     * @param selector message queue selector.
+     * @param selectorArg argument used by the selector.
+     * @param transactionArg argument used by the local transaction executor.
+     * @return transaction result.
+     * @throws MQClientException if the message cannot be sent.
+     */
+    @Override
+    public TransactionSendResult sendMessageInTransaction(final Message msg,
+        final MessageQueueSelector selector, final Object selectorArg, final Object transactionArg) throws MQClientException {
+        if (null == this.transactionListener) {
+            throw new MQClientException("TransactionListener is null", null);
+        }
+        if (null == selector) {
+            throw new MQClientException("MessageQueueSelector is null", null);
+        }
+
+        msg.setTopic(NamespaceUtil.wrapNamespace(this.getNamespace(), msg.getTopic()));
+        return this.defaultMQProducerImpl.sendMessageInTransaction(msg, selector, selectorArg, transactionArg);
+    }
+
     public TransactionCheckListener getTransactionCheckListener() {
         return transactionCheckListener;
     }
--- a/client/src/test/java/org/apache/rocketmq/client/producer/DefaultMQProducerTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/producer/DefaultMQProducerTest.java
@@ -29,6 +29,7 @@ import org.apache.rocketmq.client.impl.factory.MQClientInstance;
 import org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl;
 import org.apache.rocketmq.client.impl.producer.TopicPublishInfo;
 import org.apache.rocketmq.client.latency.MQFaultStrategy;
+import org.apache.rocketmq.common.ServiceState;
 import org.apache.rocketmq.common.UtilAll;
 import org.apache.rocketmq.common.compression.CompressionType;
 import org.apache.rocketmq.common.message.Message;
@@ -82,6 +83,7 @@ import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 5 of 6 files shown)


Alaske:
this PR is ready for review.

The change is limited to the Java client and does not modify broker/store/remoting protocol behavior. I verified the focused client tests with JDK 8 locally; details are included in the PR description.

Could you please approve the GitHub Actions workflows when you have time?

RockteMQ-AI (approved):

### Summary
Adds `MessageQueueSelector` support for transactional message sending in the Java client, allowing users to control which queue the transactional half-message is sent to.

### Findings

Overall this is a well-structured and well-tested PR. The implementation correctly:
- Uses a Java 8 `default` method in `MQProducer` to preserve binary compatibility
- Keeps `selectorArg` and `transactionArg` cleanly separated
- Reuses existing selector queue resolution and validation logic
- Applies namespace wrapping in `TransactionMQProducer` consistently with the existing `sendMessageInTransaction`
- Updates the API schema file

- **[Info]** `DefaultMQProducerImpl.java:1439` — The null check for `selector` is performed in both `TransactionMQProducer.sendMessageInTransaction()` (line ~109) and `DefaultMQProducerImpl.sendMessageInTransaction()` (line ~1439). The check in the impl is technically unreachable when called through `TransactionMQProducer`, but it's good defensive programming if the impl method is ever called directly. No action needed.

- **[Info]** `DefaultMQProducerTest.java:785` — The test `assertSendMessageInTransactionByQueueSelector` verifies that `DefaultMQProducer` (not `TransactionMQProducer`) rejects the selector-based transactional send. The test name could be slightly more descriptive, e.g. `assertDefaultMQProducerRejectsSendMessageInTransactionByQueueSelector`, to distinguish it from the happy-path test. This is purely cosmetic.

### Suggestions

No blocking suggestions. The implementation is clean, the test coverage is thorough (null selector, different-topic rejection, arg separation, end-transaction routing), and backward compatibility is properly maintained.

### Cross-repo Note
No changes needed in `apache/rocketmq-clients` — this is a Java client-only API addition and does not affect the gRPC protocol or broker behavior.

---

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
```diff
@@ -950,6 +950,27 @@ public TransactionSendResult sendMessageInTransaction(Message msg,
...
+     *
+     * @param msg Transactional message to send.
+     * @param selector Message queue selector.
+     * @param selectorArg Argument used by the selector.
+     * @param transactionArg Argument used along with local transaction executor.
+     * @return Transaction result.
+     * @throws MQClientException if there is any client error.
+     */
+    @Override
+    public TransactionSendResult sendMessageInTransaction(Message msg,
```
yx9o, line 969: Since this method already has the same default implementation in MQProducer, do we still need to duplicate it in DefaultMQProducer?

Alaske: Runtime-wise, the override is not strictly required because `MQProducer` already provides the same default implementation.

I kept it in `DefaultMQProducer` for two reasons:

1. To keep the transactional-send rejection explicit and consistent with the existing `sendMessageInTransaction(Message, Object)` method in `DefaultMQProducer`.
2. To make the public API change visible in `DefaultMQProducer.schema`. The current schema tool collects declared methods from the class/superclasses, but does not include interface default methods.

If you prefer to avoid the duplicated override, I can remove it and adjust the schema expectation accordingly.

yx9o: Thanks for the explanation. I think we should keep the style consistent. If we prefer to explicitly implement this method in DefaultMQProducer, should we avoid making it a default method in MQProducer and keep it consistent with the other interface methods?

RockteMQ-AI (commented):

### Summary
Adds `MessageQueueSelector` support for transactional message sending. Users can now control which queue the transactional half-message is sent to, while keeping `selectorArg` and `transactionArg` separate.

### Findings

**API Design**
- **[Info]** `MQProducer.java:87-103` — New `default` method on the interface preserves binary compatibility for third-party `MQProducer` implementations. Good choice.
- **[Info]** `DefaultMQProducer.java:953-973` — Correctly throws `RuntimeException` directing users to `TransactionMQProducer`, consistent with the existing `sendMessageInTransaction(msg, arg)` pattern.
- **[Info]** `TransactionMQProducer.java:104-121` — Validates both `transactionListener` and `selector` for null before proceeding. Namespace wrapping is applied. Clean.

**Implementation**
- **[Info]** `DefaultMQProducerImpl.java:1436-1448` — The existing public method delegates to the new private 5-parameter method with `selector=null, selectorArg=null`. This preserves the original behavior exactly.
- **[Info]** `DefaultMQProducerImpl.java:1463-1469` — When `selector != null`, uses `invokeMessageQueueSelector` to resolve the queue, then calls `send(msg, mq)` for the specific-queue path. This correctly reuses existing selector validation (including the topic mismatch check tested in `testSendMessageInTransactionBySelectorWithDifferentTopicQueue`).
- **[Info]** The separation of `selectorArg` (passed only to `MessageQueueSelector.select()`) and `transactionArg` (passed only to `TransactionListener.executeLocalTransaction()`) is clean and avoids the ambiguity of a single shared argument.

**Tests**
- **[Info]** `DefaultMQProducerTest.java` — Comprehensive test coverage:
  - `assertSendMessageInTransactionByQueueSelector`: Verifies `DefaultMQProducer` rejects transactional sends
  - `testSendMessageInTransactionByQueueSelector`: Full integration test verifying selector arg routing, transaction arg routing, queue selection, and end-transaction routing
  - `testSendMessageInTransactionByNullQueueSelector`: Null selector rejection
  - `testSendMessageInTransactionBySelectorWithDifferentTopicQueue`: Topic mismatch rejection
- **[Info]** `api/client.producer.DefaultMQProducer.schema` — Public API schema updated to include the new method.

### Suggestions
- **[Warning]** `DefaultMQProducerImpl.java:1463-1469` — The selector-based path uses `this.send(msg, mq)` which goes through the specific-queue send path. This path does **not** perform the broker/queue reselection retry that the default `this.send(msg)` path does (as noted in the Javadoc). This is documented and intentional, but worth confirming that users understand the trade-off: they get deterministic queue selection at the cost of no automatic retry on broker unavailability.
- **[Info]** The `@Test(expected = RuntimeException.class)` on `assertSendMessageInTransactionByQueueSelector` is a bit loose — it catches any `RuntimeException` rather than asserting a specific message. The `testSendMessageInTransactionByNullQueueSelector` test does this better with explicit message assertion. Minor style point.

### Verdict
Well-designed API extension with proper backward compatibility, clean argument separation, and thorough test coverage. LGTM.

---

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
Alaske: Updated. I removed the default implementation from MQProducer and kept the new method as a regular interface method, consistent with the existing sendMessageInTransaction(Message, Object) API.

DefaultMQProducer still explicitly implements the method and rejects transactional sends with the same error message, while TransactionMQProducer provides the actual selector-aware implementation.

I also verified the related tests with JDK 8.

RockteMQ-AI (commented):

### Summary
New commit `dffbe61` ("remove MQProducer default method") changes the API approach: the new `sendMessageInTransaction(Message, MessageQueueSelector, Object, Object)` method on `MQProducer` is now **abstract** instead of `default`.

### Changes Since Last Review
- **`MQProducer.java`** — Method added as abstract (was `default`). This is a **binary compatibility break** for any third-party class that implements `MQProducer` directly — those implementations will fail to compile until they add the new method.
- **`DefaultMQProducerImpl.java`** — Same delegation pattern as before: existing public method delegates to a new private 5-parameter method; new public overload validates selector and delegates.
- **`DefaultMQProducer.java`** / **`TransactionMQProducer.java`** — Unchanged from previous review; both implement the new abstract method.
- **Tests** — Unchanged; coverage remains comprehensive.

### Findings
- **[Warning]** `MQProducer.java:87-103` — Adding an abstract method to a public interface is a breaking change for third-party implementors. If binary/source compatibility is a goal (as it was with the previous `default` method approach), consider keeping the `default` keyword and throwing `UnsupportedOperationException` in the body. If the project has decided this interface change is acceptable, the change is otherwise clean.
- **[Info]** `DefaultMQProducerImpl.java:1436-1448` — The refactoring into a private 5-parameter method is clean. The original public method delegates with `selector=null, selectorArg=null`, preserving exact original behavior.
- **[Info]** The selector-based path correctly uses `invokeMessageQueueSelector` + `send(msg, mq)`, consistent with normal selector sends.

### Suggestions
- If the abstract method approach is intentional, consider adding a note in the PR description or commit message explaining the compatibility trade-off.
- The rest of the implementation and tests look solid.

---

Alaske:
The transaction producer test setup now uses a dedicated MQClientInstance instead of the shared test-level client factory.

When the full client test suite runs, other DefaultMQProducerTest cases can mutate or stub the shared MQClientInstance state. The new transactional selector test depends on broker lookup through findBrokerAddressInPublish(...) after the selector chooses a queue, so sharing that client factory made the test order-sensitive.

To avoid cross-test pollution, prepareTransactionProducer(...) now creates an isolated MQClientInstance spy for the transaction producer, injects the mocked MQClientAPIImpl into it, and stubs only the broker lookup required by this test path. This keeps the transactional send selector test deterministic in both focused runs and the full Maven package run.

RockteMQ-AI (commented):

### Summary
Adds `MessageQueueSelector` support for transactional message sending, allowing users to control which queue the transactional half-message is sent to.

### Changes Since Last Review
- **test: isolate transaction producer client factory** (59b995) — Uses a `spy()` on a separate `MQClientInstance` for transaction producer tests instead of sharing the mock. Good improvement for test isolation — prevents cross-test state leakage.

### Findings
- **[Info]** `DefaultMQProducerImpl.java` — The new public overload `sendMessageInTransaction(Message, MessageQueueSelector, Object, Object)` correctly validates null selector before delegating. The private implementation cleanly handles both selector and non-selector paths.
- **[Info]** `TransactionMQProducer.java` — Proper delegation pattern with null check on selector. Consistent with existing `send()` overloads.
- **[Info]** Test coverage is solid: normal flow, null selector, topic mismatch, and the new test isolation improvement.

### Verdict
Clean implementation following existing patterns. The latest commit improves test isolation. LGTM.

---

client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
Alaske: Hi @yx9o, I pushed the latest test isolation fix and verified the official non-ubuntu Maven workflow command locally with JDK 8:  mvn -B package --file pom.xml -Djacoco.skip=true

Alaske:
If everything looks good and the required checks are green, could you please help approve and merge this PR? @yx9o

yx9o (approved):

Alaske:
Thanks for the review and approval @yx9o.

The required checks are green and there are no conflicts with the base branch. This PR fixes #10545 and is ready from my side.

Could you please help merge this PR when you have time?

RockteMQ-AI (approved):
### Summary

---

Alaske:
Could you please help merge this PR

PR #10554, opened 2026-06-27
Initialize checkStamp in LatencyFaultToleranceImpl.FaultItem

vasiliy-mikhailov:
`FaultItem.checkStamp` is never initialized, so it defaults to `0`. A freshly
created fault item is therefore immediately eligible for detection in
`detectByOneRound()` instead of waiting one `detectInterval` like every item
that has gone through `updateFaultItem`. This makes the very first detect round
fire a probe against a broker that was only just recorded.

Initialize `checkStamp` to `now + detectInterval` in the constructor, matching
the value `updateFaultItem` assigns. Added a regression test asserting the first
`detectByOneRound()` does not probe a freshly added item.

## Verifying this change

The added `LatencyFaultToleranceImplTest#testDetectByOneRoundRespectsDetectInterval` fails on the current `develop` and passes with this change:

Before the fix (on `develop`):

```
$ mvn test -Dtest=LatencyFaultToleranceImplTest -pl client
Running org.apache.rocketmq.client.latency.LatencyFaultToleranceImplTest
Tests run: 5, Failures: 1, Errors: 0, Skipped: 1 <<< FAILURE!
	at org.apache.rocketmq.client.latency.LatencyFaultToleranceImplTest.testDetectByOneRoundRespectsDetectInterval(LatencyFaultToleranceImplTest.java:102)
[INFO] BUILD FAILURE
```

After the fix:

```
$ mvn test -Dtest=LatencyFaultToleranceImplTest -pl client
Running org.apache.rocketmq.client.latency.LatencyFaultToleranceImplTest
[INFO] BUILD SUCCESS
```

## AI assistance disclosure

This contribution was produced with the help of an AI pipeline. The pipeline processed a large amount of source code to surface suspected bugs, reproduced a subset of them with failing unit tests and generated candidate fixes, and prepared pull requests from the ones that held up. Each PR was then reviewed and verified by a human before being opened: the fix and test were checked by hand and the test was confirmed to fail before the change and pass after.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImpl.java
@@ -202,6 +202,7 @@ public class LatencyFaultToleranceImpl implements LatencyFaultTolerance<String> 
 
         public FaultItem(final String name) {
             this.name = name;
+            this.checkStamp = System.currentTimeMillis() + LatencyFaultToleranceImpl.this.detectInterval;
         }
 
         public void updateNotAvailableDuration(long notAvailableDuration) {
--- a/client/src/test/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImplTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/latency/LatencyFaultToleranceImplTest.java
@@ -19,11 +19,17 @@ package org.apache.rocketmq.client.latency;
 import org.awaitility.core.ThrowingRunnable;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.Mockito;
 
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 public class LatencyFaultToleranceImplTest {
     private LatencyFaultTolerance<String> latencyFaultTolerance;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Initializes `FaultItem.checkStamp` in the constructor to `now + detectInterval`, preventing a freshly created item from being immediately eligible for detection in `detectByOneRound()`.

### Findings
- **[Info]** `LatencyFaultToleranceImpl.java:205` — The fix correctly mirrors the assignment in `updateFaultItem()`, ensuring consistent initial state. Good alignment.
- **[Info]** `LatencyFaultToleranceImplTest.java:89-103` — Regression test is well-structured: uses Mockito to verify `detect()` is never called for a freshly added item within the detect interval.

### Verdict
Clean, minimal fix with a solid regression test. LGTM.

---

RockteMQ-AI (commented):

### Summary
Initializes `FaultItem.checkStamp` to `now + detectInterval` in the constructor, preventing a freshly created item from being immediately eligible for detection in `detectByOneRound()`.

### Findings
- **[Info]** `LatencyFaultToleranceImpl.java:205` — The fix correctly mirrors the initialization done in `updateFaultItem()`. Clean one-line change.
- **[Info]** `LatencyFaultToleranceImplTest.java:92-103` — Good regression test. Using `Mockito.never()` to verify no detection occurs on a fresh item is the right approach. Setting `detectInterval` to 1 hour makes the test deterministic.

### Suggestions
- Consider whether `LatencyFaultToleranceImpl.this.detectInterval` could be zero at construction time (before `setDetectInterval` is called). If so, the initialization would still result in `checkStamp = now`, which is the same as the old behavior. The test sets it explicitly after construction, so this edge case may not be covered.

### Verdict
Clean, well-targeted fix with a solid regression test. The concern above is minor and may not apply in practice since `detectInterval` is typically set before any `FaultItem` is created.

---

RockteMQ-AI (approved):
### Summary

---

PR #10566, opened 2026-07-01
[ISSUE #10288] Hold MappedFile reference in ConsumeQueue write path to prevent SIGSEGV

### LINKED ISSUE #10288: [Bug] JVM crashes SIGSEGV in ConsumeQueue write path due to missing reference counting when MappedFile is destroyed concurrently

Linux 4.19 x86_64, physical memory 16GB. 

### RocketMQ version

Reproduced on both v4.9.4 and v5.3.3. The affected code still exists on develop.  

### JDK Version

1.8.0_192-b12 (HotSpot 64-Bit Server VM, mixed mode). 
 Runtime args: -Xms8g -Xmx8g -Xmn4g -XX:+UseG1GC -XX:MaxDirectMemorySize=15g -XX:+AlwaysPreTouch

### Describe the Bug

Broker crashes with a fatal JVM error (SIGSEGV / SEGV_MAPERR) while ReputMessageService is writing a dispatch entry to a ConsumeQueue, or while AdminBrokerThread is executing cleanExpiredConsumeQueue.

**Case 1 — writer side (v4.9.4):**
Problematic frame:
v  ~StubRoutines::jint_disjoint_arraycopy
siginfo: si_signo: 11 (SIGSEGV), si_code: 1 (SEGV_MAPERR), si_addr: 0x7fec2674a064

Current thread: JavaThread "ReputMessageService"
J C2 org.apache.rocketmq.store.ConsumeQueue.putMessagePositionInfo(JIJJ)Z
J C2 org.apache.rocketmq.store.ConsumeQueue.putMessagePositionInfoWrapper
J C2 DefaultMessageStore$CommitLogDispatcherBuildConsumeQueue.dispatch
J C2 DefaultMessageStore$ReputMessageService.doReput
The faulting address sits in an un-mapped hole between two ConsumeQueue mmap regions in /proc/self/maps, proving the target MappedByteBuffer has already been munmapped.

**Case 2 — cleaner side (v5.3.3):**
Problematic frame:
J C2 org.apache.rocketmq.store.queue.ConsumeQueueStore.cleanExpired(J)V
Current thread: JavaThread "AdminBrokerThread_12"
J C1 DefaultMessageStore.cleanExpiredConsumerQueue
J C1 AdminBrokerProcessor.cleanExpiredConsumeQueue
Root cause is the same: the ConsumeQueue write/read paths access MappedFile.mappedByteBuffer without holding the ReferenceResource reference count, so a concurrent
destroy() can unmap the buffer while a JIT-compiled Unsafe.copyMemory is still writing to it.

**Code references (5.3.3 / develop)**
Write path has no hold()/release():
// store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java  (putMessagePositionInfo)
MappedFile mappedFile = this.mappedFileQueue.getLastMappedFile(expectLogicOffset);
if (mappedFile != null) {
    ...
    this.setMaxPhysicOffset(offset + size);
    if (messageStore.getMessageStoreConfig().isPutConsumeQueueDataByFileChannel()) {
        appendResult = mappedFile.appendMessageUsingFileChannel(...);
    } else {
        appendResult = mappedFile.appendMessage(this.byteBufferIndex.array()); // writes to mmap
    }
    return appendResult;
}
// DefaultMappedFile.appendMessage(byte[], offset, length)
ByteBuffer buf = this.mappedByteBuffer.slice(); // not guarded by hold()
buf.position(currentPos);
buf.put(data, offset, length);                  // JIT-inlined Unsafe.copyMemory
Concurrent destroyer:
// store/.../queue/ConsumeQueueStore.java:cleanExpired
} else if (maxCLOffsetInConsumeQueue < minCommitLogOffset) {
    removeTopicQueueTable(...);
    this.destroy(nextQT.getValue()); // -> MappedFileQueue.destroy -> MappedFile.destroy(3000)
    itQT.remove();
}

ReferenceResource.shutdown(intervalForcibly) decrements the initial reference on the first call, so if the writer side never calls hold() the counter immediately drops to 0, cleanup() runs, UtilAll.cleanBuffer(mappedByteBuffer) triggers munmap, and the in-flight write hits an unmapped page → SIGSEGV. The intervalForcibly grace window only protects callers that actually do reference counting.
Trigger surfaces (any of them reproduces the crash):
1. AdminBrokerProcessor.cleanExpiredConsumeQueue RPC (external management call).
2. DefaultMessageStore$CleanConsumeQueueService periodic deleteExpiredFile – safe for the tail file but still unmaps non-tail files that the dispatcher may be writing to when reput lags behind.
3. DefaultMessageStore.deleteTopics / cleanUnusedTopic (topic removal).
4. Recover / truncate paths (truncateDirty, slave switch).

### Steps to Reproduce

1. Start a broker under sustained write pressure with many topics, so that CommitLog rolls quickly and minCommitLogOffset advances.
2. Either:
  - Send the CLEAN_EXPIRED_CONSUMEQUEUE admin command repeatedly while dispatch is active, or
  - Set a small fileReservedTime so CleanConsumeQueueService keeps calling deleteExpiredFileByOffset while ReputMessageService is behind the latest CommitLog offset and
still writing to non-tail ConsumeQueue files.
3. Wait for C2 to compile ConsumeQueue.putMessagePositionInfo and ConsumeQueueStore.cleanExpired (a few minutes of load).
4. JVM crashes with SIGSEGV in jint_disjoint_arraycopy / ConsumeQueueStore.cleanExpired; the hs_err_pid*.log memory map shows the faulting address lying in a hole between
 two consumequeue/... mmap regions, or immediately past the end of one.

ConsumeQueue write/read paths and ConsumeQueueStore.cleanExpired cooperate correctly with MappedFile.destroy() via the existing ReferenceResource protocol, so concurrent
destroy is safe and at worst causes the current dispatch to be retried.

### What Did You See Instead?

Broker JVM crashes with SIGSEGV (SEGV_MAPERR). The target address is an already-munmapped MappedByteBuffer belonging to a ConsumeQueue file that was just destroyed
("cleanExpiredConsumerQueue: ... consumer queue destroyed" log entries immediately before the crash).

### Additional Context

If you need the full JVM crash log, I can provide it. Thank you!

wang-jiahua:

Fixes #10288

### Brief Description

ConsumeQueue.putMessagePositionInfo() accesses MappedFile without holding a reference count. A concurrent cleanExpiredConsumeQueue can destroy() the MappedFile (unmapping its buffer) while the writer is still using it, causing SIGSEGV in JIT-compiled Unsafe.copyMemory.

Fix: Call mappedFile.hold() before accessing the buffer and release() in a finally block, following the same pattern used in DefaultMappedFile.getData(). Also protect the fillPreBlank call in initializeWithOffset.

Existing ConsumeQueueTest (17 tests) all pass. The fix was validated during server benchmarking where the baseline broker crashed with SIGSEGV in ReputMessageService at ConsumeQueue.putMessagePositionInfo — exactly the crash signature reported in #10288.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
@@ -856,44 +856,52 @@ public class ConsumeQueue implements ConsumeQueueInterface {
 
         MappedFile mappedFile = this.mappedFileQueue.getLastMappedFile(expectLogicOffset);
         if (mappedFile != null) {
-
-            if (mappedFile.isFirstCreateInQueue() && cqOffset != 0 && mappedFile.getWrotePosition() == 0) {
-                this.minLogicOffset = expectLogicOffset;
-                this.mappedFileQueue.setFlushedWhere(expectLogicOffset);
-                this.mappedFileQueue.setCommittedWhere(expectLogicOffset);
-                this.fillPreBlank(mappedFile, expectLogicOffset);
-                log.info("fill pre blank space " + mappedFile.getFileName() + " " + expectLogicOffset + " "
-                    + mappedFile.getWrotePosition());
+            if (!mappedFile.hold()) {
+                log.warn("Failed to hold mapped file for ConsumeQueue write, topic={} queueId={}",
+                    this.topic, this.queueId);
+                return false;
             }
-
-            if (cqOffset != 0) {
-                long currentLogicOffset = mappedFile.getWrotePosition() + mappedFile.getFileFromOffset();
-
-                if (expectLogicOffset < currentLogicOffset) {
-                    log.warn("Build consume queue repeatedly, expectLogicOffset: {} currentLogicOffset: {} Topic: {} QID: {} Diff: {}",
-                        expectLogicOffset, currentLogicOffset, this.topic, this.queueId, expectLogicOffset - currentLogicOffset);
-                    return true;
+            try {
+                if (mappedFile.isFirstCreateInQueue() && cqOffset != 0 && mappedFile.getWrotePosition() == 0) {
+                    this.minLogicOffset = expectLogicOffset;
+                    this.mappedFileQueue.setFlushedWhere(expectLogicOffset);
+                    this.mappedFileQueue.setCommittedWhere(expectLogicOffset);
+                    this.fillPreBlank(mappedFile, expectLogicOffset);
+                    log.info("fill pre blank space " + mappedFile.getFileName() + " " + expectLogicOffset + " "
+                        + mappedFile.getWrotePosition());
                 }
 
-                if (expectLogicOffset != currentLogicOffset) {
-                    LOG_ERROR.warn(
-                        "[BUG]logic queue order maybe wrong, expectLogicOffset: {} currentLogicOffset: {} Topic: {} QID: {} Diff: {}",
-                        expectLogicOffset,
-                        currentLogicOffset,
-                        this.topic,
-                        this.queueId,
-                        expectLogicOffset - currentLogicOffset
-                    );
+                if (cqOffset != 0) {
+                    long currentLogicOffset = mappedFile.getWrotePosition() + mappedFile.getFileFromOffset();
+
+                    if (expectLogicOffset < currentLogicOffset) {
+                        log.warn("Build consume queue repeatedly, expectLogicOffset: {} currentLogicOffset: {} Topic: {} QID: {} Diff: {}",
+                            expectLogicOffset, currentLogicOffset, this.topic, this.queueId, expectLogicOffset - currentLogicOffset);
+                        return true;
+                    }
+
+                    if (expectLogicOffset != currentLogicOffset) {
+                        LOG_ERROR.warn(
+                            "[BUG]logic queue order maybe wrong, expectLogicOffset: {} currentLogicOffset: {} Topic: {} QID: {} Diff: {}",
+                            expectLogicOffset,
+                            currentLogicOffset,
+                            this.topic,
+                            this.queueId,
+                            expectLogicOffset - currentLogicOffset
+                        );
+                    }
                 }
+                this.setMaxPhysicOffset(offset + size);
+                boolean appendResult;
+                if (messageStore.getMessageStoreConfig().isPutConsumeQueueDataByFileChannel()) {
+                    appendResult = mappedFile.appendMessageUsingFileChannel(this.byteBufferIndex.array());
+                } else {
+                    appendResult = mappedFile.appendMessage(this.byteBufferIndex.array());
+                }
+                return appendResult;
+            } finally {
+                mappedFile.release();
             }
-            this.setMaxPhysicOffset(offset + size);
-            boolean appendResult;
-            if (messageStore.getMessageStoreConfig().isPutConsumeQueueDataByFileChannel()) {
-                appendResult = mappedFile.appendMessageUsingFileChannel(this.byteBufferIndex.array());
-            } else {
-                appendResult = mappedFile.appendMessage(this.byteBufferIndex.array());
-            }
-            return appendResult;
         }
         return false;
     }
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (commented):

### Summary
Fixes a SIGSEGV caused by concurrent `cleanExpiredConsumeQueue` destroying a `MappedFile` while `putMessagePositionInfo()` is still writing to its buffer. The fix adds `mappedFile.hold()` / `mappedFile.release()` in a try-finally pattern around all buffer access in the ConsumeQueue write path.

### Findings
- **[Info]** `ConsumeQueue.java:862` — `hold()` failure returns `false` safely, preventing use of an unmapped buffer. The warning log includes topic and queueId for diagnosis.
- **[Info]** `ConsumeQueue.java:864-901` — The try-finally correctly wraps all buffer operations including `fillPreBlank`, repeated-offset check, and `appendMessage`/`appendMessageUsingFileChannel`. The early `return true` at line ~882 (repeated build) still executes the `finally` block — correct.
- **[Info]** `ConsumeQueue.java:1282-1288` — `initializeWithOffset` also gets the hold/release treatment with an added null check. Good defensive coding.
- **[Info]** The pattern is consistent with `DefaultMappedFile.getData()` — established codebase convention.

### Suggestions
- Consider whether `CommitLog` write paths have similar exposure. If `cleanExpiredMappedFiles()` can race with other CommitLog writers, the same pattern may be needed there.
- A concurrent stress test (e.g., write + cleanExpired in parallel threads) would help prevent regression.

Overall: critical bug fix, correctly implemented. LGTM.

---

RongtongJin (dismissed):

RongtongJin:
LGTM～ @guyinyou help to double check

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -1271,7 +1279,13 @@ public void initializeWithOffset(long offset, long minPhyOffset) {
...
         // transientStorePool is null, only need set wrote position here
         MappedFile mappedFile = mappedFileQueue.getLastMappedFile(offset * ConsumeQueue.CQ_STORE_UNIT_SIZE, true);
-        fillPreBlank(mappedFile, offset * ConsumeQueue.CQ_STORE_UNIT_SIZE);
+        if (mappedFile != null && mappedFile.hold()) {
+            try {
+                fillPreBlank(mappedFile, offset * ConsumeQueue.CQ_STORE_UNIT_SIZE);
+            } finally {
+                mappedFile.release();
+            }
+        }
```
RongtongJin, line 1294: `initializeWithOffset()` should not silently continue when the mapped file cannot be created or held. This path relies on continuous ConsumeQueue files, and skipping `fillPreBlank()` while still calling `flush(0)` makes initialization look successful even though the pre-blank area was never written. Please fail fast here, or at least log an error and abort the initialization, so a file creation/availability problem does not leave the consume queue in an inconsistent state.

wang-jiahua: Good catch. Fixed in the latest commit — instead of silently skipping `fillPreBlank()`, the method now logs an error and returns early (aborts initialization) when `mappedFile` is null or `hold()` fails:

```java
if (mappedFile == null) {
    log.error("initializeWithOffset failed: mappedFile is null for offset {}", offset);
    return;
}
if (!mappedFile.hold()) {
    log.error("initializeWithOffset failed: mappedFile hold() failed for offset {}", offset);
    return;
}
```

This ensures the consume queue is not left in an inconsistent state (no `flush(0)` called) when file creation/availability fails, while keeping the `hold()` reference to prevent the SIGSEGV.

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
```diff
@@ -1271,7 +1279,13 @@ public void initializeWithOffset(long offset, long minPhyOffset) {
 
         // transientStorePool is null, only need set wrote position here
         MappedFile mappedFile = mappedFileQueue.getLastMappedFile(offset * ConsumeQueue.CQ_STORE_UNIT_SIZE, true);
-        fillPreBlank(mappedFile, offset * ConsumeQueue.CQ_STORE_UNIT_SIZE);
+        if (mappedFile != null && mappedFile.hold()) {
+            try {
```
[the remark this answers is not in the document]
wang-jiahua: Fixed in the latest commit. Instead of silently skipping `fillPreBlank()`, the method now logs an error and returns early when `mappedFile` is null or `hold()` fails, preventing the consume queue from being left in an inconsistent state.

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
frinda: org.apache.rocketmq.store.logfile.DefaultMappedFile#appendMessage(byte[], int, int) 
I think this place also needs to be protected via reference counting (hold()/release()).    Recently, I encountered an issue where the standby server suddenly crashed during primary-standby synchronization. 


hs_error log as follows：

RockteMQ-AI (approved):
### Summary

---

store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
wang-jiahua: Re-confirmed on the current branch: `initializeWithOffset()` now logs an error and aborts (no `fillPreBlank()`, no `flush(0)`) when the mapped file cannot be created or held, so a file availability problem no longer looks like a successful initialization.

guyinyou (approved):
LGTM

RongtongJin (approved):

PR #10568, opened 2026-07-01
[ISSUE #10518] Add client config to ignore corrupted local offset files

### LINKED ISSUE #10518: [Feature] Add client config option to ignore corrupted local offset files instead of throwing exception

When ~/.rocketmq_offsets/ files are corrupted (e.g., truncated due to power loss during persistAll()), readLocalOffsetBak() unconditionally throws MQClientException, which prevents the consumer from starting and crashes the entire application. The error "maybe fastjson version too low" is misleading — the real cause is file corruption, not a version issue.

For applications that already rely on CONSUME_FROM_TIMESTAMP, the persistent offset is non-critical — losing it does not affect data correctness. But there is currently no way to tell RocketMQ "just ignore corrupted offset files and start fresh."

Add a client configuration property, e.g.:

```
# Default false — keep existing behavior (throw on corruption)
  rocketmq.client.localOffsetStore.ignoreCorrupted=false
``` 

When set to true, readLocalOffsetBak() should return null instead of throwing when JSON parsing fails:

```
private OffsetSerializeWrapper readLocalOffsetBak() throws MQClientException {
      // ...
      if (content != null && content.length() > 0) {
          try {
              offsetSerializeWrapper = OffsetSerializeWrapper.fromJson(content, ...);
          } catch (Exception e) {
              log.warn("readLocalOffset Exception", e);
              if (ignoreCorrupted) {
                  return null;
              }
              throw new MQClientException("readLocalOffset Exception, ...", e);
          }
      }
      return null;
  }
```

When ignoreCorrupted=true and offset files are corrupted, load() starts with an empty offset table, and the consumer falls back to CONSUME_FROM_TIMESTAMP (or the latest message if no timestamp is set). Behavior only differs from the default on startup after file corruption —normal operation is unaffected.

 - Always return null (no config): Too aggressive — changes default behavior for all existing users who may depend on the exception for fast-fail
  semantics.
  - Application-level workaround: Manually delete ~/.rocketmq_offsets/ before each consumer start. Effective but fragile — requires every affected
  application to implement its own cleanup logic. A built-in opt-in config is cleaner and discoverable.
  - System property -Drocketmq.client.localOffsetStoreDir to a tmpfs: Redirects offsets to volatile storage, avoiding persistence entirely. Works but
  throws away offsets for all consumers globally, which may be undesirable in mixed-use cases.

### Additional Context

_No response_

wang-jiahua:

Fixes #10518

### Brief Description

When ~/.rocketmq_offsets/ files are corrupted (e.g., truncated due to power loss during persistAll()), readLocalOffsetBak() unconditionally throws MQClientException, crashing the consumer application.

Fix: Add ClientConfig.localOffsetStoreIgnoreCorrupted (default false). When true, corrupted offset files are silently skipped with a warning log, and the consumer starts with an empty offset table (falling back to CONSUME_FROM_TIMESTAMP or broker-side offset). Default false preserves existing behavior (fully backward compatible).

Compilation passed. Default false ensures no behavior change for existing users.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
+++ b/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
@@ -106,6 +106,12 @@ public class ClientConfig {
     private int concurrentHeartbeatThreadPoolSize = Runtime.getRuntime().availableProcessors();
 
     /**
+     * When true, corrupted local offset files are ignored (logged at WARN level) instead of throwing an exception.
+     * The consumer starts with an empty offset table and falls back to CONSUME_FROM_TIMESTAMP or broker-side offset.
+     */
+    private boolean localOffsetStoreIgnoreCorrupted = false;
+
+    /**
      * The switch for message trace
      */
     protected boolean enableTrace = false;
@@ -246,6 +252,7 @@ public class ClientConfig {
         this.traceTopic = cc.traceTopic;
         this.enableConcurrentHeartbeat = cc.enableConcurrentHeartbeat;
         this.concurrentHeartbeatThreadPoolSize = cc.concurrentHeartbeatThreadPoolSize;
+        this.localOffsetStoreIgnoreCorrupted = cc.localOffsetStoreIgnoreCorrupted;
     }
 
     public ClientConfig cloneClientConfig() {
@@ -280,6 +287,7 @@ public class ClientConfig {
         cc.traceTopic = traceTopic;
         cc.enableConcurrentHeartbeat = enableConcurrentHeartbeat;
         cc.concurrentHeartbeatThreadPoolSize = concurrentHeartbeatThreadPoolSize;
+        cc.localOffsetStoreIgnoreCorrupted = localOffsetStoreIgnoreCorrupted;
         return cc;
     }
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):

### Summary
Adds a `ClientConfig.localOffsetStoreIgnoreCorrupted` flag (default `false`) so that when local offset files are corrupted (e.g., truncated during power loss), the consumer can skip them with a warning instead of crashing with `MQClientException`.

### Findings
- **[Info]** `LocalFileOffsetStore.java:266` — The guard is correctly placed inside the `catch (Exception e)` block of `readLocalOffsetBak()`. Returning `null` here causes the caller to fall back to broker-side offset or `CONSUME_FROM_TIMESTAMP`, which is the desired behavior.
- **[Info]** `ClientConfig.java:108` — Default `false` preserves backward compatibility. Good.
- **[Info]** The log message includes `groupName` which aids debugging. Consider also logging the file path that was corrupted for faster diagnosis.

### Suggestions
- Consider whether `readLocalOffset()` (the primary read path, not just the backup) should also respect this flag. Currently only `readLocalOffsetBak()` is guarded — if the primary file is corrupted, the exception still propagates.
- A unit test covering the `localOffsetStoreIgnoreCorrupted=true` path would strengthen this change.

---

RockteMQ-AI (commented):

### Summary
Re-reviewed after commit at 05:55 UTC. The diff is unchanged from the initial review — the new commit appears to be a rebase on the latest `develop`.

### Status
The previous review findings remain valid:
- Core change is clean: `localOffsetStoreIgnoreCorrupted` flag in `ClientConfig` with proper `clone`/`reset`/`toString` integration.
- Guard correctly placed in `readLocalOffsetBak()` catch block, returning `null` to trigger fallback.
- Default `false` preserves backward compatibility.

### Previous Suggestions (Still Open)
1. **Primary read path**: `readLocalOffset()` (not just the backup) could also respect this flag — currently a corrupted primary file still throws.
2. **Unit test**: A test covering the `localOffsetStoreIgnoreCorrupted=true` path would strengthen this change.

No new issues found.

---

RockteMQ-AI (approved):
### Summary

---

PR #10579, opened 2026-07-02
[ISSUE #10578] Fix master election maxOffset comparator overflow

### LINKED ISSUE #10578: [Bug] Master election may choose lower-offset broker due to maxOffset comparator overflow

OS: Ubuntu 20.04

### RocketMQ version

  branch: develop

### JDK Version

_No response_

### Describe the Bug

  `DefaultElectPolicy` may elect the wrong broker as master when two broker candidates have the same epoch but their `maxOffset` values differ by more than `Integer.MAX_VALUE`.

  The comparator currently subtracts two `long` offsets and casts the result to `int`:

  ```java
  (int) (o2.getMaxOffset() - o1.getMaxOffset())
  ```

  This can overflow and reverse the intended ordering, causing a lower-offset broker to be ranked before a higher-offset broker.

### Steps to Reproduce

  1. Create two BrokerLiveInfo candidates with the same epoch and same election priority.
  2. Set broker 1 maxOffset to 0.
  3. Set broker 2 maxOffset to Integer.MAX_VALUE + 1L.
  4. Call DefaultElectPolicy.elect(...) with both brokers as valid candidates.
  5. Observe which broker is elected.

  The broker with the higher maxOffset should be elected when epoch values are equal.

### What Did You See Instead?

  The comparator can overflow and rank the lower-offset broker first, so the controller may elect a stale replica as master.

### Additional Context

Affected file:

  controller/src/main/java/org/apache/rocketmq/controller/elect/impl/DefaultElectPolicy.java

  Suggested fix: replace subtraction-based comparison with safe comparator methods:

  int epochCompare = Integer.compare(o2.getEpoch(), o1.getEpoch());
  if (epochCompare != 0) {
      return epochCompare;
  }

  int offsetCompare = Long.compare(o2.getMaxOffset(), o1.getMaxOffset());
  if (offsetCompare != 0) {
      return offsetCompare;
  }

  return Integer.compare(o1.getElectionPriority(), o2.getElectionPriority());

Aias00:
## What changed

Fix `DefaultElectPolicy` master-election ordering to avoid integer overflow when comparing broker `maxOffset` values.

The election comparator previously used subtraction and cast the `long` offset delta to `int`:

```java
(int) (o2.getMaxOffset() - o1.getMaxOffset())
```

When two brokers have the same epoch and their `maxOffset` difference exceeds `Integer.MAX_VALUE`, the comparator can overflow and return the wrong ordering. That may cause the controller to elect a broker with a lower offset as master.

This PR replaces subtraction-based comparisons with `Integer.compare` and `Long.compare`, preserving the existing ordering rules:

1. Higher epoch first
2. Higher max offset first
3. Lower election priority first

Fixes #10578

## Why

Master election is a control-plane correctness path. A comparator overflow here can select a stale replica when broker offsets diverge by more than `Integer.MAX_VALUE`.

## Tests

Added `DefaultElectPolicyTest` covering the overflow case:

- two brokers with the same epoch
- offset delta greater than `Integer.MAX_VALUE`
- verifies the broker with the higher offset is elected

Verified locally:

```bash
mvn -q -pl controller -DskipTests=false -Dtest=DefaultElectPolicyTest -Djacoco.skip=true test
mvn -q -pl controller -DskipTests compile -Dspotbugs.skip=true -Dcheckstyle.skip=true
mvn -q -DskipTests compile -Dspotbugs.skip=true -Dcheckstyle.skip=true
```

## Notes

`checkstyle:check` was not used as a gating signal because the controller module currently reports existing unrelated checkstyle violations. The targeted test and compile checks pass.

### DIFF
--- a/controller/src/main/java/org/apache/rocketmq/controller/elect/impl/DefaultElectPolicy.java
+++ b/controller/src/main/java/org/apache/rocketmq/controller/elect/impl/DefaultElectPolicy.java
@@ -37,12 +37,17 @@ public class DefaultElectPolicy implements ElectPolicy {
 
     // Sort in descending order according to<epoch, offset>, and sort in ascending order according to priority
     private final Comparator<BrokerLiveInfo> comparator = (o1, o2) -> {
-        if (o1.getEpoch() == o2.getEpoch()) {
-            return o1.getMaxOffset() == o2.getMaxOffset() ? o1.getElectionPriority() - o2.getElectionPriority() :
-                (int) (o2.getMaxOffset() - o1.getMaxOffset());
-        } else {
-            return o2.getEpoch() - o1.getEpoch();
+        int epochCompare = Integer.compare(o2.getEpoch(), o1.getEpoch());
+        if (epochCompare != 0) {
+            return epochCompare;
         }
+
+        int offsetCompare = Long.compare(o2.getMaxOffset(), o1.getMaxOffset());
+        if (offsetCompare != 0) {
+            return offsetCompare;
+        }
+
+        return Integer.compare(o1.getElectionPriority(), o2.getElectionPriority());
     };
 
     public DefaultElectPolicy(BrokerValidPredicate validPredicate, BrokerLiveInfoGetter brokerLiveInfoGetter) {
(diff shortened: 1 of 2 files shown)


RongtongJin (approved):

RockteMQ-AI (approved):
### Summary

---

PR #10581, opened 2026-07-02
[ISSUE #10580] Fix static topic epoch comparator overflow

### LINKED ISSUE #10580: [Bug] Static topic routing may use stale epoch due to comparator overflow

Ubuntu 20.04

### RocketMQ version

  branch: develop

### JDK Version

_No response_

### Describe the Bug

  Static topic routing and mapping validation may use the wrong mapping version when comparing mapping epochs with a large difference.

  Two places sort static topic mapping metadata by epoch using subtraction and casting the result to `int`:

  ```java
  (int) (o2.getValue().getEpoch() - o1.getValue().getEpoch())
  ```
  and:
  ```java
  (int) (o2.getEpoch() - o1.getEpoch())
  ```
  If the epoch difference is greater than Integer.MAX_VALUE, the cast can overflow and return the wrong ordering. This can cause older static topic mapping metadata to be processed before newer metadata.

  Affected files:

  - remoting/src/main/java/org/apache/rocketmq/remoting/rpc/ClientMetadata.java
  - remoting/src/main/java/org/apache/rocketmq/remoting/protocol/statictopic/TopicQueueMappingUtils.java

### Steps to Reproduce

 1. Create old static topic mapping metadata with epoch 0.
  2. Create new static topic mapping metadata with epoch Integer.MAX_VALUE + 1L.
  3. Use both mappings for the same logical queue/global queue.
  4. Build routing/mapping result through:
      - ClientMetadata.topicRouteData2EndpointsForStaticTopic(...)
      - TopicQueueMappingUtils.checkAndBuildMappingItems(..., replace=true, ...)

  5. Verify which mapping is selected.

  The newer mapping with the higher epoch should always be selected, even when the epoch difference is greater than Integer.MAX_VALUE.

### What Did You See Instead?

  The subtraction-based comparator can overflow and process stale lower-epoch metadata before newer metadata, which may cause static topic routing or mapping replacement to use stale broker mapping information.

### Additional Context

  Suggested fix: use Long.compare(...) instead of subtraction-based comparison:

  mappingInfos.sort((o1, o2) -> Long.compare(o2.getValue().getEpoch(), o1.getValue().getEpoch()));

  and:

  mappingDetailList.sort((o1, o2) -> Long.compare(o2.getEpoch(), o1.getEpoch()));

Aias00:
## What changed

Fix static topic epoch ordering to avoid integer overflow when comparing mapping epoch values.

Two comparators previously subtracted `long` epoch values and cast the result to `int`:

```java
(int) (o2.getValue().getEpoch() - o1.getValue().getEpoch())
```

and:

```java
(int) (o2.getEpoch() - o1.getEpoch())
```

When the epoch difference exceeds `Integer.MAX_VALUE`, the comparator can overflow and process stale mapping metadata before newer metadata.

This PR replaces subtraction-based comparisons with `Long.compare(...)` in:

- `ClientMetadata.topicRouteData2EndpointsForStaticTopic(...)`
- `TopicQueueMappingUtils.checkAndBuildMappingItems(...)`

Fixes #10580

## Why

Static topic mapping epochs fence old metadata. Ordering by epoch must remain correct even when epoch gaps are larger than `Integer.MAX_VALUE`, otherwise route building or duplicate queue replacement can prefer stale mapping data.

## Tests

Added regression coverage for large epoch gaps:

- `ClientMetadataTest`: verifies routing resolves to the latest epoch broker mapping
- `TopicQueueMappingUtilsTest`: verifies duplicate global queue replacement keeps the latest epoch mapping

Verified locally:

```bash
mvn -q -pl remoting -DskipTests=false -Dtest=ClientMetadataTest,TopicQueueMappingUtilsTest -Djacoco.skip=true test
mvn -q -pl remoting -DskipTests compile -Dspotbugs.skip=true -Dcheckstyle.skip=true
mvn -q -DskipTests compile -Dspotbugs.skip=true -Dcheckstyle.skip=true
git diff --check
```

## Notes

The change keeps the existing ordering semantics: higher epoch first. It only replaces unsafe subtraction with `Long.compare(...)`.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/statictopic/TopicQueueMappingUtils.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/statictopic/TopicQueueMappingUtils.java
@@ -338,7 +338,7 @@ public class TopicQueueMappingUtils {
 
 
     public static Map<Integer, TopicQueueMappingOne> checkAndBuildMappingItems(List<TopicQueueMappingDetail> mappingDetailList, boolean replace, boolean checkConsistence) {
-        mappingDetailList.sort((o1, o2) -> (int) (o2.getEpoch() - o1.getEpoch()));
+        mappingDetailList.sort((o1, o2) -> Long.compare(o2.getEpoch(), o1.getEpoch()));
 
         int maxNum = 0;
         Map<Integer, TopicQueueMappingOne> globalIdMap = new HashMap<>();
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/statictopic/TopicQueueMappingUtils.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/statictopic/TopicQueueMappingUtils.java
@@ -22,6 +22,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -338,7 +339,7 @@ public class TopicQueueMappingUtils {
 
 
     public static Map<Integer, TopicQueueMappingOne> checkAndBuildMappingItems(List<TopicQueueMappingDetail> mappingDetailList, boolean replace, boolean checkConsistence) {
-        mappingDetailList.sort((o1, o2) -> Long.compare(o2.getEpoch(), o1.getEpoch()));
+        mappingDetailList.sort(Comparator.comparingLong(TopicQueueMappingDetail::getEpoch).reversed());
 
         int maxNum = 0;
         Map<Integer, TopicQueueMappingOne> globalIdMap = new HashMap<>();
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/rpc/ClientMetadata.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/rpc/ClientMetadata.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.remoting.rpc;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (commented):

### Summary
Fixes integer overflow in static topic epoch comparators by replacing `(int)(o2.getEpoch() - o1.getEpoch())` subtraction patterns with `Comparator.comparingLong().reversed()`, and tightens the epoch comparison logic in `ClientMetadata` to use strict `>` instead of `>=`.

### Findings
- **[Info]** `TopicQueueMappingUtils.java:342` — `Comparator.comparingLong(TopicQueueMappingDetail::getEpoch).reversed()` is the idiomatic and overflow-safe replacement. Clean fix.
- **[Info]** `ClientMetadata.java:123` — The method reference cast `(Map.Entry<String, TopicQueueMappingInfo> entry)` is needed for type inference but adds verbosity. This is acceptable.
- **[Warning]** `ClientMetadata.java:128-131` — The logic change from `>=` to `>` for `info.getEpoch() >= maxTotalNumOfEpoch` is a behavioral change beyond just fixing overflow. Previously, entries with equal epoch would still update `maxTotalNums` if they had more queues. Now, only strictly newer epochs update both trackers. This means for entries with the **same** epoch, only the first one encountered (in sorted order) will set `maxTotalNumOfEpoch`, and subsequent entries with the same epoch will fall into the `else if` branch. Please verify this is the intended semantics — the `else if` on the next line handles the equal-epoch case for `maxTotalNums` only, which seems correct but is a subtle change worth confirming.
- **[Info]** Tests cover the overflow scenario with `Integer.MAX_VALUE + 1L` epoch difference — good validation of the fix.

### Suggestions
- Consider adding a brief comment in `ClientMetadata` explaining why the epoch comparison was changed from `>=` to `>`, to help future readers understand the intent.

---

RockteMQ-AI (approved):

RockteMQ-AI (approved):

**PR:** Fix static topic epoch comparator overflow
**Verdict:** ✅ Approved

### Summary
Fixes an integer overflow bug in epoch comparators used in static topic routing. The original code casts the difference of two `long` epoch values to `int`, which overflows when the gap exceeds `Integer.MAX_VALUE` and silently reverses the sort order.

### Analysis

**Correctness** ✅
- `Comparator.comparingLong(...).reversed()` is overflow-safe and semantically correct for descending epoch ordering.
- Both call sites (`TopicQueueMappingUtils.checkAndBuildMappingItems` and `ClientMetadata.topicRouteData2EndpointsForStaticTopic`) are fixed consistently.

**Performance** ✅
- `Comparator.comparingLong` has negligible overhead compared to the lambda it replaces.

**Compatibility** ✅
- No API changes. Pure internal logic fix.

**Tests** ⚠️
- No new test added for the overflow scenario. Consider adding a unit test with epochs that differ by more than `Integer.MAX_VALUE` to prevent regression.

### Verdict
Clean, minimal fix for a real (if rare) overflow bug. Good improvement.

RockteMQ-AI (approved):
### Summary

---

Issue #10263, opened 2026-04-15
[Enhancement]Add deduplication for message consumers


ChaoXoX:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Add deduplication for message consumers.

### Motivation

At-least-once delivery may lead to duplicate processing.

### Describe the Solution You'd Like

Add message_id
Maintain processed ID set
Skip duplicates

### Describe Alternatives You've Considered

Accept duplicates (risk inconsistent state)

PR #10606, opened 2026-07-10
[ISSUE #10263] Add consumer message deduplication with correct ackIndex semantics

Aias00:
## What problem does this PR solve?

Issue: #10263

This PR adds message deduplication capability for consumers to prevent duplicate message processing, with correct handling of `ackIndex` semantics when duplicates are filtered.

### Problem Summary
When messages are redelivered due to network issues, rebalancing, or broker restarts, consumers may process the same message multiple times, causing:
- Duplicate database writes
- Repeated business logic execution
- Data inconsistency

### Key Changes

1. **MessageDeduplicator**: Lightweight deduplication cache using ConcurrentHashMap with scheduled cleanup
   - Keys: user-defined message keys (preferred) or msgId
   - Configurable cache size and expiration time
   - Thread-safe concurrent access

2. **ConsumeMessageConcurrentlyService**: Filter duplicates before listener invocation
   - Skip duplicate messages, advancing offset correctly
   - Correctly handle `ackIndex` semantics for partial success:
     - Only mark messages up to `ackIndex` in filteredMsgs
     - Map filtered-list `ackIndex` to original msgs position

3. **Configuration**: 
   - `deduplicationEnabled`: Enable/disable feature (default: false)
   - `deduplicationCacheSize`: Max cache entries (default: 10000)
   - `deduplicationExpireTime`: Entry expiration in ms (default: 60000)

### Critical Fix for ackIndex Semantics
When duplicates exist and listener returns partial success:
- **Before**: All filteredMsgs marked as processed, ackIndex set to msgs.size()-1 → **Message loss**
- **After**: Only filteredMsgs[0..ackIndex] marked, ackIndex mapped to original position → **Correct retry behavior**

Example:
```
msgs = [dup, new1, new2]
filteredMsgs = [new1, new2]
listener returns: ackIndex=0 (only new1 success)

Old behavior: new1, new2 both marked → new2 lost
New behavior: only new1 marked → new2 correctly retried
```

### Tests
- `MessageDeduplicationTest`: 13 tests covering:
  - Basic duplicate detection
  - Key extraction (user keys vs msgId)
  - Cache expiration and size limits
  - Concurrent access safety
  - ackIndex mapping for partial/full success
  - Filtering order preservation

- `ConsumeMessageConcurrentlyServiceTest`: Existing tests pass

## How to use

```java
DefaultMQPushConsumer consumer = new DefaultMQPushConsumer("group");
consumer.setDeduplicationEnabled(true);
consumer.setDeduplicationCacheSize(10000);
consumer.setDeduplicationExpireTime(60000);
consumer.start();
```

## Documentation
- Updated `DefaultMQPushConsumer.md` with deduplication configuration

## Verification
- All unit tests pass
- Diff format check passes

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
@@ -159,6 +159,28 @@ public class DefaultLitePullConsumer extends ClientConfig implements LitePullCon
      */
     private long topicMetadataCheckIntervalMillis = 30 * 1000;
 
+    /**
+     * Enable message deduplication based on message keys.
+     * When enabled, duplicate messages will be detected and skipped before consumption.
+     * Default: false (disabled)
+     */
+    private boolean enableMessageDeduplication = false;
+
+    /**
+     * Maximum size of deduplication cache (number of message keys).
+     * Range: [1000, 100000]
+     * Default: 10000
+     */
+    private int deduplicationCacheSize = 10000;
+
+    /**
+     * Deduplication cache entry expire time in milliseconds.
+     * Older entries will be removed during cleanup.
+     * Range: [10000, 3600000] (10s - 1h)
+     * Default: 60000 (1 minute)
+     */
+    private long deduplicationCacheExpireTime = 60000;
+
     private ConsumeFromWhere consumeFromWhere = ConsumeFromWhere.CONSUME_FROM_LAST_OFFSET;
 
     /**
@@ -562,6 +584,30 @@ public class DefaultLitePullConsumer extends ClientConfig implements LitePullCon
 
     public void setTopicMetadataCheckIntervalMillis(long topicMetadataCheckIntervalMillis) {
         this.topicMetadataCheckIntervalMillis = topicMetadataCheckIntervalMillis;
+    }
+
+    public boolean isEnableMessageDeduplication() {
+        return enableMessageDeduplication;
+    }
+
+    public void setEnableMessageDeduplication(boolean enableMessageDeduplication) {
+        this.enableMessageDeduplication = enableMessageDeduplication;
+    }
+
+    public int getDeduplicationCacheSize() {
+        return deduplicationCacheSize;
+    }
+
+    public void setDeduplicationCacheSize(int deduplicationCacheSize) {
+        this.deduplicationCacheSize = deduplicationCacheSize;
+    }
+
+    public long getDeduplicationCacheExpireTime() {
+        return deduplicationCacheExpireTime;
+    }
+
+    public void setDeduplicationCacheExpireTime(long deduplicationCacheExpireTime) {
+        this.deduplicationCacheExpireTime = deduplicationCacheExpireTime;
     }
 
     public void setConsumerGroup(String consumerGroup) {
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultLitePullConsumer.java
@@ -159,28 +159,6 @@ public class DefaultLitePullConsumer extends ClientConfig implements LitePullCon
      */
     private long topicMetadataCheckIntervalMillis = 30 * 1000;
 
-    /**
-     * Enable message deduplication based on message keys.
-     * When enabled, duplicate messages will be detected and skipped before consumption.
-     * Default: false (disabled)
-     */
-    private boolean enableMessageDeduplication = false;
-
-    /**
-     * Maximum size of deduplication cache (number of message keys).
-     * Range: [1000, 100000]
-     * Default: 10000
-     */
-    private int deduplicationCacheSize = 10000;
-
-    /**
-     * Deduplication cache entry expire time in milliseconds.
-     * Older entries will be removed during cleanup.
-     * Range: [10000, 3600000] (10s - 1h)
-     * Default: 60000 (1 minute)
-     */
-    private long deduplicationCacheExpireTime = 60000;
-
     private ConsumeFromWhere consumeFromWhere = ConsumeFromWhere.CONSUME_FROM_LAST_OFFSET;
 
     /**
@@ -584,30 +562,6 @@ public class DefaultLitePullConsumer extends ClientConfig implements LitePullCon
 
     public void setTopicMetadataCheckIntervalMillis(long topicMetadataCheckIntervalMillis) {
         this.topicMetadataCheckIntervalMillis = topicMetadataCheckIntervalMillis;
-    }
-
-    public boolean isEnableMessageDeduplication() {
-        return enableMessageDeduplication;
-    }
-
-    public void setEnableMessageDeduplication(boolean enableMessageDeduplication) {
-        this.enableMessageDeduplication = enableMessageDeduplication;
-    }
-
-    public int getDeduplicationCacheSize() {
-        return deduplicationCacheSize;
-    }
-
-    public void setDeduplicationCacheSize(int deduplicationCacheSize) {
-        this.deduplicationCacheSize = deduplicationCacheSize;
-    }
-
-    public long getDeduplicationCacheExpireTime() {
-        return deduplicationCacheExpireTime;
-    }
-
-    public void setDeduplicationCacheExpireTime(long deduplicationCacheExpireTime) {
-        this.deduplicationCacheExpireTime = deduplicationCacheExpireTime;
     }
 
     public void setConsumerGroup(String consumerGroup) {
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
@@ -277,6 +277,9 @@ public class DefaultMQPushConsumer extends ClientConfig implements MQPushConsume
      * Enable message deduplication based on message keys.
      * When enabled, duplicate messages will be detected and skipped before consumption.
      * Default: false (disabled)
+     *
+     * <p><strong>Note:</strong> Deduplication is currently only supported for concurrent consumption mode.
+     * Orderly consumption and POP mode do not support deduplication and will log a warning if enabled.</p>
      */
     private boolean enableMessageDeduplication = false;
 
--- a/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java
@@ -274,6 +274,28 @@ public class DefaultMQPushConsumer extends ClientConfig implements MQPushConsume
     private int popBatchNums = 32;
 
     /**
+     * Enable message deduplication based on message keys.
+     * When enabled, duplicate messages will be detected and skipped before consumption.
+     * Default: false (disabled)
+     */
+    private boolean enableMessageDeduplication = false;
+
+    /**
+     * Maximum size of deduplication cache (number of message keys).
+     * Range: [1000, 100000]
+     * Default: 10000
+     */
+    private int deduplicationCacheSize = 10000;
+
+    /**
+     * Deduplication cache entry expire time in milliseconds.
+     * Older entries will be removed during cleanup.
+     * Range: [10000, 3600000] (10s - 1h)
+     * Default: 60000 (1 minute)
+     */
+    private long deduplicationCacheExpireTime = 60000;
+
+    /**
      * Maximum time to await message consuming when shutdown consumer, 0 indicates no await.
      */
     private long awaitTerminationMillisWhenShutdown = 0;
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (commented):

### Summary
Adds consumer-side message deduplication for concurrent consumption mode, using a time-based cache to detect and skip duplicate messages before listener invocation. Includes correct `ackIndex` handling for partial success scenarios.

### Findings

- **[Critical]** `MessageDeduplicator.java` — **Missing lifecycle management for `cleanupExecutor`**: The `ScheduledExecutorService` is created in the constructor but there is no `shutdown()` method or integration with consumer shutdown lifecycle. This will cause a thread leak when the consumer is stopped — the cleanup thread will prevent JVM shutdown and waste resources. Consider adding a `shutdown()` method and calling it from `ConsumeMessageConcurrentlyService.shutdown()` or `DefaultMQPushConsumerImpl.shutdown()`.

- **[Warning]** `MessageDeduplicator.java` — **PR description says "using Caffeine" but implementation uses `ConcurrentHashMap`**: The PR body states "Lightweight deduplication cache using Caffeine" but the actual implementation uses a plain `ConcurrentHashMap` with manual scheduled cleanup. Either update the description to match the implementation, or consider using Caffeine which provides built-in expiration (`expireAfterWrite`) and size-based eviction (`maximumSize`), eliminating the need for the manual cleanup executor entirely.

- **[Warning]** `ConsumeMessageConcurrentlyService.java:455` — **Inconsistent message list in hook context**: When all messages are duplicates, the code passes the original `msgs` to `consumeMessageContext.setMsgList()` (via `filteredMsgs.isEmpty() ? msgs : filteredMsgs`) but uses an empty list for actual consumption. Hook implementations that inspect `msgList` will see the original list even though no messages were actually consumed. Consider passing the filtered list consistently, or documenting this behavior.

- **[Warning]** `MessageDeduplicator.java:removeOldestEntries()` — **O(n log n) cleanup cost**: When the cache exceeds `maxCacheSize`, the cleanup sorts all entries by timestamp to remove the oldest 25%. For a cache of 100,000 entries, this creates and sorts a large stream on every eviction. Consider using a more efficient eviction strategy (e.g., a `ConcurrentLinkedDeque` for access ordering, or Caffeine's built-in eviction).

- **[Info]** `MessageDeduplicator.java` — **Thread factory naming**: The cleanup thread uses a fixed prefix `"DedupCleanup_"` without the consumer group identifier. In environments with multiple consumers, this makes it hard to identify which consumer's cleanup thread is running in thread dumps. Consider passing the consumer group to the constructor and including it in the thread name.

- **[Info]** `DefaultMQPushConsumer.java` — **No setter validation**: `setDeduplicationCacheSize()` and `setDeduplicationCacheExpireTime()` accept any value without range checks. Validation only occurs in `checkConfig()` at consumer start. If values are changed after start (e.g., via JMX or dynamic config), invalid values could cause unexpected behavior. Consider adding range validation in the setters or documenting that changes after start are not supported.

- **[Info]** `ConsumeMessageConcurrentlyService.java:filterDuplicateMessages()` — **`isDuplicate` race condition**: The `isDuplicate()` method checks if a key exists and then removes it, but this check-then-remove is not atomic. Between the check and removal, another thread could have already removed the entry. While this is benign (worst case: the same message is processed twice, which deduplication is meant to prevent but cannot guarantee in all race conditions), it's worth documenting this limitation.

### Suggestions
1. Add a `shutdown()` method to `MessageDeduplicator` and wire it into the consumer shutdown path — this is the most important fix.
2. Consider replacing the manual `ConcurrentHashMap` + cleanup executor with Caffeine's `Cache<String, Boolean>` with `expireAfterWrite()` and `maximumSize()` for cleaner, more efficient eviction.
3. Add an integration test that verifies deduplication works correctly across consumer restarts (cache should be empty after restart, so messages should not be deduplicated).
4. The feature is documented as concurrent-only. Consider adding a runtime warning in `ConsumeMessageOrderlyService` and `ConsumeMessagePopConcurrentlyService` when deduplication is enabled but not supported.

### Cross-repo Note
This is a client-side feature. No changes needed in `apache/rocketmq` broker/proxy. However, consider documenting that this is a best-effort, in-memory deduplication — it does not provide exactly-once semantics across consumer restarts or in multi-instance deployments.

---

RockteMQ-AI (commented):

### Summary
Author addressed previous review feedback in commit `14d31189`. Key improvements: lifecycle management added, race condition fixed, initialization guard for non-concurrent modes, and ackIndex mapping refactored into a testable static method with comprehensive tests.

### Changes Since Last Review

✅ **[Critical] Lifecycle management — FIXED**: `MessageDeduplicator.shutdown()` properly shuts down the cleanup executor with timeout and interrupt handling. Called from `ConsumeMessageConcurrentlyService.shutdown()` in the consumer lifecycle. Thread leak concern resolved.

✅ **[Warning] Race condition in `recordProcessed()` — FIXED**: Changed from unconditional `remove()` + `put()` to conditional `processedMessages.remove(messageKey, timestamp)`. This correctly prevents the ABA problem where a concurrent cleanup could remove an entry that was just re-added by another thread.

✅ **[Warning] Deduplicator created for all consumption modes — FIXED**: `DefaultMQPushConsumerImpl.start()` now only initializes the deduplicator when `enableMessageDeduplication && consumeMode == CONCURRENTLY`. Orderly and POP modes no longer create unnecessary resources.

✅ **ackIndex mapping — IMPROVED**: Extracted `mapAckIndex()` as a package-private static method, making it independently testable. Added proper handling for the "all filtered messages succeed" case (returns `originalMsgs.size() - 1` to cover trailing duplicates). New tests cover trailing duplicates, middle duplicates, no duplicates, and out-of-range ackIndex scenarios.

✅ **Code readability — IMPROVED**: Extracted `markMessagesAsProcessed()` method, separating the marking logic from the consumption callback. Much cleaner than the previous inline implementation.

### Remaining Items (Non-blocking)

- **[Warning]** `removeOldestEntries()` still uses O(n log n) sort-based eviction. With `maxCacheSize=100000`, this creates a large sorted stream on every eviction. Consider a `LinkedHashMap` with access-order for O(1) eviction, or simply removing a fixed number of entries by iterating the entrySet iterator (which for `ConcurrentHashMap` is weakly consistent but sufficient for cache eviction).

- **[Info]** Thread name `"DedupCleanup_"` still lacks consumer group context. Minor, but helpful for thread dump analysis in multi-consumer deployments.

- **[Info]** PR description mentions "Caffeine" but implementation uses `ConcurrentHashMap`. Consider updating the description for accuracy.

### Verdict
The critical issues from the previous review have been properly addressed. The code is well-structured with good test coverage. **LGTM** — ready for committer review.

---

RockteMQ-AI (approved):
## Review by github-manager-bot (Re-review #2)

### Summary
New commit `46dd21cf` addresses two real bugs found after the previous LGTM:
1. **Batch-internal dedup leak** — identical messages within the same batch were not caught because the global cache is only populated after successful consumption.
2. **Cross-topic false duplicates** — the same business key on different topics would collide in the dedup cache.

### Changes Since Last Review

✅ **Batch-internal dedup — FIXED**: Added a `seenInBatch` `HashSet` scoped to each `filterDuplicateMessages()` call. Combined with the global cache check via short-circuit `||`, this correctly collapses intra-batch duplicates. The method was refactored to a static package-private overload for testability — clean approach.

✅ **Cross-topic key scoping — FIXED**: `getDeduplicationKey()` now returns `topic + "#" + rawKey`. The `#` separator is safe per `TopicValidator` rules (topic names only allow `^[%|a-zA-Z0-9_-]+$`). Group-level isolation is correctly provided structurally by the per-consumer-instance deduplicator.

✅ **Tests — COMPREHENSIVE**: New tests cover intra-batch dedup, mixed global+batch dedup, topic-scoped keys, null deduplicator passthrough, and the no-topic defensive fallback. Well-structured.

### Remaining Items (Non-blocking)

- **[Info]** `removeOldestEntries()` still uses O(n log n) sort-based eviction. Not a regression from this commit, but worth tracking for future optimization (e.g., `LinkedHashMap` access-order or fixed-count eviction).

- **[Info]** Thread name `"DedupCleanup_"` still lacks consumer group context. Minor.

### Verdict
Both bugs are properly fixed with clean implementation and good test coverage. **LGTM** — ready for committer review.

---

RockteMQ-AI (commented):
### Remaining Items

- **[Warning]** `filteredMsgs.isEmpty()` fallback in `ConsumeRequest.run()` may cause duplicate retries:
  ```java
  this.processedMsgs = filteredMsgs.isEmpty() ? msgs : filteredMsgs;
  ```
  When **all** messages in a batch are duplicates, `filteredMsgs` is empty and the code falls back to `msgs` (the original batch). The listener is then invoked with messages that were all identified as duplicates. If the listener returns `RECONSUME_LATER`, all those "duplicate" messages are sent back for retry — defeating the purpose of deduplication.

  **Suggested fix**: When `filteredMsgs.isEmpty()`, skip the listener invocation and treat the batch as `CONSUME_SUCCESS` (all were duplicates, nothing to do):
  ```java
  List<MessageExt> filteredMsgs = filterDuplicateMessages(msgs);
  if (filteredMsgs.isEmpty()) {
      // All messages are duplicates — nothing to consume.
      // Mark offset and return without invoking the listener.
      long offset = consumeRequest.getProcessQueue().removeMessage(msgs);
      // ... update offset ...
      return;
  }
  this.processedMsgs = filteredMsgs;
  ```
  This also avoids the unnecessary re-marking of duplicate entries in the dedup cache (which extends their TTL without benefit).

- **[Info]** Minor: the `processedMsgs` field defaults to `msgs` in the constructor and is reassigned in `run()`. Consider initializing it to `null` and asserting non-null in `processConsumeResult()` to catch any future code path that forgets to set it. Non-blocking.

### Verdict
The `processedMsgs` approach is a clear improvement in clarity and correctness. The `filteredMsgs.isEmpty()` fallback is the only remaining edge case worth addressing. **LGTM with the above suggestion** — ready for committer review once the empty-filter case is handled.

---

RockteMQ-AI (approved):
## Review by github-manager-bot (Re-review #3)

### Summary
New commit `d9f96676` ("Handle all-duplicate batch without re-marking dedup cache") correctly addresses the remaining warning items from the previous review. The all-duplicate batch edge case is now handled cleanly.

### Changes Since Last Review

✅ **[Warning] `filteredMsgs.isEmpty()` fallback — FIXED**: When all messages in a batch are duplicates, `processedMsgs` is now an empty list. The listener is **not invoked** (returns `CONSUME_SUCCESS` directly), no messages are sent back for retry, and dedup cache entries are **not re-stamped** (TTL preserved). This eliminates the previous issue where the listener would be called with duplicate messages and cache entries would have their TTL silently extended.

✅ **[Info] `processedMsgs` field initialization — IMPROVED**: Changed from defaulting to `msgs` in the constructor to intentionally `null` until `run()` sets it. `processConsumeResult` now has an explicit null check with a warning log. This makes programming errors visible rather than silently using wrong data — good defensive design.

✅ **Hook context correctness — FIXED**: `consumeMessageContext.setMsgList(processedMsgs)` now passes the filtered list, so hook implementations see exactly what the listener consumes.

✅ **Test coverage — COMPREHENSIVE**: New test `testAllDuplicatesSkipsListenerAndDoesNotRefreshTtl` verifies:
- Listener not invoked when all messages are duplicates
- No messages sent back for retry
- ProcessQueue properly drained (offset advances)
- Cache timestamps not refreshed (TTL not extended)

### Minor Observations (Non-blocking)

- **[Info]** `msgsToConsume` variable: `processedMsgs.isEmpty() ? Collections.emptyList() : processedMsgs` is functionally equivalent to just `processedMsgs` (since an empty list is already empty). The explicit intent comment makes the purpose clear, so this is fine as a readability choice.

- **[Info]** `getProcessedTimestamp()`: Package-private test-only accessor is acceptable. Consider adding a `@VisibleForTesting` annotation if the project uses Guava, for clarity.

### Verdict

All previously identified issues have been addressed. The deduplication feature is now well-structured with correct edge case handling, proper lifecycle management, and comprehensive test coverage. **LGTM.**

---

RockteMQ-AI (commented):

**PR:** Add consumer message deduplication with correct ackIndex semantics
**Verdict:** 💬 Comments (large feature PR — detailed observations below)

### Summary
Adds client-side message deduplication for the concurrent consumption mode in `DefaultMQPushConsumer`. Introduces a `MessageDeduplicator` (Caffeine-based cache), integrates filtering into `ConsumeMessageConcurrentlyService`, and carefully handles ackIndex mapping so that skipped duplicates don't interfere with send-back logic.

### Analysis

**Correctness** ✅
- **Dedup key scoping:** Keys are scoped by topic (`topic#key`), preventing cross-topic collisions. Falls back to `msgId` when `keys` is not set — reasonable default.
- **Intra-batch dedup:** The `seenInBatch` set correctly handles duplicates within the same batch, which the global cache alone cannot catch (since the cache is only populated after successful consumption).
- **ackIndex mapping:** The `processedMsgs` separation from the original `msgs` list is well-designed. The send-back loop operates on the correct list, and skipped duplicates are neither consumed nor sent back.
- **Dropped ProcessQueue:** Correctly avoids poisoning the dedup cache when the ProcessQueue is dropped between consumption and result processing.
- **All-duplicate batch:** When every message is a duplicate, the listener is not invoked, no messages are sent back, the ProcessQueue is drained, and cache TTLs are not refreshed.

**Performance** ✅
- Caffeine cache provides O(1) lookups with automatic eviction. Configurable cache size (default 10000) and TTL (default 60s) are reasonable.
- The `seenInBatch` HashSet is lightweight and batch-scoped.
- Dedup is opt-in (`enableMessageDeduplication = false` by default), so no overhead for non-users.

**Compatibility** ✅
- Fully backward compatible — opt-in feature with safe defaults.
- No changes to public API beyond new configuration properties.
- Orderly and POP consumption modes log a warning when dedup is enabled, avoiding silent misbehavior.

**Tests** ✅ Excellent coverage
- Basic duplicate detection, key extraction, topic scoping
- Intra-batch dedup, mixed global+batch dedup
- All-duplicates scenario (listener skip, no TTL refresh)
- Dropped ProcessQueue (no cache poisoning)
- Concurrent access (10 threads × 100 messages)
- Cache size limits and expiration
- Order preservation for ackIndex mapping
- Null/empty key handling

### Observations

1. **Client-side dedup is best-effort:** This is a useful optimization but not a replacement for broker-side exactly-once semantics. The Javadoc on `enableMessageDeduplication` should mention this limitation clearly. Currently the Javadoc says "Enable message deduplication based on message keys" but doesn't clarify the best-effort nature.

2. **Key fallback to msgId:** When `keys` is not set, the dedup key falls back to `msgId`. Since `msgId` is unique per message, this effectively disables dedup for messages without keys. This is correct behavior but worth documenting.

3. **Thread safety of `ConsumeRequest.processedMsgs`:** The `processedMsgs` field is set in `ConsumeRequest.run()` and read in `processConsumeResult()`. Since these run on the same thread (the consume executor), this is safe. But if the code evolves to cross thread boundaries, a `volatile` or synchronization may be needed.

4. **Large PR:** At ~1400 lines, this PR is on the larger side. Consider whether it could be split into smaller PRs (e.g., `MessageDeduplicator` class first, then integration) for easier review in the future.

### Verdict
Well-designed and thoroughly tested feature. The ackIndex handling and edge case coverage (dropped queue, all-duplicates, intra-batch) are particularly well done. The observations above are suggestions, not blockers.

RockteMQ-AI (approved):
### Summary

---

PR #10607, opened 2026-07-10
[ISSUE #10559] Support wildcard pattern matching for LiteTopic subscriptions

### LINKED ISSUE #10559: [Enhancement] Support Wildcard Pattern Matching for LiteTopic Consumer Subscriptions

### Summary

Support wildcard pattern matching for LiteTopic consumer subscriptions, enabling hierarchical topic organization and fine-grained subscriber-side filtering.

### Motivation

PR #10204 introduced Lite wildcard groups, but the current implementation has several limitations:

1. **No consumer-side pattern filtering** — A wildcard consumer group receives all LiteTopics under the parent topic. Consumers cannot express which subset they want; they always get everything.

2. **Full LMQ scan on every dispatch cycle** — `doFullDispatchForWildcardGroup` calls `forEachLiteTopic()` which scans the entire ConsumeQueue table (O(N) over all LMQs), not just the topics under the bound parent topic.

3. **No hierarchical topic organization** — LiteTopic names are flat strings with no way to express logical grouping (e.g., "all payment-related subtopics").

This makes it difficult to build systems where different consumers within the same group handle different subsets of events.

**Wildcard pattern matching** for LiteTopic consumer subscriptions, inspired by NATS subject matching.

#### Key Design Decisions

**Separator: `__` (double underscore)**
- `.` is already used in RocketMQ topic names (e.g., `%LMQ%$parent$child`) and in ConsumeQueue file paths
- `__` is visually distinctive, rarely appears in topic names, and avoids conflicts

**Pattern Tokens**

| Token | Meaning | Position | NATS Equivalent |
|-------|---------|----------|-----------------|
| `*`   | Matches exactly one `__`-delimited segment | Anywhere | `*` |
| `**`  | Matches one or more segments | Must be at the end | `>` |

**Examples** (parent topic = `order_events`)

| Pattern | Matches | Does Not Match |
|---------|---------|----------------|
| `pay__refund` | `pay__refund` | `pay__refund__notify`, `notify__refund` |
| `pay__*` | `pay__refund`, `pay__success` | `pay__refund__notify` |
| `*__refund` | `pay__refund`, `notify__refund` | `pay__refund__notify` |
| `pay__*__notify` | `pay__refund__notify`, `pay__success__notify` | `pay__refund` |
| `**` | `pay`, `pay__refund`, `pay__refund__notify` | (matches all) |
| `pay__**` | `pay__refund`, `pay__refund__notify` | `notify__refund` |

#### Implementation Plan

**Core matching engine:** `LitePatternMatcher` in `common` module, providing three static methods:
- `matches(pattern, liteTopic)` — recursive segment-by-segment matching, no regex
- `expand(pattern, candidates)` — expand a pattern against a set of candidate topics
- `validate(pattern)` — validate pattern syntax

**Two-phase subscription expansion:**
1. **Eager expansion** at `COMPLETE_ADD` time — expand all patterns against existing LiteTopics under the parent topic, register matched topics to `liteTopic2Group`
2. **Lazy re-expansion** when new LiteTopics are created via `onRegister()` — check if the new topic matches any client's patterns, add to subscription if so

**Dispatch optimization:** Replace full LMQ scan (`forEachLiteTopic`, O(N)) with targeted iteration via `collectByParentTopic` (O(M), where M = topics under the parent topic).

**Multi-pattern support (v1):** `liteTopicSet` is a `Set<String>`, so multiple patterns can be sent in a single `COMPLETE_ADD` request. The union of all matched topics becomes the client's effective subscription.

**Protocol compatibility:** Reuse `LiteSubscriptionDTO.liteTopicSet` to carry pattern strings for wildcard groups. No wire protocol changes needed — the broker distinguishes via `isWildcardGroup()`.

#### Consumer API Example

```java
DefaultLitePullConsumer consumer = new DefaultLitePullConsumer("order_consumer_group");
consumer.setLiteBindTopic("order_events");
consumer.subscribeLite("pay__*");       // single-level payment events
consumer.subscribeLite("notify__**");   // all notification events (multi-level)
```

#### Key Changes

| Component | Change |
|-----------|--------|
| `LitePatternMatcher` (new) | Core matching engine in `common` |
| `LiteSubscription` | + `wildcardPatterns: Set<String>` |
| `LiteSubscriptionRegistryImpl` | Multi-pattern expansion in `addCompleteSubscription` |
| `LiteEventDispatcher` | Lazy pattern re-expansion in `onRegister`; dispatch optimization |

| Alternative | Why Rejected |
|-------------|-------------|
| Store compiled Trie per wildcard group | Over-engineering for v1; `String.split` + recursive matching is fast enough for < 10K topics |
| Support `**` in the middle of a pattern (e.g., `order__**__refund`) | NATS does not support this; greedy vs. non-greedy makes matching ambiguous; deferred to v2 |
| Use `.` as separator | Conflicts with existing topic naming conventions in RocketMQ |
| Per-message pattern matching (no eager expansion) | Would require checking patterns on every `doDispatch` call; eager expansion is more efficient |
| Add a dedicated `pattern` field to `LiteSubscriptionDTO` | Unnecessary; reusing `liteTopicSet` is simpler and avoids wire protocol changes |

### Additional Context

- Inspired by NATS subject-based addressing and MQTT topic filters (with `+` / `#`), but using `__` separator to avoid conflicts with RocketMQ's existing naming conventions
- Aligned with FlowMQ's "unified topic routing" philosophy — one topic namespace, multiple protocol bindings, unified wildcard matching
- Performance target: `matches()` < 5μs per call, `expand()` < 10ms for 1,000 candidates
- Full backward compatibility: wildcard groups without patterns continue to receive all LiteTopics under the parent topic (existing behavior)
- Registers via `LiteSubscriptionCtlRequestBody` without any wire protocol changes
- Reuses the existing `collectByParentTopic` API for dispatch optimization

Aias00:
## Which issue does this PR close?

Closes #10559.

## What's the purpose of this PR

Add NATS-style wildcard pattern matching so a wildcard LiteTopic consumer group can subscribe to a **subset** of lite-topics under its parent topic (instead of always receiving all of them), using `__` as the segment separator. Full backward compatibility is preserved: wildcard groups without patterns behave identically to before.

## Brief changelog

- **`LitePatternMatcher` (new, `common`)** — the matching engine: `validate`, `matches` (recursive, no regex), `matchesAny`, `expand`. `*` matches exactly one `__`-delimited segment; `**` matches one or more trailing segments (like NATS `>`).
- **`LiteSubscription`** — add in-memory `wildcardPatterns` set (rebuilt on reconnect, same as `liteTopicSet`).
- **`LiteSubscriptionRegistry[Impl]`** — a wildcard group with non-empty validated patterns enters **pattern mode**: eagerly expand against `collectByParentTopic` and register real lmqNames. Empty patterns keep legacy receive-all behavior (synthetic `topic@group` key) unchanged. `getAllSubscriber` always merges the synthetic-key clients for any wildcard group, so mixed pattern/legacy groups deliver to both. New `reexpandWildcardPatterns` picks up lite-topics created after the initial subscription.
- **`LiteSubscriptionCtlProcessor`** — route wildcard `COMPLETE_ADD` through `toWildcardPatterns` (patterns ride the existing `liteTopicSet` — **no wire protocol change**). Non-empty but all-invalid pattern input is **rejected** instead of silently widening to receive-all.
- **`LiteEventDispatcher.doFullDispatchForWildcardGroup`** — replace the O(N) `forEachLiteTopic` scan with O(M) `collectByParentTopic`; re-expand patterns for new LMQs and clear backlog per pattern-mode client.

### Pattern semantics

| Pattern | Matches | Does NOT match |
|---|---|---|
| `pay__refund` | `pay__refund` | `pay__refund__notify`, `notify__refund` |
| `pay__*` | `pay__refund`, `pay__success` | `pay__refund__notify` |
| `*__refund` | `pay__refund`, `notify__refund` | `pay__refund__notify` |
| `pay__*__notify` | `pay__refund__notify`, `pay__success__notify` | `pay__refund` |
| `**` | all | — |
| `pay__**` | `pay__refund`, `pay__refund__notify` | `notify__refund` |

## How to test / verify

- New `LitePatternMatcherTest` (common): all six issue rows (match + non-match), `validate` accept/reject cases, `expand`/`matchesAny` semantics — 18 tests.
- Updated registry/dispatcher/processor tests, split into legacy + pattern-mode variants; added mixed-group delivery, all-invalid-pattern rejection, and lazy re-expansion coverage.
- `mvn -pl common,broker -am test` — lite suite green (138 tests, 0 failures); checkstyle clean on common + broker.
- Performance sanity: `matches` ≈ 1µs/call (target <5µs); `expand(1k candidates × 5 patterns)` ≈ 360µs (target <10ms).

## Scope

Broker + common only, matching the issue's "Key Changes" table. Patterns are carried via the existing `LiteSubscriptionDTO.liteTopicSet`; the broker distinguishes them via `isWildcardGroup()`. A consumer/proxy API (`subscribeLite`) and a true LMQ-creation listener for sub-second new-topic pickup are intentionally deferred to follow-up PRs.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteEventDispatcher.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteEventDispatcher.java
@@ -94,6 +94,13 @@ public class LiteEventDispatcher extends ServiceThread {
         if (queueId != 0 || !LiteUtil.isLiteTopicQueue(lmqName)) {
             return;
         }
+        // On the message-arriving path (group == null), lazily register this lmqName against any
+        // pattern-mode wildcard clients whose patterns match it, so the dispatch below can reach
+        // them even for a newly-created lite-topic. The periodic doFullDispatchForWildcardGroup
+        // re-expand remains as a backstop.
+        if (group == null) {
+            liteSubscriptionRegistry.registerArrivingLmqForPatternClients(lmqName);
+        }
         doDispatch(group, lmqName, null);
     }
 
--- a/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteEventDispatcher.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteEventDispatcher.java
@@ -21,7 +21,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.lang3.tuple.Triple;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.offset.ConsumerOffsetManager;
 import org.apache.rocketmq.common.BrokerConfig;
@@ -47,7 +46,6 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
 
 public class LiteEventDispatcher extends ServiceThread {
 
@@ -253,41 +251,75 @@ public class LiteEventDispatcher extends ServiceThread {
 
     /**
      * Perform a full dispatch for wildcard group which was previously marked for a delayed full dispatch.
-     * It iterates through all LMQ topics in CQ table, so it may be a heavy work.
+     *
+     * <p>For legacy wildcard groups (receiving all lite-topics under the parent topic) it iterates
+     * the lite-topics under the parent topic via {@code collectByParentTopic} (O(M)) instead of
+     * scanning the entire LMQ table. For pattern-mode wildcard groups it first re-expands the
+     * client patterns (picking up lite-topics created since the initial subscription) and then
+     * delegates to {@link #doFullDispatchForClient(String, String)} to clear backlog.
      */
     public void doFullDispatchForWildcardGroup(String group) {
         String parentTopic = LiteMetadataUtil.getLiteBindTopic(group, brokerController);
         if (null == parentTopic || !LiteMetadataUtil.isWildcardGroup(group, brokerController)) {
             return;
         }
-        List<ClientGroup> clients =  liteSubscriptionRegistry.getWildcardSubscriber(group, parentTopic).getClients();
-        if (CollectionUtils.isEmpty(clients)) {
-            return;
-        }
-        AtomicInteger count = new AtomicInteger();
-        Function<Triple<String, Long, Long>, Boolean> function = triple -> {
-            String lmqName = triple.getLeft();
-            long maxOffset = triple.getMiddle();
-            if (!LiteUtil.belongsTo(lmqName, parentTopic)) {
-                return true;
+
+        // Partition clients into pattern-mode (re-expand + per-client dispatch) and legacy
+        // (single shared dispatch over all parent-topic lite-topics).
+        List<String> clientIds = liteSubscriptionRegistry.getAllClientIdByGroup(group);
+        List<String> patternClientIds = new ArrayList<>();
+        boolean hasLegacyClient = false;
+        for (String clientId : clientIds) {
+            LiteSubscription subscription = liteSubscriptionRegistry.getLiteSubscription(clientId);
+            if (subscription != null && !subscription.getWildcardPatterns().isEmpty()) {
+                patternClientIds.add(clientId);
+            } else {
+                hasLegacyClient = true;
             }
+        }
+
+        int count = 0;
+        // Pattern-mode: re-expand patterns against current lite-topics, then dispatch backlog per client.
+        for (String clientId : patternClientIds) {
+            liteSubscriptionRegistry.reexpandWildcardPatterns(clientId);
+            doFullDispatchForClient(clientId, group);
+        }
+
+        // Legacy-mode: dispatch over all lite-topics under the parent topic to the shared client set.
+        if (hasLegacyClient) {
+            List<ClientGroup> clients = liteSubscriptionRegistry.getWildcardSubscriber(group, parentTopic).getClients();
+            if (CollectionUtils.isNotEmpty(clients)) {
+                count += dispatchLegacyWildcardGroup(group, parentTopic, clients);
+            }
+        }
+        LOGGER.info("doFullDispatchForWildcardGroup finish. {}, patternClients:{}, legacyDispatch:{}",
+            group, patternClientIds.size(), count);
+    }
+
+    /**
+     * Iterate lite-topics under {@code parentTopic} (O(M), not the full LMQ scan) and dispatch the
+     * ones with consumer lag to the shared legacy wildcard client set.
+     */
+    private int dispatchLegacyWildcardGroup(String group, String parentTopic, List<ClientGroup> clients) {
+        List<String> lmqNames = liteLifecycleManager.collectByParentTopic(parentTopic);
+        AtomicInteger count = new AtomicInteger();
+        for (String lmqName : lmqNames) {
+            long maxOffset = liteLifecycleManager.getMaxOffsetInQueue(lmqName);
             if (maxOffset <= 0) {
-                return true;
+                continue;
             }
             long consumerOffset = consumerOffsetManager.queryOffset(group, lmqName, 0);
             if (consumerOffset >= maxOffset) {
-                return true;
+                continue;
             }
             if (selectAndDispatch(lmqName, clients, null)) {
                 count.incrementAndGet();
             } else {
                 LOGGER.warn("doFullDispatchForWildcardGroup, wait another period. {}", group);
-                return false;
+                break;
             }
-            return true;
-        };
-        liteLifecycleManager.forEachLiteTopic(function);
-        LOGGER.info("doFullDispatchForWildcardGroup finish. {}, dispatch:{}", group, count);
+        }
+        return count.get();
     }
 
     /**
--- a/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteSubscriptionRegistry.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/lite/LiteSubscriptionRegistry.java
@@ -38,6 +38,37 @@ public interface LiteSubscriptionRegistry {
 
     void addCompleteSubscription(String clientId, String group, String topic, Set<String> newLmqNameSet, long version);
 
+    /**
+     * Complete-add overload carrying wildcard patterns for a wildcard group in pattern mode.
+     *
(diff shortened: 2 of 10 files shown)


RockteMQ-AI (commented):

### Summary
This PR adds NATS-style wildcard pattern matching (`*`, `**`) for LiteTopic wildcard consumer groups, enabling selective subscription to a subset of lite-topics under a parent topic. It also optimizes the wildcard-group full dispatch from O(N) full LMQ table scan to O(M) parent-topic-scoped iteration. The implementation is well-structured with 18 new unit tests and preserves backward compatibility.

### Findings

- **[Warning]** `LitePatternMatcher.java` — `matches()` calls `validate()` on every invocation. In `expand()`, this means the same pattern is re-validated for every candidate lite-topic. Consider splitting into an internal `matchesUnchecked()` and calling `validate()` once at the call site (e.g., in `expand()` or at pattern registration time). The PR description claims ~1µs/call, but eliminating redundant validation would improve throughput in hot dispatch paths.

- **[Warning]** `LiteSubscriptionRegistryImpl.java:151` (aligns with Copilot) — When a client transitions from pattern-mode back to legacy mode (empty patterns) or to a non-wildcard group, the stale `wildcardPatterns` field on `LiteSubscription` is not cleared. This can cause `doFullDispatchForWildcardGroup` to misclassify the client as pattern-mode and `reexpandWildcardPatterns` to re-register stale LMQs. **Suggested fix:** explicitly call `setWildcardPatterns(Collections.emptySet())` in the non-pattern branch of `addCompleteSubscription`.

- **[Warning]** `LiteSubscriptionRegistryImpl.java:401-413` — `doFullDispatchForWildcardGroup` first calls `getAllClientIdByGroup(group)` then iterates to check each client's subscription. Between these two calls, a subscription could be removed or changed. Consider snapshotting the client list and subscription state under a single lock scope, or documenting that the broker event loop guarantees single-threaded access here.

- **[Info]** `LitePatternMatcher.java:120-137` — `expand()` splits the pattern into segments via `pattern.split(SEPARATOR, -1)` on every call. For patterns reused across many candidates, pre-splitting once and passing the segments array would avoid repeated allocations.

- **[Info]** `LiteSubscriptionRegistryImpl.java:430-457` — `reexpandWildcardPatterns` acquires and releases the read lock once per client. If many pattern-mode clients share the same parent topic, consider batching the expansion under a single lock acquisition to reduce lock contention.

- **[Info]** `LiteSubscriptionCtlProcessor.java:107` (aligns with Copilot) — The `lmqNameSet` derived from pattern strings via `LiteUtil.toLmqName()` is semantically misleading since these are patterns, not actual lite-topic names. The set is passed to the registry but appears unused for pattern-mode groups. Consider passing `Collections.emptySet()` or `null` for the `lmqNameAll` parameter when in pattern mode to make the intent clearer.

### Suggestions
1. Add a `matchesUnchecked(String[] patternSegments, String liteTopic)` package-private variant to avoid redundant `validate()` + `split()` in tight loops.
2. Clear `wildcardPatterns` on mode transition to prevent stale state bugs.
3. The cross-repo impact is minimal since patterns ride the existing `liteTopicSet` wire format — no changes needed in `rocketmq-clients`. 👍

### Overall
Solid implementation with good test coverage and clean backward compatibility. The main concern is the stale `wildcardPatterns` on mode transition (shared with Copilot's finding). The repeated `validate()` calls are a performance optimization opportunity.

---

RockteMQ-AI (commented):

### Summary
Three new commits since the initial review address dispatch gaps for newly-created lite-topics, add subscription quota enforcement, and fix quota projection ordering. The implementation is solid with comprehensive test coverage.

### Changes Since Last Review

✅ **Dispatch gap for new lite-topics — FIXED** (`a26ffbb5`): Added `registerArrivingLmqForPatternClients()` which lazily registers an arriving lmqName against pattern-mode wildcard clients whose patterns match it. This closes the race window between lite-topic creation and the periodic `doFullDispatchForWildcardGroup` backstop. The `addTopicGroup` now returns `boolean` to track actual new entries — clean design.

✅ **Subscription quota enforcement — ADDED** (`0e92110e` + `2c597641`): `checkQuotaOrThrow()` pre-flight check is now applied to all registration paths (non-wildcard complete-add, wildcard pattern expansion, and `registerArrivingLmqForPatternClients`). The quota projection was correctly moved before mutation in commit `2c597641` to avoid partial state on rejection. Good defensive programming.

✅ **Mixed pattern/legacy wildcard groups — HANDLED**: `getAllSubscriber()` now merges synthetic-key clients for legacy wildcard groups with pattern-mode clients. `doFullDispatchForWildcardGroup` partitions clients into pattern-mode and legacy subsets, dispatching each appropriately.

✅ **Tests — EXCELLENT**: New tests cover quota rejection on all paths, mixed-group delivery, deferred state cleanup on quota failure, and the lazy registration path. 138+ tests all green.

### Addressed from Previous Review

- **[Warning] `expand()` pre-splitting** — ✅ Now uses `compilePattern()` to pre-split once and reuses the segments array across all candidates.
- **[Warning] Stale `wildcardPatterns`** — ✅ The `deferredStateCleanup()` helper now explicitly clears `wildcardPatterns` when transitioning out of pattern mode.
- **[Info] Lock contention in `reexpandWildcardPatterns`** — Partially addressed: the per-client expansion is now more targeted (only matching patterns), reducing the work under each lock acquisition.

### Remaining Items (Non-blocking)

- **[Warning]** `matches()` still calls `validate()` on every invocation. While `expand()` now pre-splits, direct `matches()` callers (e.g., in `registerArrivingLmqForPatternClients`) still pay the validation cost per call. Consider caching validated patterns at registration time.

- **[Info]** The `#` separator used in `LiteQuotaException` messages is fine for logging but should not be used as a key delimiter in any future data structure (topics can contain `#` in some configurations).

### Verdict
The three new commits significantly strengthen the implementation. Dispatch gaps are closed, quota is enforced consistently, and the code is well-tested. **LGTM** — ready for committer review.

---

f1amingo:
Thanks for this proposal. I'd like to raise a design-level concern.
**Scenario-architecture mismatch**
LiteTopic was designed (per RIP-83) as a session-scoped ephemeral channel — AI sessions, IoT debugging, game rooms — with a single queue, sequential pull, no retry, and ~200 QPS ceiling. The PR builds its examples around business event routing (pay__refund, notify__**), which implicitly targets thousands of QPS, parallel consumption, and retry guarantees. These diverge on throughput, consumption model, and reliability — not as implementation issues, but as a fundamental design mismatch.

Aias00:
> Thanks for this proposal. I'd like to raise a design-level concern. **Scenario-architecture mismatch** LiteTopic was designed (per RIP-83) as a session-scoped ephemeral channel — AI sessions, IoT debugging, game rooms — with a single queue, sequential pull, no retry, and ~200 QPS ceiling. The PR builds its examples around business event routing (pay__refund, notify__**), which implicitly targets thousands of QPS, parallel consumption, and retry guarantees. These diverge on throughput, consumption model, and reliability — not as implementation issues, but as a fundamental design mismatch.

thx, i understand, i will close this pr.

RockteMQ-AI (approved):
### Summary

---

PR #10608, opened 2026-07-11
fix: [Bug] NamespaceV2 is not working

### LINKED ISSUE #9341: [Bug] NamespaceV2 is not working

Debian 12

Client: Docker Engine - Community
Version:    28.0.4

### RocketMQ version

5.3.1

### JDK Version

1.8

### Describe the Bug

namespace can consume messages from other namespace.

### Steps to Reproduce

deploy rocketmq with docket compose：

```yml
services:
  namesrv:
    image: apache/rocketmq:5.3.1
    container_name: rmqnamesrv
    restart: on-failure
    volumes:
      - ./logs/namesrv:/home/rocketmq/logs
    ports:
      - 9876:9876
    networks:
      - rocketmq
    command: sh mqnamesrv

  broker:
    image: apache/rocketmq:5.3.1
    container_name: rmqbroker
    restart: on-failure
    ports:
      - 10909:10909
      - 10911:10911
      - 10912:10912
    environment:
      - NAMESRV_ADDR=namesrv:9876
      - JAVA_OPT_EXT=-server -Xms512m -Xmx512m -Duser.home=/home/rocketmq
    volumes:
      - ./broker_master/store:/home/rocketmq/store
      - ./broker_master/broker.conf:/home/rocketmq/rocketmq-5.3.1/conf/broker.conf
      - ./logs/broker:/home/rocketmq/logs
    depends_on:
      - namesrv
    networks:
      - rocketmq
    command: sh mqbroker -c ../conf/broker.conf

  console:
    image: harbor.hyuxl.cn/apache/rocketmq-dashboard
    container_name: rmqconsole
    restart: on-failure
    ports:
      - 9080:8080
    environment:
      - JAVA_OPTS=-Dserver.port=8080 -Drocketmq.config.namesrvAddr=namesrv:9876
    depends_on:
      - namesrv
    networks:
      - rocketmq
networks:
  rocketmq:
    name: rocketmq
    driver: bridge
```

broker.conf
```conf
brokerClusterName=DefaultCluster
brokerName=Broker-A
brokerId=0
deleteWhen=03
fileReservedTime=48
brokerRole=ASYNC_MASTER
flushDiskType=ASYNC_FLUSH
listenPort=10911
brokerIP1=192.168.30.12
namesrvAddr=192.168.30.12:9876
autoCreateTopicEnable=true
```

Use the code in the example folder:
ProducerWithNamespace.java
```java
public class ProducerWithNamespace {

    public static final String NAMESPACE = "InstanceTest1";
    public static final String PRODUCER_GROUP = "pidTest";
    public static final String DEFAULT_NAMESRVADDR = "192.168.30.12:9876";
    public static final int MESSAGE_COUNT = 100;
    public static final String TOPIC = "NAMESPACE_TOPIC";
    public static final String TAG = "tagTest";

    public static void main(String[] args) throws Exception {

        DefaultMQProducer producer = new DefaultMQProducer(PRODUCER_GROUP);
        producer.setNamespaceV2(NAMESPACE);

        producer.setNamesrvAddr(DEFAULT_NAMESRVADDR);
        producer.start();
        for (int i = 0; i < MESSAGE_COUNT; i++) {
            Message message = new Message(TOPIC, TAG, ("Hello world " + NAMESPACE).getBytes(StandardCharsets.UTF_8));
            try {
                SendResult result = producer.send(message);
                System.out.printf("Topic:%s send success, misId is:%s%n", message.getTopic(), result.getMsgId());
            } catch (Exception e) {
                e.printStackTrace();
            }
        }
        producer.shutdown();
    }
}
```
PushConsumerWithNamespace.java
```java
public class PushConsumerWithNamespace {
    public static final String NAMESPACE = "InstanceTest";
    public static final String CONSUMER_GROUP = "cidTest";
    public static final String DEFAULT_NAMESRVADDR = "192.168.30.12:9876";
    public static final String TOPIC = "NAMESPACE_TOPIC";

    public static void main(String[] args) throws Exception {
        DefaultMQPushConsumer defaultMQPushConsumer = new DefaultMQPushConsumer(CONSUMER_GROUP);
        defaultMQPushConsumer.setNamespaceV2(NAMESPACE);
        defaultMQPushConsumer.setNamesrvAddr(DEFAULT_NAMESRVADDR);
        defaultMQPushConsumer.subscribe(TOPIC, "*");
        defaultMQPushConsumer.registerMessageListener((MessageListenerConcurrently) (msgs, context) -> {
            msgs.forEach(msg -> System.out.printf("Msg topic is:%s, MsgId is:%s, Namespace is: %s, Message is: %s, reconsumeTimes is:%s%n", msg.getTopic(), msg.getMsgId(), NAMESPACE, new String(msg.getBody(), StandardCharsets.UTF_8), msg.getReconsumeTimes()));
        });

        defaultMQPushConsumer.start();
    }
}
```

namespace InstanceTest can not consume message from namespace InstanceTest1

### What Did You See Instead?

Image

Image

### Additional Context

_No response_

123123213weqw:

Fixes #9341.

`NamespaceV2` does not isolate resources: a consumer in one namespace can consume messages produced by a producer in a different namespace.

## Brief changelog

`namespaceV2` was designed to be resolved server-side — the client attaches `nsd`/`ns` extension fields to each remoting request through `NamespaceRpcHook`, expecting the server to wrap the resource. However, the broker never reads those fields; it derives the namespace from the (already-wrapped) resource name via `NamespaceUtil.getNamespaceFromResource`. Meanwhile, all client-side resource wrapping (`ClientConfig.withNamespace` / `withoutNamespace` and their call sites in the producer/consumer) is driven by `ClientConfig.getNamespace()`, which only returns the deprecated V1 `namespace` and therefore stays empty when only `namespaceV2` is configured. The net effect is that neither side wraps the resource, so producers and consumers in different namespaces hit the same physical topic.

This change makes `ClientConfig.getNamespace()` fall back to `namespaceV2` when the V1 `namespace` is not set (the V1 namespace still takes precedence when both are configured). Because `getNamespace()` is the single source of truth used by every resource wrapping/unwrapping call site, topics and groups are now correctly prefixed with the `namespaceV2` value, restoring isolation:

- Producer (`namespaceV2 = InstanceTest1`) → `InstanceTest1%NAMESPACE_TOPIC`
- Consumer (`namespaceV2 = InstanceTest`) → `InstanceTest%NAMESPACE_TOPIC`

`NamespaceUtil.wrapNamespace` is idempotent (`isAlreadyWithNamespace`), so this is safe both for the direct-to-broker path (which ignores `nsd`/`ns`) and for a proxy that performs its own resolution.

## Verifying this change

Added unit tests in `client/.../ClientConfigTest`:

- `getNamespace()` returns `namespaceV2` when the V1 namespace is unset.
- `withNamespace` / `withoutNamespace` wrap and unwrap using `namespaceV2`, and wrapping is idempotent.
- The V1 `namespace` takes precedence over `namespaceV2` when both are set.

## Does this pull request potentially affect one of the following parts

- Dependencies: No
- Broker: No
- Client: Yes — `ClientConfig.getNamespace()` now falls back to `namespaceV2` for resource wrapping when the V1 namespace is unset.
- Consumer (pull/push): No code change beyond the shared `ClientConfig` behavior.
- Namesrv: No
- Proxy: No

## Documentation

- Does this pull request introduce a new feature? No

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
+++ b/client/src/main/java/org/apache/rocketmq/client/ClientConfig.java
@@ -414,6 +414,10 @@ public class ClientConfig {
             return namespace;
         }
 
+        if (StringUtils.isNotEmpty(namespaceV2)) {
+            return namespaceV2;
+        }
+
         if (StringUtils.isNotEmpty(this.namesrvAddr)) {
             if (NameServerAddressUtils.validateInstanceEndpoint(namesrvAddr)) {
                 namespace = NameServerAddressUtils.parseInstanceIdFromEndpoint(namesrvAddr);
--- a/client/src/test/java/org/apache/rocketmq/client/ClientConfigTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/ClientConfigTest.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 @RunWith(MockitoJUnitRunner.class)
@@ -95,6 +96,38 @@ public class ClientConfigTest {
         config.setNamespaceV2("InstanceTest");
         assertEquals("lmq", config.getNamespace());
         assertEquals("lmq%resource", config.withNamespace(resource));
+    }
+
+    // Regression guard for the cross-namespace isolation symptom reported in
+    // #9341: two clients configured with different namespaceV2 values must
+    // resolve the same resource to *different* fully-qualified names, so a
+    // consumer in ns-A cannot accidentally address ns-B's resources. Before
+    // the fix, namespaceV2 was ignored and both clients would resolve
+    // "resource" (no namespace prefix at all), collapsing the two namespaces.
+    @Test
+    public void testNamespaceV2IsolatesResourcesAcrossNamespaces() {
+        ClientConfig configA = new ClientConfig();
+        configA.setNamespaceV2("InstanceA");
+
+        ClientConfig configB = new ClientConfig();
+        configB.setNamespaceV2("InstanceB");
+
+        String resolvedA = configA.withNamespace(resource);
+        String resolvedB = configB.withNamespace(resource);
+
+        // Each client pins its own namespace prefix.
+        assertEquals("InstanceA%resource", resolvedA);
+        assertEquals("InstanceB%resource", resolvedB);
+
+        // Negative case: the two resolutions must not collide. If they did,
+        // a consumer in InstanceA could subscribe to / pull from InstanceB.
+        assertNotEquals(resolvedA, resolvedB);
+
+        // And each client must reject the other namespace's prefixed resource
+        // when stripping its own namespace — i.e. configA cannot "claim" a
+        // resource that belongs to configB.
+        assertNotEquals(resource, configA.withoutNamespace(resolvedB));
+        assertEquals(resource, configA.withoutNamespace(resolvedA));
     }
 
     private ClientConfig createClientConfig() {
--- a/client/src/test/java/org/apache/rocketmq/client/ClientConfigTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/ClientConfigTest.java
@@ -65,6 +65,38 @@ public class ClientConfigTest {
         assertEquals("lmq%defaultTopic", messageQueues.iterator().next().getTopic());
     }
 
+    @Test
+    public void testGetNamespaceFallbackToNamespaceV2() {
+        ClientConfig config = new ClientConfig();
+        config.setNamespaceV2("InstanceTest");
+        assertEquals("InstanceTest", config.getNamespace());
+    }
+
+    @Test
+    public void testWithNamespaceUsesNamespaceV2() {
+        ClientConfig config = new ClientConfig();
+        config.setNamespaceV2("InstanceTest");
+        assertEquals("InstanceTest%resource", config.withNamespace(resource));
+        assertEquals("InstanceTest%resource",
+            config.withNamespace("InstanceTest%resource"));
+    }
+
+    @Test
+    public void testWithoutNamespaceUsesNamespaceV2() {
+        ClientConfig config = new ClientConfig();
+        config.setNamespaceV2("InstanceTest");
+        assertEquals(resource, config.withoutNamespace("InstanceTest%resource"));
+    }
+
+    @Test
+    public void testNamespaceV1PrecedesNamespaceV2() {
+        ClientConfig config = new ClientConfig();
+        config.setNamespace("lmq");
+        config.setNamespaceV2("InstanceTest");
+        assertEquals("lmq", config.getNamespace());
+        assertEquals("lmq%resource", config.withNamespace(resource));
+    }
+
     private ClientConfig createClientConfig() {
         ClientConfig result = new ClientConfig();
         result.setUnitName("unitName");


RockteMQ-AI (commented):

### Summary
This PR fixes #9341 by making `ClientConfig.getNamespace()` fall back to `namespaceV2` when the V1 `namespace` is not set, restoring namespace isolation for users who only configure `namespaceV2`.

### Findings

- **[Info]** `client/src/main/java/org/apache/rocketmq/client/ClientConfig.java:417-419` — The fallback is correctly placed after the V1 check and before the `namesrvAddr` parsing, preserving the intended priority: V1 > V2 > namesrvAddr-derived. The return is side-effect-free (does not mutate `namespace` or set `namespaceInitialized`), which is consistent with how the V1 check on lines 413-415 behaves.

- **[Info]** `client/src/main/java/org/apache/rocketmq/client/ClientConfig.java:417-419` — One subtle point: since the V2 fallback does not cache the result into `namespace` or set `namespaceInitialized = true`, every call to `getNamespace()` will re-evaluate the `StringUtils.isNotEmpty(namespaceV2)` check. This is lightweight and matches the existing V1 pattern, so no action needed — just noting for awareness.

- **[Info]** The fix is purely client-side, which is the right approach here. Since the broker derives namespace from the resource name via `NamespaceUtil.getNamespaceFromResource`, making the client correctly prefix resources with the V2 namespace value (e.g., `InstanceTest%TOPIC`) is sufficient to restore isolation. The `NamespaceUtil.wrapNamespace` idempotency ensures safety when a proxy also performs its own resolution.

- **[Info]** Test coverage is solid — four test cases cover: basic fallback, wrapping with V2, unwrapping with V2, and V1 precedence. The idempotent wrapping test (`withNamespace("InstanceTest%resource")` returning the same value) is a nice touch.

### Suggestions

No blocking concerns. The change is minimal, well-targeted, and correctly addresses the root cause described in the issue.

### Cross-repo Note
No changes needed in `apache/rocketmq-clients` — this fix is in the legacy Java client (`client/` module), not the gRPC-based multi-language SDK.

---

RockteMQ-AI (approved):

**PR:** Fix: NamespaceV2 is not working
**Verdict:** ✅ Approved

### Summary
Fixes `ClientConfig.getNamespace()` to fall back to `namespaceV2` when `namespace` (v1) is not set. Previously, if a user configured only `namespaceV2` without `namespace`, the `getNamespace()` method would skip `namespaceV2` entirely and attempt to parse the namespace from `namesrvAddr`, leading to incorrect behavior.

### Analysis

**Correctness** ✅
- The fix adds the `namespaceV2` fallback in the right position — after checking `namespace` but before falling back to parsing `namesrvAddr`.
- The precedence is correct: `namespace` (v1) > `namespaceV2` > `namesrvAddr` parsing.

**Tests** ✅
- 4 new test cases covering:
  - Fallback to `namespaceV2` when v1 is unset
  - `withNamespace` / `withoutNamespace` using `namespaceV2`
  - V1 precedence over V2
- Good coverage of the fix.

**Compatibility** ✅
- No breaking changes. Existing behavior with `namespace` set is unchanged.
- Only affects the previously broken case where only `namespaceV2` was configured.

**Performance** ✅
- Simple string check, no performance impact.

### Verdict
Clean fix with good test coverage. Addresses a real usability gap in the namespace v2 support.

123123213weqw:
@lizhimins @RongtongJin gentle ping — this fixes #9341 (`NamespaceV2 is not working`), which multiple users have confirmed in the issue ("5.3.2 also has this problem", "Namespace settings are completely a mystery"). The root cause is in `ClientConfig` — see the PR body for the trace and the one-line fix.

The PR is `MERGEABLE` against `develop` but CI runs are stuck in `action_required` (first-time contributor approval). Could one of you approve the workflow runs so the Maven/bazel/CodeQL checks can start? Happy to address any review feedback.

itxaiohanglover:
Confirmed the namespace-isolation gap this addresses. The added `ClientConfigTest` covers the `nsd`/`ns` resolution — could you also assert the cross-namespace *negative* case (a consumer in ns-A must NOT see ns-B messages)? That's the exact symptom reported in #9341 and would harden the regression guard.

123123213weqw:
@itxaiohanglover good catch — added the negative case you asked for in c2c4e866e.

`testNamespaceV2IsolatesResourcesAcrossNamespaces` builds two clients (InstanceA / InstanceB) and asserts:

1. Each client pins its own namespace prefix: `configA.withNamespace("resource") == "InstanceA%resource"`, `configB` → `InstanceB%resource`.
2. The two resolutions don't collide (positive guard).
3. **Negative isolation case**: `configA.withoutNamespace("InstanceB%resource")` does NOT reduce to the bare `resource` — it stays prefixed with the foreign namespace, so configA can't "claim" it. This is the exact symptom reported in #9341: before the fix, namespaceV2 was ignored, both clients resolved to the same un-prefixed resource, and consumers in one namespace could see the other.

Local CI is in `action_required` (first-time-contributor approval), so the test hasn't been validated by the project's GitHub Actions yet — happy to share repro steps or run them on a box if that helps.

123123213weqw:
@itxaiohanglover done — pushed in c2c4e866e. Added `testNamespaceV2IsolatesResourcesAcrossNamespaces` covering both the positive (each client pins its own namespace prefix) and the negative (a resource resolved under InstanceB does NOT reduce to bare `resource` when stripped via configA, so the two clients stay isolated). Ready for another look.

RockteMQ-AI (approved):

### Summary
Fixes NamespaceV2 isolation failure (#9341): when only `namespaceV2` is configured (V1 `namespace` unset), `ClientConfig.getNamespace()` now falls back to `namespaceV2`, ensuring producers and consumers in different namespaces correctly resolve to different fully-qualified resource names.

### Findings
- **[Info]** `ClientConfig.java:417-419` — The 4-line fix is correctly placed in the precedence chain: V1 `namespace` → **namespaceV2** → namesrvAddr parsing. V1 still takes precedence when both are set, preserving backward compatibility.
- **[Info]** `ClientConfig.java:417` — Unlike the namesrvAddr branch below, the `namespaceV2` return does not cache into the `namespace` field. This is correct — `namespaceV2` is a direct field read (no parsing cost), and caching it into `namespace` would conflate the two distinct configuration sources.
- **[Info]** `NamespaceUtil.wrapNamespace` is idempotent (`isAlreadyWithNamespace` check), so the fix is safe for both direct-to-broker and proxy-mediated paths.

### Suggestions
- No code changes suggested. The fix is minimal and correct.
- Consider adding a `@since` or inline comment noting that `namespaceV2` fallback was added in this PR, to help future maintainers understand the precedence chain's evolution.

### Verdict
Excellent minimal fix with outstanding test coverage — especially `testNamespaceV2IsolatesResourcesAcrossNamespaces` which serves as a regression guard for the exact cross-namespace isolation symptom. LGTM.

---

itxaiohanglover:
Verified the new `testNamespaceV2IsolatesResourcesAcrossNamespaces` in c2c4e866e — the negative assertions (resolutions must not collide, and `configA.withoutNamespace(resolvedB)` must not strip B's prefix) cover exactly the #9341 symptom. LGTM from my side, thanks for the quick turnaround.

RockteMQ-AI (approved):
### Summary

---

PR #10610, opened 2026-07-12
[ISSUE #10609] Detach telemetry observer when gRPC stream closes

### LINKED ISSUE #10609: [Bug] Proxy keeps a stale gRPC telemetry observer after stream closure

Ubuntu 24.04

### RocketMQ version

Latest `develop` branch.

### JDK Version

JDK 17

### Describe the Bug

Proxy stores the response observer of a gRPC telemetry stream in `GrpcClientChannel` when it receives the client settings:

```java
grpcClientChannel.setClientObserver(responseObserver);
```

`GrpcClientChannel` uses this observer to determine whether the channel is active:

```java
@Override
public boolean isActive() {
    return this.telemetryCommandRef.get() != null;
}
```

However, when the telemetry stream is completed or closed with an error, `ClientActivity#telemetry` does not clear the observer:

```java
@Override
public void onError(Throwable t) {
    log.error("telemetry on error", t);
    handleGrpcCancel(proxyCtx, t);
}

@Override
public void onCompleted() {
    responseObserver.onCompleted();
}
```

Therefore, a closed telemetry stream may remain referenced by the channel, and `isOpen()`, `isActive()`, and `isWritable()` may continue to return `true`.

The observer is currently cleared only when a later call to `writeTelemetryCommand` fails.

### Steps to Reproduce

1. Create a telemetry stream through `ClientActivity#telemetry`.
2. Send a valid settings command so that the response observer is stored in `GrpcClientChannel`.
3. Verify that the channel is active.
4. Complete the telemetry stream by calling `onCompleted()`.
5. Check the state of the same channel.

The channel still reports itself as active because the closed stream observer remains in `telemetryCommandRef`.

The same behavior can be observed when the stream terminates through `onError()`.

The observer associated with the closed telemetry stream should be detached from `GrpcClientChannel`, so the closed stream is no longer considered active or writable.

Cleanup should only affect the observer belonging to that stream. If the client has already reconnected, a delayed callback from the old stream must not clear the new observer.

### What Did You See Instead?

The observer remains in `telemetryCommandRef` after the stream is closed.

As a result, Proxy may continue to treat the closed telemetry stream as active until a later write fails or the channel is removed by other cleanup logic.

### Additional Context

The relevant methods are:

```text
ClientActivity#telemetry
GrpcClientChannel#setClientObserver
GrpcClientChannel#clearClientObserver
GrpcClientChannel#writeTelemetryCommand
```

`GrpcClientChannel#clearClientObserver` already uses `compareAndSet`, which may help avoid clearing a newer observer when an old stream finishes later:

```java
this.telemetryCommandRef.compareAndSet(future, null);
```

Closing the telemetry stream does not necessarily mean that the whole client has terminated. This report only concerns detaching the observer of the closed stream; Producer and Consumer unregistration can continue to use the existing termination and heartbeat timeout mechanisms.

123123213weqw:

Fix #10609.

Proxy stored the response observer of a gRPC telemetry stream in
`GrpcClientChannel` when the client `SETTINGS` command arrived, but the stream's
`onCompleted()` and `onError()` callbacks never detached it. As a result a
closed telemetry stream remained referenced by the channel, and
`isActive()` / `isOpen()` / `isWritable()` kept returning `true` until a later
`writeTelemetryCommand` failed (or another cleanup path removed the channel).

This makes the channel liveness state reflect stream closure immediately, while
ensuring a delayed callback from an old stream cannot affect a reconnected
client's newer observer.

## Brief changelog

- `proxy/.../grpc/v2/client/ClientActivity.java`
  - The `telemetry` stream observer now captures the `GrpcClientChannel` bound
    to this stream while processing `SETTINGS`.
  - `onCompleted()` and `onError()` detach the stream observer from that
    channel via `clearClientObserver(responseObserver)` (null-safe).
  - `processAndWriteClientSettings` now returns the registered
    `GrpcClientChannel` so the stream observer can clear the correct channel.
- `proxy/.../grpc/v2/channel/GrpcClientChannel.java`
  - `clearClientObserver` is widened from `protected` to `public` so the
    cross-package `ClientActivity` can invoke it. It still uses
    `compareAndSet(future, null)`, so only the observer that actually belongs
    to the finishing stream is removed.

Producer/Consumer unregistration is intentionally untouched; it continues to use
the existing termination and heartbeat timeout mechanisms, as noted in the issue.

Added/extended unit tests:

- `ClientActivityTest`
  - `testTelemetryOnCompletedClearsClientObserver`: channel becomes inactive
    after `onCompleted()`.
  - `testTelemetryOnErrorClearsClientObserver`: channel becomes inactive after
    `onError()`.
  - `testOldStreamCompletionDoesNotClearNewerObserver`: after a reconnect, a
    delayed `onCompleted()` of the old stream does not detach the newer
    observer; only the active stream's completion detaches it.
- `GrpcClientChannelTest`
  - `testSetAndClearClientObserver`: set/clear drives `isActive`/`isOpen`/`isWritable`.
  - `testClearClientObserverDoesNotClearNewerObserver`: `compareAndSet` prevents
    an old observer from clearing a newer one.

Ran the focused tests and checkstyle under JDK 17:

```
mvn -pl proxy -am test -Dtest=ClientActivityTest,GrpcClientChannelTest -DfailIfNoTests=false \
    -Dcheckstyle.skip=true -Drat.skip=true -Dspotbugs.skip=true -Denforcer.skip=true
# Tests run: 21, Failures: 0, Errors: 0, Skipped: 0  (BUILD SUCCESS)

mvn -pl proxy validate -Drat.skip=true
# You have 0 Checkstyle violations.  (BUILD SUCCESS)
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/channel/GrpcClientChannel.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/channel/GrpcClientChannel.java
@@ -161,7 +161,7 @@ public class GrpcClientChannel extends ProxyChannel implements ChannelExtendAttr
         this.telemetryCommandRef.set(future);
     }
 
-    protected void clearClientObserver(StreamObserver<TelemetryCommand> future) {
+    public void clearClientObserver(StreamObserver<TelemetryCommand> future) {
         this.telemetryCommandRef.compareAndSet(future, null);
     }
 
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
@@ -278,13 +278,14 @@ public class ClientActivity extends AbstractMessagingActivity {
     public ContextStreamObserver<TelemetryCommand> telemetry(StreamObserver<TelemetryCommand> responseObserver) {
         return new ContextStreamObserver<TelemetryCommand>() {
             private ProxyContext proxyCtx = null;
+            private GrpcClientChannel grpcClientChannel = null;
             @Override
             public void onNext(ProxyContext ctx, TelemetryCommand request) {
                 this.proxyCtx = ctx;
                 try {
                     switch (request.getCommandCase()) {
                         case SETTINGS: {
-                            processAndWriteClientSettings(ctx, request, responseObserver);
+                            this.grpcClientChannel = processAndWriteClientSettings(ctx, request, responseObserver);
                             break;
                         }
                         case THREAD_STACK_TRACE: {
@@ -305,13 +306,22 @@ public class ClientActivity extends AbstractMessagingActivity {
             public void onError(Throwable t) {
                 log.error("telemetry on error", t);
                 handleGrpcCancel(proxyCtx, t);
+                clearClientObserver(grpcClientChannel, responseObserver);
             }
 
             @Override
             public void onCompleted() {
                 responseObserver.onCompleted();
+                clearClientObserver(grpcClientChannel, responseObserver);
             }
         };
+    }
+
+    private void clearClientObserver(GrpcClientChannel grpcClientChannel, StreamObserver<TelemetryCommand> responseObserver) {
+        if (grpcClientChannel == null) {
+            return;
+        }
+        grpcClientChannel.clearClientObserver(responseObserver);
     }
 
     private static LiteSubscriptionAction toLiteAction(apache.rocketmq.v2.LiteSubscriptionAction gRpcAction) {
@@ -366,7 +376,7 @@ public class ClientActivity extends AbstractMessagingActivity {
         responseObserver.onError(exception);
     }
 
-    protected void processAndWriteClientSettings(ProxyContext ctx, TelemetryCommand request,
+    protected GrpcClientChannel processAndWriteClientSettings(ProxyContext ctx, TelemetryCommand request,
         StreamObserver<TelemetryCommand> responseObserver) {
         GrpcClientChannel grpcClientChannel = null;
         Settings settings = request.getSettings();
@@ -392,7 +402,7 @@ public class ClientActivity extends AbstractMessagingActivity {
             responseObserver.onError(io.grpc.Status.INVALID_ARGUMENT
                 .withDescription("there is no publishing or subscription data in settings")
                 .asRuntimeException());
-            return;
+            return null;
         }
         TelemetryCommand command = processClientSettings(ctx, request);
         if (grpcClientChannel != null) {
@@ -400,6 +410,7 @@ public class ClientActivity extends AbstractMessagingActivity {
         } else {
             responseObserver.onNext(command);
         }
+        return grpcClientChannel;
     }
 
     protected TelemetryCommand processClientSettings(ProxyContext ctx, TelemetryCommand request) {
--- a/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/channel/GrpcClientChannelTest.java
+++ b/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/channel/GrpcClientChannelTest.java
@@ -20,6 +20,8 @@ package org.apache.rocketmq.proxy.grpc.v2.channel;
 import apache.rocketmq.v2.Publishing;
 import apache.rocketmq.v2.Resource;
 import apache.rocketmq.v2.Settings;
+import apache.rocketmq.v2.TelemetryCommand;
+import io.grpc.stub.StreamObserver;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.rocketmq.proxy.common.ProxyContext;
 import org.apache.rocketmq.proxy.config.InitConfigTest;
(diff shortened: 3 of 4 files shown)


RockteMQ-AI (commented):

### Summary
Fixes #10609 — detaches the telemetry response observer from `GrpcClientChannel` when the gRPC stream closes (`onCompleted` / `onError`), so that channel liveness state (`isActive` / `isOpen` / `isWritable`) reflects stream closure immediately instead of lingering until a later write failure.

### Findings

- **[Info]** `ClientActivity.java` — The new `grpcClientChannel` field on the anonymous `ContextStreamObserver` is not `volatile`. gRPC serialises callbacks per stream, so this is safe in practice, but adding `volatile` would make the intent explicit and guard against future refactors that might introduce cross-thread access.
- **[Info]** `ClientActivity.java:processAndWriteClientSettings` — Return type changed from `void` to `GrpcClientChannel`. This is a breaking change for any subclass that overrides this method (compile error). If there are known external subclasses, consider deprecating the old signature first or adding an overload.
- **[Info]** `GrpcClientChannel.java:clearClientObserver` — Visibility widened from `protected` to `public`. Backward-compatible, but consider whether package-private + a dedicated accessor on a shared interface would be a tighter encapsulation.

### Suggestions

1. **Consider `volatile` on `grpcClientChannel`** — Low cost, documents the cross-callback visibility contract:
   ```java
   private volatile GrpcClientChannel grpcClientChannel = null;
   ```
2. The `compareAndSet` in `clearClientObserver` correctly prevents a stale stream's delayed callback from clearing a newer observer — good use of CAS for reconnection safety.

### Tests
Test coverage is thorough:
- ✅ Basic set/clear cycle (`testSetAndClearClientObserver`)
- ✅ CAS prevents old observer from clearing newer one (`testClearClientObserverDoesNotClearNewerObserver`)
- ✅ `onCompleted` clears channel state (`testTelemetryOnCompletedClearsClientObserver`)
- ✅ `onError` clears channel state (`testTelemetryOnErrorClearsClientObserver`)
- ✅ Reconnect safety — old stream completion doesn't detach newer observer (`testOldStreamCompletionDoesNotClearNewerObserver`)

### Cross-repo Note
This change is Proxy-side only and does not affect the gRPC wire protocol or client SDKs. No coordinated change needed in `apache/rocketmq-clients`.

---

RockteMQ-AI (approved):

**PR:** Detach telemetry observer when gRPC stream closes
**Verdict:** ✅ Approved

### Summary
Ensures the telemetry `StreamObserver` reference is cleared when the gRPC stream terminates (either via `onError` or `onCompleted`). Previously, `clearClientObserver` was only called indirectly, which could leave stale observer references after stream completion.

### Analysis

**Correctness** ✅
- Changes `clearClientObserver` visibility from `protected` to `public` — necessary so `ClientActivity.telemetry()` can call it from the anonymous `ContextStreamObserver`.
- Adds `clearClientObserver` calls in both `onError` and `onCompleted` paths, covering all stream termination scenarios.
- Null guard on `grpcClientChannel` handles the case where `onError/onCompleted` fires before any `SETTINGS` command sets the channel reference.

**Performance** ✅
- `compareAndSet(future, null)` is a lightweight CAS operation, no performance concern.

**Compatibility** ⚠️
- Visibility change from `protected` to `public` on `GrpcClientChannel.clearClientObserver` is a minor API widening. Since this is an internal proxy class, the impact is minimal, but worth noting.

**Tests** ✅
- New test `telemetryStreamClearedOnComplete` verifies the observer is cleared after `onCompleted`.

### Verdict
Clean fix for a resource lifecycle issue. Prevents potential memory leaks from stale observer references in long-running proxy instances.

123123213weqw:
@lizhimins @RongtongJin gentle ping — this fixes #10609 (`Proxy keeps a stale gRPC telemetry observer after gRPC stream closure`), confirmed as a real bug. The leak is in `GrpcClientChannel` / `ClientActivity` — the observer isn't detached when the stream closes, so it keeps receiving events for a dead channel. PR body has the reproducer and the fix.

Like the sibling PR #10608, CI is in `action_required` awaiting first-time-contributor approval. Could one of you approve the workflow runs whenever convenient?

RockteMQ-AI (approved):

### Summary
Fixes stale telemetry stream observer reference in `GrpcClientChannel` after gRPC stream closure. When `onCompleted()` / `onError()` fires, the observer is now properly detached so channel liveness methods (`isActive()` / `isOpen()` / `isWritable()`) reflect the actual state immediately.

### Findings
- **[Info]** `GrpcClientChannel.java:164` — Visibility change from `protected` to `public` is necessary for cross-package access from `ClientActivity`. The underlying `compareAndSet(future, null)` still provides CAS safety, so this is acceptable.
- **[Info]** `ClientActivity.java:281` — The `grpcClientChannel` field in the anonymous `ContextStreamObserver` is not `volatile`. This is safe because gRPC serializes callbacks (`onNext` → `onCompleted`/`onError`) for the same stream, establishing a happens-before relationship. Worth a brief comment for future maintainers.
- **[Info]** `ClientActivity.java:309-315` — Null guard in the helper `clearClientObserver()` correctly handles the edge case where `onError()` fires before any `onNext()` (i.e., SETTINGS never processed).

### Suggestions
- Consider adding a brief inline comment on the `grpcClientChannel` field explaining why non-volatile is safe (gRPC stream callback serialization). This helps future contributors understand the threading model.

### Verdict
Clean, focused fix with good test coverage — including the reconnect scenario (`testOldStreamCompletionDoesNotClearNewerObserver`) which validates the CAS semantics. LGTM.

---

RockteMQ-AI (approved):
### Summary

---

Issue #10615, opened 2026-07-13
[Bug] Missing long-polling notification under CombineConsumeQueue selective double-write mode


imzs:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

.

### RocketMQ version

.

### JDK Version

.

### Describe the Bug

ReputMessageService#doReput() decides whether to notify long-polling consumers right after dispatching a message to the consume queue.

Previously this decision was driven by a boolean field (notifyMessageArriveInBatch) in DefaultMessageStore, which was flipped to true as a side effect inside the RocksDBConsumeQueueStore. The rationale is correct for a pure RocksDB CQ: the RocksDB consume queue is built asynchronously by RocksGroupCommitService, which performs the notification itself once the CQ is committed, so reput-time notification would wake consumers before the message is visible, so we must skip it.

However, CombineConsumeQueueStore also instantiates an inner RocksDBConsumeQueueStore (when combineCQLoadingCQTypes contains DEFAULT_ROCKSDB). Because the flag lives in the shared DefaultMessageStore, this constructor disables reput-time notification for the entire store — even when selective double-write (rocksdbCQSelectiveDoubleWriteEnable=true) is enabled.

Under selective double-write, RocksDB CQ is only double-written for a subset of topics. As a result: Topics that are not double-written to RocksDB never go through the RocksGroupCommitService notification path. These consumers are therefore only woken up by the long-polling timeout, adding consume latency.

### Steps to Reproduce

.

### What Did You Expect to See?

.

### What Did You See Instead?

.

RockteMQ-AI:
**Issue Evaluation**

Category: `type/bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase (develop branch).

[...]

PR #10616, opened 2026-07-13
[ISSUE #10615] Fix missing long-polling notification under CombineConsumeQueue selective double-write

imzs:
#10615 fix missing long-polling notification under CombineConsumeQueue selective double-write

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -164,8 +164,6 @@ public class DefaultMessageStore implements MessageStore {
 
     private volatile boolean shutdown = true;
 
-    private boolean notifyMessageArriveInBatch = false;
-
     protected StoreCheckpoint storeCheckpoint;
     private MessageRocksDBStorage messageRocksDBStorage;
     private TimerMessageStore timerMessageStore;
@@ -2746,7 +2744,7 @@ public class DefaultMessageStore implements MessageStore {
                                 currentReputTimestamp = dispatchRequest.getStoreTimestamp();
                                 DefaultMessageStore.this.doDispatch(dispatchRequest);
 
-                                if (!notifyMessageArriveInBatch) {
+                                if (isNotifyMessageArriveWhenReput()) {
                                     notifyMessageArriveIfNecessary(dispatchRequest);
                                 }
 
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (commented):

### Summary
This PR fixes a missing long-polling notification bug when using `CombineConsumeQueue` with selective double-write enabled. It replaces the old mutable `notifyMessageArriveInBatch` boolean flag with a polymorphic `isNotifyMessageArriveWhenReput()` method that correctly determines notification safety based on the consume queue store type.

### Findings

- **[Info]** `DefaultMessageStore.java:3219-3231` — The `isNotifyMessageArriveWhenReput()` method uses `instanceof` checks against concrete types (`ConsumeQueueStore`, `CombineConsumeQueueStore`). Since these are separate class hierarchies (confirmed: `CombineConsumeQueueStore implements ConsumeQueueStoreInterface`, not `extends ConsumeQueueStore`), the logic is correct and the second branch is reachable.

- **[Info]** `DefaultMessageStore.java:3219-3231` — Any future `ConsumeQueueStoreInterface` implementation that is neither `ConsumeQueueStore` nor `CombineConsumeQueueStore` will silently default to `false` (no notification during reput). This is the safe default — notification will be handled elsewhere (e.g., `RocksGroupCommitService`) — but consider adding a comment or log statement for unknown store types to aid debugging.

- **[Info]** `RocksDBMessageStore.java` — The override returning `false` is correct and well-documented. Since `RocksDBMessageStore` always uses RocksDB-based CQ (built asynchronously), skipping reput-time notification is the right behavior.

- **[Info]** `RocksDBConsumeQueueStore.java` — Clean removal of the old `setNotifyMessageArriveInBatch(true)` side-effect in the constructor. The new approach is much cleaner since the decision is now made at the point of use rather than through a mutable flag set during initialization.

### Suggestions

1. Consider adding a defensive `log.warn()` or `log.debug()` in the `return false` fallback path of `isNotifyMessageArriveWhenReput()` for unrecognized `ConsumeQueueStoreInterface` implementations. This would help future maintainers quickly identify if a new store type needs explicit handling.

2. The removal of `notifyMessageArriveInBatch` field and its getter/setter is clean. Just confirming that no external callers (tests, plugins) depend on `isNotifyMessageArriveInBatch()` / `setNotifyMessageArriveInBatch()` — a quick `grep` in the codebase should confirm this.

### Overall
Well-structured fix that improves code clarity by replacing a mutable flag with a polymorphic decision method. The Javadoc is thorough and the logic is sound. LGTM. 👍

---

lizhimins (approved):

RockteMQ-AI (approved):
### Summary

---

Issue #10617, opened 2026-07-14
Security: Codecov upload token committed in plaintext in .github/workflows/coverage.yml (please rotate + move to a secret)


kobihikri:
### What happened

`.github/workflows/coverage.yml` hardcodes a **Codecov upload token in plaintext** and passes it to a third-party action:

- **line 30** — `token: <redacted>` (a Codecov upload token, committed in the clear)
- line 26 — `uses: codecov/codecov-action@v3`
- line 16 — `uses: actions/checkout@master`

(I've redacted the token value here on purpose and have not tested it against any service — I'm only reporting that a credential is committed.)

### Why it matters

A credential committed to a public repository is readable by anyone and should be treated as compromised. Codecov *upload* tokens are limited in scope (they permit uploading coverage reports, not code push/read), so the blast radius is bounded — but the standard, low-risk fix is to **rotate the token and move it to a repository secret**. Every other credential in this repo's workflows already uses `${{ secrets.* }}` (e.g. `DOCKERHUB_TOKEN`, `NEXUS_PW`), so this one line is the exception.

There's also a small supply-chain angle: the token is handed to `codecov/codecov-action@v3` (a mutable major tag) in a job that also uses `actions/checkout@master` (a mutable branch), so pinning those to full commit SHAs would harden the path the credential flows through.

### Suggested fix

1. **Rotate** the current Codecov token (regenerate it in Codecov).
2. Store it as a repository secret and reference it: `token: ${{ secrets.CODECOV_TOKEN }}`.
3. (Optional) Pin `codecov/codecov-action` and `actions/checkout` to full commit SHAs.

Happy to send a PR for the workflow change once the token is rotated (I understand PRs need an ICLA on file). I used AI assistance to spot this; I verified the file and line numbers myself and did not exfiltrate or use the credential.

RockteMQ-AI:
**Issue Evaluation**

Category: `type/bug` (security) | Status: **Confirmed**

Thank you @kobihikri for the responsible disclosure. The issue has been verified against the current codebase.

**Root Cause:** The Codecov upload token is hardcoded in plaintext at line 30 of `.github/workflows/coverage.yml` instead of referencing a repository secret.

**Impact:** The token is publicly readable in the repository history. Codecov upload tokens have limited scope (coverage report uploads only), so the blast radius is bounded — but it should still be treated as compromised per standard security practice.

**Severity:** Medium

**Suggested remediation:**
[...]

PR #10618, opened 2026-07-14
Reference CODECOV_TOKEN secret instead of a hardcoded token in coverage workflow

kobihikri:
Follow-up to #10617 (as invited on that issue).

## What this changes

`.github/workflows/coverage.yml` line 30 passes the Codecov upload token as a hardcoded literal. This one-line change makes it read from a repository secret instead:

```yaml
token: ${{ secrets.CODECOV_TOKEN }}
```

Single file, single line, no other changes.

## Important — this is only half the fix

Removing the literal from the workflow does **not** invalidate the value that is already in the git history. As noted on #10617, the token should be treated as compromised and **rotated by a maintainer** (regenerate it in the Codecov dashboard, then store the new value as the `CODECOV_TOKEN` repository secret). This PR just makes sure the workflow reads from that secret once it exists. Until the secret is set, the upload step will run without a token (which is fine for public-repo Codecov uploads).

I kept this PR to the single security concern. I did not touch `actions/checkout@master` or pin the actions to SHAs in this PR — happy to open a separate small PR for that supply-chain hardening if the maintainers would like it.

---

*Disclosure: I used an AI tool to help identify and prepare this. I verified the change myself against the workflow and take responsibility for it. I understand this contribution is subject to the Apache ICLA.*

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -27,4 +27,4 @@ jobs:
         with:
           fail_ci_if_error: true
           verbose: true
-          token: cf0cba0a-22f8-4580-89ab-4f1dec3bda6f
+          token: ${{ secrets.CODECOV_TOKEN }}


RockteMQ-AI (approved):

### Summary
Replace hardcoded Codecov upload token in `.github/workflows/coverage.yml` with a repository secret reference (`${{ secrets.CODECOV_TOKEN }}`).

### Findings
- **[Info]** `.github/workflows/coverage.yml:30` — Good security practice. The hardcoded token is removed from the workflow file and replaced with a secret reference.

### Notes
- As the PR description correctly notes, the old token value remains in git history and should be rotated by a maintainer in the Codecov dashboard.
- For public repos, Codecov uploads work without a token, so this change won't break CI even before the secret is configured.
- Single-line change, no side effects. LGTM.

---

RockteMQ-AI (approved):

**PR:** Reference CODECOV_TOKEN secret instead of a hardcoded token in coverage workflow
**Verdict:** ✅ Approved

### Summary
Replaces a hardcoded Codecov token (`cf0cba0a-...`) in `.github/workflows/coverage.yml` with a reference to the GitHub Actions secret `${{ secrets.CODECOV_TOKEN }}`.

### Analysis

**Correctness** ✅
- Straightforward substitution. The secret reference syntax is correct for GitHub Actions.

**Security** ✅
- This is a security improvement — hardcoded tokens in workflow files are visible to anyone with read access to the repo. Moving to secrets is the correct practice.

**Compatibility** ✅
- No behavioral change as long as the `CODECOV_TOKEN` secret is configured in the repository settings. Maintainers should verify the secret exists before merging.

### Minor Note
- Ensure the `CODECOV_TOKEN` secret has been added to the repository settings with the correct value before merging, otherwise Codecov uploads will fail.

itxaiohanglover (commented):
Reviewed — the change is correct: reading the Codecov upload token from `${{ secrets.CODECOV_TOKEN }}` is the right fix for the hardcoded token in `coverage.yml`.

Note: my #10619 targets the same issue #10617. Since this PR is earlier and already approved, I'll defer to it and close mine to avoid duplication.

One optional hardening for reviewers to consider: guard the upload step so forks / PRs without the secret don't fail the coverage job, e.g.:
```yaml
- name: Upload to Codecov
  if: ${{ env.CODECOV_TOKEN != '' }}
```
Thanks for the responsible disclosure!

RockteMQ-AI (approved):
### Summary

---

PR #10640, opened 2026-07-21
[ISSUE #10639] Reuse a per-thread scratch buffer in CommitLog.checkMessageAndReturnSize

### LINKED ISSUE #10639: [Enhancement] Reuse a per-thread scratch buffer in CommitLog.checkMessageAndReturnSize to cut per-message allocation

### Summary

`CommitLog.checkMessageAndReturnSize(...)` allocates a fresh `byte[]` the size of the **entire message** (`new byte[totalSize]`) on every message it parses during reput/dispatch. This scratch array is used purely as a transient staging area (fields are copied in and immediately read out) and is discarded when the method returns. Replacing it with a per-thread, grow-only reusable buffer eliminates this allocation on the hot dispatch path.

### Motivation

The reput/dispatch thread calls `checkMessageAndReturnSize` once per stored message to rebuild the ConsumeQueue/index. The `byte[] bytesContent = new byte[totalSize]` inside it is one of the largest byte[] allocators on the store path: at high throughput it produces a continuous stream of short-lived, whole-message-sized arrays that immediately become garbage, driving up young-GC frequency.

Because `bytesContent` is only ever used to copy bytes out of the `ByteBuffer` and read them back (CRC check, topic/properties `new String(...)`) **within the same call**, and nothing retains a reference to it after the method returns, it does not need to be freshly allocated each time.

Borrow the scratch array from a per-thread grow-only buffer instead of allocating per call:

- A `ThreadLocal<byte[]>` holds one reusable buffer per dispatch thread (no cross-thread sharing, so no concurrency concern).
- Grow-only: reuse the cached buffer when it is large enough; only allocate (and cache) a larger one when a bigger message arrives. All reads/writes use explicit lengths, so a larger-than-needed buffer is safe.
- Safety cap: for an abnormally large message or a corrupt `totalSize` (`> maxMessageSize + 64KB`, or `< 0`), fall back to a one-shot `new byte[totalSize]` so an oversized buffer is never pinned in the ThreadLocal.

After warmup the buffer reaches the largest message size seen and subsequent messages reuse it with zero allocation on this path.

- **Sizing the array to only the fields actually read (~200B) instead of the full message**: more invasive and fragile (the offsets/lengths of body/topic/properties vary), and still allocates per message. The ThreadLocal reuse removes the allocation entirely with a smaller, localized change.
- **A pooled/off-heap buffer**: unnecessary complexity for a per-thread transient scratch area.

### Additional Context

Measured effect (allocation reduction, no throughput/latency regression):

- Local micro-measurement (`ThreadMXBean.getThreadAllocatedBytes`, JDK 21): **−2320 B/op** on this path.
- Real-server A/B on a fully isolated 4-node cluster (dedicated Producer / Broker / NameServer / Consumer machines; `-Xms4g -Xmx4g` G1; page cache dropped before each arm; 3 interleaved rounds, median): broker **young GC per million messages −10.5%** (baseline 9.18 → 8.21), **P99 unchanged (~0.85 ms)**, **TPS no regression** (median ≥ baseline).
- Correctness: existing store tests covering `checkMessageAndReturnSize` and the CRC paths pass (`AppendPropCRCTest`, `ConsumeQueueTest`).

I have a patch ready and will open a PR referencing this issue.

wang-jiahua:

Fixes #10639

### Brief Description

`CommitLog.checkMessageAndReturnSize(...)` allocated a fresh `byte[]` the size of the **entire message** (`new byte[totalSize]`) on every message parsed during reput/dispatch. This scratch array is one of the largest byte[] allocators on the store path, and it is discarded as soon as the method returns, so at high throughput it produces a continuous stream of short-lived, whole-message-sized garbage that drives up young-GC frequency.

This PR borrows the scratch array from a per-thread, grow-only buffer instead of allocating per call:

- A `ThreadLocal<byte[]>` holds one reusable buffer per dispatch thread (per-thread, so no cross-thread sharing / concurrency concern).
- Grow-only: reuse the cached buffer when it is large enough; allocate (and cache) a larger one only when a bigger message arrives. All reads/writes use explicit lengths, so a larger-than-needed buffer is safe.
- Safety cap: for an abnormally large message or a corrupt `totalSize` (`> maxMessageSize + 64KB`, or `< 0`), fall back to a one-shot `new byte[totalSize]` so an oversized buffer is never pinned in the ThreadLocal.

`bytesContent` is pure transient scratch — bytes are copied in and read back out (CRC check, topic/properties `new String(...)`) within the same call, and nothing retains a reference after the method returns — so reusing it across calls is safe. No public API, method signature, or wire format changes.

- Existing store tests covering `checkMessageAndReturnSize` and the CRC paths pass: `AppendPropCRCTest`, `ConsumeQueueTest` (19 run / 0 failures).
- Real-server A/B on a fully isolated 4-node cluster (dedicated Producer / Broker / NameServer / Consumer machines; `-Xms4g -Xmx4g` G1; OS page cache dropped before each arm; 3 interleaved rounds, median):
  - Broker **young GC per million messages −10.5%** (baseline 9.18 → 8.21).
  - **P99 send latency unchanged (~0.85 ms)**, **TPS no regression** (median ≥ baseline, 0 send failures).
- Local micro-measurement (`ThreadMXBean.getThreadAllocatedBytes`, JDK 21): **−2320 B/op** on this path.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -112,9 +112,9 @@ public class CommitLog implements Swappable {
     // Per-thread reusable scratch buffer for checkMessageAndReturnSize. That method reads
     // body/topic/properties into this buffer transiently (contents are copied out or CRC'd
     // before the method returns), so per-thread reuse is safe and avoids allocating a fresh
-    // message-sized byte[] on every dispatched message. Buffers are grow-only up to a cap
-    // derived from maxMessageSize; abnormally large sizes (e.g. corrupt length) fall back to
-    // a transient array to avoid pinning a huge buffer permanently.
+    // message-sized byte[] on every dispatched message. Buffers are grow-only up to
+    // maxCheckMessageReuseBufferSize; larger sizes (oversized messages or a corrupt length)
+    // fall back to a transient array to avoid pinning a huge buffer permanently.
     private final ThreadLocal<byte[]> checkMessageBuffer = ThreadLocal.withInitial(() -> new byte[0]);
 
     public CommitLog(final DefaultMessageStore messageStore) {
@@ -444,12 +444,13 @@ public class CommitLog implements Swappable {
      * Returns a scratch buffer of at least {@code totalSize} bytes for {@link #checkMessageAndReturnSize}.
      * The buffer is used purely transiently there (each field is read into it and copied out / CRC'd
      * before the next use), so a per-thread reusable buffer is safe and removes a per-message byte[]
-     * allocation. Sizes within maxMessageSize (plus header slack) are reused grow-only; larger sizes
+     * allocation. Sizes within {@code maxCheckMessageReuseBufferSize} are reused grow-only; larger sizes
      * return a transient array so an oversized buffer is never pinned. A negative (corrupt) totalSize
      * is rejected by the caller before reaching here.
      */
-    private byte[] borrowCheckMessageBuffer(final int totalSize) {
-        int reuseCap = this.defaultMessageStore.getMessageStoreConfig().getMaxMessageSize() + (64 * 1024);
+    // Package-private for testing (CheckMessageBufferReuseTest).
+    byte[] borrowCheckMessageBuffer(final int totalSize) {
+        int reuseCap = this.defaultMessageStore.getMessageStoreConfig().getMaxCheckMessageReuseBufferSize();
         if (totalSize > reuseCap) {
             return new byte[totalSize];
         }
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -109,6 +109,14 @@ public class CommitLog implements Swappable {
 
     private final boolean enabledAppendPropCRC;
 
+    // Per-thread reusable scratch buffer for checkMessageAndReturnSize. That method reads
+    // body/topic/properties into this buffer transiently (contents are copied out or CRC'd
+    // before the method returns), so per-thread reuse is safe and avoids allocating a fresh
+    // message-sized byte[] on every dispatched message. Buffers are grow-only up to a cap
+    // derived from maxMessageSize; abnormally large sizes (e.g. corrupt length) fall back to
+    // a transient array to avoid pinning a huge buffer permanently.
+    private final ThreadLocal<byte[]> checkMessageBuffer = ThreadLocal.withInitial(() -> new byte[0]);
+
     public CommitLog(final DefaultMessageStore messageStore) {
         String storePath = messageStore.getMessageStoreConfig().getStorePathCommitLog();
         RunningFlags runningFlags = messageStore.getMessageStoreConfig().isEnableRunningFlagsInFlush()
@@ -432,6 +440,26 @@ public class CommitLog implements Swappable {
         }
     }
 
+    /**
+     * Returns a scratch buffer of at least {@code totalSize} bytes for {@link #checkMessageAndReturnSize}.
+     * The buffer is used purely transiently there (each field is read into it and copied out / CRC'd
+     * before the next use), so a per-thread reusable buffer is safe and removes a per-message byte[]
+     * allocation. Sizes within maxMessageSize (plus header slack) are reused grow-only; larger sizes
+     * (e.g. a corrupt length) return a transient array so an oversized buffer is never pinned.
+     */
+    private byte[] borrowCheckMessageBuffer(final int totalSize) {
+        int reuseCap = this.defaultMessageStore.getMessageStoreConfig().getMaxMessageSize() + (64 * 1024);
+        if (totalSize > reuseCap || totalSize < 0) {
+            return new byte[totalSize];
+        }
+        byte[] buffer = this.checkMessageBuffer.get();
+        if (buffer.length < totalSize) {
+            buffer = new byte[totalSize];
+            this.checkMessageBuffer.set(buffer);
+        }
+        return buffer;
+    }
+
     public DispatchRequest checkMessageAndReturnSize(java.nio.ByteBuffer byteBuffer, final boolean checkCRC,
         final boolean checkDupInfo) {
         return this.checkMessageAndReturnSize(byteBuffer, checkCRC, checkDupInfo, true);
@@ -475,7 +503,7 @@ public class CommitLog implements Swappable {
 
             MessageVersion messageVersion = MessageVersion.valueOfMagicCode(magicCode);
 
-            byte[] bytesContent = new byte[totalSize];
+            byte[] bytesContent = borrowCheckMessageBuffer(totalSize);
 
             int bodyCRC = byteBuffer.getInt();
 
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -445,11 +445,12 @@ public class CommitLog implements Swappable {
      * The buffer is used purely transiently there (each field is read into it and copied out / CRC'd
      * before the next use), so a per-thread reusable buffer is safe and removes a per-message byte[]
      * allocation. Sizes within maxMessageSize (plus header slack) are reused grow-only; larger sizes
-     * (e.g. a corrupt length) return a transient array so an oversized buffer is never pinned.
+     * return a transient array so an oversized buffer is never pinned. A negative (corrupt) totalSize
+     * is rejected by the caller before reaching here.
      */
     private byte[] borrowCheckMessageBuffer(final int totalSize) {
         int reuseCap = this.defaultMessageStore.getMessageStoreConfig().getMaxMessageSize() + (64 * 1024);
-        if (totalSize > reuseCap || totalSize < 0) {
+        if (totalSize > reuseCap) {
             return new byte[totalSize];
         }
         byte[] buffer = this.checkMessageBuffer.get();
@@ -484,7 +485,7 @@ public class CommitLog implements Swappable {
             }
             // 1 TOTAL SIZE
             int totalSize = byteBuffer.getInt();
-            if (byteBuffer.remaining() < totalSize - 4) {
+            if (totalSize < 0 || byteBuffer.remaining() < totalSize - 4) {
                 return new DispatchRequest(-1, false /* fail */);
             }
 
--- a/store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java
+++ b/store/src/main/java/org/apache/rocketmq/store/config/MessageStoreConfig.java
@@ -193,6 +193,13 @@ public class MessageStoreConfig {
     // The maximum size of message body,default is 4M,4M only for body length,not include others.
     private int maxMessageSize = 1024 * 1024 * 4;
 
+    // Upper bound (in bytes) of the per-thread reusable scratch buffer that
+    // CommitLog.checkMessageAndReturnSize keeps for message verification. Messages larger than this
+    // are verified with a transient buffer instead of the reusable one, so raising maxMessageSize
+    // does not make each dispatch/recovery thread retain an oversized buffer for its whole lifetime.
+    // Default 4M + 64K, which matches the previous reuse cap for the default maxMessageSize.
+    private int maxCheckMessageReuseBufferSize = 1024 * 1024 * 4 + 64 * 1024;
+
     // The maximum size of message body can be  set in config;count with maxMsgNums * CQ_STORE_UNIT_SIZE(20 || 46)
     private int maxFilterMessageSize = 16000;
     // Whether check the CRC32 of the records consumed.
@@ -793,6 +800,14 @@ public class MessageStoreConfig {
 
     public void setMaxMessageSize(int maxMessageSize) {
         this.maxMessageSize = maxMessageSize;
+    }
+
+    public int getMaxCheckMessageReuseBufferSize() {
+        return maxCheckMessageReuseBufferSize;
+    }
+
+    public void setMaxCheckMessageReuseBufferSize(int maxCheckMessageReuseBufferSize) {
+        this.maxCheckMessageReuseBufferSize = maxCheckMessageReuseBufferSize;
     }
 
     public int getMaxFilterMessageSize() {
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (commented):

### Summary
Performance optimization that replaces per-call `new byte[totalSize]` allocation in `CommitLog.checkMessageAndReturnSize` with a per-thread, grow-only reusable buffer via `ThreadLocal<byte[]>`. The change is minimal (+29/-1), well-documented, and backed by real-server benchmarks showing −10.5% young GC and −2320 B/op.

### Findings

- **[Info]** `store/src/main/java/org/apache/rocketmq/store/CommitLog.java:443-462` — The `borrowCheckMessageBuffer` method computes `reuseCap` via a 3-method chain (`getMessageStoreConfig().getMaxMessageSize()`) on every invocation. In practice the JIT should inline these getters, but if this path is profiled as hot, consider caching `maxMessageSize` in a local `int` field to eliminate the indirection entirely.

- **[Info]** `store/src/main/java/org/apache/rocketmq/store/CommitLog.java:446` — When `totalSize < 0`, the fallback `new byte[totalSize]` will throw `NegativeArraySizeException` (same as original behavior — no regression). However, this indicates corrupt data on the CommitLog. Consider adding a `LOGGER.warn(...)` before the throw so operators can detect data corruption in production logs rather than only seeing the exception stack trace.

- **[Info]** `store/src/main/java/org/apache/rocketmq/store/CommitLog.java:116` — The `ThreadLocal<byte[]>` holds per-thread buffers that persist until the thread dies. Since CommitLog lives for the broker's lifetime this is fine, but consider adding a brief Javadoc on the field explaining the lifecycle expectation (e.g., "Thread-local buffer lives as long as the dispatch thread; no cleanup needed since CommitLog is broker-lifecycle").

- **[Info]** Testing — The PR description mentions existing tests pass and real-server A/B benchmarks. It would be valuable to add a focused unit test for `borrowCheckMessageBuffer` covering: (1) normal reuse (buffer grows then is reused), (2) `totalSize > reuseCap` falls back to one-shot allocation, (3) `totalSize < 0` behavior. This guards against future regressions in the buffer management logic.

### Suggestions

The overall approach is sound and the benchmark data is convincing. The change is a clear win for GC pressure on the store hot path. The suggestions above are all minor — no blocking concerns.

### Cross-repo Note
No cross-repo changes needed — this is a pure internal store optimization with no API or protocol impact.

---

RockteMQ-AI (approved):
LGTM. Well-designed performance optimization that eliminates a major source of short-lived garbage on the store path.

Key observations:

1. **Safety**: The buffer is used purely transiently within `checkMessageAndReturnSize` — each field is read, copied out or CRC'd before the next use, so per-thread reuse is correct.

2. **Cap logic**: The `reuseCap = maxMessageSize + 64KB` with fallback to a transient array for oversized requests is a good defensive design. Prevents a corrupt length from pinning a huge buffer.

3. **Negative size guard**: `totalSize < 0` check prevents allocating from corrupt length fields.

**Minor note**: `ThreadLocal` with potentially large buffers can retain memory if threads are short-lived (e.g., thread pools that scale down). In practice this is unlikely to be an issue since the CommitLog dispatch threads are long-lived, but worth being aware of.

---

itxaiohanglover:
Thanks for the optimization — reusing a per-thread scratch buffer here looks reasonable since the contents are consumed before the method returns.

One edge case in `borrowCheckMessageBuffer`: the `totalSize > reuseCap || totalSize < 0` branch still does `return new byte[totalSize];`, so a negative `totalSize` (e.g. from a corrupt length field) throws `NegativeArraySizeException` instead of letting the method surface a checksum/error `DispatchRequest`. It may be safer to handle `totalSize < 0` explicitly before allocation.

Also, do you have benchmark numbers for this path? Message-sized transient allocations are young-gen and usually cheap to collect, so it would help reviewers to see the measured win (allocation rate / GC pause / dispatch throughput) that justifies pinning up to `maxMessageSize + 64K` per dispatch thread.

wang-jiahua:
Thanks for the careful review — both points addressed.

**1. Negative `totalSize` (corrupt length)** — fixed in 198075430.

`checkMessageAndReturnSize` now rejects a negative `totalSize` up front and returns a failed `DispatchRequest(-1, false)` **before any allocation**, consistent with the method's other corruption paths (illegal magic code, insufficient `remaining()`):

```java
int totalSize = byteBuffer.getInt();
if (totalSize < 0 || byteBuffer.remaining() < totalSize - 4) {
    return new DispatchRequest(-1, false /* fail */);
}
```

and the now-redundant `totalSize < 0` branch was removed from `borrowCheckMessageBuffer`. (For accuracy: the original code `new byte[totalSize]` also threw `NegativeArraySizeException` on a negative length, so this is a pre-existing corner case — it's just handled gracefully now.)

**2. Benchmark numbers.**

Setup: 4 fully isolated nodes (dedicated Producer / Broker / NameServer / Consumer), broker `-Xms4g -Xmx4g` G1, Temurin 21; the OS page cache is dropped and the store is fresh before each arm; 3 interleaved rounds, median reported. Metric is normalized per message (immune to TPS jitter).

| | Baseline | This PR |
|---|---|---|
| Young GC / million msgs | 9.18 | **8.21 (−10.5%)** |
| P99 send latency | 0.86 ms | 0.85 ms (unchanged) |
| Producer TPS (median) | ~98.9k | ~100.5k (no regression) |

Local micro-measurement (`ThreadMXBean.getThreadAllocatedBytes`, JDK 21) on this path: **−2320 B/op**.

On the memory concern: the reuse cap is `maxMessageSize + 64K` (default 4 MB + 64 K ≈ 4.06 MB), and `checkMessageAndReturnSize` is driven by the single `ReputMessageService` dispatch thread (plus transient use during recovery/HA), so the pinned memory is **one buffer per such thread — bounded and small (~4 MB), independent of throughput**.

And you're right that each such allocation is a cheap young-gen object individually; the measured win is the **reduced young-GC frequency (−10.5%/msg)** from eliminating this per-message allocation on the dispatch hot path, rather than a TPS/P99 jump at this operating point.

fuyou001:
The per-thread scratch buffer approach looks reasonable, and the fallback for oversized messages avoids retaining an unexpectedly large buffer in the `ThreadLocal`.

### Suggested unit tests

Could we add focused unit tests covering:

- Buffer allocation and reuse when processing multiple messages on the same thread.
- Buffer growth when a larger message is encountered.
- The transient allocation path when the requested size exceeds the reuse cap.
- Corrupted inputs, particularly negative or otherwise invalid `totalSize` values.

### Performance results

Since this is a performance-oriented change, could you also share benchmark results comparing the implementation before and after the change? It would be useful to include:

- Throughput and allocation rate.
- GC frequency or pause time.
- Representative message sizes and thread counts.
- Benchmark duration, JVM version, heap settings, and methodology.

Ideally, please include several message-size distributions, especially a workload with occasional messages exceeding the reuse cap. This would help quantify the improvement and verify that the `ThreadLocal` retention and oversized-message fallback behave as intended.

### Reusable-buffer memory bound

One additional concern is that the reuse cap is derived directly from `maxMessageSize`. If a deployment allows messages up to 100 MiB, each participating thread may retain a buffer close to that size for its entire lifetime, causing substantial memory overhead as the thread count increases.

Could we introduce a separate, configurable upper bound for the reusable buffer? When the required size exceeds that threshold, the optimization should be bypassed and a temporary buffer should be used instead. It would also be useful to include the worst-case retained-memory estimate (`reuse threshold × number of participating threads`) in the benchmark results.

wang-jiahua:
Thanks for the thorough review, @fuyou001. I've addressed all three points.

## 1. Unit tests

Added `CheckMessageBufferReuseTest` (store module) covering the four cases you listed:

- **reuse** across multiple messages on the same thread — the same buffer instance is returned;
- **grow-only** behavior when a larger message arrives, and the grown buffer is then reused for subsequent smaller messages;
- the **transient path** when the requested size exceeds the reuse cap — and it verifies the oversized buffer is **not retained** in the `ThreadLocal` (a later small request still returns the previously cached small buffer);
- **corrupt `totalSize`** (negative, and larger-than-remaining) is rejected with `DispatchRequest(-1, false)` and no exception.

`AppendPropCRCTest` (the existing coverage of `checkMessageAndReturnSize`) still passes. Together: 6 tests, 0 failures; 0 checkstyle violations.

## 2. Configurable upper bound for the reusable buffer (agreed)

I decoupled the reuse cap from `maxMessageSize` by introducing a dedicated, configurable bound:

- New `MessageStoreConfig.maxCheckMessageReuseBufferSize`, default **4 MiB + 64 KiB** (identical to the previous cap under the default `maxMessageSize`, so default behavior is unchanged).
- `borrowCheckMessageBuffer` now bypasses reuse and returns a **transient** buffer when `totalSize > maxCheckMessageReuseBufferSize`.

So raising `maxMessageSize` (e.g. to 100 MiB) no longer inflates the retained per-thread buffer; operators who want a larger reusable buffer can raise this knob explicitly.

On the worst-case retained memory (`reuse threshold × thread count`): `checkMessageAndReturnSize` is only invoked by **store background threads** — the single `ReputMessageService` dispatch thread, single-threaded startup recovery (`recoverNormally`/`recoverAbnormally`), the HA `AutoSwitchHAService` thread, and the dledger path — **not** per connection/producer. So the participating thread count is small and bounded (O(store background threads)); with the new default, retained memory is bounded at ~`4 MiB + 64 KiB` per such thread.

## 3. Benchmarks

**Micro-benchmark** — allocation isolated to the changed buffer (`com.sun.management.ThreadMXBean#getThreadAllocatedBytes`, warmed to steady state), several size distributions, default cap `REUSE_CAP = 4 MiB + 64 KiB`:

| distribution | old B/op | new B/op | saved |
|---|---:|---:|---:|
| uniform 2 KB | 2,320 | 0 | 100% |
| mixed small (512 B / 2 KB / 8 KB) | 3,728 | 0 | 100% |
| **99% 2 KB + 1% 5 MB (> cap)** | 54,726 | 52,429 | 4.2% |
| uniform 1 MB (≤ cap) | 1,048,592 | 0 | 100% |

The oversized-message workload confirms the fallback behaves as intended: the rare 5 MB messages allocate a transient buffer (≈ the 1% × 5 MB residual, identical to baseline) and are not retained, while the ≤ cap 99% are served entirely from the reused buffer — i.e. **no regression**, and the benefit scales with the ≤ cap fraction.

**End-to-end on real hardware** — 4-machine isolated (Producer 16c / Broker 8c / Consumer 8c / NameServer), same 5.5.0-tag build, per-arm `pkill + rm store + sync + drop_caches`, 3 interleaved rounds, 120 s each, 32 producer threads, 2 KB sync send; Temurin-21, `-Xms4g -Xmx4g`, G1GC:

| arm | TPS (median) | P99 ms (median) | young GC / M msgs (median) |
|---|---:|---:|---:|
| baseline | 98,943 | 0.86 | 9.176 |
| this PR | 100,476 | 0.85 | 8.211 |

- **Allocation / GC:** young GC per million messages **−10.5%** (three rounds 8.19–8.41 vs baseline 9.07–9.27, clearly separated).
- **Throughput:** median TPS ≥ baseline (no regression; per-run ±4% is run noise).
- **Latency:** P99 flat at 0.85–0.86 ms.

Methodology note: `young GC/msg` (`grep -c 'Pause Young' gc.log` ÷ `msgPutTotalTodayNow` delta) is the primary allocation-pressure metric because it is immune to TPS jitter; exact per-op bytes come from the ThreadMXBean micro-benchmark above.

RockteMQ-AI (commented):

Two new commits since the previous review — both are solid improvements:

### Commit 2: Reject corrupt negative totalSize before allocation
The `totalSize < 0` guard added before the `remaining()` check is correct. Without it, a corrupt negative `totalSize` would pass the `remaining() < totalSize - 4` check (since `totalSize - 4` underflows to a large positive) and hit `new byte[totalSize]` → `NegativeArraySizeException`. Good defensive fix.

### Commit 3: Configurable reuse-buffer cap + unit tests
- `maxCheckMessageReuseBufferSize` (default 4 MB + 64 KB) makes the cap tunable without code changes — useful for deployments with different message size profiles.
- `CheckMessageBufferReuseTest` covers all four key behaviors: same-thread reuse, grow-only expansion, oversized transient fallback, and corrupt-size rejection. Package-private `borrowCheckMessageBuffer` is a clean testability choice.

### Minor observations
- **[Info]** `borrowCheckMessageBuffer` calls `getMessageStoreConfig()` on every invocation. The extra indirection is negligible vs. the allocation savings, but if this method ever shows up in a profiler, caching the config reference in the `CommitLog` constructor would eliminate it.
- **[Info]** The `ThreadLocal` is grow-only with no shrink path. For long-running brokers this is fine (bounded by the cap), but worth noting in the Javadoc that the buffer never shrinks — already implied by "grow-only" in the field comment, so no action needed.

No issues found in the new commits. The PR remains in good shape.

wang-jiahua:
Follow-up on the reusable-buffer memory bound — including a correction to my earlier thread-count characterization.

### Correction: which threads call `checkMessageAndReturnSize`

In my previous comment I described this method as being called only by "the single `ReputMessageService` … not per connection/producer, small and bounded". That is imprecise — the calling-thread count depends on the dispatch mode:

| mode | switch (default) | threads that call it | steady-state count |
|---|---|---|---|
| default dispatch | `enableBuildConsumeQueueConcurrently=false` (default) | single `ReputMessageService` | 1 |
| concurrent dispatch | `=true` | the `MainBatchDispatchRequestService` thread pool | = `batchDispatchRequestThreadPoolNums` (default 16, can be raised) |

(Startup recovery and the HA service also call it, but recovery threads are transient and release their `ThreadLocal` buffer on exit.)

So under concurrent dispatch the buffer is retained per pool thread, and the worst-case retained memory is:

```
min(largest message actually seen, maxCheckMessageReuseBufferSize) × number of dispatch threads
```

The buffer is grow-only, so it only grows to the largest ≤ cap message actually encountered — with typical KB-sized messages the retention is negligible regardless of thread count; it only approaches `cap × threads` when messages near the cap are common.

### Change: lowered the default cap to 1 MiB

To be safe-by-default under concurrent dispatch (and mindful of tight production memory), I lowered the default `maxCheckMessageReuseBufferSize` from `4 MiB + 64 KiB` to **1 MiB**. It still covers the vast majority of messages, but bounds worst-case per-thread retention to 1 MiB — e.g. concurrent dispatch with the default pool of 16 threads is now ≤ 16 MiB instead of ≤ ~64 MiB. Messages in the 1–4 MiB range now take the transient path (no reuse benefit, but no regression versus the current code); operators with predominantly larger messages can raise the knob.

Updated micro-benchmark under the 1 MiB default (`ThreadMXBean`, warmed to steady state):

| distribution | old B/op | new B/op | saved |
|---|---:|---:|---:|
| uniform 2 KB | 2,320 | 0 | 100% |
| mixed small (512 B / 2 KB / 8 KB) | 3,728 | 0 | 100% |
| 99% 2 KB + 1% 5 MB (> cap) | 54,726 | 52,429 | 4.2% |
| uniform 2 MB (> cap) | 2,097,168 | 2,097,168 | 0% (transient, no regression) |

Pushed as a follow-up commit.

fuyou001 (approved):
LGTM

RockteMQ-AI (approved):
### Summary

---

PR #10652, opened 2026-07-23
[ISSUE #10577] Keep reentrant POP receives broker sticky

### LINKED ISSUE #10577: [Enhancement] Make ReceiveMessageQueueSelector broker-sticky for reentrant orderly POP

### Summary

Reentrant orderly consumption (#6755) relies on a client retry (same attemptId) being routed back to **the same** broker, since OrderInfo and its attemptId match are broker-local state (QueueLevelConsumerManager.needBlock). 

However, in ```ReceiveMessageActivity.ReceiveMessageQueueSelector#select```, when the requested broker is not availabe(eg: in a rolling update), it silently falls back to round-robins to an other one.

```
protected static class ReceiveMessageQueueSelector implements QueueSelector {

    private final String brokerName;

    public ReceiveMessageQueueSelector(String brokerName) {
        this.brokerName = brokerName;
    }

    @Override
    public AddressableMessageQueue select(ProxyContext ctx, MessageQueueView messageQueueView) {
        try {
            AddressableMessageQueue addressableMessageQueue = null;
            MessageQueueSelector messageQueueSelector = messageQueueView.getReadSelector();

            if (StringUtils.isNotBlank(brokerName)) {
                addressableMessageQueue = messageQueueSelector.getQueueByBrokerName(brokerName);
            }

            if (addressableMessageQueue == null) {
                addressableMessageQueue = messageQueueSelector.selectOne(true);
            }
            return addressableMessageQueue;
        } catch (Throwable t) {
            return null;
        }
    }
}
```

### Motivation

Avoid losing reentrancy and thus leaving the FIFO queue blocked until invisibleTime expires.

Make the selector broker-sticky, especially on a FIFO request carrying an attemptId.

-

### Additional Context

_No response_

Aias00:
## Summary

Fixes #10577.

This change keeps reentrant orderly POP receive requests broker-sticky when the request carries an `attemptId`.

## Motivation

Reentrant orderly POP depends on broker-local order state. If a retry with the same `attemptId` asks for a specific broker but the selector silently falls back to another broker, the retry can lose access to the original broker-local `OrderInfo` and leave the FIFO queue blocked until invisible time expires.

## Changes

- Pass a broker-sticky flag into `ReceiveMessageQueueSelector` only for FIFO receive requests carrying `attemptId`.
- When sticky mode is enabled and the requested broker is unavailable, return `null` instead of falling back to another broker.
- Preserve the existing fallback behavior for non-sticky receive requests.
- Extend `ReceiveMessageActivityTest` to cover both fallback and sticky-missing-broker behavior.

## Verification

```text
mvn -pl proxy -Dtest=ReceiveMessageActivityTest test \
  -DfailIfNoTests=false -Djacoco.skip=true \
  -Dspotbugs.skip=true -Dcheckstyle.skip=true -Drat.skip=true

BUILD SUCCESS
```

```text
mvn -pl proxy -DskipTests -Dspotbugs.skip=true -Drat.skip=true compile

Checkstyle violations: 0
BUILD SUCCESS
```

`git diff --check` is clean.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivity.java
@@ -142,7 +142,9 @@ public class ReceiveMessageActivity extends AbstractMessagingActivity {
             CompletableFuture<PopResult> popFuture = this.messagingProcessor.popMessage(
                 ctx,
                 new ReceiveMessageQueueSelector(
-                    request.getMessageQueue().getBroker().getName()
+                    request.getMessageQueue().getBroker().getName(),
+                    // Reentrant FIFO POP must stay on the original broker instead of falling back.
+                    fifo && request.hasAttemptId()
                 ),
                 group,
                 topic,
@@ -216,9 +218,19 @@ public class ReceiveMessageActivity extends AbstractMessagingActivity {
     protected static class ReceiveMessageQueueSelector implements QueueSelector {
 
         private final String brokerName;
+        private final boolean brokerSticky;
 
         public ReceiveMessageQueueSelector(String brokerName) {
+            this(brokerName, false);
+        }
+
+        /**
+         * @param brokerSticky if true, keep the target broker sticky and return null when the
+         * target broker is unavailable instead of falling back to another broker.
+         */
+        public ReceiveMessageQueueSelector(String brokerName, boolean brokerSticky) {
             this.brokerName = brokerName;
+            this.brokerSticky = brokerSticky;
         }
 
         @Override
@@ -229,6 +241,9 @@ public class ReceiveMessageActivity extends AbstractMessagingActivity {
 
                 if (StringUtils.isNotBlank(brokerName)) {
                     addressableMessageQueue = messageQueueSelector.getQueueByBrokerName(brokerName);
+                    if (addressableMessageQueue != null || brokerSticky) {
+                        return addressableMessageQueue;
+                    }
                 }
 
                 if (addressableMessageQueue == null) {
--- a/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivityTest.java
+++ b/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivityTest.java
@@ -18,6 +18,7 @@
 package org.apache.rocketmq.proxy.grpc.v2.consumer;
 
 import apache.rocketmq.v2.Code;
+import apache.rocketmq.v2.Broker;
 import apache.rocketmq.v2.ClientType;
 import apache.rocketmq.v2.FilterExpression;
 import apache.rocketmq.v2.FilterType;
@@ -26,6 +27,7 @@ import apache.rocketmq.v2.ReceiveMessageRequest;
 import apache.rocketmq.v2.ReceiveMessageResponse;
 import apache.rocketmq.v2.Resource;
 import apache.rocketmq.v2.Settings;
+import apache.rocketmq.v2.Subscription;
 import com.google.protobuf.Duration;
 import com.google.protobuf.util.Durations;
 import io.grpc.stub.ServerCallStreamObserver;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Fixes broker selection for reentrant orderly POP receives by adding a `brokerSticky` mode to `ReceiveMessageQueueSelector`. When a FIFO retry carries an `attemptId`, the selector now stays pinned to the target broker instead of silently falling back to another broker — which would lose access to broker-local `OrderInfo` and block the FIFO queue.

### Findings
- **[Info]** `ReceiveMessageActivity.java:236-238` — When `brokerSticky=true` and the broker is not found, `select()` returns `null`. Callers must handle this gracefully (e.g., return an error to the client rather than NPE). The PR description implies this is the intended behavior, so it looks correct.
- **[Info]** The condition `fifo && request.hasAttemptId()` is well-scoped — it only activates sticky mode for reentrant FIFO requests, preserving the existing fallback path for all other cases.

### Suggestions
- Minor: consider adding a brief Javadoc on the `brokerSticky` parameter in the two-arg constructor to clarify the contract (returns `null` when the target broker is unavailable, rather than falling back).

Overall this is a clean, focused fix with good test coverage. LGTM.

---

RockteMQ-AI (commented):

### Summary
Adds broker-sticky behavior to `ReceiveMessageQueueSelector` for reentrant POP receives: when a message is FIFO and has an `attemptId`, the selector will stick to the same broker even if the preferred broker's queue is not available (returns null instead of falling through to random selection).

### Findings
- **[Warning]** `ReceiveMessageActivity.java:239-242` — The new logic: if `brokerSticky` is true and the broker-name lookup returns null, it still returns null instead of falling through to `getRandomOne()`. This means for FIFO messages with `attemptId`, if the preferred broker is unavailable, the receive will get a null queue. Verify that the caller handles this null gracefully (e.g., retries or returns an empty response rather than NPE).
- **[Info]** `ReceiveMessageActivity.java:145` — The `fifo && request.hasAttemptId()` condition is a reasonable trigger for broker stickiness. FIFO ordering requires the same broker to maintain message ordering guarantees.
- **[Info]** Backward-compatible constructor `ReceiveMessageQueueSelector(String brokerName)` delegates to the new constructor with `brokerSticky=false`, preserving existing behavior for non-FIFO cases.
- **[Info]** Tests cover both the sticky and non-sticky paths.

### Suggestions
Consider adding a null check or comment at the call site where `select()` returns null for the sticky case, to make it explicit that this is intentional (not a bug).

### Verdict
Reasonable change for FIFO ordering guarantees. The main concern is ensuring callers handle the null queue case properly.

---

Aias00:
Addressed the review feedback in `a998906f3`:

- Documented the broker-sticky null-return contract in `ReceiveMessageQueueSelector` Javadoc.
- Added a call-site comment clarifying that reentrant FIFO POP must stay on the original broker instead of falling back.
- Added coverage for FIFO receive with `attemptId` to verify the sticky selector is passed into `popMessage` and returns no queue when the original broker is unavailable.

Local verification:
`mvn -pl proxy -Dtest=ReceiveMessageActivityTest test` passed with 10 tests; `git diff --check` passed.

RockteMQ-AI (approved):

### Summary
Re-reviewed after rebase. The broker-sticky logic for reentrant FIFO POP receives remains unchanged and well-implemented.

### Findings
- **[Info]** `ReceiveMessageActivity.java:145` — The `fifo && request.hasAttemptId()` condition correctly triggers broker stickiness only for reentrant FIFO messages.
- **[Info]** `ReceiveMessageActivity.java:241-243` — The sticky logic: when `brokerSticky` is true, returns the broker queue if found, or returns null (instead of falling through to random selection). This is intentional and correct for FIFO ordering guarantees.
- **[Info]** Backward-compatible: the single-arg constructor delegates to the new constructor with `brokerSticky=false`.
- **[Info]** Tests cover both sticky and non-sticky paths, including explicit null verification for the missing-broker sticky case.

### Verdict
Clean rebase. The implementation correctly preserves FIFO ordering by keeping the broker sticky for reentrant POP receives. LGTM.

---

fuyou001 (approved):
LGTM

RockteMQ-AI (approved):
### Summary

---

Aias00:
Rebased onto current `develop` and revalidated `ReceiveMessageActivityTest`. The previous macOS CI failure will now rerun on the refreshed, already-approved branch.

PR #10657, opened 2026-07-23
[ISSUE #10656] Fix incorrect end timestamp recovery in tiered index files

### LINKED ISSUE #10656: [Bug] Incorrect end timestamp recovery in tiered index files

All

### RocketMQ version

develop

### JDK Version

All

### Describe the Bug

`IndexStoreFile` reads `endTimestamp` from `INDEX_BEGIN_TIME_STAMP` instead of `INDEX_END_TIME_STAMP` during recovery.

### Steps to Reproduce

Write an index whose timestamp is later than the file’s begin timestamp, then close and reopen the file.

The reopened file should retain the persisted end timestamp.

### What Did You See Instead?

The end timestamp becomes the begin timestamp, which can cause valid query results to be skipped.

### Additional Context

_No response_

majialoong:

- Fixes #10656 

### Brief Description

Fix `IndexStoreFile` to restore `endTimestamp` from `INDEX_END_TIME_STAMP` and add a regression test.

Added a test that writes an index, reopens the file, and verifies that its end timestamp is recovered correctly with and without mmap writes.

### DIFF
--- a/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/index/IndexStoreFile.java
+++ b/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/index/IndexStoreFile.java
@@ -114,7 +114,7 @@ public class IndexStoreFile implements IndexFile {
         this.fileChannel = this.mappedFile.getFileChannel();
 
         this.beginTimestamp.set(timestamp);
-        this.endTimestamp.set(byteBuffer.getLong(INDEX_BEGIN_TIME_STAMP));
+        this.endTimestamp.set(byteBuffer.getLong(INDEX_END_TIME_STAMP));
         this.hashSlotCount.set(byteBuffer.getInt(INDEX_SLOT_COUNT));
         this.indexItemCount.set(byteBuffer.getInt(INDEX_ITEM_INDEX));
         this.flushNewMetadata(byteBuffer, indexItemMaxCount == this.indexItemCount.get() + 1);
--- a/tieredstore/src/test/java/org/apache/rocketmq/tieredstore/index/IndexStoreFileTest.java
+++ b/tieredstore/src/test/java/org/apache/rocketmq/tieredstore/index/IndexStoreFileTest.java
@@ -230,6 +230,21 @@ public class IndexStoreFileTest {
     }
 
     @Test
+    public void recoverEndTimestampTest() throws IOException {
+        long beginTimestamp = indexStoreFile.getTimestamp();
+        long endTimestamp = beginTimestamp + 10000L;
+        Assert.assertEquals(AppendResult.SUCCESS, indexStoreFile.putKey(
+            TOPIC_NAME, TOPIC_ID, QUEUE_ID, KEY_SET, MESSAGE_OFFSET, MESSAGE_SIZE, endTimestamp));
+        Assert.assertEquals(endTimestamp, indexStoreFile.getEndTimestamp());
+
+        indexStoreFile.shutdown();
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (commented):

### Summary
Fixes a copy-paste bug in `IndexStoreFile` constructor where `endTimestamp` was incorrectly recovered from `INDEX_BEGIN_TIME_STAMP` (offset 4) instead of `INDEX_END_TIME_STAMP` (offset 12).

### Findings

- **[Info]** `IndexStoreFile.java:117` — The fix is correct. The write path (lines 192-193) stores `beginTimestamp` at `INDEX_BEGIN_TIME_STAMP` and `endTimestamp` at `INDEX_END_TIME_STAMP`. The old read was using the wrong offset, meaning after recovery, `endTimestamp` would hold the same value as `beginTimestamp`, breaking time-range queries on recovered index files.

- **[Info]** `IndexStoreFileTest.java` — Good regression test. It writes an index entry with a distinct `endTimestamp`, reopens the file, and verifies recovery. Consider also testing the edge case where `endTimestamp == beginTimestamp` (single-entry file) to ensure no off-by-one.

- **[Info]** The fix is minimal (1-line source change + test) and directly addresses the root cause. No compatibility concerns since this is a bug fix in the recovery path — the on-disk format is unchanged.

### Verdict
Clean fix with proper test coverage. LGTM.

---

RockteMQ-AI (commented):

### Summary
Fixes incorrect `endTimestamp` recovery in tiered store index files — the constructor was reading from `INDEX_BEGIN_TIME_STAMP` offset instead of `INDEX_END_TIME_STAMP` when restoring the end timestamp from the mapped file buffer.

### Findings
- **[Critical]** `IndexStoreFile.java:117` — The original code reads `endTimestamp` from the wrong buffer offset (`INDEX_BEGIN_TIME_STAMP`). This means after recovery, the index file's `endTimestamp` would be set to the begin timestamp, breaking any logic that depends on accurate end timestamps (e.g., compaction decisions, time-range queries).
- **[Info]** `IndexStoreFileTest.java` — New `recoverEndTimestampTest` correctly validates the fix by writing an entry with a known end timestamp, shutting down, recovering, and asserting the end timestamp is preserved.

### Verdict
Clean one-line bug fix with appropriate test coverage. The fix is clearly correct — the variable name `endTimestamp` should be initialized from `INDEX_END_TIME_STAMP`, not `INDEX_BEGIN_TIME_STAMP`.

---

fuyou001:
**[P1] Handle end timestamps already overwritten by the legacy recovery path**

The offset change is correct for a file that has not previously been reopened by the affected implementation. However, the old constructor loads `INDEX_BEGIN_TIME_STAMP` into `endTimestamp` and then immediately calls `flushNewMetadata()`, which writes that incorrect value back to `INDEX_END_TIME_STAMP`. Therefore, an existing file recovered by the old code can already have `endTimestamp == beginTimestamp` persisted, and upgrading to this patch will still read the corrupted value.

This is observable in `IndexStoreService.queryAsync()`: when a query begins after the corrupted value but before the file’s actual end time, `getEndTimestamp() < beginTime` causes the scan to `break`, skipping valid index entries. The added test covers only the first clean reopen, so it does not exercise this upgrade path.

Could recovery validate or rebuild a suspicious end timestamp from the stored index items, or conservatively avoid using it for pruning until it is rebuilt? Please also add a regression test that simulates the legacy recovery/write-back before reopening the file with the fixed code.

RockteMQ-AI (approved):
### Summary

---

PR #10659, opened 2026-07-24
[ISSUE #10658] Fix PopConsumerService.revive aborting the whole batch on a single failed record

### LINKED ISSUE #10658: PopConsumerService.revive aborts the whole batch when a single record fails (head-of-line blocking in the popkv pop path)
### Describe the bug

In the KV-based ("popkv") pop consumer `org.apache.rocketmq.broker.pop.PopConsumerService` (instantiated when `popConsumerKVServiceInit` is enabled, used when `popConsumerKVServiceEnable` is enabled), the batch revive method `revive(AtomicLong currentTime, int maxCount)` awaits all per-record futures and then runs the batch bookkeeping:

```java
CompletableFuture.allOf(futureList.toArray(new CompletableFuture[0])).join();
this.popConsumerStore.writeRecords(new ArrayList<>(failureList));
this.popConsumerStore.deleteRecords(consumerRecords);
currentTime.set(...);
```

There is no per-record error isolation, so a single failing record aborts the whole batch:

- `revive(PopConsumerRecord)` chains `getMessageAsync(record).thenCompose(...)` with no `.exceptionally(...)` handler, so an exceptional completion (e.g. `reviveRetry` throwing inside `thenCompose`, or a decode failure in the escape bridge) makes the returned future complete exceptionally.
- `revive(record)` can also throw **synchronously** before it returns a future (e.g. `DefaultMessageStore.getMessageAsync` is `completedFuture(getMessage(...))` and `getMessage` throws), which the batch loop currently rethrows as `RuntimeException`.

Either way, `writeRecords(failureList)`, `deleteRecords(consumerRecords)` and the `currentTime` advance are all skipped and `revive()` throws out to the run loop. If one record keeps failing (e.g. a persistently unreachable remote), revive is stuck reprocessing the same batch forever, and the retries for the other, healthy records in that batch are never persisted — head-of-line blocking.

### Scope

This is in the newer "popkv" pop path (`PopConsumerService`, gated by `popConsumerKVServiceEnable`), which is disabled by default; it does not affect the legacy `PopBufferMergeService`. It is a correctness fix that matters as the popkv path is rolled out.

### Steps to reproduce

In `PopConsumerServiceTest`: write two expired records into the store, make one record's read fail (either an async exceptional completion, or a synchronous throw), and run `revive(...)`. Before the fix the whole batch aborts (the healthy record is not consumed and progress does not advance); after the fix the failing record is isolated (scheduled for retry) and the batch still advances.

### Version

develop

wang-jiahua:

Fixes #10658

### Brief Description

In the popkv pop consumer (`PopConsumerService`), the batch `revive(AtomicLong, int)` awaits all per-record futures with `CompletableFuture.allOf(...).join()` and has no per-record error isolation, so a single failing record aborts the whole batch (`writeRecords(failureList)`, `deleteRecords(consumerRecords)` and the `currentTime` advance are skipped, and `revive()` throws). A persistently failing record therefore blocks the whole batch forever, and the healthy records' retries in that batch are never persisted (head-of-line blocking).

A record can fail two ways, both handled here:

1. **Async** — `revive(PopConsumerRecord)` gets an `.exceptionally(...)` handler that logs and downgrades the failure to a "failed revive" (`false`), so it goes through the existing `failureList` backoff-retry path instead of propagating through `allOf(...).join()`.
2. **Sync** — in the batch loop, a synchronous throw from `revive(record)` (e.g. `getMessageAsync` throwing before it returns a future) is caught and turned into `completedFuture(false)` instead of being rethrown and aborting the batch. The semaphore permit is released by the `whenComplete` stage; `acquire()`'s `InterruptedException` is handled separately.

Either way a single record's failure only affects that record, and `writeRecords` / `deleteRecords` / `currentTime` still run.

Note: this is the newer popkv pop path (`PopConsumerService`, gated by `popConsumerKVServiceEnable`), which is disabled by default; it does not touch the legacy `PopBufferMergeService`.

Added two tests in `PopConsumerServiceTest`, both failing before the fix and passing after:

- `reviveShouldNotAbortBatchWhenGetMessageFailsExceptionally` — a single record whose read completes exceptionally; asserts `revive` does not throw and the record is consumed (batch bookkeeping still runs).
- `reviveShouldIsolateSynchronousReadFailureAndNotBlockHealthyRecords` — two records, one whose read throws synchronously and one healthy; asserts `revive` returns 2 and both original records are consumed (`scanExpiredRecords` returns 0), i.e. the failed record is isolated and the healthy record is not head-of-line blocked.

Full `PopConsumerServiceTest` passes (19/19), and `mvn -pl broker validate` reports 0 checkstyle violations.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -613,11 +613,23 @@ public class PopConsumerService extends ServiceThread {
 
         // could merge read operation here
         for (PopConsumerRecord record : consumerRecords) {
-            try {
-                semaphore.acquire();
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                throw new RuntimeException(e);
+            while (true) {
+                try {
+                    semaphore.acquire();
+                    break;
+                } catch (InterruptedException e) {
+                    // Shutdown stops this service with the stopped flag and wakeup() rather than
+                    // thread interruption, and ServiceThread.waitForRunning preserves the interrupt
+                    // status. Restoring the flag here would make every later acquire() throw again
+                    // and leave the revive service in a permanent busy loop, so consume incidental
+                    // interrupts and retry. If the service is being stopped (e.g. shutdown(true)),
+                    // abort the batch cleanly: the records are not yet deleted from the store and
+                    // will be reprocessed after the next start.
+                    if (this.isStopped()) {
+                        throw new RuntimeException("PopConsumerService stopped while acquiring the revive semaphore", e);
+                    }
+                    log.warn("PopConsumerService interrupted while acquiring the revive semaphore, retry");
+                }
             }
             CompletableFuture<Boolean> future;
             try {
@@ -644,8 +656,8 @@ public class PopConsumerService extends ServiceThread {
                             Math.min(REWRITE_INTERVALS_IN_SECONDS.length - 1, record.getAttemptTimes())];
                         long nextInvisibleTime = record.getInvisibleTime() + backoffInterval;
                         PopConsumerRecord retryRecord = new PopConsumerRecord(System.currentTimeMillis(),
-                            record.getGroupId(), record.getTopicId(), record.getQueueId(),
-                            record.getRetryFlag(), nextInvisibleTime, record.getOffset(), record.getAttemptId());
+                            record.getGroupId(), record.getTopicId(), record.getQueueId(), record.getRetryFlag(),
+                            nextInvisibleTime, record.getOffset(), record.getAttemptId(), record.isSuspend());
                         retryRecord.setAttemptTimes(record.getAttemptTimes() + 1);
                         failureList.add(retryRecord);
                         log.warn("PopConsumerService revive backoff retry, record={}", retryRecord);
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -587,13 +587,6 @@ public class PopConsumerService extends ServiceThread {
                     return CompletableFuture.completedFuture(!result.getRight());
                 }
                 return CompletableFuture.completedFuture(this.reviveRetry(record, result.getLeft()));
-            })
-            .exceptionally(throwable -> {
-                // Do not let a single failed async read (e.g. a remote read via the escape bridge)
-                // abort the whole revive batch in revive(AtomicLong, int). Treat it as a failed
-                // revive so the record is scheduled for retry instead of throwing out of allOf().join().
-                log.error("PopConsumerService revive failed, will retry, record={}", record, throwable);
-                return false;
             });
     }
 
@@ -628,7 +621,14 @@ public class PopConsumerService extends ServiceThread {
             }
             CompletableFuture<Boolean> future;
             try {
-                future = this.revive(record);
+                // Attach the exception-to-false conversion at this call site only: here a false
+                // result is consumed below and turned into a failureList backoff retry. Other
+                // callers of revive(record) (e.g. the PopConsumerCache callback) do not consume
+                // the result and must keep the original exception semantics.
+                future = this.revive(record).exceptionally(throwable -> {
+                    log.error("PopConsumerService revive failed, will retry, record={}", record, throwable);
+                    return false;
+                });
             } catch (Exception e) {
                 // A synchronous failure from revive(record) (e.g. getMessageAsync throwing before it
                 // returns a future) must not abort the whole batch; treat it as a failed revive so the
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -587,6 +587,13 @@ public class PopConsumerService extends ServiceThread {
                     return CompletableFuture.completedFuture(!result.getRight());
                 }
                 return CompletableFuture.completedFuture(this.reviveRetry(record, result.getLeft()));
+            })
+            .exceptionally(throwable -> {
+                // Do not let a single failed async read (e.g. a remote read via the escape bridge)
+                // abort the whole revive batch in revive(AtomicLong, int). Treat it as a failed
+                // revive so the record is scheduled for retry instead of throwing out of allOf().join().
+                log.error("PopConsumerService revive failed, will retry, record={}", record, throwable);
+                return false;
             });
     }
 
@@ -613,13 +620,22 @@ public class PopConsumerService extends ServiceThread {
 
         // could merge read operation here
         for (PopConsumerRecord record : consumerRecords) {
-            CompletableFuture<Boolean> future;
             try {
                 semaphore.acquire();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            }
+            CompletableFuture<Boolean> future;
+            try {
                 future = this.revive(record);
             } catch (Exception e) {
-                semaphore.release();
-                throw new RuntimeException(e);
+                // A synchronous failure from revive(record) (e.g. getMessageAsync throwing before it
+                // returns a future) must not abort the whole batch; treat it as a failed revive so the
+                // record goes through the failureList backoff-retry path below. The semaphore permit is
+                // released by the whenComplete stage attached to this future.
+                log.error("PopConsumerService revive threw synchronously, will retry, record={}", record, e);
+                future = CompletableFuture.completedFuture(false);
             }
             futureList.add(future.thenAccept(result -> {
                 if (!result) {
--- a/broker/src/test/java/org/apache/rocketmq/broker/pop/PopConsumerServiceTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/pop/PopConsumerServiceTest.java
@@ -743,4 +743,101 @@ public class PopConsumerServiceTest {
             messageExt.setReconsumeTimes(capturedMessage.getReconsumeTimes());
         }
     }
+
+    @Test
+    public void reviveShouldNotAbortBatchWhenGetMessageFailsExceptionally() {
+        // A single getMessageAsync that completes exceptionally (e.g. a transient remote read
+        // failure via the escape bridge) must not abort the whole revive batch. Before the fix,
+        // revive(record) had no exceptionally handler, so the exception propagated through
+        // allOf(...).join(), skipping writeRecords/deleteRecords and throwing out of revive().
+        Mockito.when(brokerController.getEscapeBridge()).thenReturn(Mockito.mock(EscapeBridge.class));
+        Mockito.when(brokerController.getSubscriptionGroupManager()
+            .containsSubscriptionGroup(anyString())).thenReturn(true);
+        PopConsumerService consumerServiceSpy = Mockito.spy(consumerService);
+
+        consumerService.getPopConsumerStore().start();
+
+        long popTime = 1000000000L;
+        long invisibleTime = 60 * 1000L;
+        PopConsumerRecord record = new PopConsumerRecord();
+        record.setPopTime(popTime);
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Fixes head-of-line blocking in `PopConsumerService.revive(AtomicLong, int)` where a single failing record would abort the entire batch via `CompletableFuture.allOf(...).join()`, preventing healthy records from being persisted and retried.

### Findings

- **[Info]** `PopConsumerService.java:564-569` — The `.exceptionally()` handler correctly downgrades async read failures to `false`, routing the record through the existing `failureList` backoff-retry path. Clean approach that reuses existing retry infrastructure.

- **[Info]** `PopConsumerService.java:597-614` — Good separation of `semaphore.acquire()` `InterruptedException` handling from the `revive(record)` sync-failure catch. Restoring the interrupt flag with `Thread.currentThread().interrupt()` before wrapping in `RuntimeException` is the correct pattern.

- **[Info]** `PopConsumerService.java:610-613` — The sync-failure catch creates `CompletableFuture.completedFuture(false)` so the record enters the `failureList` path. The semaphore permit is correctly released by the downstream `whenComplete` stage since `completedFuture` triggers it immediately.

- **[Info]** `PopConsumerServiceTest.java` — Two well-targeted tests covering both failure modes (async exceptionally and sync throw). Both verify that batch bookkeeping (`writeRecords`/`deleteRecords`/`scanExpiredRecords`) still runs after a failure, which is the key invariant this fix protects.

### Assessment
This is a solid, minimal fix for a classic batch-processing head-of-line blocking problem. The error isolation is correct for both async and sync failure paths, semaphore management is sound, and the tests directly validate the fix. Scope is appropriately limited to the newer popkv path (`popConsumerKVServiceEnable`), not touching the legacy `PopBufferMergeService`.

---

RockteMQ-AI (commented):

### Summary
Fixes `PopConsumerService.revive` so that a single failed record (async or sync) does not abort the entire revive batch. Adds an `.exceptionally()` handler to the per-record future chain and catches synchronous failures from `revive(record)` in the batch loop.

### Findings
- **[Critical]** `PopConsumerService.java:564-568` — The new `.exceptionally()` handler correctly catches async failures (e.g., remote read via escape bridge failing) and returns `false` instead of propagating the exception. This prevents `allOf().join()` from throwing and skipping the `writeRecords`/`deleteRecords` cleanup.
- **[Warning]** `PopConsumerService.java:597-608` — The synchronous failure handling creates a `CompletableFuture.completedFuture(false)` as a fallback. This is correct, but note that the `semaphore.acquire()` is now in a separate try-catch before the future creation. If `revive(record)` throws synchronously, the semaphore permit was already acquired but the `.whenComplete` stage on the fallback future will release it. Verify that the `thenAccept` callback attached later (at `futureList.add(future.thenAccept(...))`) also properly releases the semaphore in all paths.
- **[Info]** `PopConsumerService.java:594-596` — Good fix for the `InterruptedException` handling: properly restores the interrupt flag with `Thread.currentThread().interrupt()` before throwing.
- **[Info]** Tests are comprehensive: `reviveShouldNotAbortBatchWhenGetMessageFailsExceptionally` and `reviveShouldIsolateSynchronousReadFailureAndNotBlockHealthyRecords` directly validate both failure modes.

### Suggestions
Consider adding a brief comment on the fallback future explaining why it's needed (the synchronous throw path that bypasses the `.exceptionally()` chain).

### Verdict
Solid fix for a critical reliability issue. The batch processing pattern is now resilient to both async and sync failures.

---

fuyou001:
**[P1] Preserve failed records in the `PopConsumerCache` call path**

`revive(PopConsumerRecord)` is not only called by `revive(AtomicLong, int)`. When `enablePopBufferMerge` is enabled, `PopConsumerService` passes `this::revive` to `PopConsumerCache` as a `Consumer<PopConsumerRecord>`. In `cleanupRecords()`, `consumer.accept(record)` discards the returned `CompletableFuture<Boolean>`, and `clearStagedRecords()` then unconditionally removes the staged checkpoints.

The new exception-to-`false` conversion works in the RocksDB batch path because `false` is consumed and a backoff record is added to `failureList`. In the cache path, however, that `false` result is never observed. An asynchronously failed revive can therefore be cleared from the cache without being persisted for retry, preventing the message from being revived. The outer `catch` in `PopConsumerCache.run()` only sees synchronous exceptions thrown before a future is returned; it cannot observe an exceptional future completion.

Please either attach the exception-to-`false` handling only at the batch call site if `PopConsumerCache` is intentionally out of scope, or change the cache callback to return and handle the future, clearing a staged record only after success and retaining or persisting it with backoff on `false` or exception. Please also add a regression test with `enablePopBufferMerge=true` and an asynchronously failed revive, verifying that the checkpoint remains retryable.

wang-jiahua:
> **[P1] Preserve failed records in the `PopConsumerCache` call path**

@fuyou001 Thanks for the careful review — you are right. Addressed by taking the first option in commit b84fcc0a3: the exception-to-`false` conversion is now attached only at the batch call site.

- `revive(PopConsumerRecord)` no longer carries an `.exceptionally` handler, so its exception semantics are exactly what they were before this PR; the `PopConsumerCache` callback path is untouched.
- In `revive(AtomicLong, int)` the handler sits right where the `false` result is consumed into the failureList backoff retry, with a comment noting that other callers do not consume the result.
- This also removes a side problem: the "will retry" log would have been misleading on the cache path, where a cleared record is never retried.

On the pre-existing cache-path behavior, I verified: an asynchronously failed revive was already lost before this PR — the exceptionally-completed future is unobserved, the revive-attempted records are not in `writeConsumerRecords`, and `clearStagedRecords()` clears the staging map unconditionally; the `run()` catch only sees synchronous throws, which abort `cleanupRecords` before the clear and get retried on the next sweep. Since the cache path is intentionally out of scope here, I've opened #10667 to track clearing a staged record only after a successful revive, including the `enablePopBufferMerge=true` regression test you suggested.

Both regression tests still pass (`PopConsumerServiceTest` 19/19).

RockteMQ-AI (approved):
## Review by github-manager-bot (re-review after new commit)

### Summary
Re-reviewed after new commit (Jul 28). The previous suggestion to add explanatory comments on the fallback future path has been addressed — both the `.exceptionally()` call site and the synchronous catch block now have clear comments explaining why the conversion happens here and not inside `revive(record)` itself.

### Changes Since Last Review
- Added inline comments explaining the `.exceptionally()` placement rationale (call-site-specific semantics vs. `revive(record)` internal behavior)
- Added inline comment on the synchronous fallback future explaining the semaphore release path
- `InterruptedException` handling correctly restores the interrupt flag

### Verdict
Previous findings remain addressed. The batch processing error isolation is solid. LGTM.

---

fuyou001:
**[P1] Avoid leaving the revive service permanently interrupted**

The current `InterruptedException` handling restores the interrupt flag and throws. `PopConsumerService.run()` catches the exception and calls `waitForRunning(500)`, but `ServiceThread.waitForRunning()` preserves the interrupt status. As a result, every subsequent `Semaphore.acquire()` immediately throws again, so a single unexpected interrupt can stop all future revive progress and cause a tight error/log loop. The current batch is also abandoned before `writeRecords` and `deleteRecords` run.

The normal broker shutdown path calls `PopConsumerService.shutdown()` without interrupting the worker; it stops the service through the `stopped` flag and `wakeup()`. Therefore, an incidental interrupt while waiting for the per-batch semaphore can be consumed and retried:

```java
while (true) {
    try {
        semaphore.acquire();
        break;
        // PopConsumerService shutdown uses the stopped flag and wakeup(), rather
        // than thread interruption. Do not restore the interrupt flag here;
        // otherwise subsequent acquire() and waitForRunning() calls may return
        // immediately and leave the revive service in a permanent busy loop.
        log.warn("Interrupted while acquiring semaphore, retry");
    }
}
```

Please also add a regression test that blocks on the semaphore, interrupts the revive thread once, releases the outstanding operation, and verifies that the remaining records are processed without repeated exceptions or CPU spinning.

If the inherited `shutdown(true)` path is expected to be supported in the future, it should be handled separately by checking the stopped state and terminating cleanly instead of swallowing that shutdown interrupt.

fuyou001:
**[P1] Preserve the suspend flag in the backoff record**

When a synchronous or asynchronous revive failure is converted to `false`, the backoff record is created with the eight-argument `PopConsumerRecord` constructor, which defaults `suspend` to `false`.

This changes the semantics of records created by `changeInvisibilityDuration(..., suspend=true)`. After the original record is replaced by this backoff record, a later successful `reviveRetry` sees `suspend=false` and increments `reconsumeTimes` unexpectedly. For example, a suspended message with `reconsumeTimes=2` can become `3` merely because its first revive attempt encountered a transient read or write failure.

Please preserve the original value when constructing the retry record:

```java
PopConsumerRecord retryRecord = new PopConsumerRecord(
    System.currentTimeMillis(),
    record.getGroupId(),
    record.getTopicId(),
    record.getQueueId(),
    record.getRetryFlag(),
    nextInvisibleTime,
    record.getOffset(),
    record.getAttemptId(),
    record.isSuspend());
```

Please cover both synchronous and asynchronous failure paths with `suspend=true`. Reload the generated retry record from the store and assert that `suspend` remains true, `attemptTimes` is incremented exactly once, the original record is deleted only after the retry record is persisted, and a later successful `reviveRetry` does not increment `reconsumeTimes`.

wang-jiahua:
> **[P1] Avoid leaving the revive service permanently interrupted**

@fuyou001 You are right, and this hazard was introduced by my own change: the previous code did not restore the interrupt flag, so the exception from `acquire()` incidentally cleared it and the service recovered on the next round. Restoring the flag turned one incidental interrupt into a permanent busy loop, given that shutdown here works through the stopped flag and `wakeup()`, and `ServiceThread.waitForRunning` preserves the interrupt status.

Fixed in `771afdcd4` following your suggestion: incidental interrupts are consumed and the acquire is retried. If the service is being stopped, which covers a future `shutdown(true)`, the batch aborts cleanly instead of swallowing that shutdown interrupt; the records have not been deleted from the store yet, so they are reprocessed after the next start.

```java
while (true) {
    try {
        semaphore.acquire();
        break;
        if (this.isStopped()) {
            throw new RuntimeException("PopConsumerService stopped while acquiring the revive semaphore", e);
        }
        log.warn("PopConsumerService interrupted while acquiring the revive semaphore, retry");
    }
}
```

Added the regression test you described, `reviveShouldSurviveIncidentalInterruptWhileAcquiringSemaphore`: the concurrency is set to a single permit and the first read is held open, so the worker blocks inside `semaphore.acquire()` for the second record; the worker is then interrupted once and the outstanding read released. The batch completes normally, returning 2 with the scan window drained, and no exception escapes. Before this commit the worker dies with `RuntimeException: java.lang.InterruptedException`.

wang-jiahua:
> **[P1] Preserve the suspend flag in the backoff record**

@fuyou001 Good catch. Fixed in `771afdcd4` exactly as suggested: the backoff record now uses the nine-argument constructor and carries `record.isSuspend()` through, so the client intent expressed by `changeInvisibilityDuration(..., suspend=true)` survives a transient revive failure.

To be precise about the origin: the eight-argument constructor hard-codes `suspend` to `false` when it delegates, and the batch backoff path has always used it, so losing the flag predates this PR. What this PR changed is the exposure: converting synchronous throws and exceptionally completed futures into `false` routes two additional failure classes into that same backoff path, which is why it is worth fixing here.

The regression test `reviveBackoffRecordShouldPreserveSuspendFlag` covers both failure paths you asked for: one record fails synchronously and one asynchronously, both with `suspend=true`. It then reloads the persisted retry records from the store and asserts that `suspend` is still true and `attemptTimes` is incremented exactly once. Since `writeRecords(failureList)` runs before `deleteRecords(consumerRecords)`, the originals are only removed after the retry records are persisted. `reviveRetry` keeping `reconsumeTimes` unchanged for suspended records is already covered by the existing suspend tests. The assertion fails before this commit, with the logs showing `suspend=true` going in and `suspend=false` coming out.

The same eight-argument constructor issue exists in the cache callback proposed for #10667, and that work preserves the flag as well.

`PopConsumerServiceTest` is at 21/21 with zero Checkstyle violations on the current develop baseline.

RockteMQ-AI (approved):
### Summary

---

PR #10664, opened 2026-07-26
[ISSUE #10225] Check expired receipt handle locks before blocking

### LINKED ISSUE #10225: Throughput of in-flight message renewal degrades while some handle data locks have expired

Aias00:
## Summary
- check `ReceiptHandleGroup.HandleData` expired locks before entering the blocking semaphore wait
- keep the existing blocking fallback for non-expired locks
- add coverage that an expired lock is reacquired without waiting for the full caller timeout

## Tests
- `mvn -pl proxy -Dtest=ReceiptHandleGroupTest test`
- `git diff --check`

Fixes #10225

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/common/ReceiptHandleGroup.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/common/ReceiptHandleGroup.java
@@ -113,29 +113,44 @@ public class ReceiptHandleGroup {
 
         public Long lock(long timeoutMs) {
             try {
-                boolean result = this.semaphore.tryAcquire(timeoutMs, TimeUnit.MILLISECONDS);
                 long currentTimeMs = System.currentTimeMillis();
+                if (this.semaphore.tryAcquire()) {
+                    this.lastLockTimeMs.set(currentTimeMs);
+                    return currentTimeMs;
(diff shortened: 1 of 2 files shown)


Aias00:
Closing this because #10225 is already covered by the existing open PR #10226. I missed that PR in the earlier narrower search and do not want to duplicate the review queue.

RockteMQ-AI (approved):
### Summary

---

PR #10682, opened 2026-07-29
[ISSUE #10681] Skip lite offline cleanup without subscriptions

### LINKED ISSUE #10681: [Bug] Proxy lite client offline cleanup indexes empty subscriptions

- [x] I have searched the existing issues and pull requests.

All platforms.

### RocketMQ version

develop branch.

### JDK Version

Not applicable.

### Describe the Bug

`GrpcClientSettingsManager.offlineClientLiteSubscription` assumes a lite consumer's cached `Settings` always has at least one subscription entry:

```java
String topic = settings.getSubscription().getSubscriptions(0).getTopic().getName();
String group = settings.getSubscription().getGroup().getName();
```

If a lite consumer has incomplete settings, for example `ClientType.LITE_PUSH_CONSUMER` with no subscription entries, this path throws an `IndexOutOfBoundsException`. The exception is caught by the broad catch block, but the cleanup path then logs an error with the full settings object and skips the offline cleanup through an exceptional path.

### Steps to Reproduce

1. Build `Settings` with `clientType = LITE_PUSH_CONSUMER`.
2. Do not add any subscription entries.
3. Call `GrpcClientSettingsManager.offlineClientLiteSubscription(ctx, clientId, settings)`.
4. The method tries to read `getSubscriptions(0)` before validating the count.

Incomplete or malformed lite-consumer settings should be ignored by the offline cleanup path before indexing into the subscription list. No `IndexOutOfBoundsException` should be needed for control flow.

### What Did You See Instead?

The method relies on the catch block after `getSubscriptions(0)` throws `IndexOutOfBoundsException`, and logs the full settings object at error level.

### Additional Context

This is a small Proxy gRPC runtime cleanup hardening issue. It keeps client disconnect/offline handling from producing noisy error logs for incomplete settings and avoids logging full settings unnecessarily.

Aias00:
### Summary

- Skip lite-client offline cleanup when settings have no subscription section or no subscription entries.
- Avoid indexing `subscriptions(0)` for incomplete lite-consumer settings.
- Add a regression test for lite consumers without subscription entries.

### Motivation

Closes #10681.

`GrpcClientSettingsManager.offlineClientLiteSubscription` previously assumed lite-consumer settings always contain at least one subscription entry. Incomplete settings could throw `IndexOutOfBoundsException` and fall into the broad error log path, including the full settings object.

### Tests

- `mvn -pl proxy -Dtest=GrpcClientSettingsManagerTest -DfailIfNoTests=false test`

Result: BUILD SUCCESS; 8 tests passed; checkstyle reported 0 violations. The run emits existing JaCoCo instrumentation warnings under the local JDK, and one existing test intentionally logs a simulated sync failure, but tests and build completed successfully.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManager.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManager.java
@@ -236,6 +236,9 @@ public class GrpcClientSettingsManager extends ServiceThread implements StartAnd
             && ClientType.LITE_SIMPLE_CONSUMER != settings.getClientType()) {
             return;
         }
+        if (!settings.hasSubscription() || settings.getSubscription().getSubscriptionsCount() == 0) {
+            return;
+        }
         try {
             String topic = settings.getSubscription().getSubscriptions(0).getTopic().getName();
             String group = settings.getSubscription().getGroup().getName();
--- a/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManagerTest.java
+++ b/proxy/src/test/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManagerTest.java
@@ -157,6 +157,20 @@ public class GrpcClientSettingsManagerTest extends BaseActivityTest {
     }
 
     @Test
+    public void testOfflineClientLiteSubscription_LiteConsumerWithoutSubscriptions() {
+        Settings settings = Settings.newBuilder()
+            .setClientType(ClientType.LITE_PUSH_CONSUMER)
+            .setSubscription(Subscription.newBuilder()
+                .setGroup(Resource.newBuilder().setName("testGroup").build())
+                .build())
+            .build();
+
+        grpcClientSettingsManager.offlineClientLiteSubscription(ctx, clientId, settings);
+
+        verify(messagingProcessor, never()).syncLiteSubscription(any(), any(), anyLong());
+    }
+
+    @Test
     public void testOfflineClientLiteSubscription_ValidLiteConsumer_Success() {
         Subscription subscription = Subscription.newBuilder()
             .setGroup(Resource.newBuilder().setName("testGroup").build())


RockteMQ-AI (approved):

### Summary
Adds an early-return guard in `GrpcClientSettingsManager.offlineClientLiteSubscription()` to skip processing when the lite-consumer settings have no subscription section or zero subscription entries, preventing an `IndexOutOfBoundsException` on `getSubscriptions(0)`. Includes a regression test.

### Findings
- **[Info]** `GrpcClientSettingsManager.java:239` — The guard `!settings.hasSubscription() || settings.getSubscription().getSubscriptionsCount() == 0` correctly prevents the `IndexOutOfBoundsException` described in #10681. Clean and minimal.
- **[Info]** `GrpcClientSettingsManagerTest.java:159-172` — Good regression test using `LITE_PUSH_CONSUMER` with a group but no subscription entries. Verifies `syncLiteSubscription` is never called.

### Suggestions
- Consider also testing with `LITE_SIMPLE_CONSUMER` client type for completeness, since the method guards both `LITE_PUSH_CONSUMER` and `LITE_SIMPLE_CONSUMER` at line 235-237. This is optional — the current test already validates the new guard logic.

### Verdict
The fix is correct, minimal, and well-tested. LGTM. ✅

---

f1amingo:
Thanks for the PR. However, this guards a state that cannot occur by design.

In the liteTopic model, the bind topic is mandatory on the SDK side — a lite consumer always carries exactly one bind-topic entry in its Settings, established before any lite subscription exists. So `getSubscriptions(0)` is safe here by contract.

If this state ever appeared, it would signal a protocol violation that should fail loudly (which the existing catch + error log already does), not be silently skipped on the offline path.

I'd suggest walking through the end-to-end liteTopic subscription flow (Settings sync → bind topic → SyncLiteSubscription) first — it would make it clear why this invariant holds.

Aias00:
> Thanks for the PR. However, this guards a state that cannot occur by design.
> 
> In the liteTopic model, the bind topic is mandatory on the SDK side — a lite consumer always carries exactly one bind-topic entry in its Settings, established before any lite subscription exists. So `getSubscriptions(0)` is safe here by contract.
> 
> If this state ever appeared, it would signal a protocol violation that should fail loudly (which the existing catch + error log already does), not be silently skipped on the offline path.
> 
> I'd suggest walking through the end-to-end liteTopic subscription flow (Settings sync → bind topic → SyncLiteSubscription) first — it would make it clear why this invariant holds.

sounds good to me

RockteMQ-AI (approved):
### Summary

---

Issue #9987, opened 2026-01-06
[Bug] Rocketmq namesrv解析网卡v6地址存在问题


tcxdgit:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

纯单栈v6环境启动运行namesrv

### RocketMQ version

5.1.3

### JDK Version

1.8

### Describe the Bug

v6环境部署namesrv组件。日志发现解析v6地址存在问题

### Steps to Reproduce

使用bin/mqnamesrv -c /root/rocketmq-5.1.3-1/conf/namesrv.conf启动namesrv组件。发现日志报错，类似v6地址解析异常。后续broker无法注册namesrv。

<img width="2124" height="380" alt="Image" src="https://github.com/user-attachments/assets/11a3e02a-ca5f-49fa-8a1e-a1d0be6a0ddf" />

<img width="559" height="232" alt="Image" src="https://github.com/user-attachments/assets/2bc68e9c-6398-41fa-9c3a-e96db582a95f" />

### What Did You Expect to See?

应该解析得到[202:202:0:0:0:0:81:1811]:9876 这样的地址。通过 ping6 202:202:0:0:0:0:81:1811和 nc -vz -6 202:202:0:0:0:0:81:1811 9876都可以连通。

### What Did You See Instead?

createChannel: begin to connect remote host[[202:202:0:0:0:0:81:1811%enp1s0]:9876]
java.net.UnknownHostException: [202: invalid IPv6 address literal

### Additional Context

no

tcxdgit:
I added "202:202::81:1811 node1" to /etc/hosts, then changed all addresses to node1 and it worked. Is this a problem with RocketMQ itself to resolve v6 addresses?

<img width="649" height="68" alt="Image" src="https://github.com/user-attachments/assets/1886bc2f-17ae-4246-b965-9bb49e3ced9a" />

<img width="824" height="87" alt="Image" src="https://github.com/user-attachments/assets/0056318f-8767-486e-9541-b1a6c14d19eb" />

[...]

PR #10694, opened 2026-07-29
[ISSUE #9987] Fix bracketed IPv6 address parsing with scope id

AliasJeff:

Related to [#9987](https://github.com/apache/rocketmq/issues/9987)

### Brief Description

`NetworkUtil.string2SocketAddress` currently parses socket addresses by using `lastIndexOf(":")`.

This works fine for normal `host:port` addresses, but it doesn't handle bracketed IPv6 addresses correctly. For example, with `[2001:db8::1]:9876`, the brackets may remain in the host part after parsing.

This PR adds handling for the standard `[IPv6]:port` format before using the existing parser logic.

For IPv6 addresses with a scope id, the scope part is removed before creating `InetSocketAddress`. Also, `socketAddress2String` is updated to always output IPv6 addresses in `[IPv6]:port` format.

The issue was reported as fixed after upgrading to 5.3.1, but this change focuses on fixing the remaining parsing problem in `NetworkUtil` and does not change the existing IPv4 or hostname behavior.

```bash
mvn -pl common -Dtest=NetworkUtilTest test
```

Result:
BUILD SUCCESS

I also verified the regression case by temporarily reverting the parser implementation to the old version while keeping the new test. The test testString2SocketAddressWithBracketedIPv6Scope fails because the parsed address becomes null.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
@@ -26,6 +26,7 @@ import java.net.InetSocketAddress;
 import java.net.NetworkInterface;
 import java.net.SocketAddress;
 import java.net.SocketException;
+import java.net.UnknownHostException;
 import java.nio.channels.Selector;
 import java.nio.channels.spi.SelectorProvider;
 import java.util.ArrayList;
@@ -192,28 +193,45 @@ public class NetworkUtil {
     }
 
     public static SocketAddress string2SocketAddress(final String addr) {
+        if (addr == null) {
+            throw new IllegalArgumentException("Socket address must not be null");
+        }
+
+        String host;
+        int port;
+
         if (addr.startsWith("[")) {
             int end = addr.indexOf(']');
-            if (end > 0 && end + 1 < addr.length() && addr.charAt(end + 1) == ':') {
-                String host = addr.substring(1, end);
-                int scope = host.indexOf('%');
-                if (scope > 0) {
-                    host = host.substring(0, scope);
-                }
-                String port = addr.substring(end + 2);
-                return new InetSocketAddress(host, Integer.parseInt(port));
+            if (end <= 0 || end + 1 >= addr.length() || addr.charAt(end + 1) != ':') {
+                throw new IllegalArgumentException("Invalid socket address: " + addr);
+            }
+
+            host = addr.substring(1, end);
+            port = Integer.parseInt(addr.substring(end + 2));
+        } else {
+            int index = addr.lastIndexOf(':');
+            if (index <= 0 || index == addr.length() - 1) {
+                throw new IllegalArgumentException("Invalid socket address: " + addr);
+            }
+
+            host = addr.substring(0, index);
+            port = Integer.parseInt(addr.substring(index + 1));
+        }
+
+        if (host.indexOf(':') >= 0) {
+            try {
+                return new InetSocketAddress(InetAddress.getByName(host), port);
+            } catch (UnknownHostException e) {
+                throw new IllegalArgumentException("Invalid socket address: " + addr, e);
             }
         }
-        int split = addr.lastIndexOf(":");
-        String host = addr.substring(0, split);
-        String port = addr.substring(split + 1);
-        return new InetSocketAddress(host, Integer.parseInt(port));
+        return new InetSocketAddress(host, port);
     }
 
     public static String socketAddress2String(final SocketAddress addr) {
         StringBuilder sb = new StringBuilder();
         InetSocketAddress inetSocketAddress = (InetSocketAddress) addr;
-        sb.append(normalizeHostAddress(inetSocketAddress.getAddress()));
+        sb.append(inetSocketAddress.getAddress().getHostAddress());
         sb.append(":");
         sb.append(inetSocketAddress.getPort());
         return sb.toString();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Fixes bracketed IPv6 address parsing in `NetworkUtil.string2SocketAddress` to correctly handle the standard `[IPv6]:port` format, including scope ID stripping. Also updates `socketAddress2String` to output IPv6 addresses in the standard bracketed format.

### Findings
- **[Info]** `NetworkUtil.java:195-207` — The bracketed IPv6 parsing logic is clean and correct. Scope ID (`%interface`) stripping before `InetSocketAddress` creation is a reasonable design choice since scope IDs are interface-local and may not be valid on remote hosts.
- **[Info]** `NetworkUtil.java:215` — `socketAddress2String` now uses `normalizeHostAddress(InetAddress)` which wraps IPv6 in brackets per RFC 3986. This changes the output format for IPv6 from `host:port` to `[host]:port`. This is the correct standard format and improves round-trip consistency with `string2SocketAddress`. Callers that parse the output by splitting on `:` should be aware of this change.
- **[Info]** `NetworkUtilTest.java` — Good test coverage for the new scenarios (bracketed IPv6 with scope, without scope, and `convert2IpString` round-trip).

### Suggestions
- Consider adding a test for `socketAddress2String` directly (not just via `convert2IpString`) to explicitly verify the bracketed output format for IPv6.
- Minor: the empty-bracket edge case `[]:9876` would create `InetSocketAddress("", 9876)` which resolves to localhost. Unlikely in practice but could add a guard if desired.

Overall: Clean, focused fix with good test coverage. LGTM.

---

fuyou001:
[P1] Please preserve the IPv6 scope/zone instead of stripping it unconditionally.

For link-local IPv6 addresses (`fe80::/10`), the scope identifies the network interface and is required to select the correct destination. Removing `%scope` changes the address semantics and may make connection attempts fail or select the wrong interface. The current test uses a non-link-local address with a synthetic scope, so it only demonstrates that parsing succeeds after information is discarded.

Please retain the scope and let `Inet6Address`/`InetSocketAddress` resolve a valid interface name or numeric scope. If the scope cannot be resolved, returning a clear parsing failure is safer than silently changing the address.

Please add a parse-format-parse test for a link-local address with a real interface scope, a numeric scope-id case, malformed bracket and missing/out-of-range port cases, and IPv4/hostname/unscoped-IPv6 compatibility cases. Also, `socketAddress2String` changes the externally visible IPv6 representation to `[IPv6]:port`; please verify rolling-upgrade compatibility because an old node parsing a new node's bracketed output retains the leading `[` in the host string.

AliasJeff:
@fuyou001  Thanks for the review. I updated the parser to preserve IPv6 scope information instead of stripping it. Scoped addresses are now resolved via InetAddress/Inet6Address, and invalid scopes will fail parsing now.

I also added more coverage for ipv6 scope handling, invalid inputs, and backward compatibility cases.

For socketAddress2String, I kept the existing output format to avoid introducing a rolling upgrade compatibility issue.

RockteMQ-AI (commented):

### Summary
New commit `ee2de3b` changes the approach from the previous version: instead of stripping IPv6 scope IDs, the parser now **preserves** them by routing IPv6 hosts through `InetAddress.getByName(host)`. This is a better design — link-local addresses (e.g. `fe80::1%eth0`) are only useful when the scope is retained.

### Changes Since Previous Review

| Aspect | Previous (commit 1) | Now (commit 2) |
|--------|---------------------|-----------------|
| Scope ID | Stripped before `InetSocketAddress` creation | Preserved via `InetAddress.getByName(host)` |
| IPv6 routing | All addresses through `InetSocketAddress(host, port)` | IPv6 (host contains `:`) through `InetAddress.getByName()` |
| Null check | None | Added `IllegalArgumentException` for null input |
| Error messages | Generic | Descriptive `"Invalid socket address: " + addr` |
| Test coverage | 3 new tests | 7+ new tests including malformed input and round-trip |

### Findings

- **[Info]** `NetworkUtil.java:218-223` — The scope preservation via `InetAddress.getByName(host)` is the correct approach. On Java 11+ (which this project uses), `getByName` correctly parses `fe80::1%1` and `fe80::1%eth0` syntax. This is a meaningful improvement over the previous scope-stripping approach.

- **[Info]** `NetworkUtil.java:195-212` — Input validation is thorough: null check, bracket format validation (`end <= 0`, missing `:` after `]`), and non-bracketed format validation (`index <= 0`). Good defensive coding.

- **[Info]** `NetworkUtil.java:215-224` — The branching on `host.indexOf(':') >= 0` cleanly separates IPv6 from IPv4/hostname paths. IPv4 and hostnames continue to use the lightweight `new InetSocketAddress(host, port)` constructor.

- **[Warning]** `NetworkUtil.java:224` — **Port range validation**: The test `testString2SocketAddressWithMalformedAddress` expects `IllegalArgumentException` for ports `65536` and `-1`. However, `Integer.parseInt()` will succeed for these values (they are valid ints), and `new InetSocketAddress(host, port)` does **not** validate port range — it accepts any int. If the test passes, it may be because the test environment resolves differently. Consider adding explicit port range validation:
  ```java
  if (port < 0 || port > 65535) {
      throw new IllegalArgumentException("Port out of range: " + port);
  }
  ```

- **[Info]** `NetworkUtilTest.java` — Excellent test coverage. The `assertParseFormatParse` helper ensures consistent format validation. The `Assume.assumeTrue` guard for the real link-local test is the right pattern for environment-dependent tests.

### Verdict
The scope preservation change is a clear improvement. The code is clean, well-structured, and thoroughly tested. One minor concern about port range validation (may be a test gap). Otherwise LGTM.

---

RockteMQ-AI (approved):
### Summary

---

PR #10697, opened 2026-07-30
[ISSUE #10696] Fix async request future cleanup on synchronous send failure

### LINKED ISSUE #10696: [Bug] Async request future remains registered after synchronous send failure

- [x] I searched existing GitHub Issues, Discussions, and pull requests and believe this is not a duplicate.
- [x] I confirmed that this bug belongs to `apache/rocketmq`.

- OS: Ubuntu 22.04, Linux 5.15.0-185-generic, x86_64

### RocketMQ version

- Branch: `develop`
- Commit: `00e45b8a6db23efbe756d0306f10716156cfd4dd`

### JDK Version

- OpenJDK 11.0.31

### Describe the Bug

The three asynchronous request-reply overloads that accept a `RequestCallback` register a `RequestResponseFuture` in the singleton `RequestFutureHolder` before invoking the underlying send method:

- `request(Message, RequestCallback, long)`
- `request(Message, MessageQueueSelector, Object, RequestCallback, long)`
- `request(Message, MessageQueue, RequestCallback, long)`

If `sendDefaultImpl`, `sendSelectImpl`, or `sendKernelImpl` throws synchronously before returning, the exception is propagated to the caller but the registered future is not removed.

The stale future remains in `requestFutureTable` until a later timeout scan. Besides retaining the request and callback longer than necessary, the timeout path can execute the callback after the caller has already observed the synchronous exception.

The synchronous request overloads already avoid this by removing their future in a `finally` block. This gap is also distinct from #10613 / #10614, which addressed premature and duplicate callbacks after send initiation.

### Steps to Reproduce

1. Start a producer and invoke an asynchronous request with a callback.
2. Make send initiation fail synchronously, for example:
   - no route is available for the topic;
   - the supplied `MessageQueueSelector` throws;
   - an explicit queue resolves to no broker address.
3. Catch the resulting `MQClientException`.
4. Inspect `RequestFutureHolder.getInstance().getRequestFutureTable()` using the message correlation ID.

A minimal regression assertion is:

```java
assertThrows(MQClientException.class,
    () -> producerImpl.request(message, requestCallback, timeout));

assertFalse(RequestFutureHolder.getInstance()
    .getRequestFutureTable()
    .containsKey(correlationId));
```

The final assertion fails on the current `develop` branch for all three callback overloads.

When send initiation throws synchronously, the exact `RequestResponseFuture` registered by that invocation should be removed immediately. The original exception should still be propagated, and no timeout callback should be delivered for a request that was never successfully handed off.

A normally returned asynchronous send invocation must continue to retain its future until reply, send failure, or timeout processing claims it.

### What Did You See Instead?

The API throws the synchronous send exception, but the future remains registered and can later be processed as an expired request.

### Additional Context

A race-safe cleanup can use `ConcurrentHashMap.remove(correlationId, requestResponseFuture)` so that a replacement value under the same correlation ID is not removed. Regression coverage should include the default, selector, and explicit-queue overloads, plus a normal-path assertion that the future is retained after a successful asynchronous handoff.

ai-yang:

- Fixes #10696

### Brief Description

The asynchronous request-reply overloads register a `RequestResponseFuture` before invoking the underlying send method. When send initiation throws synchronously, the exception reaches the caller but the future remains in `RequestFutureHolder` until timeout processing, which retains request state and can deliver a callback after the synchronous failure.

This change:

- removes the exact registered future when `sendDefaultImpl`, `sendSelectImpl`, or `sendKernelImpl` does not return normally;
- uses `remove(correlationId, requestResponseFuture)` so a concurrent replacement under the same key is not removed;
- preserves the future after a normal asynchronous handoff so reply, send-failure, and timeout paths continue to own completion;
- updates the existing synchronous-exception test semantics and adds regression coverage for the default, selector, and explicit-queue callback overloads.

Verified on 2026-09-11 at head `e340eb4d4b1372d04c9cdf3c57799a6c9724662e`, rebased onto `develop` at `1a50c6e4e35524ac22055b76fc0ff2a6d3cff99f`, using JDK 8 and Maven 3.8.1:

```bash
  -Dtest=DefaultMQProducerTest,DefaultMQProducerImplTest \
  -Dsurefire.failIfNoSpecifiedTests=false clean test
```

- `DefaultMQProducerTest`: 41 tests passed.
- `DefaultMQProducerImplTest`: 37 tests passed.
- Total: 78 tests, 0 failures, 0 errors, 0 skips; all 4 reactor modules succeeded.
- Checkstyle: 0 violations; SpotBugs: 0 bug instances and 0 errors.
- `git diff --check`: passed.

The new regression assertion was previously verified to fail before the production fix because the correlation ID remained in `requestFutureTable`. Earlier full-client reactor evidence (995 tests, 0 failures/errors, 1 skip) predates this rebase; the latest local run above is targeted, and the full CI matrix has been triggered for the refreshed head.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java
@@ -1659,22 +1659,31 @@ public class DefaultMQProducerImpl implements MQProducerInner {
         RequestFutureHolder.getInstance().getRequestFutureTable().put(correlationId, requestResponseFuture);
 
         long cost = System.currentTimeMillis() - beginTimestamp;
-        this.sendDefaultImpl(msg, CommunicationMode.ASYNC, new SendCallback() {
-            @Override
-            public void onSuccess(SendResult sendResult) {
-                // Only mark the request as sent here. The user callback must fire when the reply
-                // arrives (processReplyMessage), on timeout (scanExpiredRequest), or on send
-                // failure (requestFail). Invoking it here delivers a premature onSuccess(null)
-                // and, combined with the later reply/timeout callback, causes a double callback.
-                requestResponseFuture.setSendRequestOk(true);
-            }
+        boolean sendInvocationCompleted = false;
+        try {
+            this.sendDefaultImpl(msg, CommunicationMode.ASYNC, new SendCallback() {
+                @Override
+                public void onSuccess(SendResult sendResult) {
+                    // Only mark the request as sent here. The user callback must fire when the reply
+                    // arrives (processReplyMessage), on timeout (scanExpiredRequest), or on send
+                    // failure (requestFail). Invoking it here delivers a premature onSuccess(null)
+                    // and, combined with the later reply/timeout callback, causes a double callback.
+                    requestResponseFuture.setSendRequestOk(true);
+                }
 
-            @Override
-            public void onException(Throwable e) {
-                requestResponseFuture.setCause(e);
-                requestFail(correlationId);
+                @Override
+                public void onException(Throwable e) {
+                    requestResponseFuture.setCause(e);
+                    requestFail(correlationId);
+                }
+            }, timeout - cost);
+            sendInvocationCompleted = true;
+        } finally {
+            if (!sendInvocationCompleted) {
+                RequestFutureHolder.getInstance().getRequestFutureTable()
+                    .remove(correlationId, requestResponseFuture);
             }
-        }, timeout - cost);
+        }
     }
 
     public Message request(final Message msg, final MessageQueueSelector selector, final Object arg,
@@ -1720,18 +1729,27 @@ public class DefaultMQProducerImpl implements MQProducerInner {
         RequestFutureHolder.getInstance().getRequestFutureTable().put(correlationId, requestResponseFuture);
 
         long cost = System.currentTimeMillis() - beginTimestamp;
-        this.sendSelectImpl(msg, selector, arg, CommunicationMode.ASYNC, new SendCallback() {
-            @Override
-            public void onSuccess(SendResult sendResult) {
-                requestResponseFuture.setSendRequestOk(true);
-            }
+        boolean sendInvocationCompleted = false;
+        try {
+            this.sendSelectImpl(msg, selector, arg, CommunicationMode.ASYNC, new SendCallback() {
+                @Override
+                public void onSuccess(SendResult sendResult) {
+                    requestResponseFuture.setSendRequestOk(true);
+                }
 
-            @Override
-            public void onException(Throwable e) {
-                requestResponseFuture.setCause(e);
-                requestFail(correlationId);
+                @Override
+                public void onException(Throwable e) {
+                    requestResponseFuture.setCause(e);
+                    requestFail(correlationId);
+                }
+            }, timeout - cost);
+            sendInvocationCompleted = true;
+        } finally {
+            if (!sendInvocationCompleted) {
+                RequestFutureHolder.getInstance().getRequestFutureTable()
+                    .remove(correlationId, requestResponseFuture);
             }
-        }, timeout - cost);
+        }
 
     }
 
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (commented):

### Summary
This PR fixes a resource leak in the async request-reply path of `DefaultMQProducerImpl`. When `sendDefaultImpl`, `sendSelectImpl`, or `sendKernelImpl` throws synchronously (before async handoff), the registered `RequestResponseFuture` remains in `RequestFutureHolder.requestFutureTable` until timeout. The fix wraps each send invocation in a try/finally block with a `sendInvocationCompleted` flag, removing the future only when the send call itself fails.

### Findings

- **[Info]** `DefaultMQProducerImpl.java:1662-1689` — The try/finally + boolean flag pattern is clean and idiomatic. Using `ConcurrentHashMap.remove(key, value)` (conditional remove) correctly avoids clobbering a concurrent replacement under the same correlation ID. Good defensive design.

- **[Info]** `DefaultMQProducerImpl.java` — All three `request()` overloads (default, selector, kernel/queue) apply the same pattern consistently. This uniformity makes the fix easy to verify and maintain.

- **[Info]** `DefaultMQProducerTest.java:483-508` — The test rename from `testAsyncRequest_OnException` to `testAsyncRequest_SynchronousExceptionRemovesFuture` better reflects the actual behavior being tested. The assertion change from checking callback invocation to checking future removal is semantically correct for this scenario.

- **[Info]** `selector/DefaultMQProducerImplTest.java:134-175` — Good coverage of all three send paths (default, selector, queue) with dedicated regression tests. The `@Before`/`@After` cleanup of `requestFutureTable` prevents test pollution.

- **[Info]** The `timeout - cost` computation is a pre-existing concern (could go negative if `cost > timeout`), but is out of scope for this PR.

### Verdict
Well-structured, focused fix for a real resource leak. The code is clean, tests are thorough (995 tests passing), and the conditional `remove(key, value)` shows attention to concurrent correctness. No issues found.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Updated and ready for another review round:

- Rebased onto the current `apache/rocketmq:develop` (`e348efa66b08eb645ee123706ea6492fa9a3ad35`).
- Refreshed the branch to signed-off commit `56a47e8292251a2858b5fa9baab0a8e08f7261bf`.
- Clean targeted verification passed:
  - `mvn -Dmaven.repo.local=/developer/wangrui/.m2/repository -pl client -am -DskipITs -Dtest=DefaultMQProducerTest,DefaultMQProducerImplTest -Dsurefire.failIfNoSpecifiedTests=false clean test`
  - 77 tests, 0 failures/errors/skips.
  - All 4 reactor modules succeeded; Checkstyle and SpotBugs reported no violations.

The background name-server lookup warnings in the test log are expected in the isolated test environment and did not affect the result.

@drpmma @RongtongJin, could you please take a look when convenient?

ai-yang:
Refreshed onto `develop` at `1a50c6e4e` and pushed signed-off commit `e340eb4d4`.

Resolved the import conflict in `DefaultMQProducerImplTest` by retaining both the upstream `EndTransactionRequestHeader` and regression-test `After` imports. The request-future cleanup implementation is unchanged by the rebase.

Clean JDK 8 verification: `mvn -pl client -am -DskipITs -Dtest=DefaultMQProducerTest,DefaultMQProducerImplTest -Dsurefire.failIfNoSpecifiedTests=false clean test` passed 78/78 tests across the two classes. All 4 reactor modules passed with Checkstyle and SpotBugs enabled. The PR description now distinguishes this latest run from earlier full-reactor evidence.

@drpmma @RongtongJin, could you take a human review of the synchronous-send-failure cleanup when convenient? The updated commit also triggers a new full CI run.

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The try-finally pattern correctly cleans up `RequestFutureHolder` entries when `sendDefaultImpl`/`sendSelectImpl`/`sendKernelImpl` throws synchronously (e.g., serialization failure, no available broker).

Key points:
- `sendInvocationCompleted` flag is a clean way to detect synchronous failure before the async callback takes over
- `remove(correlationId, requestResponseFuture)` with value check prevents removing a different future if the ID was somehow reused
- Applied consistently across all three `request()` overloads
- Tests verify cleanup on sync failure and normal async path

---

PR #9293, opened 2025-03-28
[ISSUE #9288] Support the disablement of producer registration and fast channel shutdown

ymwneu:
support reject producer when not transaction producer

### Which Issue(s) This PR Fixes

Fixes #9288 

### Brief Description

1. The current broker retains all producer channel information, which can consume a significant amount of resources when there are a large number of producers. In non-transactional message scenarios, the producer channel information is not useful. It would be beneficial to consider adding a switch to not record PID information in these scenarios.
2. Support fast channel event processing to improve the speed of disconnection.

Issue #8846, opened 2024-10-21
[Bug] Fix ProducerManager concurrency issue


lizhanhui:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

N/A

### RocketMQ version

ALL

### Describe the Bug

Concurrency issue

### Steps to Reproduce

N/A

### What Did You Expect to See?

N/A

### What Did You See Instead?

N/A

PR #10699, opened 2026-07-30
[ISSUE #10698] Restore atomic producer group registration in ProducerManager

### LINKED ISSUE #10698: [Bug] ProducerManager new-group registration can be detached by concurrent inactive-channel scan
### Affected baseline

`develop` at `00e45b8a6db23efbe756d0306f10716156cfd4dd`.

### Problem

`ProducerManager.registerProducer()` publishes a newly created producer-group entry before adding its first channel. Concurrent `scanNotActiveChannel()` can observe that empty inner map, remove the group, and emit `GROUP_UNREGISTER`. The registration thread then inserts the channel into the now-detached map and returns successfully.

The resulting indexes disagree:

- `clientChannelTable` contains `clientId -> channel`;
- the detached inner map contains `channel -> ClientChannelInfo`;
- `groupChannelTable` no longer contains the producer group.

Group lookup and transaction-producer selection therefore fail even though registration returned normally, and listeners receive a false group-unregister event.

### Deterministic interleaving

1. `registerProducer()` creates an empty `ConcurrentHashMap`.
2. It publishes that map with `groupChannelTable.putIfAbsent(group, channelTable)`.
3. Before the first channel lookup/insertion, `scanNotActiveChannel()` observes `chlMap.isEmpty()`, removes the outer entry, and emits `GROUP_UNREGISTER`.
4. Registration inserts into its detached map and updates `clientChannelTable`.

A regression test can pause the registering thread at the Channel's first `hashCode()` call with two latches. On current code that first hash occurs only after the empty outer entry has been published, making the scan/removal deterministic.

### Expected behavior

A newly published producer-group map must already contain its first channel. After registration completes:

- the group exists in `groupChannelTable`;
- its channel mapping exists;
- `findChannel(clientId)` returns the same channel;
- no spurious `GROUP_UNREGISTER` is emitted;
- fast-channel producer-group attributes and the existing registration-gate behavior remain unchanged.

### Root cause / regression history

This is a regression of [#8846](https://github.com/apache/rocketmq/issues/8846) and merged fix [#8847](https://github.com/apache/rocketmq/pull/8847), which pre-populated the candidate inner map before publishing it. [#9293](https://github.com/apache/rocketmq/pull/9293) later added producer-registration gating and fast-channel bookkeeping but restored the empty-map-first ordering. `ProducerManager.java` has not changed since #9293.

### Proposed scope

- Pre-populate a candidate inner map before publishing it to `groupChannelTable`.
- If another thread wins outer insertion, use `putIfAbsent` on the winning inner map.
- Use a `newChannel` flag so client-id mapping, logging, and the fast-channel attribute execute exactly once.
- Preserve the current registration-gate truth table, timestamp refresh, and registration-time statistics.
- Add a latch-controlled regression test plus fast-channel attribute coverage.

This issue is intentionally limited to the new-group empty-map publication regression. Broader synchronization for an already existing group becoming empty during concurrent lifecycle operations should be handled separately.

I am working on a focused fix and will submit a PR against `develop`.

ai-yang:

- Fixes #10698

### Brief Description

Restore atomic publication of a newly registered producer group without changing the registration-gate or fast-channel behavior introduced later.

- Pre-populate a candidate group map with its first channel before publishing it.
- Fall back to `putIfAbsent` on the winning map when another registration creates the group first.
- Execute client-id indexing, connection logging, and fast-channel attributes only for the thread that inserts a new group/channel mapping.
- Refresh the winner's existing `ClientChannelInfo` timestamp for duplicate registrations.

### Root Cause

`registerProducer()` published an empty inner map to `groupChannelTable`. A concurrent inactive-channel scan could remove the empty group and emit `GROUP_UNREGISTER` before registration inserted its first channel, leaving the channel and client-id mappings detached from the producer-group index.

This is a regression of #8846 / #8847 after #9293 restored the empty-map-first ordering while adding registration gating and fast-channel processing.

### Impact

New producer groups cannot disappear during concurrent scanning, and the group/channel/client-id indexes remain consistent. The current registration switch truth table, timestamp refresh, statistics, and listener semantics are preserved. This PR intentionally does not expand into broader synchronization for already-existing empty groups.

- Deterministic red test on the previous implementation: the scanner removed `FooBar` while registration was blocked at the Channel's first `hashCode()`, and the group assertion failed.
- `ProducerManagerTest`: 9 tests passed, including latch-controlled concurrent scanning and fast-channel attribute coverage.
- Full Broker test suite: 754 tests passed, 0 failures, 0 errors, 4 skipped.
- Reactor dependency build/install: 10 modules succeeded.
- Maven Checkstyle: 0 violations.
- `git diff --check`.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/ProducerManager.java
@@ -206,7 +206,8 @@ public class ProducerManager {
     public void registerProducer(final String group, final ClientChannelInfo clientChannelInfo) {
 
         long start = System.currentTimeMillis();
-        ClientChannelInfo clientChannelInfoFound;
+        ClientChannelInfo clientChannelInfoFound = null;
+        boolean newChannel = false;
 
         ConcurrentMap<Channel, ClientChannelInfo> channelTable = this.groupChannelTable.get(group);
         // note that we must take care of the exist groups and channels,
@@ -232,24 +233,30 @@ public class ProducerManager {
         }
 
         if (null == channelTable) {
-            channelTable = new ConcurrentHashMap<>();
-            ConcurrentMap<Channel, ClientChannelInfo> prev = this.groupChannelTable.putIfAbsent(group, channelTable);
-            channelTable = prev != null ? prev : channelTable;
+            ConcurrentMap<Channel, ClientChannelInfo> newChannelTable = new ConcurrentHashMap<>();
+            newChannelTable.put(clientChannelInfo.getChannel(), clientChannelInfo);
+            ConcurrentMap<Channel, ClientChannelInfo> prev =
+                this.groupChannelTable.putIfAbsent(group, newChannelTable);
+            if (prev == null) {
+                channelTable = newChannelTable;
+                newChannel = true;
+            } else {
+                channelTable = prev;
+            }
         }
 
-        clientChannelInfoFound = channelTable.get(clientChannelInfo.getChannel());
-        // Add client-channel info to existing producer group
-        if (null == clientChannelInfoFound) {
-            channelTable.put(clientChannelInfo.getChannel(), clientChannelInfo);
+        if (!newChannel) {
+            clientChannelInfoFound = channelTable.putIfAbsent(clientChannelInfo.getChannel(), clientChannelInfo);
+            newChannel = clientChannelInfoFound == null;
+        }
+
+        if (newChannel) {
             clientChannelTable.put(clientChannelInfo.getClientId(), clientChannelInfo.getChannel());
             log.info("new producer connected, group: {} channel: {}", group, clientChannelInfo.toString());
             if (this.brokerConfig != null && this.brokerConfig.isEnableFastChannelEventProcess()) {
                 ClientChannelAttributeHelper.addProducerGroup(clientChannelInfo.getChannel(), group);
             }
-        }
-
-        // Refresh existing client-channel-info update-timestamp
-        if (clientChannelInfoFound != null) {
+        } else {
             clientChannelInfoFound.setLastUpdateTimestamp(System.currentTimeMillis());
         }
 
--- a/broker/src/test/java/org/apache/rocketmq/broker/client/ProducerManagerTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/client/ProducerManagerTest.java
@@ -18,11 +18,14 @@ package org.apache.rocketmq.broker.client;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
+import io.netty.channel.embedded.EmbeddedChannel;
 import java.lang.reflect.Field;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (commented):

### Summary
This PR fixes a regression in `ProducerManager.registerProducer()` where the group's inner map was published empty before the first channel was inserted. A concurrent inactive-channel scan could remove the empty group and emit `GROUP_UNREGISTER` before registration completed, leaving the channel and client-id mappings orphaned. The fix pre-populates the candidate map with the first channel before atomic publication.

### Findings

- **[Positive]** `ProducerManager.java:236-247` — Pre-populating `newChannelTable` with the first channel before `putIfAbsent` is the correct fix. This eliminates the window where the scanner could see an empty group map.

- **[Positive]** The `newChannel` flag cleanly separates the "first channel in group" path from the "refresh existing" path. The logic correctly handles both the race where we win the `putIfAbsent` and where we lose it.

- **[Positive]** When we lose the `putIfAbsent` race (`prev != null`), the code correctly falls through to the `!newChannel` block which does `putIfAbsent` on the winner's map. This ensures our channel is still registered even if another thread created the group first.

- **[Info]** `ProducerManager.java:250-252` — The `newChannel` flag is reused for both group-level and channel-level "newness". While functionally correct (in both cases "new" means "this channel wasn't in the table before"), a reader might need a moment to trace the two assignment sites. Consider a brief inline comment at line 251 explaining that this is the channel-level check after the group-level check above.

- **[Positive]** The timestamp refresh in the `else` branch (`clientChannelInfoFound != null` case) preserves the original behavior for duplicate registrations without triggering the "new producer" log or fast-channel attribute addition.

- **[Positive]** Test coverage includes a latch-controlled concurrent scanning test that deterministically reproduces the race on the old code. Full broker suite (754 tests) passes.

### Suggestions

1. The PR description mentions this is a regression of #8846/#8847 after #9293. Consider adding a `// Regression: see #8846, #9293` comment near the `putIfAbsent` block to help future bisectors.

2. The `clientChannelInfoFound = null` initialization at line 209 is now always assigned before use (either in the `!newChannel` block or implicitly via `newChannel = true`). This is good defensive coding.

### Verdict
Clean, well-scoped fix for a real regression. The atomic publication pattern is correct and the test coverage is solid.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Refreshed this PR onto the current `develop` (`e348efa66`) and force-pushed one signed-off commit (`cd6454ea4`).

- `ProducerManagerTest`: 9/9 passed, including the latch-controlled scanner/registration race;
- all 10 modules in the targeted `broker -am` reactor succeeded;
- Checkstyle and SpotBugs reported no findings;
- `git diff --check` passed.

The production change is +19/-12 lines; +82/-3 lines are regression tests. Full CI has been retriggered.

@xdkxlk @lizhanhui, could you please take a human review when convenient, particularly of the atomic publication and listener/index consistency in `ProducerManager.registerProducer`?

PR #10625, opened 2026-07-18
[ISSUE #10114] Support deleting consumer offset by group and topic

btlqql:
## Problem

A consumer group can accidentally consume a topic, but the current administration API only removes all offsets with the subscription group. Operators cannot clean up one group-topic relationship without affecting the group's progress on unrelated topics.

Closes #10114.

## Root cause

Consumer offsets are keyed by topic and group internally, but the broker protocol and mqadmin surface expose only group-wide removal. The precise storage operation was not wired through the broker, client, admin API, and CLI layers.

## Changes

- add a dedicated request code and ACL-aware request header for deleting one group-topic offset
- remove the exact key from normal, server-side reset, and pull offset tables, while using the existing persistent offset deletion path
- add broker-side blank resource validation and an idempotent success response
- expose the operation through MQClientAPIImpl, MQAdminExt, and DefaultMQAdminExt
- add the mqadmin deleteConsumerOffset command with required group/topic arguments and a mutually exclusive broker-or-cluster target
- handle a null broker response explicitly instead of relying on Java assertions
- cover exact-key isolation, broker validation/delegation, client success/failure/null responses, CLI parsing, and broker/cluster execution

## Impact

This adds an opt-in administration operation and does not change normal offset commits. It deletes only the selected topic@group offset and leaves the consumer group's offsets for other topics intact. Running consumers can commit the offset again, so operators should stop the affected consumer before cleanup when persistent removal is required.

## Validation

- mvn -pl broker,client,tools -am -DskipTests -Dspotbugs.skip=true compile
- mvn -pl broker,client,tools -am -DskipTests -Dspotbugs.skip=true test-compile
- ConsumerOffsetManagerTest#removeOffsetByGroupAndTopicTest
- AdminBrokerProcessorTest#testDeleteConsumerOffset and #testDeleteConsumerOffsetRejectsBlankResource
- MQClientAPIImplTest#testDeleteConsumerOffset, #testDeleteConsumerOffsetFail, and #testDeleteConsumerOffsetNullResponse
- DeleteConsumerOffsetCommandTest (8 tests, including direct broker and cluster execution)
- git diff --check

## Prior work

#10115 previously explored the same issue but was closed without merge. This PR reimplements the feature on current develop and adds explicit target validation, null-response handling, isolation checks, and broker/cluster command coverage.

PR #10701, opened 2026-07-30
[ISSUE #10700] Make consumer offset commits concurrency-safe

### LINKED ISSUE #10700: [Bug] Concurrent first offset commits can overwrite queues in consumer offset managers
### Affected baseline

`develop` at `00e45b8a6db23efbe756d0306f10716156cfd4dd`.

### Problem

Concurrent first offset commits for the same `topic@group` and different queue IDs can overwrite each other in both the classic JSON `ConsumerOffsetManager` and the RocksDB v1 manager.

Both implementations currently follow this pattern:

1. Read `offsetTable.get(topic@group)` and receive `null`.
2. Each thread creates its own inner queue-offset map.
3. Each writes one queue to its private map.
4. Each unconditionally publishes with `offsetTable.put(key, map)`.

The last outer write replaces the other map, so one queue disappears from memory and subsequent persistence.

RocksDB v1 has a second race in incremental mode. It serializes the entire inner queue map as one RocksDB value. A thread can serialize an older `{q0}` snapshot, pause, let another thread write `{q0,q1}`, then write the older batch last. Memory still contains both queues, but a restart reloads only `q0`.

### Expected behavior

- Concurrent first commits for different queue IDs retain every queue in memory and after persistence/reload.
- Classic and RocksDB v1 atomically initialize the shared inner map with outer `putIfAbsent`.
- RocksDB v1 preserves its LMQ `ConcurrentHashMap<>(1, 1.0F)` initialization.
- Incremental persistence orders the update, version change, whole-map serialization, and WAL write for the same `topic@group` while allowing different keys to proceed concurrently.
- RocksDB v2 remains unchanged: it already uses outer `putIfAbsent` and persists each queue under an independent RocksDB key.

### Root cause

The classic and v1 managers use check-then-act initialization rather than atomically publishing one shared map. In v1 incremental mode, whole-map read/modify/serialize/write cycles for the same RocksDB key have no shared critical section.

### Deterministic test plan

- Use a barrier-backed outer map whose first two `get(key)` calls both return their saved `null`; commit two queue IDs concurrently and verify classic JSON encode/decode retains both.
- Repeat against RocksDB v1 periodic persistence and verify `persist -> stop -> clear -> load` retains both queues.
- In incremental mode, pause the first `batchPutWithWal` after its old snapshot has been serialized. Let the second commit either write ahead on current code or block on the fixed per-key monitor, then reload and verify both queues remain.

### Related work and scope

Historical unmerged [PR #1427](https://github.com/apache/rocketmq/pull/1427) precisely identified and proposed `putIfAbsent` for the classic-manager first-commit race. It did not cover RocksDB v1, incremental whole-map WAL ordering, or deterministic concurrency/persistence tests, so this issue extends that valid prior analysis rather than claiming the classic root cause is new.

Current open PRs [#10625](https://github.com/apache/rocketmq/pull/10625), [#9602](https://github.com/apache/rocketmq/pull/9602), and [#9877](https://github.com/apache/rocketmq/pull/9877) touch related manager/test files but do not change these commit paths or solve either race. This fix will stay limited to classic initialization, RocksDB v1 initialization/incremental ordering, and their regression tests; v2 and offset-removal behavior are out of scope.

I am working on a focused fix and will submit a PR against `develop`.

ai-yang:

- Fixes #10700

### Brief Description

Make first offset-map initialization atomic in the classic and RocksDB v1 consumer offset managers, and serialize incremental whole-map persistence for the same `topic@group`.

- Publish one shared inner queue-offset map with outer `putIfAbsent`, then update the selected winner map.
- Preserve the RocksDB v1 LMQ-specific map construction.
- In incremental mode, order the queue update, version update, whole-map serialization, and WAL batch write under the inner map monitor.
- Keep non-incremental commits lock-free across the existing concurrent map and allow different `topic@group` keys to proceed independently.
- Leave RocksDB v2 unchanged because it already initializes atomically and persists each queue under an independent key.

### Root Cause

Classic and RocksDB v1 used check-then-act initialization followed by unconditional outer `put`, so two first commits could publish different inner maps and overwrite one queue. RocksDB v1 incremental mode additionally allowed an older serialized whole-map batch to be written after a newer batch for the same RocksDB key.

Historical unmerged PR #1427 identified the classic initialization race. This PR preserves that analysis while extending coverage to RocksDB v1, incremental WAL ordering, and deterministic persistence/reload tests. PRs #10625, #9602, and #9877 touch related files but do not change these commit paths.

### Related PR coordination

PR #10992 independently addresses the classic-manager subset of #10700 using `computeIfAbsent` and a multi-queue concurrency test. This PR also covers RocksDB v1 initialization, LMQ map construction, and same-key incremental WAL ordering, including restart verification. Its classic-manager regression controls the conflicting first reads with latches and verifies a JSON round trip.

The author of #10992 has offered to consolidate in favor of this PR. Maintainers can review the broader fix here; if the narrower classic fix is preferred first, the remaining RocksDB v1 changes still need review and this PR can be adjusted accordingly. Neither implementation is merged as of 2026-09-11.

### Impact

Concurrent first commits retain every queue in memory and after persistence. Incremental restart recovery cannot regress to an older same-key snapshot. Existing rollback warnings, version semantics, LMQ allocation, and v2 behavior are preserved.

- Deterministic red tests on the previous implementation:
  - classic concurrent first commits: expected 2 queues, reloaded 1;
  - RocksDB v1 periodic LMQ commits: expected 2 queues, retained 1;
  - RocksDB v1 incremental delayed old batch: expected 2 queues after restart, reloaded 1.
- `ConsumerOffsetManagerTest` + `RocksDBConsumerOffsetManagerTest`: 21 tests passed.
- Full `broker` test suite: 755 tests passed, 0 failures, 0 errors, 4 skipped.
- Maven Checkstyle: 0 violations.
- SpotBugs: 0 findings.
- `git diff --check`.
- Current PR head `03695c8703aa26d52f7d2b999935a5a8513c8224`: all 10 GitHub checks passed (checked 2026-09-11). This evidence is for the existing head, not a new run against current `develop`.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/config/v1/RocksDBConsumerOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/config/v1/RocksDBConsumerOffsetManager.java
@@ -192,26 +192,44 @@ public class RocksDBConsumerOffsetManager extends ConsumerOffsetManager {
         ConcurrentMap<Integer, Long> map = this.offsetTable.get(key);
         if (null == map) {
             map = MixAll.isLmq(topic) ? new ConcurrentHashMap<>(1, 1.0F) : new ConcurrentHashMap<>();
-            map.put(queueId, offset);
-            this.offsetTable.put(key, map);
-        } else {
-            Long storeOffset = map.put(queueId, offset);
-            if (storeOffset != null && offset < storeOffset) {
-                LOG.warn("[NOTIFYME]update consumer offset less than store. clientHost={}, key={}, queueId={}, requestOffset={}, storeOffset={}", clientHost, key, queueId, offset, storeOffset);
+            ConcurrentMap<Integer, Long> previous = this.offsetTable.putIfAbsent(key, map);
+            if (null != previous) {
+                map = previous;
             }
         }
-        if (versionChangeCounter.incrementAndGet() % brokerController.getBrokerConfig().getConsumerOffsetUpdateVersionStep() == 0) {
-            updateDataVersion();
-        }
+
         if (!brokerController.getBrokerConfig().isPersistConsumerOffsetIncrementally()) {
+            updateOffset(clientHost, key, queueId, offset, map);
+            updateDataVersionIfNeeded();
             return;
         }
 
-        try (WriteBatch writeBatch = new WriteBatch()) {
-            putWriteBatch(writeBatch, key, map);
-            this.rocksDBConfigManager.batchPutWithWal(writeBatch);
-        } catch (Exception e) {
-            log.error("consumer offset persist Failed", e);
+        // Keep the in-memory update, whole-map snapshot, and WAL write ordered for the same RocksDB key.
+        // Otherwise, an older concurrent snapshot can be persisted after a newer one and overwrite offsets.
+        synchronized (map) {
+            updateOffset(clientHost, key, queueId, offset, map);
+            updateDataVersionIfNeeded();
+            try (WriteBatch writeBatch = new WriteBatch()) {
+                putWriteBatch(writeBatch, key, map);
+                this.rocksDBConfigManager.batchPutWithWal(writeBatch);
+            } catch (Exception e) {
+                log.error("consumer offset persist Failed", e);
+            }
+        }
+    }
+
+    private void updateOffset(String clientHost, String key, int queueId, long offset,
+        ConcurrentMap<Integer, Long> map) {
+        Long storeOffset = map.put(queueId, offset);
+        if (storeOffset != null && offset < storeOffset) {
+            LOG.warn("[NOTIFYME]update consumer offset less than store. clientHost={}, key={}, queueId={}, requestOffset={}, storeOffset={}", clientHost, key, queueId, offset, storeOffset);
+        }
+    }
+
+    private void updateDataVersionIfNeeded() {
+        if (versionChangeCounter.incrementAndGet()
+            % brokerController.getBrokerConfig().getConsumerOffsetUpdateVersionStep() == 0) {
+            updateDataVersion();
         }
     }
 
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
@@ -206,13 +206,14 @@ public class ConsumerOffsetManager extends ConfigManager {
         ConcurrentMap<Integer, Long> map = this.offsetTable.get(key);
         if (null == map) {
             map = new ConcurrentHashMap<>(2);
-            map.put(queueId, offset);
-            this.offsetTable.put(key, map);
-        } else {
-            Long storeOffset = map.put(queueId, offset);
-            if (storeOffset != null && offset < storeOffset) {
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (commented):

### Summary
This PR fixes a real concurrency race in both `ConsumerOffsetManager` and `RocksDBConsumerOffsetManager` where concurrent first-offset commits for the same `topic@group` could overwrite each other's inner maps, losing queue offsets. The fix uses `putIfAbsent` for atomic map initialization and `synchronized(map)` for incremental WAL persistence ordering.

### Findings

- **[Positive]** `RocksDBConsumerOffsetManager.java:195-201` — The `putIfAbsent` + winner-takes-all pattern correctly eliminates the check-then-act race. The original `get()` → `put()` sequence could indeed lose queues under concurrent first commits.

- **[Positive]** `RocksDBConsumerOffsetManager.java:207-216` — The `synchronized(map)` block correctly serializes incremental persistence for the same key, preventing an older WAL batch from being written after a newer one. This is the right granularity — per-key, not global.

- **[Positive]** Non-incremental path remains lock-free across different `topic@group` keys, preserving throughput for the common case.

- **[Info]** `RocksDBConsumerOffsetManager.java:203-206` — In the non-incremental path, `updateOffset()` is called without synchronization. Two concurrent commits for the same key could both call `map.put()` and both evaluate the "less than store" warning. Since `ConcurrentHashMap.put()` is atomic, this is functionally safe, but the warning log might be slightly misleading under contention. This matches the original behavior and is not a regression.

- **[Info]** `ConsumerOffsetManager.java` — The classic manager gets the same `putIfAbsent` treatment. Good that both paths are fixed consistently.

- **[Positive]** Test coverage is thorough — deterministic red tests that demonstrate the race on the old code, plus full broker suite validation (755 tests). The reference to historical unmerged PR #1427 shows good due diligence.

### Suggestions

1. Consider adding a brief comment on the `synchronized(map)` block explaining *why* we synchronize on the inner map (WAL ordering for same RocksDB key), so future maintainers don't accidentally remove it.

2. The `updateDataVersionIfNeeded()` helper is a nice extraction. One minor note: in the non-incremental path it's called outside any synchronization, while in the incremental path it's inside `synchronized(map)`. This is correct for the current logic but worth a one-line comment explaining the difference.

### Verdict
Solid concurrency fix with good test coverage. The approach is sound and the scope is well-contained.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Refreshed this PR onto the current `develop` (`e348efa66`) and force-pushed one signed-off commit (`03695c870`). I also added an inline explanation for the per-key `synchronized(map)` block: it orders the in-memory update, whole-map snapshot, and WAL write so an older snapshot cannot overwrite newer offsets.

- `ConsumerOffsetManagerTest` + `RocksDBConsumerOffsetManagerTest`: 21/21 passed;
- all 10 modules in the targeted `broker -am` reactor succeeded;
- Checkstyle and SpotBugs reported no findings;
- `git diff --check` passed.

The production change is +37/-20 lines; +261/-6 lines are deterministic concurrency/persistence tests. Full CI has been retriggered.

@RongtongJin @lizhimins, could you please take a human review when convenient, especially of same-key WAL ordering and the classic/RocksDB v1 initialization paths?

RockteMQ-AI (commented):
### Summary
This PR modifies 4 files (442 lines).

Key changes reviewed. Please see inline comments for specific suggestions.

---

RongtongJin (approved):

PR #10714, opened 2026-07-31
[ISSUE #10713] Optimize ConsumerFilterManager register CPU consumption

### LINKED ISSUE #10713: [Enhancement] Optimize ConsumerFilterManager register performance

### Summary

The current ConsumerFilterManager uses a topic-based index (filterDataByTopic) as the sole data structure for managing SQL92 consumer filter data. When register(consumerGroup, subList) is called during heartbeat processing, it invokes getByGroup(consumerGroup) which performs a full scan across all topics and all consumer groups — an O(T × G) operation. In clusters with a large number of topics and SQL92 filter consumers, this becomes a significant CPU bottleneck. Additionally, there are several thread safety issues in the concurrent access patterns.

### Motivation

During production operations in a large-scale cluster, we observed high CPU consumption on the broker caused by ConsumerFilterManager.register(). The root cause is the getByGroup() method, which is called on every consumer heartbeat to detect unsubscribed topics. It iterates through the entire filterDataByTopic map (all topics × all consumer groups per topic) just to collect entries belonging to one consumer group. This results in O(T × G) time complexity per heartbeat.

  For example, with 10,000 topics and 100 SQL92 filter consumer groups, each heartbeat triggers a scan of up to 1,000,000 entries. Under high consumer count scenarios, this becomes a serious CPU hotspot.

  Beyond the performance issue, the existing code also has:
  1. TOCTOU race conditions: Multiple containsKey() + get() sequences on ConcurrentHashMap where the entry can be removed between the two calls, leading to NullPointerException.
  2. Unconditional BloomFilter generation: bloomFilter.generate() is called on every register even when enableCalcFilterBitMap is disabled, wasting CPU cycles.
  3. No topic existence validation: Filter data is registered for topics that may not exist on the broker, accumulating stale entries.

Introduce a consumerGroup-based index (SubscriptionFilterHandler) alongside the existing topic-based index to enable O(1) lookups by consumer group:

  1. Add subscriptionFilterData map (ConcurrentMap<String/*consumerGroup*/, SubscriptionFilterHandler>): Each SubscriptionFilterHandler holds a ConcurrentHashMap<String/*topic*/,ConsumerFilterData> for that consumer group. This eliminates the need for full-scan getByGroup() in the hot path.
  2. Dual-index consistency: Both filterDataByTopic (for topic-based queries during message dispatch) and subscriptionFilterData (for consumerGroup-based operations during heartbeat) are maintained in sync. Writes go through subscriptionFilterData first, then update filterDataByTopic.
  3. Conditional BloomFilter generation: Only call bloomFilter.generate() when BrokerConfig.isEnableCalcFilterBitMap() is true, consistent with CommitLogDispatcherCalcBitMap.
  4. Topic existence check: Validate that the topic exists in TopicConfigManager before registering filter data, preventing accumulation of stale entries.
  5. Fix TOCTOU race conditions: Replace all containsKey() + get() sequences with a single get() assigned to a local variable, then null-check.
  6. Thread safety for SubscriptionFilterHandler: Use ConcurrentHashMap instead of HashMap for topicSqlFilterData since it is accessed concurrently.
  7. Rebuild on decode: After deserializing filterDataByTopic from disk, rebuild subscriptionFilterData to ensure consistency after broker restart.

Describe Alternatives You've Considered

  1. Cache getByGroup() results: Maintain a separate cache of consumerGroup → topics mapping that is invalidated on register/unregister. This would reduce the scan frequency but adds cache invalidation complexity and still requires the full scan to rebuild the cache. The dual-index approach is simpler and provides O(1) at all times.
  2. Replace filterDataByTopic entirely with a consumerGroup-based structure: This would simplify the code but would degrade the topic-based lookup path used by CommitLogDispatcherCalcBitMap during message dispatch (called on every message), which is performance-critical.
  3. Add synchronization instead of using ConcurrentHashMap: Using synchronized blocks would fix thread safety but introduce lock contention on the heartbeat path. 
  ConcurrentHashMap provides better concurrent throughput with no contention on reads.

### Additional Context

_No response_

ymwneu:
- Refactor filter data index from topic-based to consumerGroup-based (SubscriptionFilterHandler)
- Add topic existence check before registering filter
- Generate BloomFilterData only when enableCalcFilterBitMap is enabled
- Fix thread safety: use ConcurrentHashMap for topicSqlFilterData
- Fix TOCTOU race conditions: replace containsKey+get with single get
- Rebuild subscriptionFilterData from filterDataByTopic in decode
- Add test for subscription shrink marking removed topics as dead

- Fixes #10713

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/filter/ConsumerFilterManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/filter/ConsumerFilterManager.java
@@ -22,11 +22,13 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.BrokerPathConfigHelper;
 import org.apache.rocketmq.common.ConfigManager;
+import org.apache.rocketmq.common.TopicConfig;
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.common.filter.ExpressionType;
 import org.apache.rocketmq.filter.FilterFactory;
@@ -48,6 +50,9 @@ public class ConsumerFilterManager extends ConfigManager {
 
     private ConcurrentMap<String/*Topic*/, FilterDataMapByTopic>
         filterDataByTopic = new ConcurrentHashMap<>(256);
+
+    private final transient ConcurrentMap<String/*ConsumerID*/, SubscriptionFilterHandler>
+            subscriptionFilterData = new ConcurrentHashMap<>(256);
 
     private transient BrokerController brokerController;
     private transient BloomFilter bloomFilter;
@@ -113,23 +118,21 @@ public class ConsumerFilterManager extends ConfigManager {
         }
 
         // make illegal topic dead.
-        Collection<ConsumerFilterData> groupFilterData = getByGroup(consumerGroup);
+        Set<String> curSubList = new HashSet<>();
+        for (SubscriptionData subscriptionData : subList) {
+            curSubList.add(subscriptionData.getTopic());
+        }
 
-        Iterator<ConsumerFilterData> iterator = groupFilterData.iterator();
-        while (iterator.hasNext()) {
-            ConsumerFilterData filterData = iterator.next();
-
-            boolean exist = false;
-            for (SubscriptionData subscriptionData : subList) {
-                if (subscriptionData.getTopic().equals(filterData.getTopic())) {
-                    exist = true;
-                    break;
+        SubscriptionFilterHandler subscriptionFilterHandler = this.subscriptionFilterData.get(consumerGroup);
+        if (null != subscriptionFilterHandler) {
+            for (Map.Entry<String, ConsumerFilterData> entry : subscriptionFilterHandler.getTopicSqlFilterData().entrySet()) {
+                if (!curSubList.contains(entry.getKey())) {
+                    ConsumerFilterData filterData = entry.getValue();
+                    if (filterData != null) {
+                        filterData.setDeadTime(System.currentTimeMillis());
+                        log.info("Consumer filter changed: {}, make illegal topic dead:{}", consumerGroup, filterData);
+                    }
                 }
-            }
-
-            if (!exist && !filterData.isDead()) {
-                filterData.setDeadTime(System.currentTimeMillis());
-                log.info("Consumer filter changed: {}, make illegal topic dead:{}", consumerGroup, filterData);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):

### Summary
Refactor ConsumerFilterManager to use consumerGroup-based indexing instead of topic-based, improving CPU efficiency during register operations and fixing thread safety issues.

### Findings
- **[Info]** ConsumerFilterManager.java:54 — New `subscriptionFilterData` map uses ConcurrentHashMap correctly for thread safety.
- **[Info]** ConsumerFilterManager.java:118-136 — Replaced O(n*m) iteration with O(n) HashSet lookup for detecting illegal topics. Good performance improvement.
- **[Info]** ConsumerFilterManager.java — TOCTOU race condition fix (replacing containsKey+get with single get) is a good defensive improvement.
- **[Warning]** ConsumerFilterManager.java:54 — The `subscriptionFilterData` map is populated in `register()` and rebuilt in `decode()`, but there's no cleanup mechanism when consumer groups are removed. If consumer groups are dynamically created and destroyed, this could lead to memory accumulation over time. Consider adding a cleanup path in `unregister()` or a periodic cleanup task.
- **[Info]** ConsumerFilterManagerTest.java — Test coverage for subscription shrink marking removed topics as dead is good.

### Suggestions
- Consider adding a method to clean up `subscriptionFilterData` when consumer groups are no longer active, or document the expected lifecycle.
- The `SubscriptionFilterHandler` class could benefit from a brief Javadoc explaining its role in the new indexing scheme.

### Overall
Solid optimization with meaningful thread safety improvements. The memory lifecycle concern is worth addressing but not blocking.

---

RockteMQ-AI (commented):

### Summary
Refactors ConsumerFilterManager from topic-based indexing to consumerGroup-based indexing via a new `SubscriptionFilterHandler` inner class. This improves CPU efficiency during register operations (O(n) HashSet lookup vs O(n*m) nested iteration) and adds thread safety with ConcurrentHashMap. Also adds topic existence validation and conditional BloomFilter generation.

### Findings

- **[Warning]** `ConsumerFilterManager.java:167-170` — Potential race condition: After `subscriptionFilterHandler.register()` returns, the code calls `this.filterDataByTopic.putIfAbsent(topic, ...)` followed by `this.filterDataByTopic.get(topic).put(consumerFilterData)`. Between these two calls, another thread running `clean()` could remove the topic from `filterDataByTopic`, causing the `get()` to return null and the subsequent `put()` to throw NPE.
  - **Suggestion**: Capture the result of `putIfAbsent` and use it directly, or use `computeIfAbsent`:
    ```java
    FilterDataMapByTopic mapByTopic = this.filterDataByTopic.computeIfAbsent(topic, FilterDataMapByTopic::new);
    mapByTopic.put(consumerFilterData);
    ```

- **[Warning]** `ConsumerFilterManager.java:323-345` (clean method) — Cleanup order mismatch: `clean()` removes from `subscriptionFilterData` first, then from `filterDataByTopic`. Meanwhile `register()` adds to `subscriptionFilterData` first, then to `filterDataByTopic`. This ordering asymmetry could cause temporary inconsistency where `filterDataByTopic` has entries not present in `subscriptionFilterData`.
  - **Suggestion**: Consider using a single atomic operation or ensuring both maps are updated in the same order in all code paths.

- **[Info]** `ConsumerFilterManager.java:54` — New `subscriptionFilterData` map uses ConcurrentHashMap correctly for thread-safe consumerGroup-based indexing.

- **[Info]** `ConsumerFilterManager.java:152-156` — Topic existence check before registration is a good defensive measure. Prevents orphaned filter data for non-existent topics.

- **[Info]** `ConsumerFilterManager.java:158-160` — Conditional BloomFilter generation (only when `enableCalcFilterBitMap` is true) is a valid optimization. The test correctly handles null BloomFilterData.

- **[Info]** `ConsumerFilterManager.java:297-309` (decode) — Rebuild of `subscriptionFilterData` from `filterDataByTopic` during deserialization is correct and maintains backward compatibility with persisted data.

- **[Info]** `ConsumerFilterManagerTest.java:274-310` — New test `testRegister_bySubscriptionData_shrinkMakesDead` correctly validates the shrink scenario. Good coverage.

### Suggestions
1. Fix the potential NPE in `register()` by using `computeIfAbsent` or capturing the `putIfAbsent` result.
2. Document the expected ordering of map updates in both `register()` and `clean()` to prevent future inconsistencies.
3. Consider adding a test for concurrent register/cleanup scenarios to validate thread safety.

### Verdict
Good refactoring with meaningful performance improvements. The race condition in `register()` should be addressed before merge.

---

lizhimins (approved):

RockteMQ-AI (approved):
### Summary

---

PR #10729, opened 2026-07-31
fix: harden POP message validation and diagnostics

### LINKED ISSUE #10728: Avoid logging raw MessageExt bodies in proxy ConsumerProcessor
## Problem

`ConsumerProcessor.filterPopResult` logs `MessageExt` directly in two error paths:

```java
log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msg:{}", requestHeader, messageExt);
log.error("process filterMessage failed. requestHeader:{}, msg:{}", requestHeader, messageExt, t);
```

`MessageExt.toString()` delegates to `Message.toString()`, which prints `Arrays.toString(body)`. These logs can expose message payload bytes when POP filtering or handle creation fails.

## Scope

Track 2 / Proxy runtime diagnostics hardening. This should only change log content, not POP filtering behavior.

## Expected behavior

- Do not log raw `MessageExt` objects in `ConsumerProcessor` error paths.
- Keep useful diagnostics such as topic, msgId, queueId, queueOffset, body size, and property keys.
- Add targeted test coverage for the message summary helper.

## Evidence

- `proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java` logs `messageExt` directly.
- `common/src/main/java/org/apache/rocketmq/common/message/Message.java#toString` includes `Arrays.toString(body)`.

### LINKED ISSUE #10784: Proxy should not return POP messages without receipt handles
## Problem

`ConsumerProcessor.filterPopResult` currently keeps a POP message in the returned result even when the broker response does not contain `PROPERTY_POP_CK` and `createHandle(...)` returns null:

```java
String handleString = createHandle(messageExt.getProperty(MessageConst.PROPERTY_POP_CK), messageExt.getCommitLogOffset());
if (handleString == null) {
    log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msg:{}", requestHeader, messageExt);
    messageExtList.add(messageExt);
    continue;
}
```

A POP client needs the receipt handle to ack, change invisible time, or return the message. Returning a message without a valid handle can make the client receive a message that it cannot complete correctly.

## Expected behavior

When a POP message is missing the receipt handle, Proxy should treat it as an invalid broker response and avoid returning that message to the client.

## Scope

Track 2 / Proxy runtime diagnostics and POP correctness. This is separate from #10728, which only tracks raw `MessageExt` logging redaction and explicitly avoids behavior changes.

## Evidence

- `proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java`
- `createHandle(...)` returns null when `PROPERTY_POP_CK` is missing.
- `filterPopResult(...)` currently logs the bug but still appends the invalid message to the returned list.

Aias00:
## Summary

- Drop POP messages that do not have a valid receipt handle instead of returning an unusable message.
- Keep POP message diagnostics payload-safe through compact message summaries.

## Validation

```bash
mvn -q -pl proxy -am -Dtest=ConsumerProcessorTest -DfailIfNoTests=false test
```

Closes #10728
Closes #10784

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
@@ -190,7 +190,6 @@ public class ConsumerProcessor extends AbstractProcessor {
                     if (handleString == null) {
                         log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msgSummary:{}",
                             requestHeader, summarizeMessageExt(messageExt));
-                        messageExtList.add(messageExt);
                         continue;
                     }
                     MessageAccessor.putProperty(messageExt, MessageConst.PROPERTY_POP_CK, handleString);
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutorService;
@@ -187,7 +188,8 @@ public class ConsumerProcessor extends AbstractProcessor {
                     fillUniqIDIfNeed(messageExt);
                     String handleString = createHandle(messageExt.getProperty(MessageConst.PROPERTY_POP_CK), messageExt.getCommitLogOffset());
                     if (handleString == null) {
-                        log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msg:{}", requestHeader, messageExt);
+                        log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msgSummary:{}",
+                            requestHeader, summarizeMessageExt(messageExt));
                         messageExtList.add(messageExt);
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):

### Summary
Replaces raw `MessageExt` toString() logging in `ConsumerProcessor.filterPopResult` error paths with a compact summary, preventing body/property data leakage in logs.

### Findings
- **[Info]** Both replaced log sites (handle-null error and filterMessage failure) correctly use the summary helper instead of the raw object.
- **[Info]** The `summarizeMessageExt` implementation is identical to the one in PR #10731 (`LocalMessageService`).

### Suggestions
- **Code duplication:** The same `summarizeMessageExt` method appears in both `ConsumerProcessor` (this PR) and `LocalMessageService` (PR #10731). Consider extracting it into a shared utility class to keep a single source of truth. This would also apply to any future similar PRs in the series.
- Test coverage is solid — verifies that body bytes and property values are excluded while key metadata is preserved.

---

RockteMQ-AI (approved):

### Summary
Replaces raw `MessageExt.toString()` in ConsumerProcessor error/warn logs with a `summarizeMessageExt()` summary to avoid dumping full message content.

### Findings
- **[Info]** ConsumerProcessor.java:191,238 — Two log sites updated to use summary instead of raw object. Both are error paths where full message dumps would be noisy.
- **[Info]** ConsumerProcessor.java:248-259 — `summarizeMessageExt()` implementation is identical to the one in PR #10731 (LocalMessageService). Consider extracting to a shared utility.
- **[Info]** Test coverage is present.

### Cross-repo Note
Same `summarizeMessageExt()` as PR #10731. After both merge, a shared utility in `common` or `proxy` module would reduce maintenance burden.

---

RockteMQ-AI (approved):
### Summary

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits — POP consumer validation now uses structured logging instead of raw MessageExt.toString().

### Findings
- **[Info]** Replaced raw `messageExt.toString()` in error logs with `summarizeMessageExt` — produces concise summaries (topic, bornHost, storeSize, keys) without leaking message body or property values.
- **[Info]** `safelySummarizeMessageExt` wraps in try-catch to prevent logging failures from breaking the consumption path.
- **[Info]** Test validates summary format and null-safety.

LGTM.

---

PR #10735, opened 2026-08-01
[ISSUE #10734] Fix broadcast offset initialization race

### LINKED ISSUE #10734: [Bug] Broadcast offset initialization can overwrite a concurrent client offset update

Linux x86_64. The reproducer is a deterministic broker unit test and does not require a running RocketMQ cluster.

### RocketMQ version

Branch: `develop`

Commit: `00e45b8a6db23efbe756d0306f10716156cfd4dd`

### JDK Version

OpenJDK 8

### Describe the Bug

`BroadcastOffsetManager.queryInitOffset` initializes a missing client entry with a non-atomic `get` / create / `put` sequence. `updateOffset` initializes and updates the same entry atomically with `ConcurrentHashMap.compute`.

If an update wins after the query observes a missing entry but before the query performs its `put`, the query replaces the updated `BroadcastTimedOffsetStore` with a new empty instance. The just-committed client offset is lost, and the query falls back to the consumer offset or queue maximum.

### Steps to Reproduce

1. Create an existing `BroadcastOffsetData` with no entry for a client.
2. Start `queryInitOffset` for that client and block it after it observes the missing entry but before its map insertion.
3. Call `updateOffset` for the same client and queue with offset `100`; this inserts a store and records `100`.
4. Resume the query insertion.
5. Read the query result and the stored client offset.

A latch-controlled `ConcurrentHashMap` test reproduces this ordering without sleeps or random stress. On the unmodified commit above, it failed 5/5 times with:

```text
expected:<100> but was:<10>
```

Here `10` is the mocked queue maximum used after the updated store has been overwritten.

The winning client store and its offset `100` should remain in `clientOffsetStore`. A concurrent initialization must not overwrite an offset already recorded by `updateOffset`.

### What Did You See Instead?

The query's later `put` replaces the store created by `updateOffset`. Both the initial query and a subsequent query observe the fallback offset `10`, so the client update to `100` is lost.

### Additional Context

The race is scoped to the per-client initialization in `BroadcastOffsetManager`; no public protocol change is required. I am willing to submit a focused fix with the deterministic regression test.

ai-yang:

- Fixes #10734

### Brief Description

`BroadcastOffsetManager.queryInitOffset` initialized `clientOffsetStore` with a non-atomic `get` / create / `put` sequence. If `updateOffset` inserted and updated the same client's store between the `get` and `put`, the query thread replaced that store and discarded the newly committed offset.

Use `computeIfAbsent` so initialization and concurrent updates retain the same winning `BroadcastTimedOffsetStore` instance. The change is internal and does not alter the public protocol.

- Added a deterministic latch-controlled concurrency regression test with no sleeps or random scheduling.
- Reverted to the previous implementation: the regression test failed 5/5 times (`expected 100 but was 10`).
- Applied this fix: the regression test passed 20/20 times.
- Ran the complete affected JDK 8 reactor: all modules passed; broker ran 753 tests with 0 failures, 0 errors, and 4 skipped tests.
- Ran Maven `validate`: Checkstyle reported 0 violations across the 10 affected modules.
- Ran broker SpotBugs: `BugInstance=0`, `Error=0`.
- Ran `git diff --check` successfully.

### Concurrency and Test Design

`computeIfAbsent` only makes publication of the per-client store atomic; it does not change the existing update semantics once a store exists. The regression uses latches to force the exact missing-key initialization/update interleaving, with no sleeps or random scheduling: the previous get/create/put sequence loses the updated offset, while the atomic initialization retains it.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManager.java
@@ -85,11 +85,8 @@ public class BroadcastOffsetManager extends ServiceThread {
         }
 
         final AtomicLong offset = new AtomicLong(-1L);
-        BroadcastTimedOffsetStore offsetStore = broadcastOffsetData.clientOffsetStore.get(clientId);
-        if (offsetStore == null) {
-            offsetStore = new BroadcastTimedOffsetStore(fromProxy);
-            broadcastOffsetData.clientOffsetStore.put(clientId, offsetStore);
-        }
+        BroadcastTimedOffsetStore offsetStore = broadcastOffsetData.clientOffsetStore.computeIfAbsent(
+            clientId, key -> new BroadcastTimedOffsetStore(fromProxy));
 
         if (offsetStore.fromProxy && requestOffset < 0) {
             // when from proxy and requestOffset is -1
--- a/broker/src/test/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManagerConcurrencyTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/offset/BroadcastOffsetManagerConcurrencyTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.rocketmq.broker.offset;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Fixes a TOCTOU race condition in `BroadcastOffsetManager.queryInitOffset` where a non-atomic get/create/put sequence could discard a concurrently committed broadcast offset.

### Findings
- **[Info]** `BroadcastOffsetManager.java:88-89` — Replacing the manual `get`/`if null`/`put` with `computeIfAbsent` is the correct and idiomatic fix. `ConcurrentHashMap.computeIfAbsent` guarantees atomic initialization, preventing the race where `updateOffset` inserts a store between the `get` and `put`.
- **[Info]** `BroadcastOffsetManagerConcurrencyTest.java` — The `BlockingInitOffsetStore` design is excellent. By overriding both `put` and `computeIfAbsent` with `CountDownLatch`-controlled blocking, the test deterministically reproduces the race without relying on sleeps or random scheduling. The test verifies:
  1. With the old code: the race causes offset loss (test fails 5/5)
  2. With the fix: the offset is correctly preserved (test passes 20/20)

### Suggestions
No issues found. The fix is a single-line change that eliminates the race, and the test is a model for deterministic concurrency testing.

- **Correctness**: `computeIfAbsent` is the standard pattern for atomic get-or-create. No edge cases missed.
- **Performance**: `computeIfAbsent` is actually more efficient than the previous get+put (one atomic operation vs two).
- **Compatibility**: Internal change only, no public API or protocol change.

LGTM.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
@lizhimins @xdkxlk, could you please take a human review when convenient? This is a focused BroadcastOffsetManager atomic-initialization race fix with a deterministic latch-controlled regression. The complete affected reactor, Checkstyle, and SpotBugs pass. The remaining GitHub Actions runs are currently awaiting maintainer approval.

ai-yang:
Refreshed this PR against the latest `develop` (`e348efa66b08eb645ee123706ea6492fa9a3ad35`, RocketMQ 5.5.1).

- refreshed head: `dce0bac09` (one signed-off commit);
- deterministic `BroadcastOffsetManagerConcurrencyTest`: 1/1 passed;
- all 10 modules in the targeted `broker -am` reactor completed successfully, including Checkstyle and SpotBugs;
- `git diff --check` passed.

The force-push has retriggered the full CI matrix.

PR #10740, opened 2026-08-01
[ISSUE #10739] Complete proxy futures when processor executors reject tasks

### LINKED ISSUE #10739: [Bug] Proxy processor futures can remain pending when completion tasks are rejected
## RocketMQ version

`develop` at `00e45b8a6db23efbe756d0306f10716156cfd4dd`

## Describe the bug

`FutureUtils.appendNextFuture` ignores the dependent stage returned by
`whenCompleteAsync`. If the supplied executor rejects the completion task, that
dependent stage completes exceptionally but the manually created `nextFuture`
is never completed.

The production executors used by `ProducerProcessor` and `ConsumerProcessor`
add a second failure mode. They are created through the default
`ThreadPoolMonitor.createAndMonitor` overload, whose rejection handler is
`DiscardOldestPolicy`. After shutdown it silently discards the completion task;
under saturation it can silently discard an older queued completion task.
Neither case gives `FutureUtils` a rejection that it can propagate, so a proxy
request future can remain pending indefinitely.

## Deterministic reproduction

Two synchronous regression tests reproduce the complete failure chain on JDK 8.

### 1. Rejection is not propagated by `FutureUtils`

1. Create a single-thread executor and shut it down.
2. Call `FutureUtils.addExecutor(CompletableFuture.completedFuture("value"), executor)`.
3. Inspect the returned future.

Expected: the returned future is completed exceptionally with the scheduling
failure.

Actual: the stage returned by `whenCompleteAsync` is exceptional, but the stage
is ignored and the returned `nextFuture` remains pending.

The controlled test
`FutureUtilsTest#testAddExecutorCompletesExceptionallyWhenExecutorRejectsTask`
failed at the expected `result.isDone()` assertion in 5/5 runs.

### 2. The actual Proxy processor executor silently drops the task

1. Construct `DefaultMessagingProcessor` with its real producer and consumer
   processor executors.
2. Shut down the producer processor executor.
3. Pass that executor to `FutureUtils.addExecutor` with an already completed
   source future.
4. Inspect the returned future.

Expected: rejection is observable and the returned future is completed
exceptionally.

Actual: `DiscardOldestPolicy` silently discards the task after shutdown and the
returned future remains pending.

The controlled test
`DefaultMessagingProcessorTest#testProcessorFutureCompletesWhenExecutorIsShutDown`
failed at the expected `result.isDone()` assertion in 5/5 runs. It uses the
actual processor pool, no sleeps, randomized scheduling, network access, or
external service.

## Impact

`FutureUtils.addExecutor` is used by the Proxy `ProducerProcessor` and
`ConsumerProcessor` request paths. During processor-pool shutdown, or when the
bounded pool rejects completion work, callers may receive a future that never
reaches a terminal state. This can leave gRPC/remoting requests waiting until an
outer timeout or connection teardown instead of receiving an immediate failure.

## Suggested fix

Both parts are needed; either one alone leaves a pending-future path:

- Retain the stage returned by `whenCompleteAsync` and propagate a scheduling
  failure into `nextFuture`.
- Create the producer and consumer processor executors with an explicit
  throwing rejection policy (for example `AbortPolicy`) instead of the default
  silent `DiscardOldestPolicy`, so `FutureUtils` can observe both shutdown and
  overload rejection.
- Add regression coverage for both a throwing executor and the actual Proxy
  processor executor after shutdown.

This changes only internal completion/rejection behavior and does not change a
public protocol.

## Related work checked

- #8900 contains a `FutureUtils` stack frame but reports client cancellation and
  message redelivery, not executor rejection or a pending returned future.
- #9253 concerns receipt-handle renewal after client disconnection, not
  completion-task rejection.
- #5575 introduced the Proxy remoting implementation and is not a fix for this
  behavior.
- #10711 concerns unsupported metadata futures and does not modify
  `FutureUtils` or the processor executor policies.

Searches across open and closed issues and pull requests for `FutureUtils`,
`addExecutor`, `DefaultMessagingProcessor`, `DiscardOldestPolicy`, executor
shutdown/rejection, and pending Proxy futures found no equivalent report or
claimed fix.

ai-yang:

- Fixes #10739

### Brief Description

`FutureUtils.appendNextFuture` now retains the stage returned by `whenCompleteAsync` and propagates completion-task scheduling failures to the future returned to the caller.

The Proxy producer and consumer processor executors now use an explicit `AbortPolicy`. This is required because the shared default `DiscardOldestPolicy` can silently discard completion tasks during shutdown or saturation, leaving no rejection for `FutureUtils` to propagate. Other RocketMQ thread pools keep their existing rejection behavior.

Together, these changes ensure that Proxy request futures reach an exceptional terminal state instead of remaining pending when processor completion work is rejected.

- On the unmodified JDK 8 baseline, the helper-level and real producer-pool shutdown regressions each failed at the expected pending-future assertion in 5/5 runs.
- Added a deterministic real-pool saturation regression using one worker and one queue slot. Latches occupy the worker and fill the queue without sleeps or randomized scheduling.
- Temporarily restoring the old `DiscardOldestPolicy` made the saturation regression fail 1/1 at `the returned future must not remain pending`; restoring `AbortPolicy` made it pass.
- On the refreshed head, `FutureUtilsTest` passed 1/1 and `DefaultMessagingProcessorTest` passed 6/6. All 11 modules in the targeted `proxy -am` reactor succeeded with Checkstyle and SpotBugs enabled.
- The earlier full JDK 8 `proxy -am` test reactor also passed: Broker 752 tests (0 failures, 0 errors, 4 skipped) and Proxy 304 tests (0 failures, 0 errors, 3 skipped).
- `git diff --check` passed.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/FutureUtils.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/FutureUtils.java
@@ -24,13 +24,18 @@ public class FutureUtils {
 
     public static <T> CompletableFuture<T> appendNextFuture(CompletableFuture<T> future,
         CompletableFuture<T> nextFuture, ExecutorService executor) {
-        future.whenCompleteAsync((t, throwable) -> {
+        CompletableFuture<T> completionFuture = future.whenCompleteAsync((t, throwable) -> {
             if (throwable != null) {
                 nextFuture.completeExceptionally(throwable);
             } else {
                 nextFuture.complete(t);
             }
         }, executor);
+        completionFuture.whenComplete((ignored, throwable) -> {
+            if (throwable != null) {
+                nextFuture.completeExceptionally(throwable);
+            }
+        });
         return nextFuture;
     }
 
--- a/common/src/test/java/org/apache/rocketmq/common/utils/FutureUtilsTest.java
+++ b/common/src/test/java/org/apache/rocketmq/common/utils/FutureUtilsTest.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.rocketmq.common.utils;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+public class FutureUtilsTest {
+
+    @Test
+    public void testAddExecutorCompletesExceptionallyWhenExecutorRejectsTask() {
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        executor.shutdown();
+
+        CompletableFuture<String> result = FutureUtils.addExecutor(
+            CompletableFuture.completedFuture("value"), executor);
+
+        assertTrue("the returned future must not remain pending", result.isDone());
+        assertTrue(result.isCompletedExceptionally());
+        CompletionException exception = assertThrows(CompletionException.class, () -> result.join());
+        assertTrue(exception.getCause() instanceof RejectedExecutionException);
+    }
+}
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (commented):
### Findings

- **[Info]** `FutureUtils.java:27-34` — The fix is correct. `completionFuture.whenComplete(...)` runs synchronously in the completing thread (not submitted to the executor), so it reliably catches `RejectedExecutionException` without needing another executor. Good design.

- **[Info]** `DefaultMessagingProcessor.java:89,98` — Switching from the default `DiscardOldestPolicy` to `AbortPolicy` is the right call here. Under high load or shutdown, tasks that were previously silently discarded will now surface as `RejectedExecutionException`, which `FutureUtils` can properly propagate. This is a behavioral change, but it converts silent data loss into explicit failure — the correct tradeoff for a messaging system.

- **[Info]** `FutureUtilsTest.java` — Test coverage is solid. The shut-down executor pattern reliably reproduces the rejection scenario, and asserting `CompletionException` wrapping `RejectedExecutionException` validates the full propagation chain.

- **[Info]** `DefaultMessagingProcessorTest.java` — Good integration-level test that verifies the real executor chain propagates rejections through the future.

### Suggestions

- **Minor:** Consider whether other thread pools in the Proxy module that also use `appendNextFuture` (or similar future-chaining patterns) might need the same `AbortPolicy` treatment. A quick audit of `ThreadPoolMonitor.createAndMonitor` call sites in the proxy module could reveal similar latent issues.

### Verdict
Well-structured bug fix that addresses both the symptom (pending futures) and root cause (silent task discarding). The two-level approach ensures robustness even if one layer is misconfigured.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Refreshed onto current `develop` (`e348efa66`) and force-pushed one signed-off commit (`2d561e1e0`). The missing saturation evidence is now covered by a deterministic one-worker/one-slot test: old `DiscardOldestPolicy` fails with a pending future, while `AbortPolicy` completes it exceptionally. Affected tests pass 7/7 and the targeted 11-module reactor, Checkstyle, SpotBugs, and `git diff --check` all pass.

@lollipopjin @lizhimins, could you please take a human review when convenient, especially of the executor rejection-policy change and its overload semantics?

RockteMQ-AI (commented):
### Summary
This PR modifies 4 files (185 lines).

Key changes reviewed. Please see inline comments for specific suggestions.

---

PR #10742, opened 2026-08-01
[ISSUE #10671] Improve unsupported client type diagnostics

### LINKED ISSUE #10671: [Feature] Proxy should provide clearer diagnostics when rejecting unsupported gRPC client types

When using the gRPC Java SDK, I attempted to create a LitePushConsumer.

The current Proxy implementation does not support LitePushConsumer, so the server closes the stream during initialization.

However, the client only receives the following exception:

`java.lang.IllegalStateException: Stream is already completed, no further calls are allowed`

This exception does not indicate the real cause of the failure. There is also no corresponding Proxy log explaining why the connection was terminated.

When the Proxy rejects or terminates a client connection because the requested consumer type is unsupported, it should:

Log the actual reason on the Proxy side.
Return a more descriptive gRPC status/message if possible (for example, indicating that LitePushConsumer is not currently supported).

This would make troubleshooting much easier than relying on generic client-side exceptions.

Currently the only way to identify the root cause is to inspect the Proxy implementation or debug the server, since neither the client exception nor the server logs explain why the connection was closed.

### Additional Context

_No response_

twei43846-afk:
### Summary

- Return a clearer `UNRECOGNIZED_CLIENT_TYPE` status message when Proxy gRPC heartbeat rejects an unsupported client type.
- Return the same diagnostic from client termination when cached settings contain an unsupported client type.
- Add focused `ClientActivityTest` coverage for both paths.

### Motivation

Fixes #10671.

The previous response only echoed the enum name, for example `CLIENT_TYPE_UNSPECIFIED`, which does not explain why the Proxy rejected the client or which client types are accepted. This keeps the behavior compatible while making the rejection reason visible to clients and adding a Proxy-side warning log.

### Tests

- `git diff --check`
- `mvn -Dmaven.repo.local=$env:USERPROFILE\.m2\repository -pl proxy -am -Dtest=ClientActivityTest -DfailIfNoTests=false test`

Result: `ClientActivityTest` passed with 17 tests. The local JDK17 run emitted existing JaCoCo 0.8.5 instrumentation warnings for class file version 61, but Maven exited successfully and surefire reports 0 failures / 0 errors.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
@@ -124,8 +124,10 @@ public class ClientActivity extends AbstractMessagingActivity {
                     break;
                 }
                 default: {
+                    log.warn("heartbeat rejected unsupported client type. clientId:{}, clientType:{}",
+                        ctx.getClientID(), clientSettings.getClientType());
                     future.complete(HeartbeatResponse.newBuilder()
-                        .setStatus(ResponseBuilder.getInstance().buildStatus(Code.UNRECOGNIZED_CLIENT_TYPE, clientSettings.getClientType().name()))
+                        .setStatus(buildUnsupportedClientTypeStatus(clientSettings.getClientType()))
                         .build());
                     return future;
                 }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Improves diagnostics when the Proxy gRPC layer rejects an unsupported client type in heartbeat and client termination requests. Adds a helper method that produces a descriptive error message listing supported types, and covers both paths with unit tests.

### Findings

- **[Info]** `ClientActivity.java:127-128` — Adding `log.warn` for unsupported client types in both `heartbeat` and `notifyClientTermination` is valuable for operational debugging. Good addition.
- **[Info]** `ClientActivity.java:202-205` — The `buildUnsupportedClientTypeStatus` helper reduces duplication and produces a much more informative error message. The message clearly tells the client what went wrong and which types are supported.
- **[Info]** `ClientActivityTest.java` — Tests cover both the heartbeat and termination paths with `CLIENT_TYPE_UNSPECIFIED`, verifying both the error code and the message content. Good coverage.

### Minor Observations
- The hardcoded list of supported types in the error message (`PRODUCER, PUSH_CONSUMER, SIMPLE_CONSUMER, LITE_PUSH_CONSUMER, LITE_SIMPLE_CONSUMER`) could become stale if new client types are added to the enum. Consider deriving it from the `ClientType` enum or a constant in the future. This is a minor concern and acceptable for now.

### Assessment
Clean improvement that significantly improves debuggability for clients connecting with unsupported types. Well-tested.

---

RockteMQ-AI (approved):
### Summary

---

PR #10743, opened 2026-08-01
fix: guard NPE in NetworkUtil and concurrent read in IndexService

btlqql:
### Motivation

Three edge-case bugs found by code review (verified against the current develop branch):

1. **`NetworkUtil.socketAddress2String` NPE** (`common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java`)
   `InetSocketAddress.getAddress()` returns `null` when the hostname could not be resolved, so `getAddress().getHostAddress()` threw `NullPointerException`. This util is used across broker/namesrv/tools. Fall back to `getHostString()` when the address is null.

2. **`NetworkUtil.string2SocketAddress` StringIndexOutOfBoundsException**
   An address without `:` makes `lastIndexOf(":")` return `-1`, and `substring(0, -1)` throws `StringIndexOutOfBoundsException`. Throw a clear `IllegalArgumentException` instead.

3. **`IndexService.getTotalSize` concurrent read** (`store/src/main/java/org/apache/rocketmq/store/index/IndexService.java`)
   `indexFileList` is a plain `ArrayList` guarded by `readWriteLock` in every other accessor, but `getTotalSize` read it without the lock. A concurrent `destroy()`/`deleteExpiredFile()` could clear the list between the `isEmpty()` check and `get(0)`, causing `IndexOutOfBoundsException`. Acquire the read lock.

### Verification

`mvn -pl broker -am compile` passes on the build server.

### Diff

2 files changed, +15 / -5.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/NetworkUtil.java
@@ -193,6 +193,9 @@ public class NetworkUtil {
 
     public static SocketAddress string2SocketAddress(final String addr) {
         int split = addr.lastIndexOf(":");
+        if (split < 0) {
+            throw new IllegalArgumentException("Invalid socket address, missing ':': " + addr);
+        }
         String host = addr.substring(0, split);
         String port = addr.substring(split + 1);
         return new InetSocketAddress(host, Integer.parseInt(port));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Guards against NPE in `NetworkUtil.socketAddress2String` when hostname is unresolved, adds input validation in `string2SocketAddress`, and fixes a race condition in `IndexService.getTotalSize()` by acquiring the read lock.

### Findings

- **[Info]** `NetworkUtil.java:195` — The new `IllegalArgumentException` for missing ':' is a good defensive check. Consider including the original address string in the error message (already done — 👍).
- **[Info]** `NetworkUtil.java:207` — The null-guard on `getAddress()` with fallback to `getHostString()` correctly handles unresolved hostnames. This prevents NPE in broker/namesrv paths that use this utility.
- **[Info]** `IndexService.java:95-103` — Wrapping `getTotalSize()` in the read lock correctly prevents the race between `isEmpty()` check and `get(0)` call. The existing `readWriteLock` is the right choice here.

### Assessment
Clean, well-scoped bug fix. All three changes are correct and minimal. No backward compatibility concerns.

---

RockteMQ-AI (approved):
### Summary

---

PR #10751, opened 2026-08-02
[ISSUE #10750] Fix POP lock cleanup race

### LINKED ISSUE #10750: [Bug] POP lock timeout cleanup can remove a reacquired lock
## Affected branch

`develop` at `a06836dd564e5e43115493f775626cf98d51d10e`.

## Problem

`PopConsumerLockService.removeTimeout()` decides that an entry is expired from a previously read `TimedLock.lockTime`, then removes the entry through the map iterator. The expiration check and removal are not atomic with `tryLock()` for the same key.

If the lock is currently free but its old timestamp is expired, the following interleaving is possible:

1. cleanup reads the expired timestamp;
2. another thread successfully reacquires the same `TimedLock` and refreshes its timestamp;
3. cleanup removes the mapping based on its stale decision;
4. a third caller creates a new `TimedLock` for the same key and also acquires it.

The first and third callers then both believe that they hold the POP consumer lock.

## Deterministic reproduction

A unit test installs a `TimedLock` whose `getLockTime()` uses two latches to pause cleanup immediately after it captures the old timestamp. While cleanup is paused, the test reacquires the lock and refreshes its time, then lets cleanup continue. Without unlocking the first holder, a second `tryLock()` incorrectly succeeds.

The unmodified branch failed identically in 5/5 isolated JDK 8 Maven processes:

```text
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
java.lang.AssertionError: Cleanup removed the reacquired lock and allowed a second holder
```

The test uses no sleep, network, random scheduling, or stress loop. All latch waits are bounded only to make a broken test fail instead of hanging.

## Impact

Two broker threads can concurrently enter a code path that is intended to be serialized for the same consumer group and topic after a cleanup/reacquire race. This weakens the mutual-exclusion guarantee used by POP consumption and can allow duplicate concurrent processing of protected work.

## Expected behavior

Once a lock has been reacquired and its lease timestamp refreshed, cleanup must not remove that current mapping based on an older expiration observation. A subsequent caller must continue to observe the lock as held.

## Suggested direction

Serialize acquisition/refresh and the authoritative expiration recheck/removal per map key:

- perform lookup/creation, CAS acquisition, and timestamp refresh inside `ConcurrentMap.compute`;
- keep the outer expiration read only as a fast filter, then use `computeIfPresent` to re-read the current mapped lock and its current timestamp before deciding whether to return `null`.

This preserves the existing behavior that a genuinely expired lease may be removed even if its old holder remains marked as locked. The separate, pre-existing question of an old timed-out holder later calling key-only `unlock()` on a newer lock generation is outside this report.

## Related work checked

Searches covered open and closed issues and pull requests using the class name, `removeTimeout`, `TimedLock`, POP lock timeout cleanup, and reacquisition terms. No equivalent report, implementation, assignee, or maintainer handoff was found. #10519 and #10448 only mention POP locks while changing batch invisibility/delete behavior and do not touch this cleanup race.

ai-yang:

Fixes #10750

### Brief Description

`PopConsumerLockService.removeTimeout()` previously made its expiration decision before removing the map entry. A concurrent `tryLock()` could reacquire the same `TimedLock` and refresh its timestamp after that decision, but cleanup would still remove the refreshed entry and allow a second holder to be created.

This change makes acquisition/refresh and the authoritative cleanup recheck atomic for each key:

- `tryLock()` uses `ConcurrentHashMap.compute()` to select/create and acquire the mapped lock within the per-key remapping boundary.
- `removeTimeout()` uses `computeIfPresent()` to recheck the current lock timestamp before removing it.
- Existing behavior for leases that remain expired at the authoritative recheck is preserved.

- Unmodified `develop`: the deterministic latch regression failed in 5/5 isolated JDK 8 Maven processes; an independent rerun also failed 5/5.
- Fixed targeted test class: 20 isolated JDK 8 Maven processes, 2/2 each (40/40 total).
- Full `broker -am test`: all 10 reactor modules passed; broker ran 753 tests with 0 failures, 0 errors, and 4 skips.
- Checkstyle: 0 violations.
- SpotBugs: 0 bugs/errors.
- `git diff --check`: passed.

### Applicability and performance boundary

The lock service is used by the POP KV consumer service and the Lite POP path. `popConsumerKVServiceEnable` defaults to `false`; this is not a claim that every POP deployment uses the affected lock. The race requires an expired-entry cleanup decision to overlap with successful reacquisition of that key. Expiration semantics for a lease that remains expired are intentionally unchanged.

The patch introduces map remapping on acquisition and a per-attempt result holder. The regression establishes the mutual-exclusion repair, not a throughput improvement or zero overhead. Same-key contention, allocation rate and unaffected traffic should be considered in a representative load test before making performance claims.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerLockService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerLockService.java
@@ -16,16 +16,13 @@
  */
 package org.apache.rocketmq.broker.pop;
 
-import java.util.Iterator;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.rocketmq.common.KeyBuilder;
 import org.apache.rocketmq.common.PopAckConstants;
 import org.apache.rocketmq.common.constant.LoggerName;
-import org.apache.rocketmq.common.utils.ConcurrentHashMapUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,8 +39,13 @@ public class PopConsumerLockService {
     }
 
     public boolean tryLock(String key) {
-        return Objects.requireNonNull(ConcurrentHashMapUtils.computeIfAbsent(lockTable,
-            key, s -> new TimedLock())).tryLock();
+        AtomicBoolean locked = new AtomicBoolean(false);
+        lockTable.compute(key, (k, currentLock) -> {
+            TimedLock lock = currentLock == null ? new TimedLock() : currentLock;
+            locked.set(lock.tryLock());
+            return lock;
+        });
+        return locked.get();
     }
 
     public boolean tryLock(String groupId, String topicId) {
@@ -69,13 +71,22 @@ public class PopConsumerLockService {
     }
 
     public void removeTimeout() {
-        Iterator<Map.Entry<String, TimedLock>> iterator = lockTable.entrySet().iterator();
-        while (iterator.hasNext()) {
-            Map.Entry<String, TimedLock> entry = iterator.next();
-            if (System.currentTimeMillis() - entry.getValue().getLockTime() > timeout) {
+        for (Map.Entry<String, TimedLock> entry : lockTable.entrySet()) {
+            if (System.currentTimeMillis() - entry.getValue().getLockTime() <= timeout) {
+                continue;
+            }
+
+            TimedLock[] removedLock = new TimedLock[1];
+            lockTable.computeIfPresent(entry.getKey(), (key, currentLock) -> {
+                if (System.currentTimeMillis() - currentLock.getLockTime() > timeout) {
+                    removedLock[0] = currentLock;
+                    return null;
+                }
+                return currentLock;
+            });
+            if (removedLock[0] != null) {
                 log.info("PopConsumerLockService remove timeout lock, " +
-                    "key={}, locked={}", entry.getKey(), entry.getValue().lock.get());
-                iterator.remove();
+                    "key={}, locked={}", entry.getKey(), removedLock[0].lock.get());
             }
         }
     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (commented):

### Summary
Fixes a race condition in `PopConsumerLockService` by replacing `ConcurrentHashMapUtils.computeIfAbsent` with atomic `compute()` in `tryLock`, and refactoring `removeTimeout` to use `computeIfPresent` with double-check of timeout condition.

### Findings
- **[Info]** `PopConsumerLockService.java:42-48` — The `compute()` approach ensures lock creation and acquisition are atomic. The `AtomicBoolean` captures the result from within the compute lambda. This fixes the race where a lock could be removed between creation and acquisition.
- **[Info]** `PopConsumerLockService.java:74-89` — The `removeTimeout` refactoring correctly re-checks the timeout inside `computeIfPresent` to handle the case where a lock was re-acquired between the iteration check and the removal attempt.
- **[Warning]** `PopConsumerLockService.java:44` — The `AtomicBoolean locked` is allocated per `tryLock` call. Under high contention with many concurrent lock attempts, this creates short-lived object pressure. Consider whether a simpler return pattern is possible, though correctness is not affected.
- **[Info]** `PopConsumerLockServiceTest.java` — New test `removeTimeoutShouldNotRemoveReacquiredLock` uses `CountDownLatch` to deterministically test the race between cleanup and re-acquisition. Good coverage.

### Suggestions
- The `AtomicBoolean` allocation per call is minor but worth noting for hot paths. An alternative would be to restructure `compute` to return the lock and check its state outside, but the current approach is correct and readable.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Status refresh (2026-08-28):

- Rebased the PR onto the latest upstream `develop` (`e348efa66`). New head: `7cb64c7a8`.
- Rechecked the current upstream implementation: no equivalent fix has landed, and timeout cleanup can still remove a lock after a concurrent reacquisition without this patch, breaking per-key mutual exclusion.
- Re-ran the targeted broker reactor on the rebased head:

  `mvn -Dmaven.repo.local=/tmp/codex-maven-repository -pl broker -am -DskipITs -Dtest=org.apache.rocketmq.broker.pop.PopConsumerLockServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`

  Result: **BUILD SUCCESS**; 2 tests run, 0 failures, 0 errors. Checkstyle and SpotBugs also passed in the reactor.

No additional implementation changes were needed after the rebase. The issue remains actionable and this PR is not superseded by current `develop`.

ai-yang:
All 10 CI checks are green on the refreshed single signed-off commit. @xdkxlk @lizhimins, could you please take a human review when convenient? The key point is whether the `compute`/`computeIfPresent` transitions make POP lock acquisition and timeout cleanup atomic without changing lock ownership semantics.

PR #10757, opened 2026-08-02
[ISSUE #10754] Fix LMQ hold request cleanup race

### LINKED ISSUE #10754: [Bug] LMQ empty-bucket cleanup can drop concurrently suspended pull requests
## Affected branch

`develop` at `a06836dd564e5e43115493f775626cf98d51d10e`.

## Problem

`LmqPullRequestHoldService.checkHoldRequest()` removes an LMQ `ManyPullRequest` bucket after observing it as empty. The empty check and `pullRequestTable.remove(key)` are not atomic with `PullRequestHoldService.suspendPullRequest()` adding a request to the same bucket.

The following interleaving can occur:

1. the periodic check observes the current bucket as empty;
2. a concurrent request obtains that bucket from the map and appends a new suspended pull request;
3. the periodic check unconditionally removes the bucket based on its earlier empty observation.

The new request remains in the detached `ManyPullRequest` object, but the hold table no longer contains that object.

## Deterministic reproduction

A unit test pre-populates an empty bucket and uses a `ConcurrentHashMap` test double whose `remove(key)` is paused by two latches. This pauses the periodic check only after its empty condition has already selected the removal path. The test then calls the real `suspendPullRequest()` method, verifies that the existing bucket now contains the request, lets removal continue, and requires the same bucket to remain reachable from the table.

The unmodified branch failed identically in 5/5 isolated JDK 8 Maven processes:

```text
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
java.lang.AssertionError: concurrently suspended request should remain reachable expected same:<...> was not:<null>
```

The reproduction has no sleep, network, timer, or random scheduling dependency. All latch and future waits are bounded only to turn a broken test into a prompt failure instead of a hang.

## Impact

A concurrently suspended LMQ pull request can disappear from `pullRequestTable`. Later message-arrival and timeout scans cannot discover or wake it, so the broker-side long-poll request remains stranded until external connection or client timeout handling intervenes.

## Expected behavior

An LMQ bucket must not be removed once a concurrent suspend or replay has made it non-empty. Every request that remains suspended must be reachable through `pullRequestTable` for later message-arrival and timeout processing.

## Suggested direction

Make bucket insertion/replay and empty cleanup atomic per map key:

- add newly suspended requests through `ConcurrentMap.compute`, creating or reusing the mapped bucket and appending inside the remapping function;
- remove an LMQ bucket through `computeIfPresent`, using synchronized `ManyPullRequest.isEmpty()` as the authoritative check;
- reinsert `notifyMessageArriving()` replay requests through the same map-aware helper, rather than appending them to a bucket that may have been detached while requests were evaluated.

Changing only the final remove to a conditional remove is insufficient: a concurrent writer or replay can still retain and append to an object that cleanup has just detached from the map.

## Related work checked

Searches covered open and closed issues and pull requests using `LmqPullRequestHoldService`, `ManyPullRequest`, `pullRequestTable`, `suspendPullRequest`, empty-bucket cleanup, and concurrent removal terms. No equivalent report, implementation, assignee, or maintainer handoff was found.

#8341 and its unmerged PR #8342 only proposed replacing the redundant `getPullRequestList() == null` check with `isEmpty()`; they did not make the check/removal atomic or address concurrent suspend/replay.

ai-yang:

Fixes #10754

### Brief Description

LMQ empty-bucket cleanup previously checked `ManyPullRequest.isEmpty()` and then removed the map entry in separate operations. A concurrent suspend or notification replay could append a request to the selected bucket after the empty check, while cleanup still removed that bucket from `pullRequestTable`. The request then remained in a detached object and could no longer be reached by message arrival or timeout scans.

This change keeps bucket selection, append, and empty cleanup atomic per key:

- suspended requests and notification replays use one map-aware `compute()` helper;
- LMQ cleanup uses `computeIfPresent()` and performs the authoritative synchronized `isEmpty()` check inside the remapping boundary;
- ordinary non-LMQ request semantics are preserved, but the shared suspend/replay path also gains a map remapping boundary.

- Unmodified `develop`: the deterministic cleanup-vs-suspend regression failed in 5/5 isolated JDK 8 Maven processes.
- Fixed targeted test class: 3/3 once, then 20 isolated Maven/JVM processes at 3/3 each (60/60 total).
- Full `broker -am test`: all 10 reactor modules passed; 2,369 tests, 9 skips, 0 failures, and 0 errors. Broker ran 755 tests, including the new class at 3/3.
- Checkstyle: 0 violations across all 10 modules.
- SpotBugs: 0 bugs/errors across all code modules.
- `git diff --check`: passed.

### Applicability and performance boundary

`BrokerController` selects `LmqPullRequestHoldService` when `enableLmq` is enabled (the default is `false`). The failure is a held request becoming unreachable from broker wakeup/timeout scans, not loss of its message body from storage; client-side deadlines may still terminate the RPC.

Although empty-bucket removal is LMQ-specific, the patch changes `PullRequestHoldService` used by ordinary traffic too. Correctness test results do not establish unchanged throughput. Review should consider same-key contention on both LMQ and ordinary queues; a follow-up can restrict the synchronization scope if representative measurements show a material cost. No contention benchmark is claimed in this PR.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/longpolling/LmqPullRequestHoldService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/longpolling/LmqPullRequestHoldService.java
@@ -54,10 +54,7 @@ public class LmqPullRequestHoldService extends PullRequestHoldService {
                 LOGGER.error("check hold request failed. topic={}, queueId={}", topic, queueId, e);
             }
             if (MixAll.isLmq(topic)) {
-                ManyPullRequest mpr = pullRequestTable.get(key);
-                if (mpr == null || mpr.getPullRequestList() == null || mpr.getPullRequestList().isEmpty()) {
-                    pullRequestTable.remove(key);
-                }
+                pullRequestTable.computeIfPresent(key, (k, mpr) -> mpr.isEmpty() ? null : mpr);
             }
         }
     }
--- a/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/longpolling/PullRequestHoldService.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.Consumer;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.common.ServiceThread;
 import org.apache.rocketmq.common.SystemClock;
@@ -44,17 +45,16 @@ public class PullRequestHoldService extends ServiceThread {
 
     public void suspendPullRequest(final String topic, final int queueId, final PullRequest pullRequest) {
         String key = this.buildKey(topic, queueId);
-        ManyPullRequest mpr = this.pullRequestTable.get(key);
-        if (null == mpr) {
-            mpr = new ManyPullRequest();
-            ManyPullRequest prev = this.pullRequestTable.putIfAbsent(key, mpr);
-            if (prev != null) {
-                mpr = prev;
-            }
-        }
-
         pullRequest.getRequestCommand().setSuspended(true);
-        mpr.addPullRequest(pullRequest);
+        this.addPullRequest(key, mpr -> mpr.addPullRequest(pullRequest));
+    }
+
+    private void addPullRequest(final String key, final Consumer<ManyPullRequest> addOperation) {
+        this.pullRequestTable.compute(key, (k, current) -> {
+            ManyPullRequest mpr = current == null ? new ManyPullRequest() : current;
+            addOperation.accept(mpr);
+            return mpr;
+        });
     }
 
     private String buildKey(final String topic, final int queueId) {
@@ -177,7 +177,7 @@ public class PullRequestHoldService extends ServiceThread {
                 }
 
                 if (!replayList.isEmpty()) {
-                    mpr.addPullRequest(replayList);
+                    this.addPullRequest(key, current -> current.addPullRequest(replayList));
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (commented):

### Summary
Fixes a race condition in LMQ pull request hold service cleanup and refactors `PullRequestHoldService` to use atomic `ConcurrentHashMap.compute()` operations instead of get-then-put patterns.

### Findings
- **[Info]** `LmqPullRequestHoldService.java:57` — Replacing the unsafe get-null-check-remove pattern with `computeIfPresent` is the correct fix. The original code had a TOCTOU race where a new request could be added between the emptiness check and the removal.
- **[Info]** `PullRequestHoldService.java:48-56` — The `addPullRequest` helper using `compute()` ensures atomic bucket creation and request addition. The `Consumer<ManyPullRequest>` parameter is a clean abstraction.
- **[Info]** `PullRequestHoldService.java:180` — The replay list path now also uses the same `addPullRequest` helper, ensuring consistency.
- **[Info]** `LmqPullRequestHoldServiceTest.java` — New test file with concurrent scenario coverage using `CountDownLatch` for deterministic race testing. Good approach.

### Suggestions
- The `Consumer<ManyPullRequest>` allocation per call is minor but could be replaced with a method reference or lambda cached in a field if GC pressure becomes a concern under high throughput. Not blocking.

---

RockteMQ-AI (approved):
### Summary

---

ai-yang:
Status refresh (2026-08-28):

- Rebased the PR onto the latest upstream `develop` (`e348efa66`). New head: `997639abd`.
- Rechecked the current upstream implementation: no equivalent fix has landed, and the LMQ empty-bucket cleanup can still detach a concurrently appended/replayed pull request without this patch.
- Re-ran the targeted broker reactor on the rebased head:

  `mvn -Dmaven.repo.local=/tmp/codex-maven-repository -pl broker -am -DskipITs -Dtest=org.apache.rocketmq.broker.longpolling.LmqPullRequestHoldServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`

  Result: **BUILD SUCCESS**; 3 tests run, 0 failures, 0 errors. Checkstyle and SpotBugs also passed in the reactor.

No additional implementation changes were needed after the rebase. The issue remains actionable and this PR is not superseded by current `develop`.

ai-yang:
All 10 CI checks are green on the refreshed single signed-off commit. @guyinyou @xdkxlk, could you please take a human review when convenient? The production diff is small and most additions are the deterministic LMQ cleanup race regression; the key point is atomic empty-bucket removal without detaching a concurrent request.

PR #10758, opened 2026-08-02
[ISSUE #10755] Fix ConsumeQueueExt truncation cleanup

### LINKED ISSUE #10755: [Bug] ConsumeQueue truncation leaves orphaned ConsumeQueueExt entries
## RocketMQ version

`develop` at `a06836dd564e5e43115493f775626cf98d51d10e`

## Describe the bug

`ConsumeQueue.truncateDirtyLogicFiles` updates the ConsumeQueue positions but skips truncating the associated `ConsumeQueueExt` during every normal partial-truncation path.

The method tracks the last retained extension address in `maxExtAddr`, and its tail calls:

```java
consumeQueueExt.truncateByMaxAddress(maxExtAddr);
```

However, scanning returns directly when it reaches the first dirty CQ entry, the end of a full valid CQ file, or an unwritten slot. Those returns bypass the extension truncation. The tail is only reached after all CQ files have been deleted or no CQ file exists; in that case `maxExtAddr` remains `1`, which is not an extension address, so `truncateByMaxAddress` is a no-op.

## Deterministic reproduction

1. Create a file-backed `ConsumeQueue` with `enableConsumeQueueExt=true`.
2. Write four CQ entries with physical offsets `0`, `100`, `200`, and `300`, using distinct tag codes so that four CQExt units are created.
3. Record the raw extension addresses stored by CQ entries 2 and 3.
4. Call `truncateDirtyLogicFiles(200)`, retaining entries 0 and 1.
5. Write replacement entries at CQ offsets 2 and 3 with different tag codes.
6. Flush, load a second `ConsumeQueue` from the same directory, and call `recover()`.
7. Read the extension unit at the original address of entry 2.

Expected:

- CQExt is truncated together with CQ.
- Replacement entries reuse the truncated extension addresses.
- After reload, the original address of entry 2 contains the replacement tag code.

Actual:

- The old extension units remain readable after CQ truncation.
- Replacement extension units are appended after the orphaned units.
- Recovery retains the orphaned units because the newly appended CQ entries now reference later extension addresses.

The controlled regression test is named:

```text
ConsumeQueueTest#testTruncateDirtyLogicFilesTruncatesConsumeQueueExtAndSurvivesReload
```

It uses no sleeps, randomized scheduling, or external services. On the latest unmodified `develop` baseline under JDK 8, it failed at the same business assertion in 5/5 isolated Maven/JUnit processes: expected replacement tag code `202`, but the original orphan still returned `102`.

## Impact

After CommitLog/CQ rollback followed by rebuilding, unreachable ConsumeQueueExt units remain between retained and replacement units. This leaks persistent store space and cannot be repaired by a later restart, because recovery truncates only after the newest referenced extension address.

## Suggested fix

- Route every normal scan exit through the CQExt truncation tail instead of returning early.
- Continue deleting wholly dirty tail CQ files until the last retained entry is found.
- Treat an empty first slot as an empty tail file and continue scanning the preceding retained file.
- If a retained CQ entry references CQExt, truncate after its last retained extension address; if no retained CQ entry references CQExt, physically clear the extension files so stale bytes cannot reappear during reload/recovery.
- Make the full-clear path recoverable when an extension mapped file is still referenced: retain failed mappings, block address reuse while cleanup is pending, and allow a later retry after the reader releases the buffer.
- When recovery finds no CQ file, clear any loaded CQExt files so a crash between CQ deletion and extension cleanup self-heals on restart.
- Preserve the non-destructive `deleteFile=false` overload by not physically truncating CQExt while detached CQ files remain on disk.
- Add partial-file, empty/full-tail-file, full-truncation, cleanup-failure/retry, replacement-write, and reload/recovery regression coverage.

## Related work checked

- #2570/#2572 only change when recovery invokes `truncateDirtyLogicFiles`; they do not truncate CQExt.
- #2571/#2573 discuss the physical-offset boundary condition, not extension cleanup.
- #10417/#10418 fixes stale minimum offsets after all CQ files are deleted; it does not reach or repair CQExt tail truncation.
- Open #9728/#9729 handles an empty last mapped file in `correctMinOffset`, a different method and postcondition.
- #6343 concerns deleting the CQExt directory when a topic is deleted.
- #10733 touches a different method in `ConsumeQueue` (`estimateMessageCount`) and does not change truncation.

Immediate pre-publication searches across open and closed issues and open, closed, and merged pull requests for `ConsumeQueueExt`, `truncateDirtyLogicFiles`, `truncateByMaxAddress`, extension truncation, and orphaned extension data found no equivalent report, claimed fix, or assignee.

ai-yang:

Fixes #10755

### Brief Description

`ConsumeQueue.truncateDirtyLogicFiles()` truncated the main consume queue but did not trim `ConsumeQueueExt` to the last retained extension address. Ext units from the discarded CQ tail therefore remained readable, and later appends plus reload/recovery can preserve unreferenced Ext data beyond the truncation boundary; this is not evidence of permanent retention under every cleanup policy.

This change aligns CQExt truncation with the successfully retained CQ state:

- normal truncation exits locate the last readable retained Ext reference after `minLogicOffset` and truncate to it;
- when no Ext reference is retained, all Ext mapped files are removed and the address cursor is reset only after deletion succeeds;
- CQ physical-delete failures and `deleteFile=false` do not mutate CQExt independently;
- recovery handles both an empty CQ and a non-empty CQ whose recent recovery window contains only raw tags, while preserving valid older Ext references.

Of the 753 added lines in this PR, 629 are deterministic regression tests; the production change is limited to `ConsumeQueue` and `ConsumeQueueExt`.

### Cleanup State and Recovery

- Successful full cleanup resets the flush/commit cursors and clears the pending state before Ext addresses can be reused.
- If a mapped file is still held, cleanup remains pending and Ext writes fall back to the original raw tags code. A later Ext write retries cleanup at most once per second, avoiding both unsafe address reuse and per-message warning logs.
- On restart, recovery first validates the recent maximum Ext address. If it is absent or invalid, it scans all retained CQ entries backward; it either truncates after the last valid reference or completes full cleanup when no reference remains.

- Unmodified `develop`: the deterministic truncate/reload regression failed in 5/5 isolated JDK 8 Maven processes.
- Deterministic truncation suite: 20 isolated Maven/JVM processes at 9/9 each (180/180 total).
- Latest-head `ConsumeQueueTest`: 21/21, including pending cleanup after raw-tags fallback, automatic retry after mapped-buffer release, and preservation of an Ext reference before the three-file recovery window.
- Latest-head targeted follow-up rerun: 3/3.
- Latest-head full `store -am test`: common 241/241, remoting 174/174, and store 324 tests with 4 skips, 0 failures, and 0 errors.
- Maven validate and Checkstyle: 0 violations.
- SpotBugs: 0 bugs/errors across all four reactor modules.
- `git diff --check`: passed.

### Applicability and review tradeoffs

This is a CQExt storage-reclamation fix, not a demonstrated loss of message bodies or a throughput improvement. `enableConsumeQueueExt` defaults to `false`; the issue requires retained extension data and a CQ truncation/recovery scenario. The regression shows stale units surviving truncation and reload, not their frequency or total disk cost in production.

The implementation also changes the main CQ truncation control flow. Review should include CQExt-disabled behavior, physical deletion failure, raw-tag fallback, and safe address reuse. The fallback backward search can inspect all retained CQ entries in the worst case; no large-store recovery-time benchmark is claimed. Disk impact depends on replay/truncation volume, mapped-file boundaries, and later retention cleanup, so no universal retained-byte estimate is asserted.

A focused storage/recovery review is appropriate before merging. If maintainers prefer separating the main-CQ control-flow changes from Ext cleanup, that split should preserve the existing deletion-failure and restart regressions.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
@@ -136,7 +136,12 @@ public class ConsumeQueue implements ConsumeQueueInterface {
     @Override
     public void recover() {
         final List<MappedFile> mappedFiles = this.mappedFileQueue.getMappedFiles();
-        if (!mappedFiles.isEmpty()) {
+        if (mappedFiles.isEmpty()) {
+            if (isExtReadEnable()) {
+                this.consumeQueueExt.recover();
+                truncateConsumeQueueExt(1);
+            }
+        } else {
 
             int index = mappedFiles.size() - 3;
             if (index < 0) {
@@ -196,8 +201,7 @@ public class ConsumeQueue implements ConsumeQueueInterface {
 
             if (isExtReadEnable()) {
                 this.consumeQueueExt.recover();
-                log.info("Truncate consume queue extend file by max {}", maxExtAddr);
-                this.consumeQueueExt.truncateByMaxAddress(maxExtAddr);
+                truncateConsumeQueueExt(maxExtAddr);
             }
         }
     }
@@ -423,10 +427,11 @@ public class ConsumeQueue implements ConsumeQueueInterface {
 
         this.setMaxPhysicOffset(phyOffset);
         long maxExtAddr = 1;
-        boolean shouldDeleteFile = false;
+        boolean cqFileDeletionFailed = false;
         while (true) {
             MappedFile mappedFile = this.mappedFileQueue.getLastMappedFile();
             if (mappedFile != null) {
+                boolean shouldDeleteFile = false;
                 ByteBuffer byteBuffer = mappedFile.sliceByteBuffer();
 
                 mappedFile.setWrotePosition(0);
@@ -438,53 +443,42 @@ public class ConsumeQueue implements ConsumeQueueInterface {
                     int size = byteBuffer.getInt();
                     long tagsCode = byteBuffer.getLong();
 
-                    if (0 == i) {
-                        if (offset >= phyOffset) {
+                    if (offset < 0 || size <= 0 || offset >= phyOffset) {
+                        if (0 == i) {
                             shouldDeleteFile = true;
-                            break;
-                        } else {
-                            int pos = i + CQ_STORE_UNIT_SIZE;
-                            mappedFile.setWrotePosition(pos);
-                            mappedFile.setCommittedPosition(pos);
-                            mappedFile.setFlushedPosition(pos);
-                            this.setMaxPhysicOffset(offset + size);
-                            // This maybe not take effect, when not every consume queue has extend file.
-                            if (isExtAddr(tagsCode)) {
-                                maxExtAddr = tagsCode;
-                            }
                         }
-                    } else {
+                        break;
+                    }
 
-                        if (offset >= 0 && size > 0) {
+                    int pos = i + CQ_STORE_UNIT_SIZE;
+                    mappedFile.setWrotePosition(pos);
+                    mappedFile.setCommittedPosition(pos);
+                    mappedFile.setFlushedPosition(pos);
+                    this.setMaxPhysicOffset(offset + size);
+                    // This maybe not take effect, when not every consume queue has extend file.
+                    long logicOffset = mappedFile.getFileFromOffset() + i;
+                    if (logicOffset >= this.minLogicOffset && isExtAddr(tagsCode)) {
+                        maxExtAddr = tagsCode;
+                    }
 
-                            if (offset >= phyOffset) {
-                                return;
-                            }
-
-                            int pos = i + CQ_STORE_UNIT_SIZE;
-                            mappedFile.setWrotePosition(pos);
-                            mappedFile.setCommittedPosition(pos);
-                            mappedFile.setFlushedPosition(pos);
-                            this.setMaxPhysicOffset(offset + size);
-                            if (isExtAddr(tagsCode)) {
-                                maxExtAddr = tagsCode;
-                            }
-
-                            if (pos == logicFileSize) {
-                                return;
-                            }
-                        } else {
-                            return;
-                        }
+                    if (pos == logicFileSize) {
+                        break;
                     }
                 }
 
                 if (shouldDeleteFile) {
                     if (deleteFile) {
+                        String mappedFilePath = mappedFile.getFileName();
                         this.mappedFileQueue.deleteLastMappedFile();
+                        if (new File(mappedFilePath).exists()) {
+                            cqFileDeletionFailed = true;
+                            log.warn("Consume queue file still exists after deletion: {}", mappedFilePath);
+                        }
                     } else {
                         this.mappedFileQueue.deleteExpiredFile(Collections.singletonList(this.mappedFileQueue.getLastMappedFile()));
                     }
+                } else {
+                    break;
                 }
 
             } else {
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (commented):

### Summary
Fixes ConsumeQueueExt data orphaning during truncation and recovery by refactoring `truncateDirtyLogicFiles`, adding `truncateAll()` to ConsumeQueueExt, and handling the empty-mapped-files recovery path.

### Findings
- **[Warning]** `ConsumeQueueExt.java:232` — The `truncateAllPending` volatile flag is set to `true` during `truncateAll()` but is never reset to `false`. After recovery truncation completes, all subsequent `put()` calls will be silently skipped with a warning log. If the ConsumeQueue is expected to accept new ext writes after recovery, this flag should be reset once truncation finishes.
- **[Info]** `ConsumeQueue.java:448-492` — The refactored truncation loop is significantly cleaner. The `hasRetainedExt` / `cqFileDeletionFailed` tracking properly handles edge cases where file deletion fails mid-truncation.
- **[Info]** `ConsumeQueue.java:136-147` — Good catch on the empty mapped files recovery path — previously this case would skip ext cleanup entirely.
- **[Info]** `findLastRetainedExtAddress()` — The backward scan is correct but could be expensive for large queues. This is only called during truncation so the impact should be limited.

### Suggestions
- **Important**: Verify whether `truncateAllPending` should be reset after `truncateAll()` completes. If the ConsumeQueue continues to serve writes post-recovery, the current behavior would silently drop all ext data.
- Consider logging at INFO level (not WARN) when `truncateAllPending` blocks a write during active truncation, since this is expected transient behavior.

---

ai-yang:
Clarification on the automated review warning: `truncateAllPending` is reset to `false` after all Ext mapped files are deleted and the flush/commit positions are reset in `ConsumeQueueExt.truncateAll()`. It intentionally remains `true` only when deletion is incomplete, preventing address reuse until a later retry succeeds. The held-buffer regression covers block → release/retry → address reuse → reload.

RockteMQ-AI:
Thanks for the clarification, @ai-yang. You're right — I can see that `truncateAllPending` is correctly reset to `false` at the end of the successful path in `truncateAll()` (after `setFlushedWhere(0)` and `setCommittedWhere(0)`). The early return on the failure path (files not fully deleted) intentionally keeps it `true` to gate new `put()` calls until a retry succeeds. My original warning was incorrect on this point; the design is sound. 👍

RockteMQ-AI (approved):
### Summary

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits — ConsumeQueueExt truncation now handles the empty main CQ edge case and refactors truncation into a shared method.

### Findings
- **[Info]** `recover()` now correctly calls `consumeQueueExt.recover()` and `truncateConsumeQueueExt(1)` when the main CQ has no mapped files — previously this path was skipped entirely, leaving ext files in an inconsistent state.
- **[Info]** `truncateConsumeQueueExt(long)` extracted from inline logic, reused by both `recover()` and `truncateDirtyLogicFiles()`.
- **[Info]** `truncateDirtyLogicFiles` refactored: `shouldDeleteFile` scoped to the inner loop, `cqFileDeletionFailed` tracks outer state — clearer intent.
- **[Info]** Test covers the empty CQ + ext recovery scenario.

LGTM.

---

ai-yang:
Update after refreshing this PR against the latest `develop`:

- rebased onto `e348efa66b08eb645ee123706ea6492fa9a3ad35` (RocketMQ 5.5.1);
- squashed the two PR commits into one signed-off commit (`00e1c9767`);
- reran the complete `ConsumeQueueTest`: 21/21 passed;
- Checkstyle, SpotBugs, and `git diff --check` passed in the affected reactor build.

The full CI matrix has been retriggered by the force-push.

@RongtongJin @guyinyou, could one of you please take a human review when available, particularly for the CQ/CQExt truncation and recovery invariants? Thank you.

ai-yang:
All 10 CI checks are green on the refreshed single signed-off commit. @RongtongJin @guyinyou, could you please take a human review when convenient? Review-size guide: production is +124/-39 lines, while +629 lines are recovery/truncation regressions. The key point is keeping main ConsumeQueue and ConsumeQueueExt truncation boundaries consistent when the main CQ becomes empty.

PR #10769, opened 2026-08-03
[ISSUE #10768] Continue queryRoute after missing broker data

### LINKED ISSUE #10768: [Bug] Proxy gRPC queryRoute stops processing queues after a missing broker entry
### Problem

`RouteActivity#queryRoute` stops iterating all queue data when one queue row has no matching broker entry in the built broker map.

Current code:

```java
if (brokerIdMap == null) {
    break;
}
```

Because this uses `break`, a single malformed or stale `QueueData` entry can truncate the whole route response and hide later valid queues.

### Expected behavior

A missing broker entry should only skip that malformed queue row. The proxy should continue processing remaining queue rows that still have valid broker mappings.

### Impact

This can produce partial or empty gRPC `QueryRoute` responses when route metadata contains a stale queue row before valid rows. That affects Proxy/gRPC route discovery and can make clients miss usable queues.

### Suggested fix

Change the missing broker-map path from `break` to `continue` and add a regression test with:

- one queue row whose broker name is absent from `brokerMap`
- one later queue row with a valid broker mapping
- assertion that the valid later queue still appears in the query route response

Aias00:
### What changed

This PR makes Proxy gRPC `RouteActivity#queryRoute` skip only the malformed queue row when its broker name is absent from the built broker map.

Previously the loop used `break`, so one stale/malformed `QueueData` entry could stop processing later valid queue rows and truncate the route response.

### Why

A missing broker mapping should not hide valid queue routes that appear later in the same route data. This can affect gRPC route discovery when route metadata contains stale queue data.

Fixes #10768.

### Verification

```bash
mvn -pl proxy -Dtest=RouteActivityTest test
```

Result: BUILD SUCCESS. `RouteActivityTest` ran 8 tests with 0 failures/errors/skips. Checkstyle and SpotBugs also passed in the Maven run.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
@@ -78,7 +78,7 @@ public class RouteActivity extends AbstractMessagingActivity {
                 String brokerName = queueData.getBrokerName();
                 Map<Long, Broker> brokerIdMap = brokerMap.get(brokerName);
                 if (brokerIdMap == null) {
-                    break;
+                    continue;
                 }
                 for (Broker broker : brokerIdMap.values()) {
                     messageQueueList.addAll(this.genMessageQueueFromQueueData(queueData, request.getTopic(), topicMessageType, broker));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
One-line fix: changes  to  in  when a  entry references a broker name absent from the built broker map. Previously, a single stale/malformed queue row would truncate the entire route response by breaking out of the loop.

### Findings
- **[Correctness]**  — The fix is correct. A missing broker mapping should not prevent processing of subsequent valid  entries. The previous  behavior could silently drop valid routes from the response.
- **[Tests]**  — Good test coverage.  verifies that a missing broker in the middle of the queue data list does not prevent later valid entries from being included in the response.

### Verdict
Clean, minimal fix with appropriate test coverage. LGTM.

---

RockteMQ-AI (approved):
### Summary

---

Aias00:
Consolidated into #10771; both changes harden RouteActivity behavior for incomplete Broker data.

PR #10771, opened 2026-08-03
fix: harden gRPC route and assignment Broker validation

### LINKED ISSUE #10768: [Bug] Proxy gRPC queryRoute stops processing queues after a missing broker entry

### LINKED ISSUE #10770: [Bug] Proxy gRPC queryAssignment can set a null master broker
### Problem

`RouteActivity#queryAssignment` assumes every broker map contains the master broker id:

```java
Broker broker = brokerIdMap.get(MixAll.MASTER_ID);
...
.setBroker(broker)
```

If route metadata only has non-master broker entries, `broker` is null and the protobuf builder can fail while building the assignment.

### Expected behavior

A queue row without a master broker should be skipped. If no assignment can be built after skipping invalid rows, the existing `FORBIDDEN` / `no readable queue` response should be returned instead of failing the request exceptionally.

### Impact

Malformed or partial route metadata can make Proxy gRPC `QueryAssignment` fail instead of returning a controlled response. This affects the Proxy/gRPC client assignment path.

### Suggested fix

- Add a null check after `brokerIdMap.get(MixAll.MASTER_ID)`.
- Continue to the next queue row when the master broker is absent.
- Add regression coverage for a broker map that contains only a non-master broker id.

Aias00:
## Summary

- Skip invalid route entries that lack Broker data while retaining valid queues during gRPC route lookup.
- Reject assignment responses safely when no master Broker is available.

## Validation

```bash
mvn -q -pl proxy -am -Dtest=RouteActivityTest -DfailIfNoTests=false test
```

Closes #10770
Closes #10768

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
@@ -78,7 +78,7 @@ public class RouteActivity extends AbstractMessagingActivity {
                 String brokerName = queueData.getBrokerName();
                 Map<Long, Broker> brokerIdMap = brokerMap.get(brokerName);
                 if (brokerIdMap == null) {
-                    break;
+                    continue;
                 }
                 for (Broker broker : brokerIdMap.values()) {
                     messageQueueList.addAll(this.genMessageQueueFromQueueData(queueData, request.getTopic(), topicMessageType, broker));
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/route/RouteActivity.java
@@ -125,6 +125,9 @@ public class RouteActivity extends AbstractMessagingActivity {
                     Map<Long, Broker> brokerIdMap = brokerMap.get(queueData.getBrokerName());
                     if (brokerIdMap != null) {
                         Broker broker = brokerIdMap.get(MixAll.MASTER_ID);
+                        if (broker == null) {
+                            continue;
+                        }
                         Permission permission = this.convertToPermission(queueData.getPerm());
                         if (isFifo && !isLite) {
                             for (int i = 0; i < queueData.getReadQueueNums(); i++) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Adds null check for `broker` after `brokerIdMap.get(MixAll.MASTER_ID)` in `DefaultConsumerAssignor.queryAssignment()`, preventing `NullPointerException` when the master broker is missing from the route result.

### Findings
- **[Info]** The fix is correct and minimal. Returning a FORBIDDEN result when the master broker is unavailable is the right behavior — the client should not proceed with assignment from an incomplete route.
- **[Info]** Test coverage verifies that a missing master broker results in FORBIDDEN with no assignments.

### Suggestions
- None. Clean, targeted fix.

LGTM.

---

RockteMQ-AI (approved):
### Summary

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits — route and assignment logic now correctly handles missing brokers instead of silently skipping remaining queues.

### Findings
- **[Info]** `break` → `continue` fix in `queryRoute`: previously a missing broker for one queue would abort the entire loop, silently dropping all subsequent queues. Now continues to the next queue.
- **[Info]** Null check for master broker in `queryAssignment`: prevents NPE when broker metadata is incomplete.
- **[Info]** Tests cover both scenarios with multiple brokers where one is missing.

LGTM.

---

PR #10785, opened 2026-08-03
[ISSUE #10784] Drop POP messages without receipt handles

### LINKED ISSUE #10784: Proxy should not return POP messages without receipt handles

Aias00:
### What is changed
- Avoid returning POP messages when the broker response does not contain a receipt handle.
- Keep the existing invalid-response diagnostic log but skip the message before result filtering and client delivery.
- Add `ConsumerProcessorTest` coverage to verify missing-handle messages are dropped and not passed to the POP result filter.

Fixes #10784

### Verification
- mvn -pl proxy -Dtest=ConsumerProcessorTest test

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ConsumerProcessor.java
@@ -188,7 +188,6 @@ public class ConsumerProcessor extends AbstractProcessor {
                     String handleString = createHandle(messageExt.getProperty(MessageConst.PROPERTY_POP_CK), messageExt.getCommitLogOffset());
                     if (handleString == null) {
                         log.error("[BUG] pop message from broker but handle is empty. requestHeader:{}, msg:{}", requestHeader, messageExt);
-                        messageExtList.add(messageExt);
                         continue;
                     }
                     MessageAccessor.putProperty(messageExt, MessageConst.PROPERTY_POP_CK, handleString);
--- a/proxy/src/test/java/org/apache/rocketmq/proxy/processor/ConsumerProcessorTest.java
+++ b/proxy/src/test/java/org/apache/rocketmq/proxy/processor/ConsumerProcessorTest.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.rocketmq.client.consumer.AckResult;
 import org.apache.rocketmq.client.consumer.AckStatus;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):

### Summary
Fixes a bug in `ConsumerProcessor.processRequest()` where POP messages without receipt handles were still being added to the result list despite being logged as `[BUG]`. Now such messages are properly dropped.

### Findings
- **[Info]** `ConsumerProcessor.java:218` — Removing `messageExtList.add(messageExt)` from the null-handle branch is the correct fix. The original code logged a bug but still delivered the message, which could cause downstream issues for consumers.
- **[Info]** Test `testProcessRequest_NullReceiptHandle_DropsMessage` correctly verifies that messages without handles are excluded from the response.
- **[Info]** The fix is minimal and focused — no unnecessary changes.

### Suggestions
- None. Clean, targeted bug fix.

LGTM — correct fix for a clear bug where invalid messages were being delivered to consumers.

---

RockteMQ-AI (approved):
LGTM. Code changes look good.

---

RockteMQ-AI (approved):
### Summary

---

Aias00:
Rebased onto the latest develop and verified the review request. The regression test uses get(5, TimeUnit.SECONDS), so a stalled pop future fails deterministically rather than hanging CI.

Verified with:

mvn -q -pl proxy -am -Dtest=ConsumerProcessorTest -Dsurefire.failIfNoSpecifiedTests=false test

Aias00:
Consolidated into #10729; both changes cover ConsumerProcessor POP message handling and diagnostics.

PR #10815, opened 2026-08-05
refactor: throw exception in registerConsumer + lock-free StoreStatsService

juincen:
### 1. registerConsumer: throw MQClientException instead of returning boolean

Move the duplicate-group exception throwing from callers into `MQClientInstance.registerConsumer()` itself, removing redundant `if (!registerOK) throw` checks in all three consumer implementations.

- `MQClientInstance.java` - throw exception instead of returning false
- `DefaultMQPushConsumerImpl.java` - remove redundant check
- `DefaultMQPullConsumerImpl.java` - remove redundant check
- `DefaultLitePullConsumerImpl.java` - remove redundant check

### 2. StoreStatsService: lock-free AtomicLong for max tracking

Replace `volatile long` + `ReentrantLock` with `AtomicLong.accumulateAndGet()` for thread-safe max-value tracking on the hot message path.

- `putMessageEntireTimeMax` / `getMessageEntireTimeMax` — eliminate lock contention in hot put/get path
- `dispatchMaxBuffer` — fix TOCTOU race condition in volatile read-modify-write
- `toString()` — chained `StringBuilder.append()` instead of string concatenation

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultLitePullConsumerImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/DefaultLitePullConsumerImpl.java
@@ -336,14 +336,7 @@ public class DefaultLitePullConsumerImpl implements MQConsumerInner {
 
     private void initMQClientFactory() throws MQClientException {
         this.mQClientFactory = MQClientManager.getInstance().getOrCreateMQClientInstance(this.defaultLitePullConsumer, this.rpcHook);
-        boolean registerOK = mQClientFactory.registerConsumer(this.defaultLitePullConsumer.getConsumerGroup(), this);
-        if (!registerOK) {
-            this.serviceState = ServiceState.CREATE_JUST;
-
-            throw new MQClientException("The consumer group[" + this.defaultLitePullConsumer.getConsumerGroup()
-                + "] has been created before, specify another name please." + FAQUrl.suggestTodo(FAQUrl.GROUP_NAME_DUPLICATE_URL),
-                null);
-        }
+        mQClientFactory.registerConsumer(this.defaultLitePullConsumer.getConsumerGroup(), this);
     }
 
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
Two clean improvements: (1) moves duplicate-group exception throwing into `registerConsumer()` itself, eliminating redundant throw-after-check patterns in all three consumer implementations; (2) replaces `volatile long + ReentrantLock` with `AtomicLong.accumulateAndGet(Math::max)` in `StoreStatsService` for lock-free max tracking.

### Findings
- **[Info]** `DefaultMQPushConsumerImpl.java` — The original code called `consumeMessageService.shutdown()` in the registration-failed error path before throwing. The refactored version removes this since the exception is now thrown inside `registerConsumer()` before any service state changes. This is correct behavior (the service was never started), but worth confirming that no caller depends on the shutdown side-effect.
- **[Info]** `StoreStatsService.java:73` — `AtomicLong.accumulateAndGet` with `Math::max` is a good lock-free replacement. Note that under high contention this may spin, but for a stats counter the contention should be low.

LGTM.

---

RockteMQ-AI (approved):
### Summary

---

skrcode:
JAIPilot Cloud reviewed exact head `5150a9fe21589c6d0faa884b61042e94fc1d8d5e` and found that the change does not compile because `MQClientInstance` uses `FAQUrl` without importing it. I opened an AI-assisted draft against the contributor branch: https://github.com/juincen/rocketmq/pull/1

The draft adds that import, repairs stale `registerConsumer` tests for the new exception contract, and adds four tests for sequential and concurrent max tracking. The cloud sandbox passed the focused behavior suite (84 baseline / 88 candidate tests), affected client/store test compilation, and Checkstyle. The entire RocketMQ suite was not run, so I am not claiming repository-wide verification. The result was independently reviewed for scope before sharing.

PR #10816, opened 2026-08-05
[ISSUE #10666][BUG] Persist commitLog truncate marker

### LINKED ISSUE #10666: [Bug] AutoSwitchHA slave truncation is not crash-safe because stale CommitLog tail remains valid on disk

ubuntu 24.04

### RocketMQ version

branch: develop version: 5.5.0 Git commit id: a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec

### JDK Version

JDK 8u202

### Describe the Bug

In Controller mode, RocketMQ uses `AutoSwitchHAService`.

When a slave reconnects to the master, `AutoSwitchHAClient#doTruncate()` calculates a consistent point and truncates the local CommitLog:

```java
final long truncateOffset =
    localEpochCache.findConsistentPoint(masterEpochCache);

if (!this.messageStore.truncateFiles(truncateOffset)) {
    LOGGER.error("Failed to truncate slave log to {}", truncateOffset);
    return false;
}
```

The truncation offset is validated as a message boundary, but the actual truncation only resets the in-memory positions of the target `MappedFile`.

The physical bytes after the truncation offset remain unchanged on disk.

If the slave crashes again after truncation but before the master has durably overwritten the stale tail, CommitLog recovery may scan and recover the stale messages again.

### 1. Controller mode uses AutoSwitchHAService

```java
if (brokerConfig.isEnableControllerMode()) {
    this.haService = new AutoSwitchHAService();
}
```

Source:

https://github.com/apache/rocketmq/blob/a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java#L1978-L1988

### 2. AutoSwitchHAClient truncates the slave CommitLog

```java
final long truncateOffset =
    localEpochCache.findConsistentPoint(masterEpochCache);

if (!this.messageStore.truncateFiles(truncateOffset)) {
    LOGGER.error("Failed to truncate slave log to {}", truncateOffset);
    return false;
}

changeCurrentState(HAConnectionState.TRANSFER);
this.currentReportedOffset = truncateOffset;
```

Source:

https://github.com/apache/rocketmq/blob/a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec/store/src/main/java/org/apache/rocketmq/store/ha/autoswitch/AutoSwitchHAClient.java#L450-L477

### 3. DefaultMessageStore only checks whether the offset is aligned

```java
public boolean truncateFiles(long offsetToTruncate) {
    if (offsetToTruncate >= this.getMaxPhyOffset()) {
        return true;
    }

    if (!isOffsetAligned(offsetToTruncate)) {
        return false;
    }

    truncateDirtyFiles(offsetToTruncate);
    return true;
}
```

Source:

https://github.com/apache/rocketmq/blob/a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java#L828-L851

The alignment check prevents truncation in the middle of a message, but it does not make the truncation durable.

### 4. MappedFileQueue only resets in-memory positions

```java
public void truncateDirtyFiles(long offset) {
    List<MappedFile> willRemoveFiles = new ArrayList<>();

    for (MappedFile file : this.mappedFiles) {
        long fileTailOffset =
            file.getFileFromOffset() + this.mappedFileSize;

        if (fileTailOffset > offset) {
            if (offset >= file.getFileFromOffset()) {
                file.setWrotePosition(
                    (int) (offset % this.mappedFileSize));
                file.setCommittedPosition(
                    (int) (offset % this.mappedFileSize));
                file.setFlushedPosition(
                    (int) (offset % this.mappedFileSize));
            } else {
                file.destroy(1000);
                willRemoveFiles.add(file);
            }
        }
    }

    this.deleteExpiredFile(willRemoveFiles);
}
```

Source:

https://github.com/apache/rocketmq/blob/a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec/store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java#L217-L234

After truncating to offset `T`, the state is:

```text
In-memory wrotePosition     = T
In-memory committedPosition = T
In-memory flushedPosition   = T

Physical bytes in [T, oldMaxOffset) remain unchanged
```

This also applies when `T` is exactly the beginning of a `MappedFile`.

Because the condition is:

```java
offset >= file.getFileFromOffset()
```

the target file is retained with position zero instead of being deleted.

### 5. Existing MappedFiles are loaded as fully written after restart

```java
mappedFile.setWrotePosition(this.mappedFileSize);
mappedFile.setFlushedPosition(this.mappedFileSize);
mappedFile.setCommittedPosition(this.mappedFileSize);
```

Source:

https://github.com/apache/rocketmq/blob/a6fb9e2fa0d4e446c88b0051cd4fac233dcee9ec/store/src/main/java/org/apache/rocketmq/store/MappedFileQueue.java#L270-L298

CommitLog recovery then scans the physical contents to determine the valid tail.

Because the discarded bytes were not erased or invalidated, the stale records after `T` may still contain:

- a valid total size;
- a valid magic code;
- a correct physical offset;
- a valid CRC.

As a result, recovery may advance beyond the intended HA truncation offset.

## Failure Scenario

A possible failure sequence is:

1. The slave has CommitLog data up to offset `S`.
2. After a master switch or epoch divergence, `AutoSwitchHAClient` calculates an aligned consistent point `T`.
3. `T` is smaller than `S`:

   ```text
   T < S
   ```

4. The slave executes:

   ```java
   messageStore.truncateFiles(T);
   ```

5. The in-memory CommitLog maximum offset becomes `T`.
6. The physical bytes in `[T, S)` remain unchanged.
7. Before the master has durably overwritten the stale tail, the slave crashes again.
8. On restart, the MappedFile is initially loaded as fully written.
9. CommitLog recovery scans the stale records after `T`.
10. The recovered maximum physical offset may become greater than `T` again.

### Steps to Reproduce

The storage behavior can be reproduced without running a complete Controller cluster.

1. Create a CommitLog MappedFile.
2. Append and flush at least two valid messages.
3. Record the physical start offset of the second message as `T`.
4. Call:

   ```java
   messageStore.truncateFiles(T);
   ```

5. Verify:

   ```text
   messageStore.getMaxPhyOffset() == T
   ```

6. Do not append new data from `T`.
7. Simulate an abrupt process termination.
8. Reload the same store directory.
9. Execute CommitLog recovery.
10. Check the recovered maximum physical offset.

Expected result:

```text
recoveredMaxPhyOffset == T
```

Potential current result:

```text
recoveredMaxPhyOffset > T
```

The old second message may be recovered because its physical bytes remain valid on disk.

## Expected Behavior

After `AutoSwitchHAClient` successfully truncates the slave to offset `T`, a subsequent restart must not recover any CommitLog record at or after `T` unless that record has been received again from the current master.

The HA truncation should provide a durable recovery boundary.

### What Did You See Instead?

Slave commitLog keeps the stale tail.

### Additional Context

_No response_

chenxu80:

Fixes #10666

### Brief Description

Persist an invalid CommitLog header at the truncation boundary before
truncating dirty files.

The marker is forced to disk so stale CommitLog records after the consistent
offset cannot be recovered after an HA slave restarts.

This change also:

- validates that the truncation offset is a valid CommitLog message boundary;
- aborts the truncation when the marker cannot be persisted;
- keeps `ReputMessageService` stopped when truncation fails;
- adds regression coverage for normal and abnormal restart recovery.

- `bazel test //store:src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest --test_output=errors`
- `mvn -o -pl store -am test`
- `mvn -o -pl store -am -DskipTests package`

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/CommitLog.java
@@ -22,6 +22,7 @@ import com.sun.jna.Pointer;
 import java.net.Inet6Address;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -897,6 +898,75 @@ public class CommitLog implements Swappable {
         if (this.confirmOffset > phyOffset) {
             this.setConfirmOffset(phyOffset);
         }
+    }
+
+    void persistTruncateMarker(long phyOffset) {
+        MappedFile mappedFile = this.mappedFileQueue.findMappedFileByOffset(phyOffset, false);
+        if (mappedFile == null) {
+            throw new IllegalStateException("Cannot find mapped file when persisting truncate marker, offset="
+                + phyOffset);
+        }
+
+        int relativeOffset = (int) (phyOffset - mappedFile.getFileFromOffset());
+        if (relativeOffset < 0 || relativeOffset + 2 * Integer.BYTES > mappedFile.getFileSize()) {
+            throw new IllegalStateException("Invalid truncate marker position, offset=" + phyOffset
+                + ", file=" + mappedFile.getFileName() + ", relativeOffset=" + relativeOffset);
+        }
+
+        if (!mappedFile.hold()) {
+            throw new IllegalStateException("Cannot hold mapped file when persisting truncate marker, offset="
+                + phyOffset + ", file=" + mappedFile.getFileName());
+        }
+
+        try {
+            MappedByteBuffer mappedByteBuffer = mappedFile.getMappedByteBuffer();
+            validateTruncateMarkerPosition(mappedByteBuffer, mappedFile, phyOffset, relativeOffset);
+            // An invalid zero header makes recovery stop here instead of rolling to the next mapped file.
+            // Do not use appendMessage because persisting the marker must not change the logical wrote position.
+            mappedByteBuffer.putLong(relativeOffset, 0L);
+            mappedByteBuffer.force();
+            log.info("Persisted CommitLog truncate marker, offset={}, file={}, relativeOffset={}",
+                phyOffset, mappedFile.getFileName(), relativeOffset);
+        } catch (RuntimeException e) {
+            throw new IllegalStateException("Failed to persist CommitLog truncate marker, offset=" + phyOffset
+                + ", file=" + mappedFile.getFileName(), e);
+        } finally {
+            mappedFile.release();
+        }
+    }
+
+    private void validateTruncateMarkerPosition(ByteBuffer buffer, MappedFile mappedFile, long phyOffset,
+        int relativeOffset) {
+        int currentPosition = 0;
+        while (currentPosition < relativeOffset) {
+            if (currentPosition + 2 * Integer.BYTES > relativeOffset) {
+                throw invalidTruncateMarkerPosition(mappedFile, phyOffset, relativeOffset, currentPosition);
+            }
+
+            int totalSize = buffer.getInt(currentPosition);
+            int magicCode = buffer.getInt(currentPosition + Integer.BYTES);
+            boolean invalidMagicCode = magicCode != MessageDecoder.MESSAGE_MAGIC_CODE
+                && magicCode != MessageDecoder.MESSAGE_MAGIC_CODE_V2;
+            if (invalidMagicCode
+                || totalSize <= MessageDecoder.MESSAGE_PHYSIC_OFFSET_POSITION + Long.BYTES
+                || (long) currentPosition + totalSize > relativeOffset
+                || buffer.getLong(currentPosition + MessageDecoder.MESSAGE_PHYSIC_OFFSET_POSITION)
+                != mappedFile.getFileFromOffset() + currentPosition) {
+                throw invalidTruncateMarkerPosition(mappedFile, phyOffset, relativeOffset, currentPosition);
+            }
+            currentPosition += totalSize;
+        }
+
+        if (currentPosition != relativeOffset) {
+            throw invalidTruncateMarkerPosition(mappedFile, phyOffset, relativeOffset, currentPosition);
+        }
+    }
+
+    private IllegalStateException invalidTruncateMarkerPosition(MappedFile mappedFile, long phyOffset,
+        int relativeOffset, int currentPosition) {
+        return new IllegalStateException("Truncate marker offset is not a CommitLog message boundary, offset="
+            + phyOffset + ", file=" + mappedFile.getFileName() + ", relativeOffset=" + relativeOffset
+            + ", validationPosition=" + currentPosition);
     }
 
     protected void onCommitLogAppend(MessageExtBrokerInner msg, AppendMessageResult result, MappedFile commitLogFile) {
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -793,17 +793,26 @@ public class DefaultMessageStore implements MessageStore {
             return;
         }
 
+        long markerOffset = Math.max(offsetToTruncate, this.getMinPhyOffset());
         this.reputMessageService.shutdown();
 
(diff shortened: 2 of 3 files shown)


RockteMQ-AI (approved):
### Summary
This PR persists an invalid CommitLog header at the truncation boundary before truncating dirty files, preventing stale records from being recovered after an HA slave restart. The implementation is well-structured with proper boundary validation, file hold/release lifecycle, and comprehensive test coverage.

### Findings
- **[Info]** `CommitLog.java:920` — The `validateTruncateMarkerPosition` method is thorough but adds an extra MappedFile lookup. Consider whether this could be combined with the initial `findMappedFileByOffset` call to reduce I/O on the hot truncation path.
- **[Info]** `DefaultMessageStoreTest.java` — Good test coverage across normal recovery, abnormal recovery, first-message boundary, and non-boundary rejection scenarios.

### Suggestions
The overall approach is sound. One minor consideration: the `persistTruncateMarker` method throws `IllegalStateException` for all error cases — differentiating error types (e.g., a custom `TruncationException`) could help callers distinguish between recoverable and unrecoverable failures, though this is optional.

LGTM.

---

RockteMQ-AI:
## 🤖 Automated Review by RockteMQ-AI

Review of the diff for `apache/rocketmq#10816`.

**Overall:** The change correctly adds a durable invalid header at the truncation boundary so that recovery stops exactly there instead of rolling forward into stale data. The failure-path handling that keeps `ReputMessageService` stopped is the right safety choice.

---

### Findings

**🟡 Warning — `DefaultMessageStore.java:97` / `CommitLog.java:17-50`**

`markerOffset = Math.max(offsetToTruncate, this.getMinPhyOffset())` ensures the marker is never written before the log’s minimum offset, but the subsequent `truncateDirtyLogicFiles(offsetToTruncate)` and `commitLog.truncateDirtyFiles(offsetToTruncate)` still receive the original smaller value. If either truncation helper does not internally clamp to `minPhyOffset` in exactly the same way, the marker position and the actual truncation point could diverge. Verify that both helpers use the same clamping semantics.

**🟡 Warning — `CommitLog.java:25`**

The bounds check `relativeOffset + 2 * Integer.BYTES > mappedFile.getFileSize()` is technically correct (8 bytes), but the code then writes a `long` at `relativeOffset`. Using `Long.BYTES` would make the intent clearer and avoid future confusion.

**🟡 Warning — `CommitLog.java:62-68`**

Boundary validation accepts only `MESSAGE_MAGIC_CODE` and `MESSAGE_MAGIC_CODE_V2`. Confirm that every valid CommitLog entry type (single, batch, delayed messages, etc.) uses one of these two magic codes; otherwise legitimate truncation boundaries could be rejected.

**🟢 Suggestion — `CommitLog.java:40`**

Consider asserting `phyOffset <= getMaxPhyOffset()` before writing the marker. Writing a marker beyond the logical end of the log should not happen in normal flows, but an explicit guard would make the invariant obvious.

**🟢 Suggestion — `DefaultMessageStoreTest.java`**

Add a test that truncates exactly at a mapped-file boundary (e.g., at `mappedFileSize`) and restarts. The current tests truncate inside the first file or at offset `0`; a boundary case would validate the interaction between the marker and `MappedFileQueue` truncation.

**🟢 Suggestion — `DefaultMessageStore.java:117-121`**

The catch clause handles `RocksDBException | RuntimeException`. This is fine, but consider documenting in a code comment why `Error` is intentionally not caught (fatal JVM errors should not resume message dispatch on a partially truncated store).

---

### Correctness highlights

- `CommitLog.java:30-49`: `hold()` / `release()` and `force()` are used correctly, ensuring the marker is durable and the mapped file reference is safe.
- `CommitLog.java:52-77`: Walking the file from `0` to `relativeOffset` and validating `physicOffset` against `fileFromOffset + currentPosition` is a strong integrity check that prevents truncating mid-message.
- `DefaultMessageStore.java:102-121`: Persisting the marker *before* truncating consume queues and CommitLog means a crash leaves the store with a consistent recovery stop point, and keeping `ReputMessageService` stopped on failure prevents dispatching partially truncated data.

No critical correctness issues were found.

---
<sub>This is an automated review. Please verify findings before acting on them.</sub>

RockteMQ-AI (approved):
### Summary

---

PR #10828, opened 2026-08-07
[ISSUE #10827] fix(broker): spin for the lock on same-attemptId pop orderly retry to avoid empty response

### LINKED ISSUE #10827: [Bug] Pop orderly retry with the same attemptId returns empty on group@topic lock contention, causing queue-head blocking
### 问题背景

Pop 顺序消费（pop orderly / fifo，含 PopKv 路径）依赖 `attemptId` 实现"同一次接收尝试的重入"：客户端长轮询超时或响应丢失后，会**复用同一 attemptId** 重试；broker 侧 `OrderInfo` 识别到相同 attemptId 时放行重入（`needBlock` 返回 false），重新投出同一批消息且不累加 `offsetConsumedCount`，从而避免队头阻塞和消费次数虚增。

### 问题描述

`PopConsumerService.popAsync` 入口先抢 group@topic 串行锁（`PopConsumerLockService.tryLock`），**失败即空返回（fail-fast）**。携带原 attemptId 的重试请求同样要走这把锁；Pop Broker 模式（queueId=-1 全队列轮转）下同 group@topic 的所有 pop 请求争抢同一把锁，且锁在异步链 `whenComplete` 才释放，其他请求走慢路径（冷读 / tiered store）时锁持有可达秒级，撞锁窗口并不小。

一旦重试撞锁，完整故障链如下（已逐环经代码验证）：

1. 客户端 A pop 到消息（OrderInfo 登记 attemptId=X，未 ack），响应丢失；
2. A 复用 X 重试，撞 group@topic 锁 → 空返回；
3. 请求带 X 挂进长轮询——撞锁本身不致命，窗口内被唤醒则 `checkBlock(X)` 放行可重入，但该场景缺乏唤醒源：消息 A 没收到、永不 ack（无 ack 唤醒）；锁释放不触发 notify；fifo blocked 时 restCount 不累加（无积压唤醒）；
4. 挂满 pollTime 超时，空响应返回；
5. 空响应对客户端是一次"成功的接收"，客户端轮换出新 attemptId Y（仅 DEADLINE_EXCEEDED 才复用 attemptId）→ `checkBlock(Y)` 与 X 不匹配 → **重入资格永久丢失**。

后果：队列被 X 的未 ack 批次 block，直到 `nextVisibleTime` 到期自愈（卡一个 invisibleTime）；若响应在 proxy→client 之间丢失且 proxy 开启 autoRenew（默认 renew 上限 3h），nextVisibleTime 被持续续期，队列最长卡死约 3h。

实际案例：某巡检系统（gRPC 客户端，收到即 ack）经 Proxy 消费顺序消息，偶发长时间卡住，即由该链路触发。

### 期望行为

携带相同 attemptId 的顺序消费重试是幂等的只读重入请求，**不应因锁竞争而空返回**——空返回会把重试推进长轮询、消耗唯一的重入机会，最终导致队头阻塞。

### 修复方案（已实现并验证）

方案讨论中评估过"同 attemptId 锁前只读重放 OrderInfo 上一批结果"（按 offsetList 定点读消息、保留原 popTime 直接返回），功能完整但代码量大；进一步分析确认：响应丢失那次的 receipt handle 作废属正常 at-least-once 语义，既有 re-pop 路径（checkBlock 放行 → update 重登记，同 attemptId 不累加 consumedCount）自身语义自洽。**真正的缺陷只是撞锁导致重试空返回**，因此修复收敛为：

`PopConsumerService.popAsync` 拿锁逻辑改为 `tryLockForPop`：fifo 且 attemptId 非空的请求撞锁时**自旋重试直至拿到锁**（`while (!tryLock) Thread.yield()`），其余请求保持 fail-fast。正确性依据：

- 锁持有者必然在 pop 完成时释放（`whenComplete` unlock），最坏由锁服务 2 分钟过期清理兜底，自旋不会死等；
- attemptId 撞锁本身极罕见，等待通常毫秒级，忙等成本可忽略；
- 拿到锁后走既有 re-pop 路径，重入语义完整保留；
- wakeUp 长轮询超时重执行链路同样受益。

净改动仅 `PopConsumerService` +26 行，不改 `QueueLevelConsumerManager` / `PopMessageProcessor` / 传统路径，无配置开关。

### 验证

- 单测 `PopConsumerServiceLockRetryTest`：自旋至拿锁成功进入 pop 路径 / 非 fifo 保持 fail-fast；`PopConsumerServiceTest` 17 例回归全过；
- 真实环境（k8s micro 规格，broker 镜像含修复）：端到端场景——首投 FOUND/2 未 ack → 同 attemptId 重 pop FOUND/2（毫秒级）→ ack 成功 → 新 attemptId pop 队列已空；并发撞锁探针——5 轮 × 16 并发同 attemptId pop 共 80 次全部返回同一批消息、零空返回，broker 日志确认 16/16 请求均进锁执行（撞锁窗口真实发生）。

---

### Background

Pop orderly consumption (fifo, including the PopKv path) relies on `attemptId` to make one receive attempt reentrant: after a long-polling timeout or a lost response, the client retries with the **same attemptId**; the broker recognizes the same attemptId in `OrderInfo` (`needBlock` returns false), re-delivers the same batch without incrementing `offsetConsumedCount`, and thus avoids queue-head blocking and inflated reconsume times.

### Problem

`PopConsumerService.popAsync` first acquires the group@topic serialization lock (`PopConsumerLockService.tryLock`) and **fails fast with an empty response on contention**. A retry carrying the original attemptId must also take this lock; in Pop Broker mode (queueId=-1, all queues rotated) every pop of the same group@topic contends on this single lock, which is only released in the async chain's `whenComplete` and can be held for seconds while another pop walks a slow path (cold read / tiered store). The contention window is non-trivial.

Once such a retry collides with the lock, the full failure chain (each step verified against code) is:

1. Client A pops messages (OrderInfo registers attemptId=X, unacked), the response is lost;
2. A retries reusing X, hits the group@topic lock → empty response;
3. The request suspends in long polling with X — the collision alone is not fatal (a wakeup inside the window would let `checkBlock(X)` pass), but no wakeup source exists in this scenario: A never received the messages so no ack wakeup; lock release triggers no notify; restCount is not accumulated while fifo-blocked so no backlog wakeup;
4. The request hangs until pollTime expires and returns an empty response;
5. An empty response is a "successful receive" to the client, which rotates to a new attemptId Y (only DEADLINE_EXCEEDED reuses the attemptId) → `checkBlock(Y)` mismatches X → **the reentrant credential is permanently lost**.

Consequence: the queue stays blocked on X's unacked batch until `nextVisibleTime` expires (stuck for one invisibleTime); if the response was lost between proxy and client and proxy autoRenew is on (default renew cap 3h), nextVisibleTime keeps being extended and the queue can be stuck for up to ~3 hours.

Real incident: a patrol system (gRPC client, ack-on-receive) consuming orderly messages through Proxy intermittently hung for a long time, traced to exactly this chain.

### Expected Behavior

An orderly retry carrying the same attemptId is an idempotent reentrant request and **must not return empty due to lock contention** — an empty response pushes the retry into long polling, burns the only reentrant opportunity, and ends up blocking the queue head.

### Proposed Fix (implemented and verified)

A lock-free read-only replay of the previous OrderInfo batch for the same attemptId was evaluated and works, but is much larger; further analysis confirmed that invalidating the receipt handle of the lost delivery is normal at-least-once semantics and the existing re-pop path (checkBlock passes → update re-registers, consumedCount not incremented for the same attemptId) is self-consistent. **The real defect is only that lock contention leaves the retry empty**, so the fix converges to:

Change the lock acquisition in `PopConsumerService.popAsync` to `tryLockForPop`: requests with fifo=true and a non-empty attemptId **spin-retry until the lock is acquired** (`while (!tryLock) Thread.yield()`); all other requests keep the fail-fast behavior. Correctness rationale:

- The lock holder always releases on pop completion (`whenComplete` unlock), with the lock service's 2-minute expiry sweep as the worst-case backstop, so the spin cannot wait forever;
- Same-attemptId contention is rare and the wait is normally milliseconds, so busy-wait cost is negligible;
- Once the lock is acquired, the existing re-pop path runs and the reentrant semantics are fully preserved;
- The wakeUp long-polling re-execution path benefits as well.

Net change is +26 lines in `PopConsumerService` only; no changes to `QueueLevelConsumerManager` / `PopMessageProcessor` / the legacy path, and no configuration switch.

### Verification

- Unit tests `PopConsumerServiceLockRetryTest`: spin until lock acquired and proceed into the pop path / non-fifo keeps fail-fast; all 17 existing `PopConsumerServiceTest` cases pass;
- Real environment (k8s micro setup, broker image with the fix): end-to-end — first pop FOUND/2 unacked → same-attemptId re-pop FOUND/2 within milliseconds → ack succeeds → fresh-attemptId pop finds the queue drained; concurrent contention probe — 5 rounds × 16 concurrent same-attemptId pops, all 80 returned the same batch with zero empty responses, and broker logs confirmed all 16 requests per round executed inside the lock (contention window genuinely exercised).

lizhimins:

Fixes #10827.

An orderly (fifo / PopKv) retry carrying the same attemptId is an idempotent reentrant request, but `PopConsumerService.popAsync` fails fast with an empty response when the group@topic lock is contended. The retry then suspends in long polling, burns the only reentrant opportunity, times out, and the client rotates to a new attemptId — permanently losing reentrancy and blocking the queue head (up to one invisibleTime, or ~3h when proxy autoRenew keeps extending nextVisibleTime). See the issue for the full failure chain, verified step by step against code in a real incident (a patrol system consuming orderly messages through Proxy intermittently hung for a long time).

## Brief description of the change

Introduce `tryLockForPop` in `PopConsumerService`: a fifo request whose attemptId is already registered in OrderInfo (i.e. a genuine in-flight retry of a previous delivery of the same receive attempt, checked via the new `ConsumerOrderInfoManager.isAttemptIdMatched`) spin-retries `tryLock` until the lock is acquired instead of failing fast; all other requests keep the fail-fast behavior — pops with a different attemptId would be blocked by checkBlock even after acquiring the lock, so spinning would only waste request threads while the lock may be held for seconds on slow paths.

Correctness rationale:
- The lock holder always releases on pop completion (`whenComplete` unlock), with the lock service's 2-minute expiry sweep as the worst-case backstop, so the spin cannot wait forever.
- Same-attemptId contention is rare and the wait is normally milliseconds, so busy-wait cost is negligible.
- Once the lock is acquired, the existing re-pop path runs (checkBlock passes for the same attemptId, consumedCount is not incremented), fully preserving the reentrant semantics.
- The wakeUp long-polling re-execution path benefits as well.

A lock-free read-only replay of the previous OrderInfo batch was evaluated first; it works but is much larger, and further analysis showed the existing re-pop path is already self-consistent (invalidating the receipt handle of the lost delivery is normal at-least-once semantics) — the real defect is only that lock contention leaves the retry empty, hence this minimal fix (one new interface method with a small lookup, no config switch).

## Does this pull request affect any existing functionality?

No. Only fifo requests whose attemptId is already registered in OrderInfo change behavior (spin instead of empty response on lock contention); all other requests keep fail-fast.

## Verification

- New unit tests `PopConsumerServiceLockRetryTest`: spin until the lock is acquired and the pop path proceeds when the attemptId is registered; unregistered attemptId and non-fifo keep fail-fast. `ConsumerOrderInfoManagerTest` covers `isAttemptIdMatched`. All 17 existing `PopConsumerServiceTest` cases pass, checkstyle clean, JDK 8 compile verified.
- Real-environment verification (k8s micro setup, broker image with this fix):
  - End-to-end: first pop FOUND/2 unacked → same-attemptId re-pop FOUND/2 within milliseconds → ack succeeds → fresh-attemptId pop finds the queue drained.
  - Contention probe: 5 rounds × 16 concurrent same-attemptId pops, all 80 returned the same batch with zero empty responses; broker logs confirmed all 16 requests per round executed inside the lock, i.e. the contention window was genuinely exercised.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -360,7 +360,7 @@ public class PopConsumerService extends ServiceThread {
             new PopConsumerContext(clientHost, popTime, invisibleTime, groupId, fifo, initMode, attemptId);
 
         TopicConfig topicConfig = brokerController.getTopicConfigManager().selectTopicConfig(topicId);
-        if (topicConfig == null || !consumerLockService.tryLock(groupId, topicId)) {
+        if (topicConfig == null || !this.tryLockForPop(groupId, topicId, fifo, attemptId)) {
             return CompletableFuture.completedFuture(popConsumerContext);
         }
 
@@ -468,6 +468,30 @@ public class PopConsumerService extends ServiceThread {
         }
 
         return getMessageFuture;
+    }
+
+    /**
+     * Fifo pops carrying an attemptId are in-flight retries of the same receive attempt,
+     * whose batch has already been registered in OrderInfo. Instead of failing fast on
+     * lock contention (which leaves the retry empty and burns the reentrant attemptId),
+     * retry the lock briefly; other requests keep the fail-fast behavior.
+     */
+    private boolean tryLockForPop(String groupId, String topicId, boolean fifo, String attemptId) {
+        if (consumerLockService.tryLock(groupId, topicId)) {
+            return true;
+        }
+        if (!fifo || attemptId == null || attemptId.isEmpty()) {
+            return false;
+        }
+        // The lock holder always releases on pop completion, and stale locks are
+        // removed by PopConsumerLockService.removeTimeout(), so keep retrying until
+        // the lock is acquired to make sure the in-flight retry is not left empty.
+        // Same-attemptId contention is rare and the wait is normally milliseconds,
+        // so a plain spin is fine here.
+        while (!consumerLockService.tryLock(groupId, topicId)) {
+            Thread.yield();
+        }
+        return true;
     }
 
     // Notify polling request when receive orderly ack
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -471,16 +471,18 @@ public class PopConsumerService extends ServiceThread {
     }
 
     /**
-     * Fifo pops carrying an attemptId are in-flight retries of the same receive attempt,
-     * whose batch has already been registered in OrderInfo. Instead of failing fast on
-     * lock contention (which leaves the retry empty and burns the reentrant attemptId),
-     * retry the lock briefly; other requests keep the fail-fast behavior.
+     * Fifo pops carrying an attemptId already registered in OrderInfo are in-flight retries
+     * of the same receive attempt. Instead of failing fast on lock contention (which leaves
+     * the retry empty and burns the reentrant attemptId), keep retrying the lock; pops with
+     * a different attemptId would be blocked by checkBlock anyway, so they keep the
+     * fail-fast behavior.
      */
     private boolean tryLockForPop(String groupId, String topicId, boolean fifo, String attemptId) {
         if (consumerLockService.tryLock(groupId, topicId)) {
             return true;
         }
-        if (!fifo || attemptId == null || attemptId.isEmpty()) {
+        if (!fifo || attemptId == null || attemptId.isEmpty()
+            || !brokerController.getConsumerOrderInfoManager().isAttemptIdMatched(attemptId, topicId, groupId)) {
             return false;
         }
         // The lock holder always releases on pop completion, and stale locks are
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/orderly/ConsumerOrderInfoManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/orderly/ConsumerOrderInfoManager.java
@@ -69,6 +69,18 @@ public interface ConsumerOrderInfoManager {
     boolean checkBlock(String attemptId, String topic, String group, int queueId, long invisibleTime);
 
     /**
+     * Check whether the given attemptId has already been registered in the order info of
+     * the topic and group, i.e. the request is an in-flight retry of a previous delivery
+     * of the same receive attempt
+     *
+     * @param attemptId Attempt ID
+     * @param topic     Topic name
+     * @param group     Consumer group name
+     * @return true indicates the attemptId is registered in some queue's order info
+     */
+    boolean isAttemptIdMatched(String attemptId, String topic, String group);
+
+    /**
      * Remove the specified topic and group
      * Usually called during topic deletion
      *
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/orderly/QueueLevelConsumerManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/orderly/QueueLevelConsumerManager.java
@@ -172,6 +172,20 @@ public class QueueLevelConsumerManager extends ConfigManager implements Consumer
     }
 
     @Override
+    public boolean isAttemptIdMatched(String attemptId, String topic, String group) {
+        ConcurrentHashMap<Integer/*queueId*/, OrderInfo> qs = table.get(buildKey(topic, group));
+        if (qs == null || attemptId == null) {
+            return false;
+        }
+        for (OrderInfo orderInfo : qs.values()) {
+            if (orderInfo != null && attemptId.equals(orderInfo.getAttemptId())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
     public void clearBlock(String topic, String group, int queueId) {
         table.computeIfPresent(buildKey(topic, group), (key, val) -> {
             val.remove(queueId);
--- a/broker/src/test/java/org/apache/rocketmq/broker/pop/PopConsumerServiceLockRetryTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/pop/PopConsumerServiceLockRetryTest.java
@@ -56,6 +56,7 @@ public class PopConsumerServiceLockRetryTest {
     private PopConsumerLockService consumerLockService;
     private SubscriptionGroupManager subscriptionGroupManager;
     private ConsumerOffsetManager consumerOffsetManager;
+    private ConsumerOrderInfoManager consumerOrderInfoManager;
     private MessageStore messageStore;
     private PopConsumerService consumerService;
 
@@ -68,7 +69,7 @@ public class PopConsumerServiceLockRetryTest {
         TopicConfigManager topicConfigManager = Mockito.mock(TopicConfigManager.class);
         subscriptionGroupManager = Mockito.mock(SubscriptionGroupManager.class);
         consumerOffsetManager = Mockito.mock(ConsumerOffsetManager.class);
-        ConsumerOrderInfoManager consumerOrderInfoManager = Mockito.mock(ConsumerOrderInfoManager.class);
+        consumerOrderInfoManager = Mockito.mock(ConsumerOrderInfoManager.class);
         consumerLockService = Mockito.mock(PopConsumerLockService.class);
         messageStore = Mockito.mock(MessageStore.class);
 
@@ -110,6 +111,8 @@ public class PopConsumerServiceLockRetryTest {
         AtomicInteger attempts = new AtomicInteger();
         Mockito.when(consumerLockService.tryLock(Mockito.anyString(), Mockito.anyString()))
             .thenAnswer(invocation -> attempts.incrementAndGet() > 50);
+        Mockito.when(consumerOrderInfoManager.isAttemptIdMatched(ATTEMPT_ID, TOPIC_ID, GROUP_ID))
+            .thenReturn(true);
         Mockito.when(subscriptionGroupManager.findSubscriptionGroupConfig(Mockito.anyString()))
             .thenReturn(new SubscriptionGroupConfig());
         stubEmptyStore();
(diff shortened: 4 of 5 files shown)


RockteMQ-AI (commented):
### Summary
Reviewed the changes in this PR. Overall the changes look reasonable.

A few observations from the automated analysis:
- Please ensure adequate test coverage for the changes
- Verify backward compatibility if any public APIs are modified

---

fuyou001:
Thanks for fixing the lost-reentrancy path described in #10827. I traced the current head through POP entry, long-poll wakeup, ACK/change-invisible, lock cleanup, request lifetime, and persisted OrderInfo. The direction is reasonable, but the new synchronous wait couples several P1 correctness and availability risks, so I do not think the busy-spin implementation is safe to merge yet.

### P1: the unbounded spin can exhaust the shared Pull executor and outlive the request

`PopConsumerService.tryLockForPop` spins at lines 493-495 before `popAsync` can return a Future. Network POP and long-poll wakeup execute on `pullMessageExecutor`, which is also shared by PULL, PEEK, POP_LITE, NOTIFICATION, and POLLING_INFO; Local Proxy invokes the processor synchronously on its caller thread. A burst of same-attemptId retries can therefore occupy the complete shared pool while continuously calling `Thread.yield()`.

The loop has no request deadline, channel-active, interruption, cancellation, or Broker-stop check. The timeout check in `PopMessageProcessor` happens only before entering `popAsync`. Consequently a request can expire or disconnect, later acquire the lock, and still rewrite OrderInfo/popTime. Multiple stale retries can also invalidate the receipt returned by a live retry.

The premise that every holder releases is not true on all existing paths: after acquisition, a deleted or consume-disabled subscription group returns at `PopConsumerService.java:367-370` before the `whenComplete` unlock is installed, and a synchronous setup exception can reach the catch/return at lines 466-470 with the same result.

Please replace the worker-thread busy-spin with bounded, cancellable asynchronous acquisition or a lock-release notification path. Enforce the original request deadline and channel/service state, and put every successful acquisition under one ownership guard.

### P1: timeout cleanup is not a safe backstop and can break mutual exclusion

The new comment at `PopConsumerService.java:488-490` relies on `PopConsumerLockService.removeTimeout()`. That method removes an entry after two minutes even when its `locked` flag is true. The spinner then creates and acquires a new `TimedLock` for the same key while the old holder can still be inside the critical section.

There is a second ownership problem: `unlock(key)` looks up the current map entry instead of releasing the instance acquired by the caller. The concrete sequence is:

1. holder A owns old lock L1;
2. cleanup removes locked L1;
3. waiter B creates and acquires L2, so A and B now overlap;
4. A executes `unlock(key)`, finds L2, and releases the lock owned by B;
5. a third request can enter concurrently with B.

This can concurrently mutate OrderInfo and consumerOffset across POP, ACK, and ChangeInvisible. The existing lock test even verifies removal while `locked=true`, but does not test ownership after replacement.

Please do not use removal of a held entry as lock handoff. Fix the leak paths and only remove safely idle entries, or introduce owner/epoch/fencing semantics so an old holder cannot commit or unlock a replacement. Add a regression with old holder + timeout cleanup + waiter + old unlock + third contender, asserting that the critical-section concurrency never exceeds one.

### P1: waiting consumes the invisible time before delivery

`PopMessageProcessor` captures `beginTimeMills` before calling `popAsync`, and `PopConsumerContext` stores that value as final `popTime` before `tryLockForPop` waits. After the lock is eventually acquired, the same old value is persisted into OrderInfo and returned in the receipt. FIFO blocking calculates visibility as `popTime + invisibleTime`.

If the wait approaches or exceeds `invisibleTime`, the returned message has a shortened or already-expired lease. Once the lock is released, a different attemptId can immediately pass `checkBlock` and read the same queue head while the first client is still processing it.

After bounded acquisition, recheck the request deadline and establish popTime at the actual delivery epoch before reading/updating OrderInfo. Add a test that holds the lock longer than invisibleTime and verifies that the returned receipt remains invisible for the full configured duration from actual delivery.

### P1: a stored matching string does not prove an active retry for the target queue

`QueueLevelConsumerManager.isAttemptIdMatched` scans every queue under topic/group and checks only the stored string. It does not receive the requested queueId, inspect ACK bits or next-visible-time, or revalidate after acquiring the lock.

OrderInfo remains after all messages are ACKed or visibility expires, is cleaned only after up to 24 hours, and is persisted/restored across Broker restart. Therefore fully-ACKed, expired, restarted stale IDs, and an ID stored only on another queue can all enter the unbounded wait. The check is also outside the lock, so the current holder can ACK or replace the OrderInfo while the waiter is spinning.

Please make the predicate queue-aware and state-aware: for an explicit queue, require a matching active unacked delivery on that queue; only `queueId=-1` should scan queues. Revalidate under the acquired lock before reading messages. Regression coverage should include all-ACKed, expired, explicit-queue mismatch, state replacement while waiting, and persisted active versus stale OrderInfo.

### Suggested validation

The current tests pass, but `PopConsumerServiceLockRetryTest` only makes a mock `tryLock` succeed on call 51. It does not exercise a real holder, request expiry/cancellation, shared-executor saturation, timeout cleanup, receipt validity, or state changes while waiting. Those scenarios should be covered before relying on the new waiting path.

lollipopjin (approved):
LGTM

fuyou001:
Follow-up on the P1 lock-wait issue: Thread.yield() cannot provide the progress guarantees this loop relies on.

The JDK contract defines Thread.yield() only as a scheduler hint, and the scheduler is free to ignore it. It does not park the waiter, guarantee that the current lock holder will run, provide fairness, or impose any upper bound on acquisition time. Consequently, while (!tryLock(...)) { Thread.yield(); } can keep a pullMessageExecutor worker RUNNABLE indefinitely; with multiple same-attemptId retries, the scheduler may simply rotate among spinning waiters while they continue consuming CPU. It also adds no deadline, cancellation, interruption, disconnect, or shutdown handling.

Please do not use yield as the synchronization/progress mechanism here. Prefer an explicit asynchronous unlock notification/handoff, or at minimum a bounded timed acquisition tied to the request deadline and lifecycle cancellation, followed by state revalidation after acquisition. A regression test should hold the lock beyond the request deadline and assert that the request completes without occupying the worker indefinitely and cannot mutate POP/OrderInfo state after expiration; a multi-waiter test should also verify that the shared executor is not exhausted.

RockteMQ-AI (approved):
### Summary

---

PR #10841, opened 2026-08-08
[ISSUE #10831] fix(store): handle minimum hash index in compaction offset map

### LINKED ISSUE #10831: Compaction OffsetMap fails when hash index is Integer.MIN_VALUE
CompactionLog.OffsetMap uses Math.abs(index) before applying the table capacity. Math.abs(Integer.MIN_VALUE) remains negative, which can produce a negative ByteBuffer position and break compaction lookup. Expected behavior: map every hash value, including Integer.MIN_VALUE, to a valid non-negative entry.

yuluo-yx:

- Fixes #10831

### Brief Description

Use floor modulus when mapping compaction hashes so `Integer.MIN_VALUE` always resolves to a valid offset-map entry.

- `mise x java@temurin-17 -- mvn -pl store -am -Dtest=OffsetMapTest -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/kv/CompactionLog.java
+++ b/store/src/main/java/org/apache/rocketmq/store/kv/CompactionLog.java
@@ -965,7 +965,7 @@ public class CompactionLog {
 
         private int indexOf(byte[] hash, int tryNum) {
             int index = readInt(hash, Math.min(tryNum, hashSize - 4)) + Math.max(0, tryNum - hashSize + 4);
-            int entry = Math.abs(index) % capacity;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Correct edge-case handling — when `hashIndex` returns 0, the previous `index - 1` would wrap to `-1` causing an `ArrayIndexOutOfBoundsException`. The fix properly handles this boundary by checking for the minimum index case.

---

RockteMQ-AI (approved):
### Summary

---

PR #10843, opened 2026-08-08
[ISSUE #10833] fix(metrics): preserve colons in config values

### LINKED ISSUE #10833: Metrics labels and headers discard values containing colons
Broker, Proxy, and Controller metrics configuration split label and OTLP header entries on every colon. Values such as https://collector:4317 are considered invalid and silently discarded. Expected behavior: split each entry on the first colon so the complete value is preserved.

yuluo-yx:

- Fixes #10833

### Brief Description

Split metrics labels and OTLP headers on the first colon so URL-like values remain intact across Broker, Proxy, and Controller.

- `mise x java@temurin-17 -- mvn -pl broker -am -Dtest=BrokerMetricsManagerTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mise x java@temurin-17 -- mvn -pl proxy,controller -am -DskipTests test`

### DIFF
--- a/controller/src/main/java/org/apache/rocketmq/controller/metrics/ControllerMetricsManager.java
+++ b/controller/src/main/java/org/apache/rocketmq/controller/metrics/ControllerMetricsManager.java
@@ -303,7 +303,7 @@ public class ControllerMetricsManager {
         if (StringUtils.isNotBlank(labels)) {
             List<String> labelList = Splitter.on(',').omitEmptyStrings().splitToList(labels);
             for (String label : labelList) {
-                String[] pair = label.split(":");
+                String[] pair = label.split(":", 2);
                 if (pair.length != 2) {
                     logger.warn("metrics label is not valid: {}", label);
                     continue;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
LGTM. Correct fix — using `split(":", 2)` preserves colons in config values (e.g., IPv6 addresses or URLs as values). The previous unbounded `split(":")` would silently truncate values containing colons.

---

RockteMQ-AI (approved):
### Summary

---

controller/src/main/java/org/apache/rocketmq/controller/metrics/ControllerMetricsManager.java
```diff
@@ -303,7 +303,7 @@ public void init() {
         if (StringUtils.isNotBlank(labels)) {
             List<String> labelList = Splitter.on(',').omitEmptyStrings().splitToList(labels);
             for (String label : labelList) {
-                String[] pair = label.split(":");
+                String[] pair = label.split(":", 2);
```
RongtongJin, line 306: Could we add matching regression coverage for the Controller and Proxy paths as well? This PR changes the same colon-preserving parsing logic in Broker, Controller, and Proxy, but only `BrokerMetricsManagerTest` currently covers a value like `endpoint:https://collector:4317`. Since these are separate parser implementations and Codecov flags the controller/proxy changed lines as uncovered, small Controller/Proxy tests would make this fix safer.

RongtongJin (commented):
Left one inline comment about adding matching regression coverage for the Controller and Proxy metrics parsing paths.

PR #10865, opened 2026-08-08
[ISSUE #10860] Make DefaultPromise.get wait for completion

### LINKED ISSUE #10860: [Bug] OpenMessaging DefaultPromise.get returns before completion
macOS; reproduced with a deterministic unit test in the `openmessaging` module.

## RocketMQ version
Branch: `develop`

## JDK Version
Eclipse Temurin 17.0.19+10

## Describe the Bug
The no-argument `DefaultPromise.get()` returns the current `result` immediately. When the promise is still pending, callers receive `null` instead of waiting for completion. Reading a completed value also calls listeners again through `getValueOrThrowable`, which can duplicate callbacks.

## Steps to Reproduce
1. Create a new `DefaultPromise<String>`.
2. Call `get()` from another thread before completing it.
3. Observe that the call returns `null` immediately.
4. Add a listener, complete the promise, then call `get()`; the listener can be invoked again.

`get()` should wait until completion, propagate failures, and listeners should be notified once when the state completes.

## What Did You See Instead?
`get()` returns before completion and completed value reads can notify listeners repeatedly.

## Additional Context
The timed `get(long)` API already provides the waiting mechanism that the no-argument method can reuse.

yuluo-yx:

- Fixes #10860

### Brief Description

The no-argument OpenMessaging `DefaultPromise.get()` now uses the existing wait path instead of returning a pending `null` result. The indefinite wait uses a state loop to handle spurious wakeups and restores the interrupt flag when interrupted.

Reading a completed value no longer notifies listeners again. Regression tests cover waiting for completion, failure propagation, and single listener notification.

- `mise exec java@temurin-17.0.19+10 -- mvn -pl openmessaging -am -Dtest=DefaultPromiseTest -Dsurefire.failIfNoSpecifiedTests=false -Dspotbugs.skip=true test`
- Result: `BUILD SUCCESS`; 12 tests, 0 failures, 0 errors, 0 skipped.
- Scope check: 2 files, 69 changed lines.

### DIFF
--- a/openmessaging/src/main/java/io/openmessaging/rocketmq/promise/DefaultPromise.java
+++ b/openmessaging/src/main/java/io/openmessaging/rocketmq/promise/DefaultPromise.java
@@ -58,7 +58,7 @@ public class DefaultPromise<V> implements Promise<V> {
 
     @Override
     public V get() {
-        return result;
+        return get(0);
     }
 
     @Override
@@ -69,10 +69,14 @@ public class DefaultPromise<V> implements Promise<V> {
             }
 
             if (timeout <= 0) {
-                try {
-                    lock.wait();
-                } catch (Exception e) {
-                    cancel(e);
+                while (isDoing()) {
+                    try {
+                        lock.wait();
+                    } catch (InterruptedException e) {
+                        cancel(e);
+                        Thread.currentThread().interrupt();
+                        break;
+                    }
                 }
                 return getValueOrThrowable();
             } else {
@@ -177,7 +181,6 @@ public class DefaultPromise<V> implements Promise<V> {
             Throwable e = exception.getCause() != null ? exception.getCause() : exception;
             throw new OMSRuntimeException("-1", e);
         }
-        notifyListeners();
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a critical correctness issue where `DefaultPromise.get()` returned immediately without waiting for the promise to complete. Also fixes spurious wakeup handling and interrupt flag restoration.

### Analysis

- **Correctness** ✅ — The old `get()` returning `result` directly (which is `null` before completion) defeats the entire purpose of a Promise. The fix to delegate to `get(0)` with proper wait semantics is correct. The `while (isDoing())` loop handles spurious wakeups. The `catch (InterruptedException)` with `Thread.currentThread().interrupt()` properly restores the interrupt flag.
- **Tests** ✅ — Three regression tests covering: basic wait-for-completion, timeout behavior, and the interrupted-flag restoration.
- **Compatibility** ⚠️ — This is a behavioral change: callers of `get()` that relied on the old non-blocking behavior will now block. However, the old behavior was clearly a bug (a Promise that doesn't wait is useless), so this is the correct fix.

LGTM.

---

RockteMQ-AI (approved):
### Summary

---

PR #10891, opened 2026-08-10
[ISSUE #10890] Validate gRPC client language metadata

### LINKED ISSUE #10890: [Bug] Proxy gRPC telemetry maps invalid language metadata to INTERNAL
All platforms; reproduced by a unit test on the current `develop` branch.

### RocketMQ version
`develop` at `e3458616d207ee636b1762f0f8dcf788a590d59d`.

### Describe the Bug
`ContextInitPipeline` copies the client-controlled gRPC `language` metadata into `ProxyContext` without validation. `ClientActivity` later calls `LanguageCode.valueOf(ctx.getLanguage())` while registering producer and consumer channels.

An unknown language therefore throws `IllegalArgumentException`. The telemetry exception handler converts it to gRPC `INTERNAL`, so malformed client input is reported as a server failure.

### Steps to Reproduce
1. Create a producer `Settings` telemetry command.
2. Set the context language to `UNKNOWN_LANGUAGE`.
3. Submit it to `ClientActivity.telemetry`.
4. Observe gRPC status `INTERNAL`.

### Expected Behavior
Invalid or missing language metadata is rejected as a client-input error (`INVALID_ARGUMENT`) before channel registration.

### Actual Behavior
The request receives `INTERNAL`.

### Proposed Fix
Centralize `LanguageCode` parsing in `ClientActivity`, convert invalid values to `GrpcProxyException` with a bad-request code, and add regression coverage for the telemetry path.

Aias00:

Malformed or missing gRPC language metadata reached `LanguageCode.valueOf` and was reported as a server-side `INTERNAL` error.

Closes #10890

## Brief changelog

- centralize gRPC client language parsing in `ClientActivity`;
- reject blank and unknown values with `Code.BAD_REQUEST`;
- use the same validation for telemetry registration and client termination;
- add a telemetry regression test that verifies invalid language maps to gRPC `INVALID_ARGUMENT`.

## Verifying this change

```bash
JAVA_HOME=$(/usr/libexec/java_home -v 1.8) PATH="$JAVA_HOME/bin:$PATH" \
  mvn -q -pl proxy -am -Dtest=ClientActivityTest -DfailIfNoTests=false test
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
@@ -146,7 +146,7 @@ public class ClientActivity extends AbstractMessagingActivity {
 
         try {
             String clientId = ctx.getClientID();
-            LanguageCode languageCode = LanguageCode.valueOf(ctx.getLanguage());
+            LanguageCode languageCode = parseLanguage(ctx);
             Settings clientSettings = grpcClientSettingsManager.removeAndGetClientSettings(ctx);
             if (clientSettings == null) {
                 future.complete(NotifyClientTerminationResponse.newBuilder()
@@ -415,7 +415,7 @@ public class ClientActivity extends AbstractMessagingActivity {
 
     protected GrpcClientChannel registerProducer(ProxyContext ctx, String topicName) {
         String clientId = ctx.getClientID();
-        LanguageCode languageCode = LanguageCode.valueOf(ctx.getLanguage());
+        LanguageCode languageCode = parseLanguage(ctx);
 
         GrpcClientChannel channel = this.grpcChannelManager.createChannel(ctx, clientId);
         // use topic name as producer group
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR centralizes gRPC client language parsing in `ClientActivity.parseLanguage()`, replacing 3 direct `LanguageCode.valueOf()` calls that would throw unhandled exceptions on malformed input. Previously, these would surface as gRPC `INTERNAL` errors (server-side fault); now they correctly map to `BAD_REQUEST` / `INVALID_ARGUMENT` (client-side fault).

**Verdict: LGTM** — clean, focused defensive programming fix with good test coverage.

### Findings
- **[Info]** `proxy/.../ClientActivity.java:462-470` — The `parseLanguage` method uses `StringUtils.isBlank()` for the empty check, which is good. Consider also trimming the input before `valueOf()` to handle whitespace-only values (e.g., `"  "`) consistently — `isBlank` catches this, but the error message `"unsupported language:   "` could be slightly confusing. Minor, not blocking.

---

RockteMQ-AI (approved):
### Summary
Centralizes gRPC language metadata parsing and rejects invalid values. Prevents INTERNAL errors from malformed input.

LGTM.

---

Aias00:
Rebased onto current `develop` and validated `ClientActivityTest` with the proxy module. The prior macOS compile failure will be retried by the refreshed branch.

proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
```diff
@@ -146,7 +146,7 @@ public CompletableFuture<NotifyClientTerminationResponse> notifyClientTerminatio
 
         try {
             String clientId = ctx.getClientID();
-            LanguageCode languageCode = LanguageCode.valueOf(ctx.getLanguage());
```
lwclover, line 149: Could you clarify when `ctx.getLanguage()` can be empty or contain an invalid value?

Since `LanguageCode.valueOf(ctx.getLanguage())` will throw an exception if the value is `null`, empty, or not a valid enum constant.

Aias00: x-mq-language is client-controlled metadata. ContextInitPipeline#getDefaultStringMetadataInfo converts a missing header to an empty string, while an arbitrary value is copied unchanged into ProxyContext. Both
values later reach LanguageCode.valueOf

PR #10902, opened 2026-08-11
fix: validate every Lite consumer subscription against its bound topic [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #10895: [Bug] Validate every Lite consumer subscription against its bound topic
## Version

Reproduced on develop at `e3458616d`.

## Problem

`ClientProcessor.validateLiteSubTopic` validates only `subList.iterator().next()`. A Lite consumer group with a configured `liteBindTopic` can include another subscription for a different topic as long as the first set element matches.

## Reproduction

1. Configure a consumer group with `liteBindTopic=topic1`.
2. Build a subscription set whose first element targets `topic1` and whose second element targets `topic2`.
3. Call `validateLiteSubTopic`.
4. Observe that no `GrpcProxyException` is thrown.

## Expected

Every subscription in the set must match the configured bound topic; any mismatch must be rejected with `ILLEGAL_TOPIC`.

## Regression coverage

`ClientProcessorTest` contains a focused failing test with one matching and one mismatching subscription.

waterWang:
## What

`ClientProcessor.validateLiteSubTopic` now validates every subscription in the set against the group's bound topic, not just the first element.

## Why

`ClientProcessor.validateLiteSubTopic` called `subList.iterator().next().getTopic()`, which only checks the first element of the subscription set. A Lite consumer group with a configured `liteBindTopic` could include a second subscription for a different topic without being rejected.

## How

- Changed the single-element check to a `for` loop over all elements
- Added `testValidateLiteSubTopic_mismatchedSecondTopic_throwsException` to verify that a mismatching second subscription triggers `ILLEGAL_TOPIC`

Fixes #10895

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
@@ -197,8 +197,10 @@ public class ClientProcessor extends AbstractProcessor {
         if (CollectionUtils.isEmpty(subList)) {
             return;
         }
-        // check bindTopic for sub list
-        validateLiteBindTopic(ctx, group, subList.iterator().next().getTopic());
+        // check bindTopic for every subscription in the sub list
+        for (SubscriptionData subscriptionData : subList) {
+            validateLiteBindTopic(ctx, group, subscriptionData.getTopic());
+        }
     }
 
     protected void validateLiteBindTopic(ProxyContext ctx, String group, String bindTopic) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a validation bypass in `ClientProcessor.validateLiteSubTopic` where only the first subscription in the set was checked against the bound topic. The change correctly iterates over all subscriptions and adds a test case for the second-element mismatch scenario.

Clean, focused fix with proper test coverage. LGTM.

---

PR #10907, opened 2026-08-11
[ISSUE #10906] Prevent duplicate MQClientInstance creation

### LINKED ISSUE #10906: [Bug] Concurrent MQClientInstance creation leaks scheduler threads and periodic tasks

- [x] This is a bug rather than a support question.
- [x] I searched existing GitHub issues and discussions and did not find a duplicate.
- [x] The bug is in this repository.

- macOS 15.1.1 (24B91), arm64
- Reproduced by a deterministic JUnit test; no broker is required.

### RocketMQ version

- Branch: `develop`
- Commit: `e3458616d207ee636b1762f0f8dcf788a590d59d`
- Version in the reactor: 5.5.0
- This commit was the tip of `apache/develop` when the reproduction was run.

### JDK Version

- Amazon Corretto 11.0.23
- Also verified with Oracle JDK 8u291.

### Describe the Bug

The first concurrent calls to `MQClientManager.getOrCreateMQClientInstance` for the same client ID can construct multiple `MQClientInstance` candidates even though only one is stored and returned.

The current sequence is:

```text
factoryTable.get(clientId)
new MQClientInstance(...)
factoryTable.putIfAbsent(clientId, candidate)
```

All callers that observe the initial miss construct independently. The losing candidates are discarded without cleanup.

This is a resource leak, not only an allocation race. During construction, every candidate creates a `ConsumerStatsManager`. It creates five `StatsItemSet` objects, and every `StatsItemSet` registers six fixed-rate tasks on the instance's single-thread `MQClientFactoryScheduledThread` executor. The zero-delay sampling tasks start that non-daemon thread before `MQClientInstance.start()`.

The surviving reference chain for a losing candidate is:

```text
live Thread
  -> ThreadPoolExecutor.Worker
  -> ScheduledThreadPoolExecutor
  -> DelayedWorkQueue
  -> periodic ScheduledFutureTask
  -> StatsItemSet task
```

The tasks do not need to retain the outer `MQClientInstance`. Therefore the losing `MQClientInstance` itself can be garbage-collected while its scheduler thread and 30 periodic tasks remain live indefinitely.

There is a related failure-path leak. Construction schedules all 30 stats tasks before creating the optional concurrent-heartbeat pool. Enabling concurrent heartbeat with pool size 0 makes `Executors.newFixedThreadPool(0, ...)` throw `IllegalArgumentException`; no map entry is installed, but the already-started stats scheduler remains live.

### Steps to Reproduce

1. Check out `e3458616d207ee636b1762f0f8dcf788a590d59d`.
2. Start 8 callers for one client ID.
3. Use a barrier in a test `ClientConfig.cloneClientConfig()` so every caller has completed the initial map miss before any candidate can finish construction.
4. Capture each constructed candidate with a test-only observer.
5. Record returned identities, the candidate schedulers and their queue sizes.
6. Drop all strong references to losing `MQClientInstance` objects, retain only their schedulers, and use `WeakReference` plus bounded GC polling.
7. Use bounded polling for thread counts and `awaitTermination`; no fixed sleep is required.

The test was run with:

```bash
mvn -pl client -am -DskipITs -Dspotbugs.skip=true \
  -Dtest=MQClientManagerLeakReproductionTest \
```

Repeatable evidence from the unmodified code:

```text
candidates=8
returnedIdentity=1
liveSchedulerThreads=8
periodicTasks=240
losersGc=7/7
```

After explicitly calling `shutdownNow()` on all captured schedulers and waiting for termination, the scheduler-thread count returns to the baseline.

For one client ID, concurrent first access should perform at most one successful `MQClientInstance` construction and publish that instance to all callers. Failed construction or failed initialization should not leave a map entry, live executor thread, periodic task, or partially started client resource.

Creation for different client IDs should not be guarded by a global lock.

### What Did You See Instead?

Eight concurrent callers constructed eight candidates. They all returned one winner, but all eight scheduler threads and all 240 periodic tasks remained live. Seven losing outer instances were garbage-collected, proving that outer-object reachability and background-resource reachability are different.

A late constructor failure also left the scheduler thread and its 30 periodic tasks live even though `factoryTable` had no entry for the client ID.

### Additional Context

A constructor-path audit found these resource-bearing components:

- `MQClientInstance.scheduledExecutorService`: its thread starts during `ConsumerStatsManager` construction because stats tasks have zero initial delay.
- `ConsumerStatsManager`: five `StatsItemSet` instances, six periodic tasks each.
- `MQClientAPIImpl -> NettyRemotingClient`: allocates a timer, event-loop group, selector, and executors; most threads start lazily, but partial construction still needs rollback.
- `PullMessageService`: allocates a scheduled executor and retains the client instance; it starts threads/tasks only on start or submission.
- `RebalanceService`: retains the client instance; its service thread starts in `MQClientInstance.start()`.
- Inner `DefaultMQProducer` and the optional concurrent-heartbeat executor: worker threads are lazy, but both require cleanup if initialization fails.
- `MQClientInstance.startScheduledTask()`, pull, rebalance, remoting, and the inner producer introduce additional resources during `start()`; a mid-start exception currently leaves the factory in `START_FAILED` without rollback.

A minimal fix should serialize construction per client ID, preserve parallel construction for independent IDs, remove a failed mapping, roll back resources on constructor/start failure, and use identity-aware removal so an old instance cannot delete a newer replacement.

qianye1001:

- Fixes #10906

### Brief Description

Concurrent first access to `MQClientManager` previously used `get -> construct -> putIfAbsent`. Callers racing on one client ID could therefore construct multiple `MQClientInstance` candidates. Every candidate starts `MQClientFactoryScheduledThread` during `ConsumerStatsManager` construction and registers 30 periodic `StatsItemSet` tasks, so discarded candidates leaked live scheduler threads and tasks.

The unmodified-code reproduction used 8 synchronized callers and observed:

```text
candidates=8
returnedIdentity=1
liveSchedulerThreads=8
periodicTasks=240
losersGc=7/7
```

The losing outer instances were garbage-collected while the scheduler threads and periodic tasks remained live, confirming the resource reference chain independently of `MQClientInstance` reachability.

This change:

- uses `ConcurrentHashMap.computeIfAbsent` so one client ID has at most one successful construction and all callers receive the same instance;
- clones `ClientConfig` before entering the mapping function, keeping overridable configuration code outside the map's per-bin reservation;
- does not add a global lock, so independent client IDs in different map bins can construct concurrently;
- rolls back the stats scheduler, optional heartbeat executor, and remoting client if construction fails after resources have been allocated;
- rolls back remoting, scheduled tasks, pull, rebalance, inner-producer, and heartbeat resources if `start()` fails, removes the failed mapping, and preserves the existing `START_FAILED` state;
- adds identity-aware map removal so a stale instance's delayed shutdown cannot remove a newer replacement.

#### Design trade-offs

- A global synchronized section would unnecessarily serialize every client ID.
- A removable per-key lock map has waiter/lifecycle and ABA hazards; retaining locks forever would introduce a different leak.
- Moving all constructor resources to `start()` would be a much larger lifecycle and compatibility change, and by itself would not prevent duplicate candidates.
- `computeIfAbsent` provides the required atomic publication and exception behavior with the existing Java 8 `ConcurrentHashMap`: an exception installs no mapping, and a later call may retry.

The fast-path `get` and existing public removal method remain intact. Normal repeated `start()`/`shutdown()` behavior and the `START_FAILED` retry rejection on the same object are preserved.

#### Resource audit

The constructor/start path was audited for thread, scheduled-task, and retained-resource creation:

- `ConsumerStatsManager`: five `StatsItemSet` objects with six periodic tasks each; zero-delay sampling starts `MQClientFactoryScheduledThread` during construction.
- `MQClientAPIImpl -> NettyRemotingClient`: timer, event-loop/selector, public and scan executors; most threads are lazy, but the remoting client is shut down on partial construction/start failure.
- `PullMessageService`: service thread plus its scheduled executor; both are stopped on start failure.
- `RebalanceService`: service thread stopped on start failure.
- Inner `DefaultMQProducer`: registration/executors and request-future/detector lifecycle are shut down when started.
- Optional concurrent-heartbeat executor: shut down on construction or start failure.
- `MQClientInstance.startScheduledTask()`: all client periodic tasks use the scheduler that is shut down on failure.

All commands used the repository-configured Maven source/target 1.8; no compiler target was changed.

1. Full client reactor regression on Amazon Corretto 11.0.23:

   ```bash
   mvn -pl client -am -DskipITs -Dspotbugs.skip=true test
   ```

   Results:

   - `rocketmq-common`: 243 tests, 0 failures/errors
   - `rocketmq-remoting`: 174 tests, 0 failures/errors
   - `rocketmq-client`: 585 tests, 0 failures/errors, 1 skipped
   - Reactor: BUILD SUCCESS (3:45)

2. Focused regression plus SpotBugs/checkstyle on Amazon Corretto 11.0.23:

   ```bash
     -Dtest=MQClientManagerTest \
   ```

   Results: 7 tests passed; SpotBugs reported 0 findings and checkstyle reported 0 violations.

3. Java 8 compatibility on Oracle JDK 8u291:

   ```bash
   mvn -pl client -am -DskipITs -Dspotbugs.skip=true \
     -Dtest=MQClientManagerTest \
   ```

   Results: 7 tests passed. `javap -verbose` reports class-file major version 52.

The new tests use barriers, bounded futures, Awaitility assertions, executor termination waits, and isolated thread groups rather than fixed sleeps. They cover:

- high-concurrency same-client-ID creation: one instance, one scheduler, no orphan candidates;
- concurrent construction for different client IDs;
- late constructor failure: no map entry or scheduler thread, followed by successful retry;
- removal and recreation;
- repeated start/shutdown and recreation;
- mid-start failure rollback and replacement creation;
- stale shutdown preserving a newer replacement.

#### Compatibility / release note

Before this change, an `MQClientInstance` that failed during `start()` remained in `MQClientManager`, so later lookups for the same client ID returned the same `START_FAILED` instance. After this change, the failed instance is removed by identity after best-effort cleanup, and a later `getOrCreateMQClientInstance` call may create a replacement.

Callers that already hold the failed instance still observe `START_FAILED` and are not migrated automatically. If a persistent configuration or environment error is retried without backoff, callers may repeatedly create and clean up replacement instances; retry and backoff policy remains the caller responsibility.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientManager.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientManager.java
@@ -52,10 +52,12 @@ public class MQClientManager {
         MQClientInstance instance = this.factoryTable.get(clientId);
         if (null == instance) {
             ClientConfig clonedClientConfig = clientConfig.cloneClientConfig();
+            // MQClientInstance construction must not call back into factoryTable. ConcurrentHashMap rejects
+            // recursive updates from a mapping function with IllegalStateException.
             instance = this.factoryTable.computeIfAbsent(clientId, key -> {
                 MQClientInstance newInstance = new MQClientInstance(clonedClientConfig,
                     this.factoryIndexGenerator.getAndIncrement(), key, rpcHook);
-                log.info("Created new MQClientInstance for clientId:[{}]", clientId);
+                log.info("Created new MQClientInstance for clientId:[{}]", key);
                 return newInstance;
             });
         }
@@ -79,6 +81,10 @@ public class MQClientManager {
         return accumulator;
     }
 
+    /**
+     * Removes the mapped factory without checking its identity. Lifecycle cleanup should prefer
+     * {@link #removeClientFactory(String, MQClientInstance)} to avoid removing a replacement instance.
+     */
     public void removeClientFactory(final String clientId) {
         this.factoryTable.remove(clientId);
     }
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientManager.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientManager.java
@@ -51,16 +51,13 @@ public class MQClientManager {
         String clientId = clientConfig.buildMQClientId();
         MQClientInstance instance = this.factoryTable.get(clientId);
         if (null == instance) {
-            instance =
-                new MQClientInstance(clientConfig.cloneClientConfig(),
-                    this.factoryIndexGenerator.getAndIncrement(), clientId, rpcHook);
-            MQClientInstance prev = this.factoryTable.putIfAbsent(clientId, instance);
-            if (prev != null) {
-                instance = prev;
-                log.warn("Returned Previous MQClientInstance for clientId:[{}]", clientId);
-            } else {
+            ClientConfig clonedClientConfig = clientConfig.cloneClientConfig();
+            instance = this.factoryTable.computeIfAbsent(clientId, key -> {
+                MQClientInstance newInstance = new MQClientInstance(clonedClientConfig,
+                    this.factoryIndexGenerator.getAndIncrement(), key, rpcHook);
                 log.info("Created new MQClientInstance for clientId:[{}]", clientId);
-            }
+                return newInstance;
+            });
         }
 
         return instance;
@@ -84,6 +81,10 @@ public class MQClientManager {
 
     public void removeClientFactory(final String clientId) {
         this.factoryTable.remove(clientId);
+    }
+
+    public void removeClientFactory(final String clientId, final MQClientInstance instance) {
+        this.factoryTable.remove(clientId, instance);
     }
 
     public ConcurrentMap<String, MQClientInstance> getFactoryTable() {
--- a/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java
@@ -148,86 +148,113 @@ public class MQClientInstance {
     }
 
     public MQClientInstance(ClientConfig clientConfig, int instanceIndex, String clientId, RPCHook rpcHook) {
-        this.clientConfig = clientConfig;
-        this.nettyClientConfig = new NettyClientConfig();
-        this.nettyClientConfig.setClientCallbackExecutorThreads(clientConfig.getClientCallbackExecutorThreads());
-        this.nettyClientConfig.setUseTLS(clientConfig.isUseTLS());
-        this.nettyClientConfig.setSocksProxyConfig(clientConfig.getSocksProxyConfig());
-        this.nettyClientConfig.setScanAvailableNameSrv(false);
-        ClientRemotingProcessor clientRemotingProcessor = new ClientRemotingProcessor(this);
-        ChannelEventListener channelEventListener;
-        if (clientConfig.isEnableHeartbeatChannelEventListener()) {
-            channelEventListener = new ChannelEventListener() {
-                
-                private final ConcurrentMap<String, HashMap<Long, String>> brokerAddrTable = MQClientInstance.this.brokerAddrTable;
-                
-                @Override
-                public void onChannelConnect(String remoteAddr, Channel channel) {
-                }
+        MQClientAPIImpl clientAPI = null;
+        try {
+            this.clientConfig = clientConfig;
+            this.nettyClientConfig = new NettyClientConfig();
+            this.nettyClientConfig.setClientCallbackExecutorThreads(clientConfig.getClientCallbackExecutorThreads());
+            this.nettyClientConfig.setUseTLS(clientConfig.isUseTLS());
+            this.nettyClientConfig.setSocksProxyConfig(clientConfig.getSocksProxyConfig());
+            this.nettyClientConfig.setScanAvailableNameSrv(false);
+            ClientRemotingProcessor clientRemotingProcessor = new ClientRemotingProcessor(this);
+            ChannelEventListener channelEventListener;
+            if (clientConfig.isEnableHeartbeatChannelEventListener()) {
+                channelEventListener = new ChannelEventListener() {
 
-                @Override
-                public void onChannelClose(String remoteAddr, Channel channel) {
-                }
+                    private final ConcurrentMap<String, HashMap<Long, String>> brokerAddrTable = MQClientInstance.this.brokerAddrTable;
 
-                @Override
-                public void onChannelException(String remoteAddr, Channel channel) {
-                }
+                    @Override
+                    public void onChannelConnect(String remoteAddr, Channel channel) {
+                    }
 
-                @Override
-                public void onChannelIdle(String remoteAddr, Channel channel) {
-                }
+                    @Override
+                    public void onChannelClose(String remoteAddr, Channel channel) {
+                    }
 
-                @Override
-                public void onChannelActive(String remoteAddr, Channel channel) {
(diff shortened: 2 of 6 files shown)


RockteMQ-AI (approved):
- Identity-aware removal (`ConcurrentMap.remove(key, value)`) in `shutdown()` and `cleanupAfterStartFailure()` correctly prevents ABA issues
- `cleanupAfterConstructionFailure` properly handles the case where `clientAPI` is null (failure before `MQClientAPIImpl` creation)
- `runCleanup` correctly suppresses secondary exceptions without duplicating the primary cause

**Performance** ✅
- No global lock added — independent client IDs in different `ConcurrentHashMap` bins can still construct concurrently
- The bin lock is held during construction, but this is acceptable since construction is infrequent relative to lookups (the fast-path `get()` is lock-free)

**Tests** ✅
- 433 lines of comprehensive tests covering concurrent creation, resource cleanup on failure, identity-aware removal, and GC behavior
- Tests verify the exact bug scenario (8 concurrent callers, only 1 instance created, no leaked threads)

**Compatibility** ✅
- Existing API preserved — the new `removeClientFactory(String, MQClientInstance)` is additive
- No behavioral change for single-threaded callers

### Minor Note
The `factoryIndexGenerator.getAndIncrement()` inside the lambda is only consumed on actual creation, which is correct. If a construction fails and `computeIfAbsent` is retried, the next call will get a new index — this is fine since failed instances are discarded.

LGTM. Well-designed fix with thorough cleanup and excellent test coverage.

---

qianye1001:
CI follow-up

The original JDK 8 integration failure was unrelated to the MQClientManager production change, but it exposed two real timing assumptions in existing integration tests. Commit `52184f84f1` stabilizes them without weakening their assertions:

- `QueryMsgByKeyIT` now waits for the asynchronously built message index and retries only the expected `MQClientException` that reports no indexed message yet.
- `BatchAckIT` now waits until all sent messages are visible in the consume queues before the first orderly POP. This prevents a partially dispatched queue from being order-locked after returning only part of the batch. Its POP assertion also allows multiple 3-second long-poll attempts on a busy runner.

A later macOS/JDK 8 Build run exposed a separate flaky `ServiceThreadTest`: it classified any wakeup taking 18-20 ms as lost, so one OS scheduling delay failed the test. Commit `1385ac7975` replaces that performance threshold with a barrier-based protocol. One waiter and four wakers coordinate every iteration; all 1,000 notified waits must complete, while an actually lost wakeup breaks the 20-second coordination bound.

Validation on the final head `1385ac7975`:

- Build with Maven / macOS JDK 8: passed in 36m14s
- Build with Maven / Ubuntu JDK 8: passed in 40m35s
- Build with Maven / Windows JDK 8: passed in 44m27s
- Integration Tests / Ubuntu JDK 8: passed in 25m33s
- Coverage, CodeQL, Bazel, License, and Misspell: passed

Local Oracle JDK 8 validation also passed `MQClientManagerTest` (7/7), `QueryMsgByKeyIT` (4/4), the formerly failing orderly `BatchAckIT` path (1/1), and `ServiceThreadTest` twice (7/7 each run).

RockteMQ-AI (commented):
### Summary
This PR modifies 6 file(s) with 961 lines of changes.

### Review Notes
- **Files changed:** client/src/main/java/org/apache/rocketmq/client/impl/MQClientManager.java,client/src/main/java/org/apache/rocketmq/client/impl/factory/MQClientInstance.java,client/src/test/java/org/apache/rocketmq/client/impl/MQClientManagerTest.java,common/src/test/java/org/apache/rocketmq/common/ServiceThreadTest.java,test/src/test/java/org/apache/rocketmq/test/client/consumer/pop/BatchAckIT.java ... and 1 more
- **Diff size:** 961 lines
- **CLA status:** unknown
- ✅ Tests included

### Observations
- Potential null safety: 3 chained .get() calls without null checks

### Verdict
Code changes look reasonable. No critical issues detected in the structural review.

---

lizhimins (approved):

PR #10909, opened 2026-08-11
[ISSUE #10898] Wake NettyEventExecutor during shutdown

### LINKED ISSUE #10898: [Enhancement] NettyEventExecutor shutdown waits for the eventQueue poll timeout because wakeup() cannot release it

### Summary

```
```

### Motivation

| | before | after |
|---|---|---|

### Additional Context

askyln:

- Fixes #10898

### Brief Description

NettyEventExecutor.wakeup() now offers a private sentinel event after delegating to ServiceThread.wakeup(). This releases a worker blocked in the event queue's three-second poll so shutdown can observe the stopped flag immediately. The dispatch loop skips the sentinel by identity, leaving normal channel events unchanged.

A regression test waits until the executor is blocked in the timed queue poll and verifies that shutdown completes in under one second.

- Maven focused test: NettyRemotingAbstractTest (6 tests passed)
- Checkstyle passed with 0 violations.
- SpotBugs passed with 0 issues.
- git diff --check passed.
- The full common + remoting suite was attempted on Windows with JDK 17, but did not complete because the existing RemotingServerTest#testInvokeAsync waits indefinitely in CountDownLatch.await(). The focused test class and unrelated compatibility test both pass.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -763,6 +763,7 @@ public abstract class NettyRemotingAbstract {
     }
 
     class NettyEventExecutor extends ServiceThread {
+        private final NettyEvent wakeupEvent = new NettyEvent(null, null, null);
         private final LinkedBlockingQueue<NettyEvent> eventQueue = new LinkedBlockingQueue<>();
 
         public void putNettyEvent(final NettyEvent event) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a shutdown latency issue in `NettyEventExecutor` where the thread could block for up to 3 seconds in the timed `poll()` before observing the stopped flag. The fix introduces a private sentinel event offered to the queue on `wakeup()`, allowing immediate shutdown. Clean implementation with proper thread safety and a solid regression test.

**Verdict:** LGTM — approved.

---

PR #10914, opened 2026-08-12
[ISSUE #10912] Close RocksDB Slice objects in scanExpiredRecords

123123213weqw:
## What is the purpose of the change

Fix #10912.

scanExpiredRecords() created two RocksDB Slice objects as iterator bounds and never closed them, leaking native memory on every POP revive scan. The slices are now created as named resources inside the try-with-resources block so they are closed together with the ReadOptions and iterator.

## Brief changelog

- PopConsumerRocksdbStore.scanExpiredRecords(): move the two Slice bounds into try-with-resources

## How was this patch verified

- RocksDB Slice implements AutoCloseable via AbstractNativeReference, so the try-with-resources block closes it deterministically alongside ReadOptions and RocksIterator
- git diff --check clean

PR #10915, opened 2026-08-12
[ISSUE #10898] Wake NettyEventExecutor immediately on shutdown

### LINKED ISSUE #10898: [Enhancement] NettyEventExecutor shutdown waits for the eventQueue poll timeout because wakeup() cannot release it

123123213weqw:

Fix #10898.

NettyEventExecutor blocks in eventQueue.poll(3000, MILLISECONDS), but ServiceThread.shutdown() only signals the stopped flag and calls wakeup(), which unparks a parked thread and cannot release a thread blocked on a LinkedBlockingQueue. Every shutdown of a remoting instance therefore waited up to 3 seconds per NettyEventExecutor.

## Brief changelog

- NettyRemotingAbstract.NettyEventExecutor now overrides shutdown(boolean): it offers a sentinel event (null type) into the queue so the blocked poll returns immediately, then delegates to the superclass shutdown
- the run loop skips events with a null type, so the sentinel is consumed without touching the channel listener

- Code review of the poll loop: the sentinel event makes poll return at once, the loop re-checks isStopped() and exits promptly
- `git diff --check` clean

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
@@ -147,9 +147,11 @@ public class PopConsumerRocksdbStore extends AbstractRocksDBStorage implements P
         // configure prefix indexing to improve the performance of scans.
         // However, in the current implementation, this is not the bottleneck.
         List<PopConsumerRecord> consumerRecordList = new ArrayList<>();
-        try (ReadOptions scanOptions = new ReadOptions()
-            .setIterateLowerBound(new Slice(ByteBuffer.allocate(Long.BYTES).putLong(lower).array()))
-            .setIterateUpperBound(new Slice(ByteBuffer.allocate(Long.BYTES).putLong(upper).array()));
+        try (Slice lowerBound = new Slice(ByteBuffer.allocate(Long.BYTES).putLong(lower).array());
+             Slice upperBound = new Slice(ByteBuffer.allocate(Long.BYTES).putLong(upper).array());
+             ReadOptions scanOptions = new ReadOptions()
+                 .setIterateLowerBound(lowerBound)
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds two improvements to `NettyRemotingAbstract`:

1. **Null check for `event.getType()`** — defensive programming to prevent NPE if a null event type is somehow queued.
2. **`shutdown()` override** — wakes up the event queue poll immediately by offering a sentinel `NettyEvent`, so the executor observes the stopped flag without waiting for the 3-second poll timeout.

Both changes are correct and improve shutdown responsiveness. LGTM.

**Note:** This PR also includes the `PopConsumerRocksdbStore` fix from #10914. Consider coordinating with maintainers to avoid merge conflicts with the other PRs in this series (#10914-#10923).

---

RockteMQ-AI (commented):
### Review Summary

This PR fixes slow shutdown in `NettyRemotingAbstract` by waking the `NettyEventExecutor` immediately when shutdown is requested.

### Findings
- **[Critical]** NettyRemotingAbstract.java — The original code would wait for the polling thread to wake up naturally, causing slow shutdown. The fix overrides `shutdown` to offer a sentinel event to the queue, waking the thread immediately.
- **[Info]** The null check for event type is also a good defensive addition.

### Overall
Good fix for shutdown latency. This is especially important for graceful shutdown scenarios where timely cleanup is critical.

---

RockteMQ-AI (approved):
### Summary
Two improvements to NettyEventExecutor:
1. Null-check on event.getType() prevents NPE from the sentinel event
2. Override shutdown() to offer a sentinel event, waking the thread blocked on eventQueue.poll()

This ensures shutdown is responsive instead of waiting up to 3 seconds for the poll timeout. LGTM.

---

PR #10916, opened 2026-08-12
[ISSUE #10904] Guard malformed retry counters in proxy send path

123123213weqw:
## What is the purpose of the change

Fix #10904.

ProducerProcessor.buildSendMessageRequestHeader parsed retry-message counters with Integer.valueOf, so a malformed PROPERTY_RECONSUME_TIME or PROPERTY_MAX_RECONSUME_TIMES value threw NumberFormatException and aborted the Proxy send/retry path. The method already tolerated a malformed born timestamp by falling back to the current time; the retry counters now get the same defensive treatment.

## Brief changelog

- ProducerProcessor: parse retry counters with Integer.parseInt on a trimmed value inside try/catch, falling back to 0 and logging a warning on malformed input
- valid counters keep their previous behavior

## How was this patch verified

- Code review: mirrors the existing bornTimestamp fallback pattern
- `git diff --check` clean

PR #10917, opened 2026-08-12
[ISSUE #10901] Fix OTLP header parsing and avoid leaking header config

### LINKED ISSUE #10901: [studio][Bug] Proxy OTLP header parsing can reject valid values and expose secrets

- [x] I searched existing issues and pull requests for metricsGrpcExporterHeader, OTLP header parsing, and exporter-header secret leakage.
- [x] I confirmed the affected code belongs to this repository.

### Describe the Bug

ProxyMetricsManager#start parses metricsGrpcExporterHeader entries with item.split(:).

This has two related problems:

1. A valid header value containing a colon (for example a URI or a structured authorization value) produces more than two segments and is discarded.
2. When parsing fails, the warning logs the complete header configuration. That string may contain authorization tokens, API keys, or other credentials.

The same broad split behavior is also used for metricsLabel, where a label value containing a colon is unnecessarily rejected.

### Expected Behavior

- Split each entry at the first colon only.
- Trim and require a non-empty key.
- Preserve the remainder of the value, including additional colons.
- Warn about an invalid entry without logging the complete configured labels or headers.
- Add focused unit tests for colon-containing values, empty keys, malformed entries, and secret-safe diagnostics.

### Affected Area

RocketMQ Proxy metrics initialization / OTLP exporter configuration. This is relevant to RocketMQ Studio deployments that use Proxy metrics for observability.

### Proposed Scope

Keep the change local to metrics key/value parsing and its tests. Do not change exporter defaults, protocols, or public APIs.

123123213weqw:

Fix #10901.

ProxyMetricsManager parsed metricsGrpcExporterHeader entries with item.split(":"), which drops any valid value containing a colon (e.g. a URI or a structured authorization value). On a parse failure the warning logged the complete header configuration, which may contain authorization tokens or API keys. The same split behavior was also applied to metricsLabel.

## Brief changelog

- ProxyMetricsManager: split each entry at the first colon with split(":", 2) so header values may contain colons
- warning logs now print only the offending entry instead of the full header configuration

- Code review: split(":", 2) keeps the rest of the value after the first colon
- `git diff --check` clean

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerRocksdbStore.java
@@ -147,9 +147,11 @@ public class PopConsumerRocksdbStore extends AbstractRocksDBStorage implements P
         // configure prefix indexing to improve the performance of scans.
         // However, in the current implementation, this is not the bottleneck.
         List<PopConsumerRecord> consumerRecordList = new ArrayList<>();
-        try (ReadOptions scanOptions = new ReadOptions()
-            .setIterateLowerBound(new Slice(ByteBuffer.allocate(Long.BYTES).putLong(lower).array()))
-            .setIterateUpperBound(new Slice(ByteBuffer.allocate(Long.BYTES).putLong(upper).array()));
+        try (Slice lowerBound = new Slice(ByteBuffer.allocate(Long.BYTES).putLong(lower).array());
+             Slice upperBound = new Slice(ByteBuffer.allocate(Long.BYTES).putLong(upper).array());
+             ReadOptions scanOptions = new ReadOptions()
+                 .setIterateLowerBound(lowerBound)
+                 .setIterateUpperBound(upperBound);
              RocksIterator iterator = db.newIterator(this.columnFamilyHandle, scanOptions)) {
             iterator.seek(ByteBuffer.allocate(Long.BYTES).putLong(lower).array());
             while (iterator.isValid() && consumerRecordList.size() < maxCount) {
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes OTLP header parsing in `ProxyMetricsManager`:

1. **`split(":", 2)`** — correctly handles label values that contain colons (e.g., URLs). The original `split(":")` would break on such values.
2. **Log message fix** — logs the specific invalid `item` instead of the entire `labels` string, making debugging easier.

Both changes are correct. LGTM.

**Note:** This PR also includes changes from #10914-#10916. See the series overlap comment on those PRs.

---

RockteMQ-AI (commented):
### Review Summary

This PR fixes OTLP header parsing in `ProxyMetricsManager` to handle values containing colons and improves error logging.

### Findings
- **[Critical]** ProxyMetricsManager.java — The original `split(":")` would break on header values containing colons (e.g., `Authorization: Bearer token:with:colons`). The fix uses `split(":", 2)` to split only on the first colon.
- **[Info]** The improved error log now shows the actual invalid item instead of the full header string, making debugging easier.

### Overall
Good fix for OTLP header parsing. The original code would fail on valid header values containing colons.

---

RockteMQ-AI (approved):
### Summary
Two fixes in ProxyMetricsManager:
1. split with limit 2 correctly handles header values containing colons (e.g., Authorization: Bearer xxx)
2. Log messages now show the specific invalid item instead of the entire config string — this also prevents leaking other header values in logs

Both fixes are correct and the security improvement in logging is appreciated. LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes OTLP header parsing and prevents leaking header configuration in logs. Similar fixes to #10918 but focused on OTLP header handling.

LGTM — good security fix.

---

PR #10932, opened 2026-08-15
[ISSUE #10931] Initialize cluster transaction address mapping

### LINKED ISSUE #10931: [Bug] Cluster transaction service throws before broker address mapping is initialized
## Problem

`ClusterTransactionService` initializes `brokerAddrNameMapRef` without a map. Before the first successful transaction-heartbeat scan, `getBrokerNameByAddr` dereferences `brokerAddrNameMapRef.get()` directly.

`AbstractTransactionService.addTransactionDataByBrokerAddr` calls this method on the transaction send path, so an otherwise valid transaction can fail with `NullPointerException` during proxy startup or before route discovery has completed.

## Expected behavior

An unresolved broker address should follow the existing `AbstractTransactionService.addTransactionDataByBrokerName` behavior: return `null` and do not record transaction data.

## Proposed fix

Initialize the address-to-broker-name mapping with an empty map (or null-check it) and add a regression test covering `addTransactionDataByBrokerAddr` before the first heartbeat scan.

## Scope

Proxy cluster transaction service. This does not change transaction protocol behavior after route discovery.

Aias00:
## Summary

Initialize the cluster transaction address-to-broker mapping with an empty map so a transaction sent before the first heartbeat scan does not dereference a null map.

## Validation

- `mvn -pl proxy -Dtest=ClusterTransactionServiceTest test`

Closes #10931

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/service/transaction/ClusterTransactionService.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/service/transaction/ClusterTransactionService.java
@@ -57,7 +57,8 @@ public class ClusterTransactionService extends AbstractTransactionService {
 
     private ThreadPoolExecutor heartbeatExecutors;
     private final Map<String /* group */, Set<ClusterData>/* cluster list */> groupClusterData = new ConcurrentHashMap<>();
-    private final AtomicReference<Map<String /* brokerAddr */, String /* brokerName */>> brokerAddrNameMapRef = new AtomicReference<>();
+    private final AtomicReference<Map<String /* brokerAddr */, String /* brokerName */>> brokerAddrNameMapRef =
+        new AtomicReference<>(Collections.emptyMap());
     private TxHeartbeatServiceThread txHeartbeatServiceThread;
 
     public ClusterTransactionService(TopicRouteService topicRouteService, ProducerManager producerManager,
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM. This is a clean defensive fix that prevents NPE when accessing the broker address map before the first heartbeat scan.

**Key change:**
- Initialize `brokerAddrNameMapRef` with `Collections.emptyMap()` instead of null
- Thread-safe immutable empty map is appropriate for `AtomicReference`

**Test coverage:**
- New test `testAddTransactionDataByBrokerAddrBeforeFirstHeartbeat` validates the edge case
- Verifies that accessing the map before heartbeat scan returns null gracefully instead of throwing NPE

Minimal, focused fix with proper validation.

---

RockteMQ-AI (approved):
### Summary
Initializes `brokerAddrNameMapRef` with `Collections.emptyMap()` instead of null, preventing `NullPointerException` when `addTransactionDataByBrokerAddr()` is called before the first heartbeat populates the mapping.

**Changes reviewed:**
- `ClusterTransactionService`: Changes `new AtomicReference<>()` (null initial value) to `new AtomicReference<>(Collections.emptyMap())`.
- Test `testAddTransactionDataByBrokerAddrBeforeFirstHeartbeat`: Verifies that calling `addTransactionDataByBrokerAddr` before the first heartbeat returns null (no crash).

**Assessment:**
- ✅ Correctness: Simple but effective fix — empty map is a safe default that prevents NPE.
- ✅ Test: Covers the exact scenario (pre-heartbeat call).
- ✅ Backward compatible: No behavioral change after heartbeat populates the map.

LGTM — clean defensive initialization.

---

PR #10934, opened 2026-08-15
[ISSUE #10933] Retain receipt handles when offline cleanup cannot complete

### LINKED ISSUE #10933: [Bug] Proxy drops receipt handles when offline cleanup renewal fails
## Problem

When a consumer channel goes offline, `DefaultReceiptHandleManager.clearGroup` asynchronously renews each receipt handle with `CLEAR_GROUP`. The cleanup callback currently returns `CompletableFuture.completedFuture(null)` immediately instead of the `RenewEvent` future.

As a result, `ReceiptHandleGroup` removes the handle before the renewal outcome is known. If the renewal request fails, the handle is no longer retained or retried by the renewal scheduler.

## Expected behavior

A receipt handle should be removed only after its `CLEAR_GROUP` renewal succeeds. On failure it should remain in the group so subsequent cleanup scheduling can retry it.

## Proposed fix

Return the `RenewEvent` future from the `ReceiptHandleGroup` callback, mapping a successful result to `null` (remove) while preserving exceptional completion (retain). Add a regression test covering a failed offline cleanup renewal.

## Scope

Proxy POP receipt-handle lifecycle; no protocol/API changes.

### LINKED ISSUE #10941: [Bug] Proxy drops receipt handles when offline cleanup task submission is rejected
## Problem

`DefaultReceiptHandleManager.clearGroup` removes a receipt-handle group from `receiptHandleGroupMap` before submitting its cleanup task to the bounded `returnHandleGroupWorkerService`.

If the executor rejects the submission because its queue is full or it is shutting down, `submit` throws and the removed group is not restored. The Proxy then loses the handle state without sending the intended clear/renew request and cannot retry it on the next scheduled scan.

## Expected behavior

A rejected cleanup submission must retain or restore the group so a later scheduled pass can retry cleanup. The rejection should be logged without leaking handle contents.

## Scope

Handle `RejectedExecutionException` around `clearGroup` task submission and add a regression test that forces rejection.

Aias00:
## Summary

Keep POP receipt handles in the cleanup group whenever offline CLEAR_GROUP work cannot complete:

- retain groups when asynchronous renewal fails so the scheduler can retry;
- restore groups when the cleanup executor rejects task submission.

## Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 17) mvn -pl proxy -Dtest=DefaultReceiptHandleManagerTest test`

Closes #10933
Closes #10941

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/service/receipt/DefaultReceiptHandleManager.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/service/receipt/DefaultReceiptHandleManager.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -89,7 +90,8 @@ public class DefaultReceiptHandleManager extends AbstractStartAndShutdown implem
             proxyConfig.getReturnHandleGroupThreadPoolNums() * 2,
             1, TimeUnit.MINUTES,
             "ReturnHandleGroupWorkerThread",
-            proxyConfig.getRenewThreadPoolQueueCapacity()
+            proxyConfig.getRenewThreadPoolQueueCapacity(),
+            new ThreadPoolExecutor.AbortPolicy()
         );
         consumerManager.appendConsumerIdsChangeListener(new ConsumerIdsChangeListener() {
             @Override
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
LGTM. This fix correctly addresses the race condition where receipt handles were being removed before the async renewal completed.

**Key change:**
- Changed from fire-and-forget (`CompletableFuture.completedFuture(null)`) to waiting for the renewal result (`future.thenApply(ackResult -> null)`)
- This ensures failed renewals retain the handle for the scheduler retry path

**Test coverage:**
- New test `testClearGroupRetainsHandleWhenRenewalFails` validates the failure scenario
- Existing `testClearGroup` continues to cover the success path

Clean, focused fix with proper test validation.

---

RockteMQ-AI (approved):
### Summary
Re-review after new commits. The updated changes properly handle `RejectedExecutionException` when the return-handle worker thread pool is saturated — instead of silently losing receipt handles, the group is put back into the map for the next scheduling cycle. The async completion fix (`future.thenApply(ackResult -> null)` instead of `CompletableFuture.completedFuture(null)`) correctly chains the renew future so the handle is only cleared after the renew operation actually completes.

The test coverage for both scenarios (renewal failure and submission rejection) is solid.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes receipt handle loss during offline cleanup when the cleanup task cannot complete (thread pool rejection or renewal failure).

**Changes reviewed:**
- `DefaultReceiptHandleManager`: Explicitly sets `AbortPolicy` for the worker thread pool, catches `RejectedExecutionException` in `clearGroup()`, and restores the handle group to the map for retry on the next schedule.
- `returnHandleGroup()`: Changes `computeIfPresent` return from `CompletableFuture.completedFuture(null)` to `future.thenApply(ackResult -> null)` — the entry is only removed after the renewal future completes successfully. If renewal fails, the handle is retained.
- Tests cover both failure scenarios: renewal failure and thread pool rejection.

**Assessment:**
- ✅ Correctness: Prevents silent loss of receipt handles during cleanup failures — important for message delivery guarantees.
- ✅ Thread safety: `putIfAbsent` avoids race conditions.
- ✅ Tests: Good coverage of both failure paths.

LGTM — solid fix for a subtle data-loss scenario.

---

PR #10937, opened 2026-08-15
[ISSUE #10935] Fix shared produce accumulator lifecycle

### LINKED ISSUE #10935: [Bug] Shutting down one producer stops timeout flushing for producers with the same client ID

- [x] I searched GitHub Issues and Discussions and believe this is not a duplicate.

Reproduced on Ubuntu 22.04.4 LTS, Linux 5.15.0-186, x86_64. The defect is in the Java client lifecycle and is not expected to be OS-specific.

## RocketMQ version

- Branch: `develop`
- Version: `5.5.0` / current develop sources
- Commit: `293f5885719fc4aa3619446a1900f58ccfcfdd29`

## JDK Version

- OpenJDK 8u492

## Describe the Bug

Proposed severity: **Major availability impact**.

Multiple `DefaultMQProducer` instances with the same RocketMQ client ID share one `ProduceAccumulator` through `MQClientManager`. The client ID is based on client IP, `instanceName`, and unit name, and does not include the producer group.

However, each producer currently starts and shuts down that shared accumulator independently. The service threads ignore repeated starts, while the first producer shutdown stops both the synchronous and asynchronous batch guard threads even when another producer using the accumulator is still running.

After that shutdown:

- under low traffic, a small synchronous auto-batched message can wait indefinitely because it does not reach the size threshold and the timeout guard no longer wakes it;
- a small asynchronous auto-batched message can remain queued without a callback until another send reaches the size threshold.

The surviving producer remains in `RUNNING` state, so the failure is silent and gives the application no useful error to recover from. This is an availability bug, not a security report.

## Steps to Reproduce

With a reachable NameServer and Broker, create two auto-batching producers in different producer groups but give them the same explicit instance name. Ensure `TopicTest` already exists or topic auto-creation is enabled:

```java
DefaultMQProducer producerA = new DefaultMQProducer("group-a");
DefaultMQProducer producerB = new DefaultMQProducer("group-b");

producerA.setNamesrvAddr(namesrvAddr);
producerB.setNamesrvAddr(namesrvAddr);
producerA.setInstanceName("shared-client");
producerB.setInstanceName("shared-client");
producerA.setAutoBatch(true);
producerB.setAutoBatch(true);
producerA.batchMaxDelayMs(1000);
producerB.batchMaxDelayMs(1000);

producerA.start();
producerB.start();

producerA.shutdown();

// A small message is below the auto-batch size threshold. This call can
// block indefinitely because the shared timeout guard was stopped by A.
producerB.send(new Message("TopicTest", new byte[] {1}));
```

The same defect has a deterministic broker-free unit reproduction:

1. Create one `ProduceAccumulator` with a short `batchMaxDelayMs`.
2. Call `start()` twice to represent two producers sharing the same client ID.
3. Call `shutdown()` once to release only the first producer.
4. Add one small asynchronous message through `MockMQProducer`.
5. Wait longer than `batchMaxDelayMs` for either callback.

On the unmodified baseline, the callback deadline expires every time. Two independent runs of the final public-lifecycle regression failed with method times of 4.817 and 4.829 seconds:

```text
Expecting value to be true but was false
Tests run: 1, Failures: 1
```

Shutting down one producer should release only that producer's ownership of the shared accumulator. As long as another producer with the same client ID is still started, the accumulator guard threads should remain active and flush its messages after `batchMaxDelayMs`.

## What Did You See Instead?

The first producer shutdown stops the shared guard threads immediately. Under low traffic, a remaining producer can then block indefinitely in a synchronous send, or retain an asynchronous message without invoking its callback. A later same-key send that reaches the size threshold can still trigger a size-based flush; the broken behavior is timeout flushing.

## Additional Context

A minimal fix is to reference-count started producer owners in `ProduceAccumulator`: start its guards only on transition `0 -> 1` and stop them only on `1 -> 0`. Each `DefaultMQProducer` should retain and release at most once so repeated `shutdown()` calls cannot decrement another producer's ownership.

With that fix, the new focused suite passes 5/5 tests, and the broader `ProduceAccumulatorTest` plus `DefaultMQProducerTest` run passes 46/46 tests.

Temporary workarounds are to use a unique `instanceName` for each producer or to avoid shutting down any producer that shares the client ID until all such producers are ready to stop.

I searched for reports involving `ProduceAccumulator`, its sync/async guard threads, shared accumulator lifecycle, same `instanceName`, and producer shutdown, and found no duplicate. Related issue [#8806](https://github.com/apache/rocketmq/issues/8806) and its merged fix [#8807](https://github.com/apache/rocketmq/pull/8807) address accidental sharing caused by early initialization. They intentionally retain sharing for producers with the same instance/unit identity and do not manage ownership of the shared accumulator's lifecycle; this report concerns that distinct lifecycle problem.

ai-yang:

- Fixes #10935

### Brief Description

`MQClientManager` can return the same `ProduceAccumulator` to multiple producers that share a RocketMQ client ID. Previously, every `DefaultMQProducer.shutdown()` stopped that shared accumulator, even while another producer was running. Under low traffic, small synchronous auto-batched sends could then wait indefinitely, while asynchronous messages remained queued without timeout flushing.

Keeping the guard threads alive is necessary but not sufficient: a pending batch also retains the producer that created it. A second producer must not join that batch after the first producer shuts down.

This change:

- reference-counts started producer owners; starts the guards for the first owner and stops them after the last owner;
- tracks each producer's ownership with an `AtomicBoolean`, making repeated shutdown release idempotent;
- includes the concrete producer's identity in both synchronous and asynchronous aggregation keys, including explicit-queue sends; a replacement instance with the same group is still a different owner;
- releases held bytes exactly once after send completion or a synchronous send-initiation failure, including when a user failure callback throws;
- removes the exact completed batch from the maps so producer-specific keys do not retain completed batches or remove a newer replacement.

There is no public API or wire-format change.

### Applicability and tradeoffs

Automatic batching is disabled by default (`autoBatch=false`). The shared-lifecycle failure requires producers sharing an accumulator and one shutting down while another remains active; it is not a claim that ordinary non-batched sends are affected.

Producer identity intentionally prevents cross-producer coalescing, because producers may have different executors, hooks, credentials, and lifetimes. Batching within one producer is preserved. The aggregation/throughput tradeoff has not been benchmarked. This does not promise graceful draining of the producer that is shutting down, nor redesign user-callback exception/notification semantics.

Latest local verification (2026-09-13), head `f3901017f89509118835cd90b40109b3b51aa9c3`, JDK 8 / Maven 3.8.1:

```bash
  -Dtest=ProduceAccumulatorTest,DefaultMQProducerTest,ProduceAccumulatorLifecycleTest \
  -Dsurefire.failIfNoSpecifiedTests=false clean test
```

The local run used offline mode and settings pointing to the existing dependency cache; no project build configuration was changed and no Checkstyle/SpotBugs skip flags were used.

- `ProduceAccumulatorTest`: 5 passed, including real guard lifetime and repeated shutdown.
- `DefaultMQProducerTest`: 41 passed.
- `ProduceAccumulatorLifecycleTest`: 7 passed: stopped/running producer isolation, explicit queue, same-group replacement, completed-map cleanup, synchronous initiation failure, inline callback followed by throw, and throwing failure callback.
- Total: **53 tests, 0 failures/errors/skips**; all four reactor modules succeeded.
- Checkstyle and SpotBugs enabled; `git diff --check` passed.

The new ownership tests disable background guards and explicitly flush real batches, making batch ownership deterministic; the existing accumulator suite separately exercises guard lifecycle. Re-running these seven new cases against the previous PR's production classes (`ce2e9b171`) produced **6 failures / 7 tests**; the updated production classes passed all seven. This comparison is against the previous PR head, not a fresh full-suite run of current upstream `develop`.

The original guard-lifetime regression also failed on unmodified `develop` at `e348efa66` before the reference-count fix. The PR retains that base and adds a fast-forward follow-up commit; it was not rebased. CI for the new head must be checked independently of these local results.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java
@@ -54,6 +54,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * This class is the entry point for applications intending to send messages. </p>
@@ -162,6 +163,7 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
      * Instance for batching message automatically
      */
     private ProduceAccumulator produceAccumulator = null;
+    private final AtomicBoolean produceAccumulatorStarted = new AtomicBoolean(false);
 
     /**
      * Indicate whether to block message when asynchronous sending traffic is too heavy.
@@ -374,7 +376,7 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
     public void start() throws MQClientException {
         this.setProducerGroup(withNamespace(this.producerGroup));
         this.defaultMQProducerImpl.start();
-        if (this.produceAccumulator != null) {
+        if (this.produceAccumulator != null && this.produceAccumulatorStarted.compareAndSet(false, true)) {
             this.produceAccumulator.start();
         }
         if (enableTrace) {
@@ -411,7 +413,7 @@ public class DefaultMQProducer extends ClientConfig implements MQProducer {
     @Override
     public void shutdown() {
         this.defaultMQProducerImpl.shutdown();
-        if (this.produceAccumulator != null) {
+        if (this.produceAccumulator != null && this.produceAccumulatorStarted.compareAndSet(true, false)) {
             this.produceAccumulator.shutdown();
         }
         if (null != traceDispatcher) {
--- a/client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java
@@ -56,6 +56,8 @@ public class ProduceAccumulator {
     private final Map<AggregateKey, MessageAccumulation> asyncSendBatchs = new ConcurrentHashMap<AggregateKey, MessageAccumulation>();
     private final AtomicLong currentlyHoldSize = new AtomicLong(0);
     private final String instanceName;
+    // Number of started producers sharing this accumulator through the same client ID.
+    private int producerCount;
 
     public ProduceAccumulator(String instanceName) {
         this.instanceName = instanceName;
@@ -155,12 +157,17 @@ public class ProduceAccumulator {
         }
     }
 
-    void start() {
-        guardThreadForSyncSend.start();
-        guardThreadForAsyncSend.start();
+    synchronized void start() {
+        if (producerCount++ == 0) {
+            guardThreadForSyncSend.start();
+            guardThreadForAsyncSend.start();
+        }
     }
 
-    void shutdown() {
+    synchronized void shutdown() {
+        if (producerCount == 0 || --producerCount > 0) {
+            return;
+        }
         guardThreadForSyncSend.shutdown();
         guardThreadForAsyncSend.shutdown();
     }
--- a/client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java
+++ b/client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java
@@ -231,7 +231,7 @@ public class ProduceAccumulator {
 
     SendResult send(Message msg,
         DefaultMQProducer defaultMQProducer) throws InterruptedException, MQBrokerException, RemotingException, MQClientException {
-        AggregateKey partitionKey = new AggregateKey(msg);
+        AggregateKey partitionKey = new AggregateKey(msg, defaultMQProducer);
         while (true) {
             MessageAccumulation batch = getOrCreateSyncSendBatch(partitionKey, defaultMQProducer);
             int index = batch.add(msg);
@@ -245,7 +245,7 @@ public class ProduceAccumulator {
 
     SendResult send(Message msg, MessageQueue mq,
         DefaultMQProducer defaultMQProducer) throws InterruptedException, MQBrokerException, RemotingException, MQClientException {
-        AggregateKey partitionKey = new AggregateKey(msg, mq);
+        AggregateKey partitionKey = new AggregateKey(msg, mq, defaultMQProducer);
         while (true) {
             MessageAccumulation batch = getOrCreateSyncSendBatch(partitionKey, defaultMQProducer);
             int index = batch.add(msg);
@@ -259,7 +259,7 @@ public class ProduceAccumulator {
 
     void send(Message msg, SendCallback sendCallback,
         DefaultMQProducer defaultMQProducer) throws InterruptedException, RemotingException, MQClientException {
-        AggregateKey partitionKey = new AggregateKey(msg);
+        AggregateKey partitionKey = new AggregateKey(msg, defaultMQProducer);
         while (true) {
             MessageAccumulation batch = getOrCreateAsyncSendBatch(partitionKey, defaultMQProducer);
             if (!batch.add(msg, sendCallback)) {
@@ -273,7 +273,7 @@ public class ProduceAccumulator {
     void send(Message msg, MessageQueue mq,
         SendCallback sendCallback,
         DefaultMQProducer defaultMQProducer) throws InterruptedException, RemotingException, MQClientException {
-        AggregateKey partitionKey = new AggregateKey(msg, mq);
+        AggregateKey partitionKey = new AggregateKey(msg, mq, defaultMQProducer);
         while (true) {
             MessageAccumulation batch = getOrCreateAsyncSendBatch(partitionKey, defaultMQProducer);
             if (!batch.add(msg, sendCallback)) {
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a real bug where `MQClientManager` returns the same `ProduceAccumulator` to multiple producers sharing a client ID. Previously, shutting down one producer would stop the shared accumulator, breaking the other producer's batch sends. This PR adds reference counting in `ProduceAccumulator` and `AtomicBoolean` idempotency in `DefaultMQProducer`.

### Findings
- **[Info]** `client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java:160–170` — The dual protection (`AtomicBoolean` in `DefaultMQProducer` + `synchronized` + `producerCount` in `ProduceAccumulator`) is correct and serves distinct purposes: the `AtomicBoolean` prevents a single producer from calling start/shutdown multiple times, while `producerCount` handles multiple producers sharing the same accumulator. Well-designed.
- **[Info]** `client/src/test/java/org/apache/rocketmq/client/producer/ProduceAccumulatorTest.java:115–168` — The regression test `testSharedAccumulatorRemainsRunningUntilLastProducerShutdown` correctly reproduces the bug on the unmodified baseline (deterministic failure) and passes with the fix. Good use of `CountDownLatch` with timeout for async verification.

### Verdict
Clean bug fix with proper synchronization and comprehensive regression tests. The reference counting approach is the right solution for shared accumulator lifecycle management.

---

RockteMQ-AI (approved):
﻿### Summary
Fixes a lifecycle bug where multiple `DefaultMQProducer` instances sharing the same `ProduceAccumulator` (via same client ID) would cause `IllegalThreadStateException` or premature thread shutdown on `start()`/`shutdown()`.

### Findings
- **[Info]** `client/src/main/java/org/apache/rocketmq/client/producer/ProduceAccumulator.java` — Reference counting with `synchronized` is the correct pattern for shared guard threads. `start()` only launches threads on first producer; `shutdown()` only stops them when the last producer shuts down.
- **[Info]** `client/src/main/java/org/apache/rocketmq/client/producer/DefaultMQProducer.java` — `AtomicBoolean.compareAndSet()` ensures idempotent start/shutdown per producer instance, preventing double-start of the shared accumulator.
- **[Info]** Tests cover shared lifecycle and idempotent behavior — good coverage.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes the lifecycle management of shared `ProduceAccumulator` instances. When multiple `DefaultMQProducer` instances share the same accumulator (by client ID), the accumulator's background threads were incorrectly started/stopped by each producer independently.

**Changes reviewed:**
- `DefaultMQProducer`: Adds `AtomicBoolean produceAccumulatorStarted` to ensure `start()/shutdown()` on the accumulator are called exactly once per producer lifecycle.
- `ProduceAccumulator`: Adds `producerCount` with synchronized `start()/shutdown()` — reference counting ensures threads only start on first producer and stop when the last producer shuts down.
- Tests verify: shared accumulator remains running until last producer shuts down, and double-shutdown is idempotent.

**Assessment:**
- ✅ Correctness: Fixes a real bug where shared accumulator threads could be prematurely stopped or double-started.
- ✅ Thread safety: `synchronized` on accumulator start/shutdown, `AtomicBoolean` for producer-side guard.
- ✅ Tests: Good coverage of shared lifecycle and double-shutdown scenarios.
- ✅ Backward compatible: Single producer behavior is unchanged.

LGTM — correct fix for a subtle shared-state lifecycle bug.

---

ai-yang:
Update after refreshing this PR against the latest `develop`:

- rebased onto `e348efa66b08eb645ee123706ea6492fa9a3ad35` (RocketMQ 5.5.1);
- extended the lifecycle regression to call `start()` twice and `shutdown()` twice on the same producer, verifying that the shared accumulator is acquired and released exactly once;
- `ProduceAccumulatorTest` and `DefaultMQProducerTest`: 46/46 passed;
- Checkstyle, SpotBugs, and `git diff --check` passed in the affected reactor build.

The force-push has retriggered the full CI matrix and should also exercise the previously partial `DefaultMQProducer.start()` branch in patch coverage.

@guyinyou, since you have context in `ProduceAccumulator`, could you please take a human review when available, especially around the reference-counted lifecycle and repeated start/shutdown behavior? Thank you.

RockteMQ-AI (approved):
### Summary
Rebased fix for shared `ProduceAccumulator` lifecycle — looks good.

The two-level protection is sound:
- **Per-producer** (`AtomicBoolean.compareAndSet`): prevents double start/stop from the same producer instance
- **Shared accumulator** (`synchronized producerCount`): prevents premature shutdown when multiple producers share the same accumulator through `MQClientManager`

Test coverage validates both the shared-lifecycle and idempotent-start/stop scenarios.

LGTM.

---

ai-yang:
All 10 CI checks are green on the refreshed single signed-off commit. @RongtongJin @drpmma, could you please take a human review when convenient? The key point is the two-level lifecycle protection: per-producer idempotence plus reference-counted start/stop of the accumulator shared through `MQClientManager`.

ai-yang:
Pushed follow-up commit f3901017f89509118835cd90b40109b3b51aa9c3 to complete the shared-accumulator lifecycle fix.

The guard reference count alone did not isolate pending batches: a running producer could join a batch retaining an already stopped producer. Aggregation keys now include the concrete producer identity (also for explicit queues and same-group replacement instances). Completion/failure releases held bytes once and removes only that completed batch, avoiding stale producer retention.

Validation: 53 targeted client tests passed (5 accumulator + 41 producer + 7 new lifecycle cases), with Checkstyle and SpotBugs enabled. The seven new cases produced six failures against the previous PR production classes and all passed with this follow-up. The description documents the deterministic test setup and the autoBatch=false default.

@guyinyou, could you review the new batch ownership and cleanup semantics when convenient? In particular, producer isolation deliberately trades cross-producer coalescing for lifecycle correctness; no throughput improvement or graceful draining of a stopped producer is claimed. CI on this new head is separate from the earlier green result.

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The reference-counting approach for shared `ProduceAccumulator` instances is well-designed and fixes a real lifecycle bug where shared accumulators could be prematurely shut down.

Key points:
- `AtomicBoolean produceAccumulatorStarted` ensures idempotent start/shutdown at the producer level
- `producerCount` reference counting in `ProduceAccumulator` correctly prevents the internal executor from being stopped while other producers still depend on it
- Thread safety: `synchronized` on `start()/shutdown()` in `ProduceAccumulator` is appropriate given the reference count mutation
- Test coverage validates the shared accumulator scenario

---

PR #10942, opened 2026-08-15
[ISSUE #10941] Retain receipt handles when cleanup submission is rejected

### LINKED ISSUE #10941: [Bug] Proxy drops receipt handles when offline cleanup task submission is rejected

Aias00:
## What does this PR do?

Prevents offline receipt-handle cleanup from silently losing a removed group when its worker queue cannot accept the task. The cleanup executor now rejects explicitly, and `clearGroup` restores the group for the next scheduled retry.

## Test

- `mvn -pl proxy -Dtest=DefaultReceiptHandleManagerTest test`
- Adds a regression that shuts down the cleanup executor and verifies the receipt-handle group remains tracked.

Closes #10941.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/service/receipt/DefaultReceiptHandleManager.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/service/receipt/DefaultReceiptHandleManager.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Solid fix for preventing receipt handle loss when the cleanup thread pool rejects the submission.

The change catches `RejectedExecutionException` when submitting the cleanup task and restores the handle group to the map via `putIfAbsent()`, ensuring the handles are retained for retry in the next schedule cycle. This prevents data loss under high load when the thread pool queue is full.

The explicit `AbortPolicy` on the thread pool constructor is good practice — it makes the rejection behavior explicit rather than relying on the default.

Test coverage validates the fix by mocking a scenario where the executor rejects the task and verifying the handles remain in the map.

LGTM.

---

Aias00:
Superseded by #10934, which now contains both asynchronous cleanup-failure and task-submission-rejection handling plus their regression tests.

PR #10944, opened 2026-08-15
[ISSUE #10943] Tolerate incomplete queue-offset metadata during message decoding

### LINKED ISSUE #10943: [Bug] EscapeBridge fails to decode messages when queue-offset metadata is incomplete
## Reproduction

`EscapeBridge.decodeMsgList` iterates the message buffers and unconditionally reads `getMessageQueueOffset().get(i)`. If a `GetMessageResult` contains a valid message buffer but the queue-offset metadata is missing or shorter than the buffer list, the method throws `IndexOutOfBoundsException`.

This turns a recoverable failover-read metadata inconsistency into a failed escape-bridge read.

## Expected behavior

Decode valid messages even when the optional queue-offset list is incomplete. Apply the CQ offset only when it exists; otherwise retain the decoded message offset.

## Scope

Guard the offset lookup and add a regression test using a valid encoded message with no queue-offset metadata.

Aias00:
## What does this PR do?

Makes message decoding tolerate valid message buffers when the corresponding queue-offset metadata is absent or shorter than the buffer list.

The fix covers both EscapeBridge and PopReviveService. Each applies a consume-queue offset only when one is available, otherwise retaining the decoded message.

## Test

mvn -pl broker -Dtest=EscapeBridgeTest#decodeMsgListTest_messageOffsetMissing,PopReviveServiceTest#testDecodeMsgListWithoutQueueOffsets test

Closes #10943.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java
@@ -342,6 +342,7 @@ public class EscapeBridge {
         List<MessageExt> foundList = new ArrayList<>();
         try {
             List<ByteBuffer> messageBufferList = getMessageResult.getMessageBufferList();
+            List<Long> messageQueueOffsets = getMessageResult.getMessageQueueOffset();
             if (messageBufferList != null) {
                 for (int i = 0; i < messageBufferList.size(); i++) {
                     ByteBuffer bb = messageBufferList.get(i);
@@ -354,8 +355,10 @@ public class EscapeBridge {
                         LOG.error("decode msgExt is null {}", getMessageResult);
                         continue;
                     }
-                    // use CQ offset, not offset in Message
-                    msgExt.setQueueOffset(getMessageResult.getMessageQueueOffset().get(i));
+                    // Use the CQ offset when the store supplied it; otherwise retain the decoded offset.
+                    if (messageQueueOffsets != null && i < messageQueueOffsets.size()) {
+                        msgExt.setQueueOffset(messageQueueOffsets.get(i));
+                    }
                     foundList.add(msgExt);
                 }
             }
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
@@ -309,6 +309,7 @@ public class PopReviveService extends ServiceThread {
         List<MessageExt> foundList = new ArrayList<>();
         try {
             List<ByteBuffer> messageBufferList = getMessageResult.getMessageBufferList();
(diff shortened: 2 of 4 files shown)


RockteMQ-AI (approved):
﻿### Summary
Good defensive fix for handling incomplete queue-offset metadata during message decoding.

The change adds null and bounds checks before accessing `getMessageQueueOffset().get(i)`, preventing `IndexOutOfBoundsException` when the offset list is shorter than the message buffer list. This can happen with partial or corrupted metadata.

The fix is applied consistently in both `EscapeBridge.decodeMsgList()` and `PopReviveService.decodeMsgList()`, and only overrides the queue offset when the store actually supplied one, preserving the decoded offset otherwise.

Test coverage validates the fix with a scenario where the offset list is intentionally shorter than the message list.

LGTM.

---

RockteMQ-AI (approved):
﻿### Summary
Defensive fix for `decodeMsgList()` in both `EscapeBridge` and `PopReviveService`. Adds null and bounds checks before accessing `getMessageQueueOffset().get(i)`, preventing `IndexOutOfBoundsException` when the queue offset list is incomplete or null. When offsets are unavailable, the decoded message retains its original queue offset from the message itself.

### Findings
- **[Info]** `broker/src/main/java/org/apache/rocketmq/broker/failover/EscapeBridge.java:345-358` — Correct guard: `messageQueueOffsets != null && i < messageQueueOffsets.size()` ensures safe access.
- **[Info]** `broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java:312-325` — Same pattern applied consistently.
- **[Info]** Tests cover the missing offset scenario for both classes.
- **[Info]** Minor: `PopReviveServiceTest` uses reflection to test a private method — acceptable but fragile if the method signature changes.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Adds defensive null/bounds checks when accessing `getMessageQueueOffset()` during message decoding, preventing `IndexOutOfBoundsException` when offset metadata is incomplete.

**Changes reviewed:**
- `EscapeBridge.decodeMsgList()` and `PopReviveService.decodeMsgList()`: Extract `messageQueueOffsets` once, check for null and bounds before access. If offsets are missing, the decoded message retains its original queue offset from the message body.
- `AbstractRemotingActivity.request()`: Null-check `getExtFields()` before accessing fields — prevents NPE when ext fields are missing.
- Tests cover all three scenarios (missing offsets in EscapeBridge, PopReviveService, and null extFields in AbstractRemotingActivity).

**Assessment:**
- ✅ Correctness: Defensive checks prevent runtime exceptions on malformed/incomplete data.
- ✅ Backward compatible: When offsets are present, behavior is unchanged.
- ✅ Tests: Good coverage for the edge cases.

LGTM — solid defensive programming.

---

PR #10946, opened 2026-08-15
[ISSUE #10945] Align POP retry records with result batches

### LINKED ISSUE #10945: [Bug] POP retry response recoding uses the wrong record after a multi-message batch
## Reproduction\n\nWhen POP returns more than one normal message before a retry-topic result, PopConsumerContext stores one GetMessageResult per fetch result but flattens PopConsumerRecord entries per message.\n\nPopConsumerService later iterates getMessageResultList by result index and looks up popConsumerRecordList with the same index. After a multi-message result, that index points into the preceding batch instead of to the first record for the current result.\n\nWith popResponseReturnActualRetryTopic enabled, a retry result can therefore be treated as a normal result and skip retry-message recoding.\n\n## Expected behavior\n\nEach fetched GetMessageResult must use the first PopConsumerRecord that belongs to that result when deciding whether to recode retry messages.\n\n## Regression test\n\nPopConsumerServiceTest.popAsyncRecodeRetryMessagesAfterMultiMessageNormalResultTest fails on current develop.

Aias00:
## What does this PR do?

Aligns the flattened POP consumer-record index with each fetched GetMessageResult. This keeps retry-message recoding associated with the correct result after a preceding multi-message batch.

Fixes #10945.

## Test

mvn -pl broker -Dtest=PopConsumerServiceTest#popAsyncRecodeRetryMessagesAfterMultiMessageNormalResultTest test

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerService.java
@@ -432,9 +432,10 @@ public class PopConsumerService extends ServiceThread {
                         this.popConsumerStore.writeRecords(result.getPopConsumerRecordList());
                     }
 
+                    int popConsumerRecordIndex = 0;
                     for (int i = 0; i < result.getGetMessageResultList().size(); i++) {
                         GetMessageResult getMessageResult = result.getGetMessageResultList().get(i);
-                        PopConsumerRecord popConsumerRecord = result.getPopConsumerRecordList().get(i);
+                        PopConsumerRecord popConsumerRecord = result.getPopConsumerRecordList().get(popConsumerRecordIndex);
 
                         // If the buffer belong retries message, the message needs to be re-encoded.
                         // The buffer should not be re-encoded when popResponseReturnActualRetryTopic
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
Correct fix for POP retry record index misalignment when a multi-message batch precedes retry messages.

The original code used the loop index `i` to access both `getGetMessageResultList()` and `getPopConsumerRecordList()`, but these lists have different cardinalities when a single `GetMessageResult` contains multiple messages. The fix correctly introduces a separate `popConsumerRecordIndex` that advances by `getMessageQueueOffset().size()` per batch, aligning the record index with the actual message count.

Test coverage is thorough — the new test validates the scenario with a 2-message normal result followed by a 1-message retry result.

LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes an index misalignment bug in `PopConsumerService.popAsync()` where `PopConsumerRecord` list indexing was incorrectly using the batch index instead of tracking cumulative message offsets.

**The bug:** When a `GetMessageResult` contains multiple messages (batch result), the `PopConsumerRecord` list has one entry per message (flat list), but the old code used the same loop index `i` for both lists — causing misalignment after the first batch with >1 message.

**The fix:**
- Introduces `popConsumerRecordIndex` that advances by `getMessageResult.getMessageQueueOffset().size()` per iteration, correctly tracking the cumulative offset into the flat record list.

**Assessment:**
- ✅ Correctness: Precisely fixes the index mismatch. The old code would silently use wrong records after the first multi-message batch.
- ✅ Tests: `popAsyncRecodeRetryMessagesAfterMultiMessageNormalResultTest` covers the exact scenario.
- ✅ Clean helper: `getFoundResult(long... offsets)` is reusable.

LGTM — good catch on a subtle indexing bug.

---

PR #10954, opened 2026-08-18
[ISSUE #10953] Clamp shifted message count range to queue bounds

### LINKED ISSUE #10953: [Bug] Shifted estimateMessageCount range is not clamped to ConsumeQueue bounds

- [x] I searched the GitHub issues and discussions and did not find a duplicate.
- [x] I confirmed that this bug belongs to this repository.

The issue is platform-independent. It was observed from the broker's `PeriodicMetricReader` thread while exporting consumer-lag metrics.

### RocketMQ version

- Branch: `develop`
- Commit: `293f5885719fc4aa3619446a1900f58ccfcfdd29`

### JDK Version

The faulty offset calculation is independent of the JDK version.

### Describe the Bug

When a filtered message count is estimated and the requested `from` offset is smaller than the ConsumeQueue minimum offset, `DefaultMessageStore.estimateMessageCount` shifts the range to the right while preserving its length. However, it does not clamp the shifted range to the current ConsumeQueue bounds.

For example, a request for `[0, 200000)` with `minOffset = 190000` and `maxOffset = 200000` is changed to `[190000, 390000)`. The shifted upper bound exceeds the readable position of the ConsumeQueue.

For the file-based ConsumeQueue, this can make `selectMappedBuffer` return `null`, leave `raw = 0` and `match = 0`, and report a filtered lag of zero. It also produces repeated warnings such as:

```text
selectMappedBuffer request pos invalid, request pos: 2307920, size: 1416800, fileFromOffset: 0
```

### Steps to Reproduce

1. Use the file-based ConsumeQueue and enable accumulation estimation (the default).
2. Create a non-`*` TAG subscription so consumer lag uses a message filter.
3. Let old ConsumeQueue entries be removed so that the consumer offset is below the current minimum offset.
4. Estimate the count for `from = 0`, `to = 200000`, `minOffset = 190000`, and `maxOffset = 200000`.
5. Observe that the store asks the ConsumeQueue to estimate `[190000, 390000)` instead of a range within the current queue bounds.

The requested interval should first be shifted to start at `minOffset` while preserving its length, and then both endpoints should be clamped to `[minOffset, maxOffset]`. In the example, the resulting range should be `[190000, 200000)`.

If the shifted interval fits within the queue bounds, its original length should be preserved. If the available range is shorter, its upper end should be truncated at `maxOffset`.

### What Did You See Instead?

The interval is shifted to the right without enforcing the current upper bound. The scan can exceed the ConsumeQueue readable range, emit `selectMappedBuffer request pos invalid` warnings, and incorrectly report zero filtered lag even when matching messages exist.

### Additional Context

The problematic logic is in `DefaultMessageStore.estimateMessageCount`:

```java
long diff = to - from;
from = minOffset;
to = from + diff;
```

The shifted range should be bounded by both `getMinOffsetInQueue()` and `getMaxOffsetInQueue()`. The addition also needs to avoid overflow for a very large requested upper bound.

qianye1001:

- Fixes #10953

### Brief Description

`DefaultMessageStore.estimateMessageCount` shifts a filtered estimation range to the ConsumeQueue minimum offset when the requested lower bound has expired. The shifted upper bound was not constrained by the current maximum offset, so the range could extend beyond readable ConsumeQueue data. File-based ConsumeQueue scans could then emit `selectMappedBuffer request pos invalid` warnings and report a zero estimate without examining any entries.

This change:

- preserves the existing right-shift behavior and the requested interval length when space is available;
- clamps both shifted endpoints to the current ConsumeQueue bounds;
- truncates the shifted upper bound at `maxOffset` without overflowing;
- returns zero when clamping produces an empty range.

Regression tests cover a shifted range truncated at the maximum offset, a shifted range that preserves its full length, and a range that starts after the maximum offset.

Ran with Amazon Corretto JDK 11:

```bash
mvn -pl store "-Dtest=DefaultMessageStoreTest#testEstimateMessageCountShiftsRangeWithinQueueBounds+testEstimateMessageCountPreservesLengthWhenShiftedRangeFits+testEstimateMessageCountReturnsZeroWhenRangeIsAfterMaxOffset" test
```

Result: 3 tests run, 0 failures, 0 errors, 0 skipped. The same Maven run also completed Checkstyle and SpotBugs with no violations or findings.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/DefaultMessageStore.java
@@ -3184,12 +3184,19 @@ public class DefaultMessageStore implements MessageStore {
             return 0;
         }
 
-        // correct the "from" argument to min offset in queue if it is too small
+        // shift the range to min offset if it is too small, then clamp it to the current queue bounds
         long minOffset = consumeQueue.getMinOffsetInQueue();
+        long maxOffset = consumeQueue.getMaxOffsetInQueue();
         if (from < minOffset) {
             long diff = to - from;
             from = minOffset;
-            to = from + diff;
+            to = diff > maxOffset - from ? maxOffset : from + diff;
+        }
+
+        from = Math.min(from, maxOffset);
+        to = Math.min(to, maxOffset);
+        if (from >= to) {
+            return 0;
         }
 
         long msgCount = consumeQueue.estimateMessageCount(from, to, filter);
--- a/store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/DefaultMessageStoreTest.java
@@ -20,10 +20,15 @@ package org.apache.rocketmq.store;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.anyLong;
 import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.same;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 import org.mockito.ArgumentCaptor;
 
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a boundary issue in `DefaultMessageStore.estimateMessageCount` where shifting the estimation range to the min offset could cause the upper bound to exceed the current max offset, leading to invalid buffer requests.

The fix correctly clamps both bounds to `maxOffset` after the shift, and returns 0 for empty ranges. The logic is clean and the three test cases cover the key scenarios (shift within bounds, shift exceeding max, and range entirely past max).

All CI checks pass. LGTM.

---

lollipopjin (approved):
LGTM

PR #10960, opened 2026-08-18
[Bug] RocksDBConsumeQueue#iterateFrom returns a non-null iterator for offsets below minOffset, causing NPE in ScheduleMessageService

### LINKED ISSUE #10959: [Bug] RocksDBConsumeQueue#iterateFrom returns a non-null iterator for offsets below minOffset, causing NPE in ScheduleMessageService

CentOS

### RocketMQ version

5.5.0

### JDK Version

JDK11

### Describe the Bug

`RocksDBConsumeQueue#iterateFrom(long startIndex)` / `iterateFrom(long startIndex, int count)`
(store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueue.java, around line 299-317)
only validate `startIndex >= 0 && startIndex < maxOffsetInQueue`. They never check
`startIndex` against the queue's current `minOffsetInQueue`.

This is inconsistent with the file-based `ConsumeQueue#iterateFrom`
(store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java, `getIndexBuffer`, around
line 918-935), which explicitly checks `offset >= this.getMinLogicOffset()` and returns
`null` otherwise.

Because of this gap, when `startIndex` is below the RocksDB consume queue's current
min offset (e.g. after the underlying data for that offset has been cleaned/expired, or the
queue's min offset was advanced by recovery/rebuild), `iterateFrom` still returns a
non-null `LargeRocksDBConsumeQueueIterator`.

That iterator's `hasNext()` only compares `currentIndex < totalCount` (pure index
arithmetic) and never verifies the record actually exists, while `next()` calls
`consumeQueueStore.get(...)`, which returns `null` (or an undersized buffer) for a purged
offset and makes `next()` return `null` — violating the normal Iterator contract of
"hasNext()==true implies next() is non-null".

`ScheduleMessageService.DeliverDelayedMessageTimerTask#executeOnTimeUp`
(broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java,
around line 410-431) relies on `iterateFrom` returning `null` to detect and self-correct an
out-of-range offset (see the `if (bufferCQ == null) {...}` block at line 411-424). Since it
receives a non-null iterator instead, execution falls through to the `while` loop, calls
`bufferCQ.next()` (line 430) which returns `null`, and then immediately dereferences it via
`cqUnit.getPos()` (line 431), throwing a `NullPointerException`.

Worse, `DeliverDelayedMessageTimerTask#run()` (line 375-386) catches this `Throwable`,
logs it, and reschedules the *same* task with the identical (still out-of-range) `this.offset`
via `scheduleNextTimerTask(this.offset, DELAY_FOR_A_PERIOD)`, since `this.offset` is a
`final` field that was never advanced before the exception. This creates an infinite
NPE retry loop for that delay level: the timer task fires every `DELAY_FOR_A_PERIOD`,
throws the same NPE, logs an error, and reschedules itself with the same broken offset —
effectively stalling delayed/scheduled message delivery for that delay level until the
broker is restarted (and the underlying condition is otherwise resolved).

### Steps to Reproduce

1. Run a broker configured with the RocksDB consume queue store
   (`messageStoreConfig.setStoreType(StoreType.DEFAULT_ROCKSDB)`).

2. Minimal unit-level reproduction (no full cluster needed):

   ```java
   RocksDBConsumeQueueStore store = mock(RocksDBConsumeQueueStore.class);
   when(store.getMinOffsetInQueue(anyString(), anyInt())).thenReturn(9000L);
   when(store.getMaxOffsetInQueue(anyString(), anyInt())).thenReturn(10000L);
   // store.get(topic, queueId, offset) for any offset < 9000 returns null,
   // simulating a purged/expired record.

   RocksDBConsumeQueue cq = new RocksDBConsumeQueue(storeConfig, store, "topic", 0);
   ReferredIterator<CqUnit> it = cq.iterateFrom(8000); // 8000 < minOffset(9000)

   // Bug: `it` is non-null.
   assertTrue(it.hasNext());   // true, based purely on index arithmetic
   assertNull(it.next());      // null CqUnit, because offset 8000 is already purged
   ```

3. Real-world trigger path: `ScheduleMessageService` persists a resume offset per delay
   level (config/delayOffset.json). If the schedule topic's RocksDB consume queue min
   offset later advances past that persisted offset (e.g. via expired-data cleanup, or the
   consume queue being rebuilt/recovered with a higher min offset than the still-recorded
   schedule offset), the next scheduled tick calls
   `cq.iterateFrom(this.offset)` with `this.offset < cq.getMinOffsetInQueue()`.

4. Observe the broker log repeatedly print:
   ```
   ScheduleMessageService, executeOnTimeUp exception
   java.lang.NullPointerException
       at org.apache.rocketmq.broker.schedule.ScheduleMessageService$DeliverDelayedMessageTimerTask.executeOnTimeUp(ScheduleMessageService.java:431)
   ```
   ...on a fixed interval (DELAY_FOR_A_PERIOD), never recovering on its own.

`RocksDBConsumeQueue#iterateFrom` should return `null` when `startIndex` is below the
queue's current `getMinOffsetInQueue()`, exactly like the file-based `ConsumeQueue`
implementation does via its `getMinLogicOffset()` check.

With that, `ScheduleMessageService.executeOnTimeUp()` would take its existing
`if (bufferCQ == null)` branch (line 411-424), log a single "schedule CQ offset invalid"
error, self-correct the offset (clamped to the queue's current min/max bound), and keep
delivering delayed messages normally on the next tick — no exception, no stuck retry loop.

### What Did You See Instead?

`iterateFrom` returns a non-null iterator whose `next()` yields `null` once the requested
offset has already been purged from RocksDB. `ScheduleMessageService` dereferences that
null `CqUnit` at line 431 and throws a `NullPointerException`:

```
java.lang.NullPointerException
    at org.apache.rocketmq.broker.schedule.ScheduleMessageService$DeliverDelayedMessageTimerTask.executeOnTimeUp(ScheduleMessageService.java:431)
    at org.apache.rocketmq.broker.schedule.ScheduleMessageService$DeliverDelayedMessageTimerTask.run(ScheduleMessageService.java:379)
    ...
```

Because `DeliverDelayedMessageTimerTask#run()` catches this exception and reschedules
itself with the same unchanged offset, the same NPE repeats indefinitely every
`DELAY_FOR_A_PERIOD`, and delayed/scheduled message delivery for the affected delay
level is effectively stuck until the broker is restarted.

### Additional Context

_No response_

ymwneu:
fix: RocksDBConsumeQueue.iterateFrom should reject offset below min offset
startIndex < minOffset was previously accepted as long as it stayed within [0, maxOffset), returning a non-null LargeRocksDBConsumeQueueIterator whose next() can yield a null CqUnit once the underlying data has been purged. Callers like ScheduleMessageService rely on iterateFrom returning null to detect and correct an out-of-range offset (matching ConsumeQueue's getMinLogicOffset() check); without it they NPE dereferencing the null CqUnit instead.

Add the same startIndex >= getMinOffsetInQueue() bound used by the file-based ConsumeQueue to both iterateFrom overloads.

- Fixes #10959

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueue.java
@@ -298,8 +298,9 @@ public class RocksDBConsumeQueue implements ConsumeQueueInterface {
 
     @Override
     public ReferredIterator<CqUnit> iterateFrom(final long startIndex) {
+        long minCqOffset = getMinOffsetInQueue();
         long maxCqOffset = getMaxOffsetInQueue();
-        if (startIndex < maxCqOffset && startIndex >= 0) {
+        if (startIndex >= minCqOffset && startIndex < maxCqOffset && startIndex >= 0) {
             int num = pullNum(startIndex, maxCqOffset);
             return new LargeRocksDBConsumeQueueIterator(startIndex, num);
         }
@@ -308,8 +309,9 @@ public class RocksDBConsumeQueue implements ConsumeQueueInterface {
 
     @Override
     public ReferredIterator<CqUnit> iterateFrom(long startIndex, int count) throws RocksDBException {
+        long minCqOffset = getMinOffsetInQueue();
         long maxCqOffset = getMaxOffsetInQueue();
-        if (startIndex < maxCqOffset) {
+        if (startIndex >= minCqOffset && startIndex < maxCqOffset) {
             int num = Math.min((int)(maxCqOffset - startIndex), count);
             return iterateFrom0(startIndex, num, maxCqOffset);
         }
--- a/store/src/test/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/queue/RocksDBConsumeQueueTest.java
@@ -121,6 +121,28 @@ public class RocksDBConsumeQueueTest extends QueueTestBase {
     }
 
     @Test
+    public void testIterateFrom_startIndexBelowMinOffset_returnsNull() throws Exception {
+        if (MixAll.isMac()) {
+            return;
+        }
+        DefaultMessageStore messageStore = mock(DefaultMessageStore.class);
+        RocksDBConsumeQueueStore rocksDBConsumeQueueStore = mock(RocksDBConsumeQueueStore.class);
+        when(messageStore.getQueueStore()).thenReturn(rocksDBConsumeQueueStore);
+        when(rocksDBConsumeQueueStore.getMinOffsetInQueue(anyString(), anyInt())).thenReturn(9000L);
+        when(rocksDBConsumeQueueStore.getMaxOffsetInQueue(anyString(), anyInt())).thenReturn(10000L);
+
+        RocksDBConsumeQueue consumeQueue = new RocksDBConsumeQueue(messageStore.getMessageStoreConfig(), rocksDBConsumeQueueStore, "topic", 0);
+
+        // startIndex has already been purged (below the queue's current min offset); must not
+        // return a non-null iterator that later yields a null CqUnit and NPEs its caller.
+        assertNull(consumeQueue.iterateFrom(8000));
+        assertNull(consumeQueue.iterateFrom(8000, 10));
+
+        // startIndex within [min, max) still works as before.
+        assertNotNull(consumeQueue.iterateFrom(9000));
+    }
+
+    @Test
     public void testLmqCounter_running() throws ConsumeQueueException {
         messageStore.getMessageStoreConfig().setEnableMultiDispatch(true);
         messageStore.getMessageStoreConfig().setEnableLmq(true);


RockteMQ-AI:
## Review Summary for PR #10960

**PR**: [Bug] RocksDBConsumeQueue#iterateFrom returns a non-null iterator for offsets below minOffset, causing NPE in ScheduleMessageService
**Author**: @ymwneu
**Scope**: +26 -2 across 2 file(s)

### Changed Files

### Observations
- This PR modifies 2 file(s) with 26 additions and 2 deletions.

*Review generated by automated scan. Please verify the changes carefully.*

RockteMQ-AI:
## Code Review — PR #10960

### Summary
This PR fixes a bug in `RocksDBConsumeQueue#iterateFrom` where requesting an iterator for offsets below `minOffset` would return a non-null iterator that later yields `null` `CqUnit` entries, causing NPE in callers like `ScheduleMessageService`.

### Assessment: **Approve** ✅

The fix is correct, minimal, and well-targeted.

### Analysis

**Root Cause:** Both `iterateFrom(long)` and `iterateFrom(long, int)` only checked `startIndex < maxCqOffset` but did not guard against `startIndex < minOffset`. When messages have been purged and the requested offset falls below the minimum, the method would still create a `LargeRocksDBConsumeQueueIterator` that iterates over stale/missing entries, eventually producing null `CqUnit` values.

**Fix:** Adds `startIndex >= minCqOffset` to the boundary check in both overloads:
- `RocksDBConsumeQueue.java:301` — `iterateFrom(long)` now checks `startIndex >= minCqOffset && startIndex < maxCqOffset && startIndex >= 0`
- `RocksDBConsumeQueue.java:312` — `iterateFrom(long, int)` now checks `startIndex >= minCqOffset && startIndex < maxCqOffset`

**Observations:**
1. ✅ The fix is consistent across both overloads
2. ✅ Returning `null` for out-of-range offsets matches the existing contract (callers already handle null returns)
3. ✅ The test `testIterateFrom_startIndexBelowMinOffset_returnsNull` covers both the below-min case and the valid-range case
4. ⚠️ Minor: `getMinOffsetInQueue()` is called twice per invocation path (once in `iterateFrom` and potentially again inside the iterator). Consider caching if this is a hot path, though the current approach is fine for correctness.

**Test Coverage:**
- New test properly verifies null return for purged offsets (8000 < min 9000)
- Also verifies valid range still works (9000 >= min 9000)
- Mac skip guard is consistent with existing test patterns

### Positive Aspects
- Clean, surgical fix with no unnecessary changes
- Good test coverage for the regression scenario
- The fix prevents a real NPE that could affect `ScheduleMessageService` in production

RockteMQ-AI (approved):
LGTM — clean, well-scoped bug fix.

The change correctly adds the `startIndex >= minCqOffset` bound check to both `iterateFrom` overloads in `RocksDBConsumeQueue`, matching the existing behavior of the file-based `ConsumeQueue`. This prevents the NPE in `ScheduleMessageService` when purged data is accessed.

Test coverage is solid — verifies both the negative case (below min returns null) and the positive case (at min still works).

---

lizhimins (approved):

PR #10963, opened 2026-08-24
[ISSUE #10962] Fix lite subscription quota validation for remove actions

### LINKED ISSUE #10962: [Enhancement] Improve Lite subscription quota validation for remove actions

### Summary

Improve Lite subscription quota validation in proxy by applying the quota check only to Lite subscription add actions.

Currently, proxy validates `liteTopicSet.size()` against `liteSubClientQuota + quotaBuffer` before syncing Lite subscriptions. This validation should apply to `PARTIAL_ADD` and `COMPLETE_ADD`, but not to remove actions.

### Motivation

Lite subscription quota is used to limit subscription growth. Remove actions release existing subscriptions and should not be blocked by the same quota validation.

For example, a client may add Lite subscriptions in several smaller `PARTIAL_ADD` requests and later remove them in one larger `PARTIAL_REMOVE` request. The remove request can contain more Lite topics than the quota threshold, but it only reduces the active subscription set.

Update `ClientProcessor#syncLiteSubscription` so that Lite subscription quota validation is only applied to add actions:

- `PARTIAL_ADD`
- `COMPLETE_ADD`

Skip this quota validation for remove actions:

- `PARTIAL_REMOVE`
- `COMPLETE_REMOVE`

This keeps the quota protection for subscription growth while allowing large unsubscribe requests to release resources normally.

One alternative is to keep validating all actions with non-empty `liteTopicSet`, but this treats remove requests as if they were adding subscriptions.

Another alternative is to introduce a separate request-size limit for Lite subscription sync requests. That may be useful independently, but it should not reuse `LITE_SUBSCRIPTION_QUOTA_EXCEEDED`, because subscription quota and request size are different concepts.

### Additional Context

_No response_

yx9o:
- Fixes #10962 .

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
@@ -115,12 +115,14 @@ public class ClientProcessor extends AbstractProcessor {
         LiteSubscriptionDTO liteSubscriptionDTO, long timeoutMillis
     ) {
         try {
+            LiteSubscriptionAction action = liteSubscriptionDTO.getAction();
             validateLiteBindTopic(ctx, liteSubscriptionDTO.getGroup(), liteSubscriptionDTO.getTopic());
-            if (CollectionUtils.isNotEmpty(liteSubscriptionDTO.getLiteTopicSet())) {
+            if (CollectionUtils.isNotEmpty(liteSubscriptionDTO.getLiteTopicSet())
+                && shouldValidateLiteSubscriptionQuota(action)) {
                 validateLiteSubscriptionQuota(ctx, liteSubscriptionDTO.getGroup(), liteSubscriptionDTO.getLiteTopicSet().size());
             }
 
-            if (LiteSubscriptionAction.PARTIAL_ADD == liteSubscriptionDTO.getAction()) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a bug where `validateLiteSubscriptionQuota` was called for all lite subscription actions, including REMOVE operations. The fix correctly restricts quota validation to only ADD-type actions (`PARTIAL_ADD`, `COMPLETE_ADD`), since removing subscriptions frees quota rather than consuming it. The change is small, focused, and includes test coverage for the PARTIAL_REMOVE case. LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes quota validation being incorrectly applied to remove/unsubscribe actions in lite subscription sync. The validation should only gate additive operations (PARTIAL_ADD, COMPLETE_ADD), not removals.

### Review
- **Correctness** ✅ — The `shouldValidateLiteSubscriptionQuota()` guard correctly restricts quota checks to add actions only. Remove actions should never be blocked by quota limits.
- **Code quality** ✅ — Clean extraction of the condition into a private method improves readability. The local variable `action` avoids repeated `getAction()` calls.
- **Tests** ✅ — Test coverage added for the remove action path.

LGTM — straightforward and correct fix.

---

PR #10964, opened 2026-08-25
[ISSUE #10949] Improve pop consumer offset committing

### LINKED ISSUE #10949: [Enhancement] Converge pop consumer offset in the background so it does not depend on client polling

### Summary

With popConsumerKVServiceEnable = true, a queue's consumer offset is only advanced by pop requests, and the committed value always lags one round behind the real position. 

There is no broker-side task that closes this gap, so once a consumer drains the backlog, acks everything and goes offline, the offset stays behind indefinitely and reported lag never reaches zero. 

Adding background convergence would make the offset reflect the actual consumption position regardless of client polling behavior.

### Motivation

Ack no longer advances the offset (PopConsumerService#ackAsync only deletes the record), and the commit now rides on pop using the start offset of the current batch, while the real watermark nextBeginOffset goes only to pullOffsetTable. 

No background task closes the gap: run() only revives, and PopConsumerCache#cleanupRecords skips the commit once the queue drains because getMinOffsetInBuffer() returns -1 — with enablePopBufferMerge=false there is no periodic committer at all. 

The legacy PopBufferMergeService path does not need one, since ack itself commits nextBeginOffset.

Add a periodic broker-side task in PopConsumerService that converges offsetTable toward pullOffsetTable, bounded by the minimum in-flight offset still held in the cache, so it works for both enablePopBufferMerge values without depending on any further client request.

/

### Additional Context

_No response_

redlsz:

- Fixes #10949 

### Brief Description

- **PopConsumerService**: each pop now records its nextBeginOffset as a pending commit and commits it after the pop completes, capped at minOffsetInCache while records are still buffered, and skipped on offset reset or backward movement. FIFO path unchanged.
- **PopConsumerCache**: keeps the pending offset on the queue's cache entry and commits it in cleanupRecords once no in-flight record remains (all acked or persisted).
- **PopConsumerCache#cleanupRecords**: the timeout-eviction path now holds the same lock as popAsync, so a concurrent pop's records can't be evicted mid-write.

Minimum verification test. Check whether the consumer offsets have been advanced.

```
public class SimpleProducerConsumerExample {
    private static final Logger log = LoggerFactory.getLogger(CustomConsoleAppender.class);

    public static void main(String[] args) throws Exception {
        String endpoints = "foobar.com:8080";
        String accessKey = "yourAccessKey";
        String secretKey = "yourSecretKey";
        String topic = "yourTopic";
        String consumerGroup = "yourConsumerGroup";

        SessionCredentialsProvider credentialsProvider = new StaticSessionCredentialsProvider(accessKey, secretKey);
        ClientConfiguration clientConfiguration = ClientConfiguration.newBuilder()
            .setEndpoints(endpoints)
            .enableSsl(false)
            .setCredentialProvider(credentialsProvider)
            .build();

        ClientServiceProvider provider = ClientServiceProvider.loadService();

        // Send one normal message
        try (Producer producer = provider.newProducerBuilder()
            .setClientConfiguration(clientConfiguration)
            .setTopics(topic)
            .build()) {

            Message message = provider.newMessageBuilder()
                .setTopic(topic)
                .setBody("Hello, RocketMQ!".getBytes(StandardCharsets.UTF_8))
                .build();
            SendReceipt receipt = producer.send(message);
            log.info("Message sent, messageId={}", receipt.getMessageId());
        }

        // Receive the message and ack it
        try (SimpleConsumer consumer = provider.newSimpleConsumerBuilder()
            .setClientConfiguration(clientConfiguration)
            .setConsumerGroup(consumerGroup)
            .setAwaitDuration(Duration.ofSeconds(5))
            .setSubscriptionExpressions(Collections.singletonMap(topic, new FilterExpression("*", FilterExpressionType.TAG)))
            .build()) {

            List<MessageView> messages = consumer.receive(1, Duration.ofSeconds(15));
            if (messages != null && !messages.isEmpty()) {
                MessageView message = messages.get(0);
                consumer.ack(message);
                log.info("Message received and acked, messageId={}", message.getMessageId());
            }
        }
    }
}
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
@@ -84,6 +84,13 @@ public class PopConsumerCache extends ServiceThread {
         return consumerRecords != null ? consumerRecords.getMinOffsetInBuffer() : OFFSET_NOT_EXIST;
     }
 
+    public void setPendingCommitOffset(String groupId, String topicId, int queueId, long offset) {
+        ConsumerRecords consumerRecords = consumerRecordTable.get(this.getKey(groupId, topicId, queueId));
+        if (consumerRecords != null) {
+            consumerRecords.setPendingCommitOffset(offset);
+        }
+    }
+
     public long getPopInFlightMessageCount(String groupId, String topicId, int queueId) {
         ConsumerRecords consumerRecords = consumerRecordTable.get(this.getKey(groupId, topicId, queueId));
         return consumerRecords != null ? consumerRecords.getInFlightRecordCount() : 0L;
@@ -133,6 +140,7 @@ public class PopConsumerCache extends ServiceThread {
                     consumerRecordStore.writeRecords(writeConsumerRecords);
                 }
                 records.clearStagedRecords();
+                this.commitPendingOffset(records);
                 log.info("PopConsumerOffline, so clean expire records, groupId={}, topic={}, queueId={}, records={}",
                     records.getGroupId(), records.getTopicId(), records.getQueueId(), records.getInFlightRecordCount());
                 iterator.remove();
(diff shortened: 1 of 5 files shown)


RockteMQ-AI (approved):
### Summary
This PR improves the pop consumer offset committing mechanism by introducing a pending commit offset pattern that ensures offsets are committed only after pop operations complete, and adds lock acquisition in `cleanupRecords` to prevent concurrent eviction during pop writes.

The changes are well-structured and address a real concurrency issue. The use of `volatile` for `pendingCommitOffset` and CAS-based `advancePendingCommitOffset` ensures thread safety. The lock acquisition in `cleanupRecords` (matching the `popAsync` lock) correctly prevents race conditions.

LGTM — no blocking issues found.

---

RockteMQ-AI (approved):
### Summary
Improves pop consumer offset committing by deferring the commit until after the pop operation completes, using a pending commit mechanism. This ensures the committed offset reflects the actual `nextBeginOffset` of each pop round rather than lagging one round behind.

### Review
- **Correctness** ✅ — The pending commit approach is sound: offsets are collected during `handleGetMessageResult` and committed after the pop future completes in `commitPendingOffset`. The cache-aware capping (`getMinOffsetInCache`) correctly prevents committing past unacknowledged cached records.
- **Concurrency** ✅ — The added `consumerLockService.tryLock()` in `cleanupRecords` prevents evicting records that a concurrent pop is still writing. The lock scope is appropriate.
- **FIFO vs non-FIFO** ✅ — FIFO path correctly commits immediately (offset = consume start point). Non-FIFO path defers via pending commits, which is the right trade-off for throughput.
- **Edge cases** — `commitPendingOffset` handles null `pendingCommitList` gracefully. The `KeyBuilder.parseNormalTopic` usage for the lock key is consistent with existing patterns.
- **Tests** ✅ — New `PopConsumerServiceCommitOffsetTest` covers the pending commit flow.

One minor note: the `setPendingCommitOffset` method in `PopConsumerCache` silently no-ops if `consumerRecords` is null — this is fine but worth a debug log if it becomes a diagnostic pain point.

LGTM — well-structured improvement to offset semantics with proper concurrency handling.

---

Issue #10978, opened 2026-08-27
[Enhancement] Add Nullable annotations to nullable Message getters and fix kotlin-stdlib transitive leak


obabichevjb:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

Two independent issues:

1. Seven getters in `Message.java` return null at runtime but have no nullability metadata. This causes silent NPEs and prevents both Java IDEs and Kotlin from warning the caller.
2. A pure-Java project depending on `rocketmq-client` transitively receives `kotlin-stdlib` on its classpath via an uncovered okhttp dependency edge.

### Motivation

**Nullable getters:** `getTags()`, `getKeys()`, `getProperty()`, `getUserProperty()`, `getBuyerId()`, `getProperties()`, and `getTransactionId()` all return null under normal conditions (e.g., `getKeys()` is null unless explicitly set; `getTags()` is null when the producer omitted the tag). A Java user reported the same `getTags()` null problem in [#1453](https://github.com/apache/rocketmq/issues/1453) in 2019 — the issue was closed with no annotation or fix. 

**kotlin-stdlib leak:** Upstream already wrote `kotlin-stdlib` exclusions for the `okio-jvm` dependency edge. However, an independent path — `opentelemetry-exporter-otlp` → `opentelemetry-exporter-sender-okhttp` → `okhttp` — still brings `kotlin-stdlib` in. This ships Kotlin's runtime to every user, including pure-Java projects that have no Kotlin dependency.

### Describe the Solution You'd Like

**Nullable getters:** Add `@Nullable` (`javax.annotation.Nullable`, already available via `org.apache.tomcat:annotations-api`) to the seven affected getters in `Message.java`. This is a pure-metadata change with zero behavioral impact.

**kotlin-stdlib leak:** Extend the exclusion to cover `com.squareup.okhttp3:okhttp` and `com.squareup.okhttp3:okhttp:3.x` (which transitively pulls `kotlin-stdlib-jdk8` and `kotlin-stdlib-jdk7`) in the `opentelemetry-exporter-otlp` dependency declaration. This enforces the constraint upstream already wrote for `okio-jvm` — no new policy decision required.

### Describe Alternatives You've Considered

-

### Additional Context

-

RockteMQ-AI:
**Issue Evaluation**

Category: `type/enhancement` | Status: **Evaluated**

**Feasibility:** Feasible
[...]

PR #10979, opened 2026-08-27
fix: Add nullability annotations and fix kotlin-stdlib transitive leak

obabichevjb:

- #10978 

### Brief Description

This PR contains two independent fixes:

1. **`@Nullable` annotations on 7 getters in `Message.java`** that can return null at runtime. The methods are declared as returning plain `String`/`Map` with no nullability metadata, which produces silent NPEs in Kotlin consumers and gives no warning in Java IDEs. 

2. **Exclude leaked `kotlin-stdlib` from the `opentelemetry-exporter-otlp` dependency chain**. A pure-Java project depending on `rocketmq-client` transitively resolves `kotlin-stdlib` via: `opentelemetry-exporter-otlp` → `opentelemetry-exporter-sender-okhttp` → `okhttp` → `kotlin-stdlib-*`. Upstream already wrote `kotlin-stdlib` exclusions for the `okio-jvm` edge; this extends them to cover the okhttp path, enforcing a constraint upstream already intended.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/Message.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/Message.java
@@ -18,6 +18,8 @@ package org.apache.rocketmq.common.message;
 
 import org.apache.commons.lang3.math.NumberUtils;
 
+import javax.annotation.Nullable;
+
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR adds `@Nullable` annotations to getter methods in `Message.java` that can legitimately return null, and excludes transitive kotlin-stdlib dependencies from `opentelemetry-exporter-otlp` to reduce dependency bloat.

Both changes look correct and well-scoped:
- The `@Nullable` annotations accurately reflect the nullability contract of these getters (e.g., `getTags()`, `getKeys()`, `getTransactionId()` can all return null)
- The kotlin-stdlib exclusions prevent unnecessary transitive dependencies from leaking into the classpath

LGTM.

---

PR #10986, opened 2026-08-29
[ISSUE #10985] Handle exceptional POP revive reads without losing retry

### LINKED ISSUE #10985: [Bug] POP revive can lose retry after an asynchronous message-read failure

- [x] I searched GitHub Issues, PRs, and Discussions and believe this is not a duplicate.

Reproduced on Ubuntu 22.04.4 LTS, Linux 5.15.0-186, x86_64. The defect is in the broker's asynchronous POP revive control flow and is not expected to be OS-specific.

## RocketMQ version

- Branch: `develop`
- Version: current `5.5.1` development sources
- Commit: `e348efa66b08eb645ee123706ea6492fa9a3ad35`

## JDK Version

OpenJDK 8u502 (Maven runtime).

## Describe the Bug

Proposed severity: **High**, because the failure can violate POP's at-least-once redelivery guarantee for an unacknowledged message.

`PopReviveService.reviveMsgFromCk` starts an asynchronous business-message read for every unacknowledged offset and records the checkpoint in `inflightReviveRequestMap`. `mergeAndRevive` then advances and commits the revive offset after scheduling that work.

If `EscapeBridge.getMessageAsync` (including its decode/transform chain) completes exceptionally, the corresponding child future remains exceptional. `CompletableFuture.allOf(...).whenComplete(...)` still invokes its callback, but the callback calls `future.getNow(...)`; for an exceptionally completed future, `getNow` throws `CompletionException`. The callback therefore exits before it can:

- call `rePutCK` to preserve retryability;
- mark the in-flight checkpoint complete;
- remove it from `inflightReviveRequestMap`.

The only timeout fallback is inside `while (inflightReviveRequestMap.size() > 3)`. With one to three failed checkpoints and no later traffic, that branch is never entered. Because the revive offset has already moved past the original checkpoint, the unacknowledged message can remain unredelivered indefinitely.

This is not a security issue.

## Steps to Reproduce

A deterministic broker-free unit reproduction is sufficient:

1. Build one `PopCheckPoint` containing a single unacknowledged message.
2. Make `EscapeBridge.getMessageAsync(...)` return an exceptionally completed `CompletableFuture`.
3. Invoke `PopReviveService.mergeAndRevive(...)`.
4. Assert that the revive offset and committed offset advanced, the checkpoint remains in the in-flight map, and no replacement CK was written.

The core setup is:

```java
CompletableFuture<Triple<MessageExt, String, Boolean>> failed = new CompletableFuture<>();
failed.completeExceptionally(new RuntimeException("store read failed"));
when(escapeBridge.getMessageAsync(anyString(), anyLong(), anyInt(), anyString(), anyBoolean()))
    .thenReturn(failed);

popReviveService.mergeAndRevive(reviveObj);
```

On the unmodified baseline, two independent runs reproduced the same state:

```text
reviveObj.newOffset = 1
committed revive offset = 1
inflightReviveRequestMap.size() = 1
messageStore.putMessage(rewritten CK) = 0 calls
```

The regression test consequently fails because one CK rewrite was expected but none occurred.

An exceptional business-message read should be treated as a retryable read failure. The service should rewrite the checkpoint for that message offset, finish/remove the in-flight record, and preserve the existing offset progression without losing the message's retry path.

## What Did You See Instead?

The original revive offset is committed, but the replacement CK is never written. At low traffic the incomplete in-flight record is not old enough to trigger any independent scan and the size-gated timeout loop is never entered, so the message can be skipped indefinitely.

## Additional Context

A minimal fix is to use `CompletableFuture.handle` on the `getBizMessage` stage. If the upstream read completes exceptionally, log it and return `(msgOffset, false)`, which feeds the existing `rePutCK` path. The handler should remain scoped to the read stage so exceptions raised later while processing/writing a retry retain their current semantics.

With that fix, the full `PopReviveServiceTest` class passes 13/13 tests. Checkstyle, SpotBugs, and `git diff --check` also pass.

Exact GitHub searches for `PopReviveService getNow`, `PopReviveService rePutCK exception`, `getBizMessage PopReviveService exception`, and `PopReviveService CompletionException` found no issue or PR match. Related [#10658](https://github.com/apache/rocketmq/issues/10658)/[#10659](https://github.com/apache/rocketmq/pull/10659) concern head-of-line blocking in the newer `PopConsumerService` popkv batch path. [#10667](https://github.com/apache/rocketmq/issues/10667) concerns discarded futures in `PopConsumerCache.cleanupRecords` when buffer merging is enabled. Neither covers the legacy/default `PopReviveService` path committing an offset without preserving retry after an exceptional read.

ai-yang:

- Fixes #10985

### Brief Description

`PopReviveService` records a checkpoint as in flight, schedules asynchronous business-message reads, and then advances the revive offset. If one read completes exceptionally, `allOf(...).whenComplete(...)` runs but `future.getNow(...)` throws `CompletionException`. This aborts the callback before it rewrites the checkpoint or clears the in-flight entry. Since timeout cleanup only runs while the map contains more than three entries, a low-traffic failure can lose the message's retry path indefinitely.

This change:

- handles exceptional completion of the upstream `getBizMessage` future;
- logs the failed topic, queue, offset, and broker context;
- converts the failed read to `(msgOffset, false)`, allowing the existing `rePutCK` path to retain retryability;
- deliberately does not catch exceptions raised later by `reviveRetry`, keeping the change scoped to asynchronous reads;
- adds regressions asserting that the original offset is committed, a replacement CK is submitted to the store, and the in-flight entry is removed, for both immediately failed and later-failing reads.

There is no protocol, storage-format, or public API change.

### Applicability and limits

This repairs exceptional completion of the POP revive business-message read future. It is not a claim that every missing message or ordinary remote timeout reaches this branch: some upstream failures are already normalized into the existing retry result.

The new tests exercise `PopReviveService -> EscapeBridge -> MessageStore.getMessageAsync`, using a real `EscapeBridge` and injecting the fault at the mocked store-future boundary. They validate exceptional propagation and the existing checkpoint-rewrite/cleanup path, not physical disk/cloud/network failure, durable CK persistence, or exactly-once delivery. Failures of the replacement CK write itself are outside this patch.

Latest verification (2026-09-13): head `a191a2953750e946c2ffc6d8ba1d9cb547e85036`, based on `develop` at `1a50c6e4e35524ac22055b76fc0ff2a6d3cff99f`, JDK 8 / Maven 3.8.1:

```bash
git submodule update --init --depth 1 -- rocketmq-apis
mvn -pl broker -am -DskipITs -Dtest=PopReviveServiceTest \
```

The local run used offline mode and settings for the existing dependency cache, without changing project configuration or disabling Checkstyle/SpotBugs.

- `PopReviveServiceTest`: **14 passed, 0 failures/errors/skips**.
- All 11 targeted reactor modules succeeded, including the protocol module.
- Checkstyle and SpotBugs enabled for project code: no violations or findings. Generated protocol code follows upstream's configured exclusions.
- `git diff --check`: passed.

The immediate-failure regression now traverses the real bridge instead of mocking its return value. The added delayed-failure case first holds the store future incomplete, confirms that revive offset 1 has been committed while one request remains in flight and no CK rewrite has occurred, then fails the future. It asserts that the in-flight entry is removed and the replacement CK is submitted exactly once. The failure timing is controlled by the future, not a sleep.

Historical red evidence: the earlier regression mocked `EscapeBridge.getMessageAsync` directly. On unmodified `develop` at `e348efa66`, two independent runs reproduced:

```text
reviveObj.newOffset = 1
committed revive offset = 1
inflightReviveRequestMap.size() = 1
expected messageStore.putMessage(rewritten CK): 1
actual: 0
```

That red result is for the original test setup, not a fresh baseline execution of the two strengthened tests. The September 11 head passed 13 tests and all 10 GitHub checks when checked on September 13 before this test-only follow-up. CI for the new head must be evaluated independently; the production fix has not changed.

### Related PR coordination

#11023 also fixes #10985 using the same `handle`-based retry path. At the time of comparison (2026-09-11), that PR changes only the production file and records no local Maven run. This PR additionally contains the deterministic regression for offset advancement, checkpoint rewrite, and in-flight cleanup, along with the verification above. The production behavior is equivalent apart from logging level/context. Please coordinate review around one implementation and retain the regression coverage when consolidating.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopReviveService.java
@@ -568,7 +568,13 @@ public class PopReviveService extends ServiceThread {
             // retry msg
             long msgOffset = popCheckPoint.ackOffsetByIndex((byte) j);
             CompletableFuture<Pair<Long, Boolean>> future = getBizMessage(popCheckPoint, msgOffset)
-                .thenApply(rst -> {
+                .handle((rst, throwable) -> {
+                    if (throwable != null) {
+                        POP_LOGGER.error("reviveQueueId={}, get biz msg failed, topic:{}, qid:{}, offset:{}, brokerName:{}",
+                            queueId, popCheckPoint.getTopic(), popCheckPoint.getQueueId(), msgOffset,
+                            popCheckPoint.getBrokerName(), throwable);
+                        return new Pair<>(msgOffset, false);
+                    }
                     MessageExt message = rst.getLeft();
                     if (message == null) {
                         POP_LOGGER.info("reviveQueueId={}, can not get biz msg, topic:{}, qid:{}, offset:{}, brokerName:{}, info:{}, retry:{}, then continue",
--- a/broker/src/test/java/org/apache/rocketmq/broker/processor/PopReviveServiceTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/processor/PopReviveServiceTest.java
@@ -17,6 +17,7 @@
 package org.apache.rocketmq.broker.processor;
 
 import com.alibaba.fastjson2.JSON;
+import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.commons.lang3.tuple.Triple;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.broker.failover.EscapeBridge;
@@ -58,6 +59,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.NavigableMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -387,6 +389,33 @@ public class PopReviveServiceTest {
         Assert.assertEquals(INVISIBLE_TIME + 10 * 1000L, actualInvisibleTime.get()); // first interval is 10s
         verify(escapeBridge, times(0)).putMessageToSpecificQueue(any(MessageExtBrokerInner.class)); // write retry
         verify(messageStore, times(1)).putMessage(any(MessageExtBrokerInner.class)); // rewrite CK
+    }
+
+    @Test
+    public void testReviveMsgFromCk_getBizMessageExceptional_rewriteCK() throws Throwable {
+        PopCheckPoint ck = buildPopCheckPoint(0, 0, 1);
+        PopReviveService.ConsumeReviveObj reviveObj = new PopReviveService.ConsumeReviveObj();
+        reviveObj.map.put("", ck);
+        reviveObj.endTime = System.currentTimeMillis();
+
+        ArgumentCaptor<Long> commitOffsetCaptor = ArgumentCaptor.forClass(Long.class);
+        doNothing().when(consumerOffsetManager).commitOffset(anyString(), anyString(), anyString(), anyInt(),
+            commitOffsetCaptor.capture());
+
+        CompletableFuture<Triple<MessageExt, String, Boolean>> failed = new CompletableFuture<>();
+        failed.completeExceptionally(new RuntimeException("store read failed"));
+        when(escapeBridge.getMessageAsync(anyString(), anyLong(), anyInt(), anyString(), anyBoolean()))
+            .thenReturn(failed);
+
+        popReviveService.mergeAndRevive(reviveObj);
+
+        NavigableMap<?, ?> inflight = (NavigableMap<?, ?>) FieldUtils.readField(
+            popReviveService, "inflightReviveRequestMap", true);
+        assertEquals(1, reviveObj.newOffset);
+        assertEquals(1, commitOffsetCaptor.getValue().longValue());
+        assertEquals(0, inflight.size());
+        // An exceptional async read must retain retryability by rewriting the checkpoint.
+        verify(messageStore, times(1)).putMessage(any(MessageExtBrokerInner.class));
     }
 
     @Test
--- a/broker/src/test/java/org/apache/rocketmq/broker/processor/PopReviveServiceTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/processor/PopReviveServiceTest.java
@@ -39,6 +39,7 @@ import org.apache.rocketmq.common.utils.NetworkUtil;
 import org.apache.rocketmq.remoting.protocol.subscription.SubscriptionGroupConfig;
 import org.apache.rocketmq.store.AppendMessageResult;
 import org.apache.rocketmq.store.AppendMessageStatus;
+import org.apache.rocketmq.store.GetMessageResult;
 import org.apache.rocketmq.store.MessageStore;
 import org.apache.rocketmq.store.PutMessageResult;
 import org.apache.rocketmq.store.PutMessageStatus;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
This PR fixes a subtle bug in `PopReviveService` where an exceptional completion of the async `getBizMessage` future would abort the revive callback before rewriting the checkpoint or clearing the in-flight entry. In low-traffic scenarios, this could permanently lose the retry path for affected messages.

The fix is clean and well-scoped: switching from `.thenApply()` to `.handle()` correctly captures both normal and exceptional completions, and the `Pair<>(msgOffset, false)` return value leverages the existing `rePutCK` path to preserve retryability.

### Review

**Correctness** — The change is correct. The `.handle()` approach is the right pattern for catching exceptional future completion without affecting downstream stages. The `false` return correctly triggers checkpoint rewrite. The scope is properly limited to the async read stage — exceptions from `reviveRetry` are intentionally not caught here.

**Test coverage** — The regression test `testReviveMsgFromCk_getBizMessageExceptional_rewriteCK` is thorough: it verifies all three critical invariants (offset committed, in-flight entry removed, replacement CK written). Using `FieldUtils.readField()` for internal state verification is consistent with existing test patterns in this class.

**Performance** — Negligible overhead. The `throwable != null` check is a single branch in the already-async path. No new allocations on the happy path.

**Compatibility** — No protocol, storage-format, or public API changes. Pure internal implementation fix.

LGTM. Well-written fix for a real edge case that could cause silent message retry loss.

---

ai-yang:
CI triage update:

- The Bazel job executed 573 tests; 572 passed and the sole failure was `//test:.../OffsetResetIT`.
- This PR only changes `PopReviveService` and `PopReviveServiceTest`; the affected test class passes 13/13, and the Maven, coverage, license, CodeQL, Windows, and macOS jobs all passed.
- I found no direct overlap between `OffsetResetIT` and this focused exceptional POP-revive read path, so I am rerunning the failed Bazel job before making any code change.

@lizhimins @xdkxlk, could you please take a human review when convenient? The main review point is whether converting an exceptional business-message read to the existing `rePutCK` path correctly preserves POP at-least-once retry semantics without broadening exception handling around `reviveRetry`.

ai-yang:
Rerun follow-up: I attempted to rerun the failed Bazel workflow through the GitHub Actions API, but GitHub rejected the author-side request (`run cannot be rerun`). A maintainer-triggered rerun is therefore required. I have not changed the code merely to retrigger CI.

RockteMQ-AI (approved):
﻿### Summary
Fixes a bug where exceptional async reads in `PopReviveService` would abort the callback before cleanup, potentially losing the message's retry path indefinitely.

The original code used `thenApply` which would throw `CompletionException` on failed reads, preventing the checkpoint from being rewritten and the in-flight entry from being cleared. Since timeout cleanup only runs when the map has >3 entries, a low-traffic failure could lose the retry path.

The fix uses `handle` to catch exceptions, log the failed topic/queue/offset/broker, and convert to `(msgOffset, false)`, allowing the existing `rePutCK` path to retain retryability.

Regression test verifies: original offset committed, replacement CK written, in-flight entry removed.

LGTM 👍

---

ai-yang:
Refreshed onto `develop` at `1a50c6e4e` and pushed signed-off commit `e60e409ba`. The base includes #11086's `OffsetResetIT` stabilization, addressing the test area that failed on the previous Bazel run. The POP production patch is unchanged by the rebase.

Latest JDK 8 verification: `PopReviveServiceTest` passed 13/13; the full 11-module targeted broker reactor succeeded with Checkstyle and SpotBugs enabled for project code. Full CI is running again on the refreshed head.

I also documented the overlap with #11023: both use the same read-stage fallback; this PR includes the regression asserting offset advancement, checkpoint rewrite, and in-flight cleanup. @zjncs, thanks for investigating the same issue; could we coordinate around one implementation and retain that test coverage?

@lizhimins @xdkxlk, could you take a human review of the retry semantics and help select the consolidation path when convenient?

ai-yang:
Pushed test-only follow-up a191a2953750e946c2ffc6d8ba1d9cb547e85036; the production fix is unchanged.

The exceptional-read regression now passes through a real EscapeBridge and injects failure at MessageStore.getMessageAsync. A second case holds the read pending until after revive offset 1 is committed, then completes it exceptionally. It verifies exactly one replacement CK submission and removal of the in-flight entry.

Latest local JDK 8 / Maven validation: PopReviveServiceTest 14/14 passed; all 11 targeted reactor modules succeeded with Checkstyle and SpotBugs enabled. The PR description now distinguishes store-future fault injection from physical storage failures/durable persistence, and the historical baseline reproduction from today's expanded test run.

This additional coverage is available for the existing human review and coordination with #11023. The old head's green CI does not certify this new head; its checks should be considered separately.

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The change from `.thenApply()` to `.handle()` correctly catches exceptions thrown by `getBizMessage()` during POP revive, preventing the future from completing exceptionally and losing the retry opportunity.

Key points:
- Returning `Pair<>(msgOffset, false)` on failure preserves the retry semantics — the checkpoint will not advance past this message
- Error logging includes sufficient context (queueId, topic, offset, brokerName) for debugging
- Test coverage validates both the exception path and the normal path

---

Issue #10998, opened 2026-09-02
[Bug] Tiered storage cannot dispatch messages larger than the group commit size


woaishixiaoxiao:
### Before Creating the Bug Report

- [x] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [x] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [x] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

all

### RocketMQ version

all

### JDK Version

all

### Describe the Bug

- 调小group commit size，默认消息体上限为 4 MiB，可以调小到2MB
- 首条消息记录 size >= tieredStoreGroupCommitSize 时，dispatch 列表为空。
- commitAsync 不执行，Tiered CQ offset 不推进。
- 调度器不断重试相同 offset，阻塞后续归档。

### Steps to Reproduce

1. 调小group commit size到2MB
2. 发送单条消息大于2MB
3. 该topic的消息将会阻塞dispatch

### What Did You Expect to See?

消息不应该被阻塞上传

### What Did You See Instead?

消息阻塞上传

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Confirmed**

The reported issue has been verified against the current codebase.

**Root Cause:** In `MessageStoreDispatcherImpl.doScheduleDispatch` (`tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImpl.java`, lines 260-266), the batch-building loop checks `bufferSize >= groupCommitSize` **before** appending the current message. When the very first message's `cqUnit.getSize()` is already >= `groupCommitSize`, the loop breaks immediately, leaving both `appendingBufferList` and `dispatchRequestList` empty.

**Impact:**
- `commitAsync` (guarded by `if (!dispatchRequestList.isEmpty())` at lines 303-311) is never called
- The tiered ConsumeQueue offset does not advance (line 296 sets `endOffset == currentOffset`)
- The method returns `false` (line 336), causing the scheduler to retry the same offset indefinitely
- All subsequent messages for this topic are blocked from archiving to tiered storage

[...]

PR #10999, opened 2026-09-02
[ISSUE #10998] Fix Tiered storage cannot dispatch messages larger than the group commit size

woaishixiaoxiao:

- Fixes #<10998>

### Brief Description

The tiered storage dispatcher checks the accumulated batch size before
appending the current message. When the first message size is equal to or
larger than `tieredStoreGroupCommitSize`, the dispatch loop exits with an
empty dispatch request list.

As a result, no commit is triggered and the tiered consume queue offset
does not advance. The dispatcher repeatedly retries the same message,
preventing it and subsequent messages from being archived.

This change ensures that at least one message is included in each dispatch
batch. A message that reaches or exceeds the group commit size is dispatched
as a single-message batch.

A regression test has also been added to verify that the tiered consume
queue offset advances when a message size equals the configured group
commit size.

1. Added `dispatchMessageAtGroupCommitSizeTest` to reproduce the issue.
   Before the fix, the test timed out because the consume queue offset
   remained at 100 instead of advancing to 101.

2. Verified the regression test after the fix:

   `mvn -pl tieredstore -Dtest=MessageStoreDispatcherImplTest#dispatchMessageAtGroupCommitSizeTest test`

3. Ran the complete tiered storage module test suite:

   `mvn -pl tieredstore test`

   Result: 128 tests run, 0 failures, 0 errors, 0 skipped.

4. Checkstyle and SpotBugs validation also completed successfully.

### DIFF
--- a/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImpl.java
+++ b/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImpl.java
@@ -261,10 +261,10 @@ public class MessageStoreDispatcherImpl extends ServiceThread implements Message
             List<DispatchRequest> dispatchRequestList = new ArrayList<>();
             for (; offset < targetOffset; offset++) {
                 cqUnit = consumeQueue.get(offset);
-                bufferSize += cqUnit.getSize();
-                if (bufferSize >= groupCommitSize) {
+                if (bufferSize > 0 && bufferSize + cqUnit.getSize() >= groupCommitSize) {
                     break;
                 }
+                bufferSize += cqUnit.getSize();
                 message = defaultStore.selectOneMessageByOffset(cqUnit.getPos(), cqUnit.getSize());
                 appendingBufferList.add(message);
 
--- a/tieredstore/src/test/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImplTest.java
+++ b/tieredstore/src/test/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImplTest.java
@@ -165,6 +165,51 @@ public class MessageStoreDispatcherImplTest {
     }
 
     @Test
+    public void dispatchMessageAtGroupCommitSizeTest() throws Exception {
+        MessageStore defaultStore = Mockito.mock(MessageStore.class);
+        Mockito.when(defaultStore.getMinOffsetInQueue(anyString(), anyInt())).thenReturn(100L);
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary

Clean and correct fix for the tiered storage dispatch stall when a message size equals or exceeds `tieredStoreGroupCommitSize`.

### Analysis

The original code adds the message size to `bufferSize` *before* the threshold check, so when the first message alone fills the buffer, the loop breaks with an empty `dispatchRequestList` — no commit is triggered, the consume queue offset never advances, and the dispatcher retries the same message indefinitely.

The fix reorders the check to happen *before* adding the current message, guarded by `bufferSize > 0`. This guarantees at least one message is always included in each dispatch batch:

- **First message, bufferSize == 0:** guard is false → skip break → add message → batch has 1 message ✓
- **Subsequent messages:** accumulate until `bufferSize + nextSize >= groupCommitSize` → break with non-empty batch ✓
- **Oversized single message:** dispatched as a single-message batch, offset advances ✓

### Test Coverage

The regression test `dispatchMessageAtGroupCommitSizeTest` correctly reproduces the stall (offset stuck at 100) and verifies the fix (offset advances to 101). Running the full tiered storage module suite (128 tests) confirms no regressions.

LGTM.

RockteMQ-AI (approved):
### Summary
Fixes a deadlock in the tiered storage dispatcher where a message equal to or larger than `tieredStoreGroupCommitSize` would cause the dispatch loop to exit with an empty batch, preventing the message (and all subsequent ones) from ever being archived.

**Root cause:** The original code incremented `bufferSize` *before* the threshold check, so the very first message could trigger the break with nothing accumulated yet.

**Fix:** Look-ahead pattern — check `bufferSize + cqUnit.getSize()` against the threshold, but only break when `bufferSize > 0` (i.e., at least one message is already in the batch). This guarantees forward progress even for oversized messages.

The included test covers the exact edge case (message size == groupCommitSize) and verifies offset advancement. Clean fix.

LGTM.

---

RockteMQ-AI (approved):
﻿### Summary
Fixes a tiered storage dispatch bug where the first message >= `tieredStoreGroupCommitSize` would cause the dispatch loop to exit with an empty batch, preventing the consume queue offset from advancing.

The fix changes the check from `bufferSize >= groupCommitSize` (after adding current message) to `bufferSize > 0 && bufferSize + cqUnit.getSize() >= groupCommitSize` (before adding), ensuring at least one message per dispatch batch.

Regression test verifies the consume queue offset advances when a message size equals the configured group commit size.

LGTM 👍

---

PR #11001, opened 2026-09-03
[ISSUE #11000] Extract user-property validation in SendMessageActivity#buildMessageProperty into an overridable checkUserProperties method

### LINKED ISSUE #11000: [Enhancement] Extract user-property validation in SendMessageActivity#buildMessageProperty into an overridable checkUserProperties method

### Summary

SendMessageActivity#buildMessageProperty() (grpc v2 producer) inlines the validation of user properties — max property count, reserved system-property keys, and control-character checks — directly inside a large method. This proposal extracts that validation block into a small, standalone protected void checkUserProperties(Map<String, String> userProperties) method that buildMessageProperty() calls, so it can be overridden independently.

### Motivation

Some deployments need custom rules for user message properties — for example, allowing certain properties to contain control characters in specific business scenarios, while keeping every other property/message-building rule untouched.

Today the only extension point is overriding the entire buildMessageProperty() method, which contains a large amount of unrelated logic (tag/keys handling, transaction properties, delay/deliver-timestamp handling, priority, message group, lite topic, trace context, born host/timestamp, etc.). Overriding it just to tweak one validation rule means:
  - Duplicating a large method that isn't the actual customization target.
  - High maintenance cost: any upstream change to buildMessageProperty() requires manually re-syncing the downstream override, risking divergence or missed fixes.

  Extracting the check logic into its own protected method gives a narrow, stable extension seam for this kind of customization without touching the rest of the message-building logic.

Add a new protected method in SendMessageActivity:

  protected void checkUserProperties(Map<String, String> userProperties) {
      ProxyConfig config = ConfigurationManager.getProxyConfig();
      if (userProperties.size() > config.getUserPropertyMaxNum()) {
          throw new GrpcProxyException(Code.MESSAGE_PROPERTIES_TOO_LARGE, "too many user properties, max is " + config.getUserPropertyMaxNum());
      }
      for (Map.Entry<String, String> userPropertiesEntry : userProperties.entrySet()) {
          if (MessageConst.STRING_HASH_SET.contains(userPropertiesEntry.getKey())) {
              throw new GrpcProxyException(Code.ILLEGAL_MESSAGE_PROPERTY_KEY, "property is used by system: " + userPropertiesEntry.getKey());
          }
          if (GrpcValidator.getInstance().containControlCharacter(userPropertiesEntry.getKey())) {
              throw new GrpcProxyException(Code.ILLEGAL_MESSAGE_PROPERTY_KEY, "the key of property cannot contain control character");
          }
          if (GrpcValidator.getInstance().containControlCharacter(userPropertiesEntry.getValue())) {
              throw new GrpcProxyException(Code.ILLEGAL_MESSAGE_PROPERTY_KEY, "the value of property cannot contain control character");
          }
      }
  }

  buildMessageProperty() calls checkUserProperties(userProperties) and keeps the property-size accumulation (used later for the total-size check) in place, unchanged. No behavior change for existing callers — this is a pure refactor/extension-point addition, verified by the existing SendMessageActivityTest suite, which passes unmodified.

- Override buildMessageProperty() entirely — works today, but forces re-implementing/copy-pasting large amounts of unrelated logic and creates ongoing merge/maintenance burden as upstream evolves.
  - Add a config switch/flag (e.g. proxyConfig.isAllowControlCharacterInUserProperty()) to toggle the control-character check — simpler for this one case, but less flexible for other custom validation rules downstream teams may want, and adds proxy-config surface for a narrow use case. An overridable method is more general and doesn't require upstream to anticipate every possible customization.
  - Introduce a pluggable validator interface/SPI for message properties — more flexible but a larger design change; extracting the method first is a minimal, low-risk step that doesn't preclude a future SPI-based approach.

### Additional Context

_No response_

ymwneu:
…ctivity

Extract the user property validation logic (property count limit, reserved-key check, control character check) from buildMessageProperty() into a standalone protected checkUserProperties() method. The property size accumulation stays in buildMessageProperty().

This allows subclasses to customize validation (e.g. allow control characters in properties for specific scenarios) by overriding checkUserProperties() alone, instead of having to override the much larger buildMessageProperty() method.

- Fixes #11000 

### Brief Description

SendMessageActivity#buildMessageProperty() (grpc v2 producer) inlines the validation of user properties — max property count, reserved system-property keys, and control-character checks — directly inside a large method. This proposal extracts that validation block into a small, standalone protected void checkUserProperties(Map<String, String> userProperties) method that buildMessageProperty() calls, so it can be overridden independently.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/producer/SendMessageActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/producer/SendMessageActivity.java
@@ -195,12 +195,8 @@ public class SendMessageActivity extends AbstractMessagingActivity {
         }
     }
 
-    protected Map<String, String> buildMessageProperty(ProxyContext context, apache.rocketmq.v2.Message message, String producerGroup) {
-        long userPropertySize = 0;
+    protected void checkUserProperties(Map<String, String> userProperties) {
         ProxyConfig config = ConfigurationManager.getProxyConfig();
-        org.apache.rocketmq.common.message.Message messageWithHeader = new org.apache.rocketmq.common.message.Message();
-        // set user properties
-        Map<String, String> userProperties = message.getUserPropertiesMap();
         if (userProperties.size() > config.getUserPropertyMaxNum()) {
             throw new GrpcProxyException(Code.MESSAGE_PROPERTIES_TOO_LARGE, "too many user properties, max is " + config.getUserPropertyMaxNum());
         }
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Extracts user-property validation logic from `buildMessageProperty()` into a standalone `protected checkUserProperties()` method. This follows the Open/Closed Principle — subclasses can now customize validation (e.g., allow control characters) by overriding just the validation method instead of the entire `buildMessageProperty()`.

The refactoring is clean: validation logic is moved as-is, no behavioral change. The property size accumulation remains in `buildMessageProperty()` as intended.

LGTM.

---

RockteMQ-AI (approved):
﻿### Summary
Extracts user-property validation logic into a standalone `protected void checkUserProperties()` method, allowing subclasses to customize validation by overriding just this method instead of the larger `buildMessageProperty()`.

Clean, minimal refactoring (12 additions, 5 deletions). No behavior change.

LGTM 👍

---

lizhimins (approved):

PR #11004, opened 2026-09-03
[ISSUE #11003] Reduce allocations in hot-path request headers' toString()

### LINKED ISSUE #11003: [Enhancement] Reduce allocations in hot-path request headers' toString() by replacing Guava MoreObjects.toStringHelper

### Summary

Replace Guava `MoreObjects.toStringHelper` with a pre-sized `StringBuilder` in the `toString()` of the high-frequency request header classes under `remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/`. The rendered output stays byte-for-byte identical; the change removes roughly N+3 short-lived allocations and an O(N) linked-list traversal per call for a header with N fields.

Scope (14 send / consume / ack / offset hot-path headers): `SendMessageRequestHeader`, `SendMessageRequestHeaderV2`, `PopMessageRequestHeader`, `PopLiteMessageRequestHeader`, `PullMessageRequestHeader`, `NotificationRequestHeader`, `AckMessageRequestHeader`, `ChangeInvisibleTimeRequestHeader`, `QueryConsumerOffsetRequestHeader`, `UpdateConsumerOffsetRequestHeader`, `GetMaxOffsetRequestHeader`, `GetMinOffsetRequestHeader`, `ConsumerSendMsgBackRequestHeader`, `RecallMessageRequestHeader`.

Measured on the module's current build target (JDK 8 / `target 1.8`): **1.62x - 2.60x faster `toString()` and 32% - 48% less allocation per call**.

### Motivation

These headers render themselves with `MoreObjects.toStringHelper`, e.g. `SendMessageRequestHeader`:

```java
return MoreObjects.toStringHelper(this)
    .add("producerGroup", producerGroup)
    .add("topic", topic)
    ... // 13 fields in total
    .toString();
```

Per invocation this allocates, before the resulting `String` is even built:

1. one `ToStringHelper` object;
2. one `ValueHolder` head plus **one `ValueHolder` node per `.add(...)`**, forming a singly linked list;
3. `getClass().getSimpleName()`, which itself allocates (`getName()` + `substring`);
4. an internal `StringBuilder(32)` that grows through repeated `Arrays.copyOf` whenever the rendering exceeds 32 characters;

and then traverses that linked list to render. For a header with N fields this is about **N + 3 short-lived objects plus an O(N) traversal per call**.

`toString()` on these headers sits on hot paths: it is invoked whenever a header is rendered for logging or diagnostics (request logging, exception messages, troubleshooting output). At high message throughput this becomes a continuous stream of short-lived garbage and a measurable amount of CPU spent purely on rendering, which also shows up as extra young-generation collection work.

Use a **pre-sized `StringBuilder`** with direct `append` calls:

- **Plain headers** - a single chained `append` sequence. The initial capacity is derived from the class's fixed skeleton (class name + field names + `=` / `, ` separators) plus an allowance for variable-length values such as `properties`, `subscription` and `extraInfo`. Pre-sizing matters here: a default-capacity builder needs 5-6 growth-and-copy steps to reach a ~350 character rendering.
- **Headers using `.omitNullValues()`** (`AckMessageRequestHeader`, `ChangeInvisibleTimeRequestHeader`, `NotificationRequestHeader`) - conditional appends that skip null values, preserving the exact omission semantics. Primitive-typed fields are always emitted, which matches today's behaviour since they autobox to a non-null value.
- **Conditional entries** such as `.add("isLiteConsumer", isLiteConsumer ? true : null)` in `NotificationRequestHeader` are preserved by evaluating the expression into a local and null-checking it, so the entry is still rendered only when `true`.

The rendered strings must remain **byte-for-byte identical**, because they end up in logs that people and external tooling parse.

**1. Plain `+` string concatenation.** More concise, and on a JDK 9+ bytecode target it compiles to `invokedynamic` / `StringConcatFactory.makeConcatWithConstants`, which beats anything hand-written - measured **4.53x** and **-73%** allocation on JDK 21.

However this module builds with `maven.compiler.source/target = 1.8`, where `javac` lowers `+` to `new StringBuilder()` with the default capacity of 16. For headers whose rendering runs to a few hundred characters (those carrying `properties` / `subscription` / `extraInfo`) that means 5-6 growth-and-copy steps, which cancels out most of the benefit of dropping Guava. Measured for `SendMessageRequestHeader` at `target 1.8`:

| form | alloc/call | throughput vs Guava |
|---|---|---|
| `+` concatenation | 3664 B (-2.6%) | 1.27x |
| pre-sized `StringBuilder` | 1944 B (-48.3%) | **2.07x** |

So plain concatenation was rejected for now: it is the worse option under the project's current build target. The pre-sized builder also stays good on newer targets (1.95x on JDK 21), so it does not become a liability. **If the project later raises the bytecode target, switching these to plain `+` concatenation would be a worthwhile follow-up.**

**2. Removing or gating the `toString()` calls at the call sites.** Out of scope: it changes observable logging behaviour rather than making the existing rendering cheaper.

**3. Also converting the remaining ~10 header classes that use `toStringHelper`.** Left out deliberately - those are admin / low-frequency headers, and keeping them out makes this change easier to review. They can follow up separately.

#### Measured effect

Microbenchmark against the real classes, JDK 8 runtime, the module's `target 1.8` bytecode, realistic field values (including a ~120 character `properties`). Allocation measured with `ThreadMXBean.getThreadAllocatedBytes`:

| Header | alloc before | alloc after | reduction | throughput |
|---|---|---|---|---|
| `SendMessageRequestHeader` (13 fields) | 3760 B | 1944 B | **-48.3%** | **2.07x** |
| `PullMessageRequestHeader` (15 fields) | 4072 B | 2328 B | -42.8% | 1.79x |
| `PopMessageRequestHeader` (12 fields) | 2456 B | 1432 B | -41.7% | 2.08x |
| `AckMessageRequestHeader` (6 fields, omitNullValues) | 1848 B | 1176 B | -36.4% | 1.80x |
| `ChangeInvisibleTimeRequestHeader` (8 fields, omitNullValues) | 2088 B | 1416 B | -32.2% | 1.62x |
| `QueryConsumerOffsetRequestHeader` (4 fields) | 1104 B | 608 B | -44.9% | **2.60x** |

#### Correctness verification

A differential test loads the original and the modified compiled classes in two separate class loaders, applies identical field values to both instances and compares `toString()`. Value plans cover: all-default, all-null, all-non-null, **each field individually null**, **each field individually set**, boolean fields in both states (this is what covers the conditional `isLiteConsumer` entry), plus 40,000 randomized adversarial value sets per class (null, empty, `\r\n\t`, other control characters, 300- and 2000-character strings, non-ASCII, `Integer`/`Long` `MAX_VALUE`/`MIN_VALUE`).

- Differential test: **560,483 cases, 0 mismatches**
- `mvn -pl remoting test`: **174 tests, 0 failures**
- checkstyle (`style/rmq_checkstyle.xml`, `validate` phase): **0 violations**

As a check on the test itself, deleting the conditional `isLiteConsumer` handling makes the differential test fail immediately and prints the exact divergence, confirming it can detect this class of regression rather than passing vacuously.

Verified against `develop` at `bee586bcd`.

qianye1001:

- Fixes #11003

### Brief Description

Replaces Guava `MoreObjects.toStringHelper` with a pre-sized `StringBuilder` in the `toString()` of the 10 request headers on the core send / consume / ack / offset-commit data path, under `remoting/.../protocol/header/`.

The helper form allocates, per call and before the resulting `String` is even built:

1. one `ToStringHelper` object;
2. one `ValueHolder` head plus **one `ValueHolder` node per `.add(...)`** (a singly linked list);
3. `getClass().getSimpleName()`, which itself allocates (`getName()` + `substring`);
4. an internal `StringBuilder(32)` that grows through repeated `Arrays.copyOf` once the rendering exceeds 32 characters;

and then traverses that list to render. For a header with N fields that is roughly N + 3 short-lived objects plus an O(N) traversal per call. These `toString()` methods are on hot paths (request logging, exception messages, troubleshooting output), so at high throughput this is a steady stream of short-lived garbage and CPU spent purely on rendering.

The replacement uses a single chained `append` sequence with a per-class initial capacity.

Semantics preserved exactly:

- **`.omitNullValues()`** (`AckMessageRequestHeader`, `ChangeInvisibleTimeRequestHeader`, `NotificationRequestHeader`) becomes conditional appends that skip null entries. Primitive-typed fields are always emitted, matching today's behaviour since they autobox to a non-null value.
- **Conditional entries** - `NotificationRequestHeader`'s `.add("isLiteConsumer", isLiteConsumer ? true : null)` is preserved by evaluating the expression into a local and null-checking it, so the entry is still rendered only when `true`.
- The unused `com.google.common.base.MoreObjects` import is dropped from each touched file.

#### Capacity estimate

The initial capacity is derived per class from the fixed skeleton (class name + field names + `=` / `, ` separators) plus a per-field value allowance based on realistic widths rather than a flat guess:

| field kind | allowance |
|---|---|
| topic / group strings | 50 |
| `properties` / `subscription` / `extraInfo` | 128 |
| `queueId`, `defaultTopicQueueNums` | 3 |
| `maxMsgNums` / `maxMsgNum` | 2 |
| `long` / `Long` (timestamps, offsets) | 20 |
| other `int` / `Integer` | 11 |
| `boolean` / `Boolean` | 5 |
| other strings | 16 |

`SendMessageRequestHeaderV2` uses compressed field names (`a`..`n`), so the allowance is applied by semantic field (`a`=producerGroup, `b`=topic, `c`=defaultTopic, `i`=properties, ...) rather than by identifier.

This matters more than it looks. Measured on a 348-character rendering, allocation and throughput by initial capacity:

| capacity | alloc vs exact-fit | throughput vs exact-fit | young GC count |
|---|---|---|---|
| 16 (what `javac` emits for `+`) | 229% | 0.56x | 40 |
| exact fit | 100% | 1.00x | 18 |
| 1.6x exact | 127% | 0.88x | 23 |
| 3.4x exact | 202% | 0.58x | 37 |
| 13.8x exact | 635% | 0.29x | 117 |

`new char[cap]` is allocated and zero-filled in proportion to the capacity, while `toString()` copies only the real length - so slack is pure waste and it accelerates eden filling. Over-sizing is therefore not free, which is why the estimate is per-field.

#### Measured effect

On the real classes, JDK 8 runtime, this module's `target 1.8` bytecode, realistic field values (~50-char topic/group, ~120-char `properties`). Allocation measured with `ThreadMXBean.getThreadAllocatedBytes`:

| Header | alloc before | alloc after | reduction | throughput |
|---|---|---|---|---|
| `SendMessageRequestHeader` (send) | 4056 B | 2432 B | -40.0% | 1.91x |
| `SendMessageRequestHeaderV2` (sendV2) | 4520 B | 2000 B | -55.8% | 2.29x |
| `PopMessageRequestHeader` (pop) | 3744 B | 1552 B | -58.5% | 2.55x |
| `PopLiteMessageRequestHeader` (popLite) | 3472 B | 1360 B | -60.8% | **2.81x** |
| `PullMessageRequestHeader` (pull) | 6624 B | 3008 B | -54.6% | 2.21x |
| `NotificationRequestHeader` (notification, omitNullValues) | 3424 B | 2632 B | -23.1% | 1.43x |
| `AckMessageRequestHeader` (ack, omitNullValues) | 3328 B | 1632 B | -51.0% | 2.14x |
| `ChangeInvisibleTimeRequestHeader` (omitNullValues) | 3568 B | 1928 B | -46.0% | 1.96x |
| `UpdateConsumerOffsetRequestHeader` (offset commit) | 1944 B | 1848 B | -4.9% | 1.46x |
| `ConsumerSendMsgBackRequestHeader` (retry) | 3288 B | 1312 B | -60.1% | 2.79x |

Note on the choice of form: plain `+` concatenation is more concise and on a JDK 9+ bytecode target compiles to `invokedynamic` / `StringConcatFactory.makeConcatWithConstants`, which beats anything hand-written - measured 4.53x and -73% allocation on JDK 21. But at this module's `target 1.8`, `javac` lowers `+` to `new StringBuilder()` with the default capacity of 16, so the growth-and-copy steps cancel out most of the benefit of dropping Guava: measured on a 348-char rendering, `+` gives 1.27x and -2.6% allocation versus 2.07x and -48.3% for the pre-sized builder. The pre-sized builder also stays good on newer targets (1.95x on JDK 21), so it is not a liability; if the project later raises the bytecode target, switching these to plain concatenation would be a worthwhile follow-up.

Scope note: `GetMaxOffsetRequestHeader`, `GetMinOffsetRequestHeader`, `QueryConsumerOffsetRequestHeader` and `RecallMessageRequestHeader` were deliberately dropped from this change and are byte-identical to `develop`. Together with the admin/low-frequency headers, the remaining `toStringHelper` users can be a follow-up.

The rendered strings end up in logs that people and external tooling parse, so the requirement is that output stays **byte-for-byte identical**. Testing focused on proving that rather than on eyeballing the diff.

**1. Differential test against the original compiled bytecode (primary evidence).** A pristine worktree at the merge base (`bee586bcd`) and the modified tree are compiled into two separate output directories, then loaded into two independent class loaders in one JVM. For each header the test instantiates both versions, applies an identical field-value plan to both via reflection, and compares `toString()`. This compares against the real original implementation, so it does not depend on any assumption about how Guava formats things.

Value plans per class: all-default (exercising field initializers such as `order = Boolean.FALSE`, `suspend = false`, `unitMode = false`), all-null, all-non-null, **each field individually null with all others set**, **each field individually set with all others null**, boolean/Boolean fields in both states (this is what covers the conditional `isLiteConsumer` entry), and 40,000 randomized adversarial value sets per class drawing from null, empty string, `\r`, `\n`, `\t`, `\r\n\t`, other control characters (`\u0001`, `\u0002`), non-ASCII and surrogate-pair text, 300- and 2000-character strings, `%RETRY%`-prefixed groups, hex message ids, and `Integer`/`Long` `MAX_VALUE`/`MIN_VALUE`.

Result: **400,380 cases, 0 mismatches** across the 10 classes.

**2. Negative control on the test itself.** An earlier iteration of this change silently dropped the conditional `isLiteConsumer` entry (its `.add(...)` value is a ternary rather than a plain field reference, so a naive rewrite lost it). Deleting that handling from the final code makes the differential test fail immediately and print the exact divergence:

```
orig=NotificationRequestHeader{consumerGroup=V, topic=V, queueId=7, pollTime=77, bornTime=77, order=true, attemptId=V, isLiteConsumer=true, clientId=V}
new =NotificationRequestHeader{consumerGroup=V, topic=V, queueId=7, pollTime=77, bornTime=77, order=true, attemptId=V, clientId=V}
```

This confirms the test detects this class of regression instead of passing vacuously.

**3. Bytecode-level scope check.** Decompiling the built classes confirms `toString()` in the 10 touched headers has zero `MoreObjects` references, while the 4 deliberately-excluded headers still show theirs (5, 4, 6 and 5 references respectively) - i.e. nothing outside the intended scope was altered. For an earlier revision of this change the same technique confirmed that all non-`toString` methods were byte-identical to the original.

**4. Existing test suite and style gates.**

- `mvn -pl remoting test`: **174 tests, 0 failures, 0 errors** (includes `SendMessageRequestHeaderV2Test`, `RpcRequestHeaderTest`, `ProxyProtocolTest`, `FastCodesHeaderTest`).
- checkstyle with the project's `style/rmq_checkstyle.xml` at the `validate` phase: **0 violations** (this also confirms dropping the now-unused `MoreObjects` import is clean).
- `mvn -pl remoting -am compile`: BUILD SUCCESS.

**5. Diff scope.** `git diff --stat origin/develop` touches exactly the 10 intended header files and nothing else; each file's change is confined to its `toString()` body plus removal of the unused import.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/AckMessageRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/AckMessageRequestHeader.java
@@ -16,7 +16,6 @@
  */
 package org.apache.rocketmq.remoting.protocol.header;
 
-import com.google.common.base.MoreObjects;
 import org.apache.rocketmq.common.action.Action;
 import org.apache.rocketmq.common.action.RocketMQAction;
 import org.apache.rocketmq.common.resource.ResourceType;
@@ -98,14 +97,52 @@ public class AckMessageRequestHeader extends TopicQueueRequestHeader {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
-            .add("consumerGroup", consumerGroup)
-            .add("topic", topic)
-            .add("queueId", queueId)
-            .add("extraInfo", extraInfo)
-            .add("offset", offset)
-            .add("liteTopic", liteTopic)
-            .omitNullValues()
-            .toString();
+        StringBuilder sb = new StringBuilder(352);
+        sb.append("AckMessageRequestHeader{");
+        boolean first = true;
+        if (consumerGroup != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("consumerGroup=").append(consumerGroup);
+            first = false;
+        }
+        if (topic != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("topic=").append(topic);
+            first = false;
+        }
+        if (queueId != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("queueId=").append(queueId);
+            first = false;
+        }
+        if (extraInfo != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("extraInfo=").append(extraInfo);
+            first = false;
+        }
+        if (offset != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("offset=").append(offset);
+            first = false;
+        }
+        if (liteTopic != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("liteTopic=").append(liteTopic);
+            first = false;
+        }
+        sb.append('}');
+        return sb.toString();
     }
 }
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/AckMessageRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/AckMessageRequestHeader.java
@@ -97,7 +97,7 @@ public class AckMessageRequestHeader extends TopicQueueRequestHeader {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder(352);
+        StringBuilder sb = new StringBuilder(391);
         sb.append("AckMessageRequestHeader{");
         boolean first = true;
         if (consumerGroup != null) {
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ChangeInvisibleTimeRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ChangeInvisibleTimeRequestHeader.java
@@ -124,7 +124,7 @@ public class ChangeInvisibleTimeRequestHeader extends TopicQueueRequestHeader {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder(384);
+        StringBuilder sb = new StringBuilder(451);
         sb.append("ChangeInvisibleTimeRequestHeader{");
         boolean first = true;
         if (consumerGroup != null) {
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ChangeInvisibleTimeRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ChangeInvisibleTimeRequestHeader.java
@@ -16,7 +16,6 @@
  */
 package org.apache.rocketmq.remoting.protocol.header;
 
-import com.google.common.base.MoreObjects;
 import org.apache.rocketmq.common.action.Action;
 import org.apache.rocketmq.common.action.RocketMQAction;
 import org.apache.rocketmq.common.resource.ResourceType;
@@ -125,16 +124,64 @@ public class ChangeInvisibleTimeRequestHeader extends TopicQueueRequestHeader {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
-            .add("consumerGroup", consumerGroup)
-            .add("topic", topic)
-            .add("queueId", queueId)
-            .add("extraInfo", extraInfo)
-            .add("offset", offset)
-            .add("invisibleTime", invisibleTime)
-            .add("liteTopic", liteTopic)
-            .add("suspend", suspend)
-            .omitNullValues()
-            .toString();
+        StringBuilder sb = new StringBuilder(384);
+        sb.append("ChangeInvisibleTimeRequestHeader{");
+        boolean first = true;
+        if (consumerGroup != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("consumerGroup=").append(consumerGroup);
+            first = false;
+        }
+        if (topic != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("topic=").append(topic);
+            first = false;
+        }
+        if (queueId != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("queueId=").append(queueId);
+            first = false;
+        }
+        if (extraInfo != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("extraInfo=").append(extraInfo);
+            first = false;
+        }
+        if (offset != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("offset=").append(offset);
+            first = false;
+        }
+        if (invisibleTime != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("invisibleTime=").append(invisibleTime);
+            first = false;
+        }
+        if (liteTopic != null) {
+            if (!first) {
+                sb.append(", ");
+            }
+            sb.append("liteTopic=").append(liteTopic);
+            first = false;
+        }
+        if (!first) {
+            sb.append(", ");
+        }
+        sb.append("suspend=").append(suspend);
+        first = false;
+        sb.append('}');
+        return sb.toString();
     }
 }
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ConsumerSendMsgBackRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ConsumerSendMsgBackRequestHeader.java
@@ -106,7 +106,7 @@ public class ConsumerSendMsgBackRequestHeader extends RpcRequestHeader {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder(384);
+        StringBuilder sb = new StringBuilder(284);
         sb.append("ConsumerSendMsgBackRequestHeader{offset=").append(offset)
             .append(", group=").append(group)
             .append(", delayLevel=").append(delayLevel)
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ConsumerSendMsgBackRequestHeader.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ConsumerSendMsgBackRequestHeader.java
@@ -17,7 +17,6 @@
 
 package org.apache.rocketmq.remoting.protocol.header;
 
-import com.google.common.base.MoreObjects;
 import org.apache.rocketmq.common.action.Action;
 import org.apache.rocketmq.common.action.RocketMQAction;
 import org.apache.rocketmq.common.resource.ResourceType;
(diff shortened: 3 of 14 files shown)


RockteMQ-AI (approved):
### Summary

Replaces Guava `MoreObjects.toStringHelper` with pre-sized `StringBuilder` in 14 hot-path request header `toString()` methods. Allocation reduced by 36–48% with 1.8–2.1x throughput improvement. Semantics are preserved correctly:

- Primitive-typed fields (`int queueId`, `long pollTime`, `long bornTime`) are always emitted — matches original behavior since they can never be null.
- Nullable fields (`String`, `Integer`, `Boolean`) are properly null-guarded with the `first` flag pattern, preserving the original `omitNullValues()` contract.
- The `isLiteConsumer ? true : null` conditional in `NotificationRequestHeader` is correctly materialized into a local `Object` and null-checked.

Benchmarks are convincing (560K+ differential test cases, per-class allocation measurements). CI is still running but `misspell-check` and `check-license` already pass.

LGTM — clean, well-tested, and a meaningful reduction in GC pressure on the request path.

---

RockteMQ-AI (approved):
### Summary
Replaces Guava `MoreObjects.toStringHelper()` with manual `StringBuilder`-based `toString()` across 10 hot-path request header classes. Clean mechanical refactor that reduces object allocation.

### Review Notes
- Pre-allocated `StringBuilder` sizes (e.g., 391, 191) are reasonable and avoid resizing
- Null-field omission logic correctly mirrors Guava's `omitNullValues()` behavior
- Output format (`ClassName{field=value, ...}`) is compatible with the original
- Primitive fields in `NotificationRequestHeader` (queueId, pollTime, bornTime) could skip the `!first` guard since they're never null, but this is cosmetic and doesn't affect correctness

LGTM — straightforward performance improvement with no behavioral changes.

---

RockteMQ-AI (approved):
### Summary
Replaces Guava MoreObjects.toStringHelper with pre-sized StringBuilder in 10 hot-path request headers. Exceptional PR quality.

**Key strengths:**
- 40-60% allocation reduction per toString() call on send/consume/ack/offset-commit paths
- Per-class capacity estimates based on realistic field widths, not flat guesses
- Semantics preserved exactly: omitNullValues() becomes conditional appends, primitive fields always emitted
- 400,380 differential test cases (including 40,000 adversarial random sets per class) with 0 mismatches
- Negative control test verifies the test catches regressions (the isLiteConsumer case)
- Clear scope: 4 deliberately-excluded headers still show their MoreObjects references

**Minor observation:**
The pre-sized StringBuilder approach is optimal for JDK 8 target. When the project eventually moves to JDK 9+ bytecode, plain concatenation would be even better (invokedynamic/StringConcatFactory). The PR body notes this as a worthwhile follow-up.

LGTM — exemplary PR with thorough testing and clear documentation 👍

---

PR #11005, opened 2026-09-03
[ISSUE #10898] Unblock the NettyEventExecutor event queue on shutdown

### LINKED ISSUE #10898: [Enhancement] NettyEventExecutor shutdown waits for the eventQueue poll timeout because wakeup() cannot release it

wang-jiahua:

Fixes #10898

### Brief Description

`NettyEventExecutor` blocks in `eventQueue.poll(3000ms)` instead of `waitForRunning`, so the `LockSupport.unpark` performed by `ServiceThread.shutdown()` cannot release it: the thread only notices the stopped flag once the poll expires, and every shutdown of a remoting instance pays up to 3 seconds for nothing.

This PR overrides `wakeup()` to offer a sentinel `NettyEvent` into the queue after `super.wakeup()`, making the blocking poll return immediately; the dispatch loop skips the sentinel. No behavior change apart from the faster exit.

- New deterministic test `testNettyEventExecutorShutdownDoesNotWaitForPollTimeout`: on current develop it fails after 3.66s (the thread waits out the full poll timeout); with this patch it passes in under 1s. `NettyRemotingAbstractTest` 6/6, checkstyle clean.
- Server-side isolated A/B on an 8C32G ECS (Dragonwell 21), driving `NettyEventExecutor` start/shutdown directly, 5 runs per arm:
  - base (current jar): 2700.3 / 2700.4 / 2700.3 / 2700.3 / 2700.3 ms
  - patched: 0.4 / 0.3 / 0.2 / 0.2 / 0.2 ms
- `NettyEventExecutor` is a package-private inner class, so there is no external subclassing surface affected by the new `wakeup()` override.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -765,6 +765,12 @@ public abstract class NettyRemotingAbstract {
     class NettyEventExecutor extends ServiceThread {
         private final LinkedBlockingQueue<NettyEvent> eventQueue = new LinkedBlockingQueue<>();
 
+        /**
+         * Sentinel used to unblock {@link #eventQueue} on shutdown. It carries no channel event and
+         * is skipped by the dispatch loop.
+         */
+        private final NettyEvent wakeupEvent = new NettyEvent(null, null, null);
+
         public void putNettyEvent(final NettyEvent event) {
             int currentSize = this.eventQueue.size();
             int maxSize = 10000;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix for the 3-second shutdown delay in `NettyEventExecutor`. The sentinel-event approach to unblock `eventQueue.poll()` is a well-known pattern, correctly implemented here: the `wakeupEvent` is `final` (safe publication), `offer()` is thread-safe on `LinkedBlockingQueue`, and the dispatch loop properly skips the sentinel via reference inequality. `super.wakeup()` is called first to preserve the existing `LockSupport.unpark` path. The deterministic test and A/B benchmarks (2700ms → 0.2ms) clearly validate the fix.

Fixes #10898. LGTM.

---

RockteMQ-AI (approved):
### Summary
Fixes a 3-second shutdown delay in  caused by  blocking the thread while  cannot release it via .

The sentinel  pattern is clean:
-  override offers the sentinel to the queue, making  return immediately
- Dispatch loop skips the sentinel via identity check ()
- No external subclassing surface (package-private inner class)

Benchmark: shutdown drops from ~2700ms to <1ms. Deterministic test verifies <1s shutdown.

LGTM 👍

---

PR #11006, opened 2026-09-03
[ISSUE #11002] Cache topic bytes in MessageExtEncoder and CRC32 property key bytes in MessageDecoder

### LINKED ISSUE #11002: [Enhancement] Avoid per-message getBytes() for topic and CRC32 property key on the broker write path

### Summary

Cache the topic bytes in `MessageExtEncoder` and the CRC32 property-key bytes in `MessageDecoder` to avoid two per-message `String.getBytes()` allocations on the broker write path.

### Motivation

- `MessageExtEncoder.encode()` / `encodeWithoutProperties()` call `msgInner.getTopic().getBytes(UTF_8)` for every message. Broker traffic is highly repetitive per thread, and the encoder instance is already `ThreadLocal`, so the same topic string is re-encoded over and over.
- `MessageDecoder.createCrc32()` re-encodes the constant property key `MessageConst.PROPERTY_CRC32` on every message when CRC32 stamping is enabled.

### Solution

- `MessageExtEncoder`: keep a single-slot `cachedTopic`/`cachedTopicData` pair (no synchronization needed — the encoder is `ThreadLocal`). Hit returns the cached bytes; miss falls back to `getBytes` and refreshes the slot, so mixed-topic traffic is never worse than today apart from one string comparison.
- `MessageDecoder`: pre-encode `PROPERTY_CRC32` into a `private static final byte[]`.

### Verification

- `AppendCallbackTest` 4/4, `AppendPropCRCTest` 2/2 (covers the CRC32 constant path), `LmqDispatchTest` 5/5, `CompactionLogTest` 4/4, `MessageDecoderTest` 7/7; checkstyle clean.
- 4-node cluster A/B (256-thread sync producer, 1KB, per-arm clean store + restart): broker TPS and young GC per million messages are flat versus baseline — the saving (one short `byte[]` per message) is below GC-count resolution at this load, so this is a cleanup-level allocation reduction, not a measurable throughput win.

wang-jiahua:

Fixes #11002

### Brief Description

Two per-message `String.getBytes()` allocations on the broker write path are avoided:

- `MessageExtEncoder`: the topic is re-encoded to UTF-8 bytes for every message. The encoder instance is already `ThreadLocal`, so a single-slot `cachedTopic`/`cachedTopicData` pair is safe without synchronization: a hit returns the cached bytes; a miss falls back to `getBytes` and refreshes the slot.
- `MessageDecoder.createCrc32()`: the constant property key `PROPERTY_CRC32` is now pre-encoded into a `private static final byte[]`.

- Functional: `AppendCallbackTest` 4/4, `AppendPropCRCTest` 2/2 (covers the CRC32 constant path), `LmqDispatchTest` 5/5, `CompactionLogTest` 4/4, `MessageDecoderTest` 7/7; checkstyle clean.
- Microbenchmark on an 8C32G ECS (Dragonwell 21), 20M ops, thread-allocated-bytes measured:
  - baseline `getBytes`: 8.14 ns/op, 56 B/op
  - cache hit (same topic, different String instance, `equals` path): 2.63 ns/op, 0 B/op
  - worst case (two topics strictly alternating, equal length differing only in the last char): 13.84 ns/op, 56 B/op — i.e. a miss costs one extra `String.equals` (~5.7ns) on top of the original `getBytes`; the break-even hit rate is ~51%, and per-thread broker traffic is typically far more repetitive than that.
- End-to-end 4-node cluster A/B (256-thread sync producer, 1KB, per-arm clean store + restart): broker TPS and young GC per million messages flat versus baseline — the saving is below GC-count resolution at this load, so this is a cleanup-level allocation reduction and is labeled as such.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
@@ -44,6 +44,9 @@ public class MessageDecoder {
     public final static int MESSAGE_PHYSIC_OFFSET_POSITION = 28;
     public final static int MESSAGE_STORE_TIMESTAMP_POSITION = 56;
 
+    // Pre-encoded constant to avoid per-message getBytes() allocation in createCrc32()
+    private static final byte[] PROPERTY_CRC32_BYTES = MessageConst.PROPERTY_CRC32.getBytes(StandardCharsets.UTF_8);
+
     // Set message magic code v2 if topic length > 127
     public final static int MESSAGE_MAGIC_CODE = -626843481;
     public final static int MESSAGE_MAGIC_CODE_V2 = -626843477;
@@ -154,7 +157,7 @@ public class MessageDecoder {
     }
 
     public static void createCrc32(final ByteBuffer input, int crc32) {
-        input.put(MessageConst.PROPERTY_CRC32.getBytes(StandardCharsets.UTF_8));
+        input.put(PROPERTY_CRC32_BYTES);
         input.put((byte) NAME_VALUE_SEPARATOR);
         for (int i = 0; i < 10; i++) {
             byte b = '0';
@@ -168,7 +171,7 @@ public class MessageDecoder {
     }
 
     public static void createCrc32(final ByteBuf input, int crc32) {
-        input.writeBytes(MessageConst.PROPERTY_CRC32.getBytes(StandardCharsets.UTF_8));
+        input.writeBytes(PROPERTY_CRC32_BYTES);
         input.writeByte((byte) NAME_VALUE_SEPARATOR);
         for (int i = 0; i < 10; i++) {
             byte b = '0';
--- a/store/src/main/java/org/apache/rocketmq/store/MessageExtEncoder.java
+++ b/store/src/main/java/org/apache/rocketmq/store/MessageExtEncoder.java
@@ -41,6 +41,11 @@ public class MessageExtEncoder {
     private final int crc32ReservedLength;
     private MessageStoreConfig messageStoreConfig;
 
+    // Cache topic bytes to avoid per-message getBytes() allocation.
+    // Safe without synchronization because MessageExtEncoder is ThreadLocal.
+    private String cachedTopic;
+    private byte[] cachedTopicData;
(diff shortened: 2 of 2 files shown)


common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
```diff
@@ -44,6 +44,9 @@ public class MessageDecoder {
     public final static int MESSAGE_PHYSIC_OFFSET_POSITION = 28;
     public final static int MESSAGE_STORE_TIMESTAMP_POSITION = 56;
 
+    // Pre-encoded constant to avoid per-message getBytes() allocation in createCrc32()
```
RockteMQ-AI, line 47: Minor consistency nit: `MessageDecoder` already declares `public final static Charset CHARSET_UTF8 = StandardCharsets.UTF_8` at line 41. Consider using `CHARSET_UTF8` here instead of `StandardCharsets.UTF_8` to match the existing convention in this file (e.g. lines 150, 198, 201, 317).

RockteMQ-AI (approved):
### Summary
Performance optimization that caches topic UTF-8 bytes in the ThreadLocal `MessageExtEncoder` and pre-encodes the CRC32 property key in `MessageDecoder`. The ThreadLocal safety argument is sound, the single-slot cache design is appropriate for per-thread broker traffic patterns, and the microbenchmark data is thorough.

Fixes #11002.

---

RockteMQ-AI (approved):
### Summary
Two per-message `String.getBytes()` allocations eliminated on the broker write path. Clean, well-benchmarked optimization.

- **MessageExtEncoder**: Single-slot `cachedTopic`/`cachedTopicData` cache is safe because the encoder is `ThreadLocal`. The `equals()` fallback handles different String instances with the same value. Worst-case (alternating topics) adds only ~5.7ns per miss on top of the original `getBytes`, with a ~51% break-even hit rate — well below typical per-thread traffic repetition.
- **MessageDecoder**: `PROPERTY_CRC32_BYTES` as `static final byte[]` is a trivial and correct optimization for the constant key.

LGTM 👍

---

PR #11014, opened 2026-09-04
[ISSUE #11013] Stop MessageRocksDBStorage flush scheduler on shutdown

### LINKED ISSUE #11013: [Bug] MessageRocksDBStorage accumulates timer-WAL flush tasks after reload

Ubuntu 22.04.5 LTS, Linux x86_64.

### RocketMQ version

Branch: `develop`  
Version: `5.5.1`  
Git commit id: `e90303810a14`

### JDK Version

OpenJDK 8u502.

### Describe the Bug

`MessageRocksDBStorage` creates a `ScheduledExecutorService` and registers a
periodic timer-WAL flush task in `postLoad()`:

```java
private final ScheduledExecutorService scheduler =
    Executors.newScheduledThreadPool(1);

scheduler.scheduleAtFixedRate(this::flushTimerWal, 5, 5, TimeUnit.MINUTES);
```

However, `preShutdown()` does not cancel the scheduled task or shut down the
scheduler. `AbstractRocksDBStorage.reloadRocksdb()` uses the existing
`shutdown()` followed by `start()` lifecycle. Because the same scheduler
remains active, every successful reload registers another periodic flush task
on the same executor.

This also leaves a non-daemon scheduler and its delayed task alive after the
RocksDB resources have been closed.

### Steps to Reproduce

1. Create a `MessageRocksDBStorage` with a temporary `MessageStoreConfig`.
2. Inspect its private scheduler as a `ScheduledThreadPoolExecutor` using a
   focused test or debugger. After construction, one periodic task is queued.
3. Call `storage.shutdown()`.
4. Inspect the scheduler again and observe that it is not shut down and that
   its queue still contains one periodic task.
5. Call `storage.start()` to simulate the lifecycle used by
   `AbstractRocksDBStorage.reloadRocksdb()`.
6. Call `storage.shutdown()` and `storage.start()` once more.
7. Inspect the scheduler after each start.

The affected code shows the following stable state transition:

| Lifecycle state | Scheduler state | Queued periodic tasks |
| --- | --- | ---: |
| After construction | active | 1 |
| After first shutdown | not shut down | 1 |
| After first start | active | 2 |
| After second shutdown | not shut down | 2 |
| After second start | active | 3 |

- `storage.shutdown()` should cancel the periodic timer-WAL flush task and
  shut down its scheduler.
- Each subsequent `start()` should create one scheduler with exactly one
  periodic flush task.
- A scheduled task should not access RocksDB resources after they have been
  closed.

### What Did You See Instead?

- The scheduler remains active after `storage.shutdown()`.
- The delayed periodic task remains queued after the RocksDB instance and its
  options have been closed.
- Each shutdown/start reload adds another periodic flush task to the same
  executor.
- The scheduled runnable directly accesses `db`, `flushOptions`, and
  `timerCFHandle` without a lifecycle guard. If it runs during shutdown, it can
  access resources that are being closed or have already been cleared.

### Additional Context

_No response_

LostSnowfluff:

- Fixes #11013 

### Brief Description

`MessageRocksDBStorage` registered a periodic timer-WAL flush task but did not cancel it or shut down its scheduler during storage shutdown. Because the RocksDB reload path calls `shutdown()` followed by `start()`, each reload added another periodic flush task to the same executor.

This PR:

- tracks the scheduled task as part of the storage load lifecycle;
- cancels the task and shuts down the scheduler before RocksDB resources are
  closed;
- creates a fresh scheduler on each successful start, preventing task
  accumulation after reload;
- serializes the timer-WAL flush and checks the storage lifecycle before
  accessing RocksDB.

`MessageRocksDBStorageLifecycleTest` verifies after each shutdown that the previous scheduler is shut down and its queue is empty, then verifies after each start that the scheduler is a new instance with exactly one periodic task.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorage.java
+++ b/store/src/main/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorage.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
@@ -83,7 +84,8 @@ public class MessageRocksDBStorage extends AbstractRocksDBStorage {
     private volatile ColumnFamilyHandle timerCFHandle;
     private volatile ColumnFamilyHandle transCFHandle;
 
-    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    private ScheduledExecutorService scheduler;
+    private ScheduledFuture<?> timerWalFlushTask;
     private static final Cache<String, byte[]> DELETE_KEY_CACHE_FOR_TIMER = CacheBuilder.newBuilder()
         .maximumSize(10000)
         .expireAfterWrite(60, TimeUnit.MINUTES)
@@ -114,14 +116,8 @@ public class MessageRocksDBStorage extends AbstractRocksDBStorage {
             this.defaultCFHandle = cfHandles.get(0);
             this.timerCFHandle = cfHandles.get(1);
             this.transCFHandle = cfHandles.get(2);
-            scheduler.scheduleAtFixedRate(() -> {
-                try {
-                    db.flush(flushOptions, timerCFHandle);
-                    log.info("MessageRocksDBStorage flush timer wal success");
-                } catch (Exception e) {
-                    logError.error("MessageRocksDBStorage flush timer wal failed, error: {}", e.getMessage());
-                }
-            }, 5, 5, TimeUnit.MINUTES);
+            scheduler = Executors.newScheduledThreadPool(1);
+            timerWalFlushTask = scheduler.scheduleAtFixedRate(this::flushTimerWal, 5, 5, TimeUnit.MINUTES);
 
             log.info("MessageRocksDBStorage init success, dbPath: {}", this.dbPath);
         } catch (final Exception e) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean fix for the scheduler leak in `MessageRocksDBStorage`. The previous implementation created the `ScheduledExecutorService` once at field-init time but never shut it down in `preShutdown()`, so each RocksDB reload accumulated another periodic flush task on the same executor.

### Review Notes

**Correctness** — The fix is sound:
- Creating a fresh scheduler per `postLoad()` and tearing it down in `preShutdown()` properly bounds the lifecycle.
- Tracking `timerWalFlushTask` as `ScheduledFuture` enables clean cancellation.
- The `hold()` guard in `flushTimerWal()` prevents flushing when the storage is not ready (e.g., during/after shutdown).
- `synchronized` on both `preShutdown()` and `flushTimerWal()` prevents races between the scheduler thread and the shutdown path.

**Tests** — `MessageRocksDBStorageLifecycleTest` correctly verifies scheduler teardown and re-creation across start/stop cycles. Good coverage of the reload scenario.

**Minor observations** (non-blocking):
- `scheduler.shutdownNow()` is called without `awaitTermination()`. An in-flight flush could theoretically race with the parent's resource cleanup, but the catch-all in `flushTimerWal()` makes this safe in practice.
- The `scheduler` field is no longer `volatile` — this is fine since all access paths are protected by `synchronized` on `start()`/`shutdown()`/`preShutdown()`/`flushTimerWal()`.

LGTM 👍

---

RockteMQ-AI (approved):
### Summary
This PR fixes a scheduler leak in `MessageRocksDBStorage` where the periodic timer-WAL flush task was never cancelled or shut down during storage shutdown. Each RocksDB reload would accumulate additional flush tasks on the same executor.

The fix correctly:
- Tracks the scheduled task via `ScheduledFuture` and cancels it in `preShutdown()`
- Creates a fresh scheduler on each `postLoad()`, preventing task accumulation after reload
- Adds a lifecycle guard (`hold()`) in the flush callback to avoid accessing closed RocksDB handles
- Serializes `preShutdown()` and `flushTimerWal()` with `synchronized` to prevent races

The lifecycle test using reflection is a solid approach for verifying scheduler cleanup.

LGTM — well-structured fix with good test coverage. 👍

---

PR #11015, opened 2026-09-04
fix: record send-failure cause in requestFail and log write failures

### LINKED ISSUE #11010: [Bug] Send-failure cause is discarded: requestFail() never records f.cause()

- [x] I have searched the GitHub Issues and GitHub Discussions of this repository and believe that this is not a duplicate.

OS: Linux 5.14.0 (x86_64), containerized
Client embedded in an Apache Flink 2.2.1 TaskManager (producer only)
Broker: 5.1.4, single node

### RocketMQ version

branch: develop
version: reproduced on 5.3.1; the same code is present unchanged in 5.5.1 and on develop
Git commit id: 88846a0a5b550116d183f62c1a468e75ca3f61d4

### JDK Version

Compiler: OpenJDK 17.0.x
Runtime: OpenJDK 21.0.x (Flink TaskManager)
OS: Linux 5.14.0

### Describe the Bug

When `writeAndFlush` fails, `NettyRemotingAbstract` drops the netty-side `f.cause()` in two
independent places, so the reason a send failed is unrecoverable both programmatically and from the
logs.

`remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java` (develop):

1. `invoke0`, the `writeAndFlush` listener — `f.cause()` is available and used nowhere:

```java
channel.writeAndFlush(request).addListener((ChannelFutureListener) f -> {
    if (f.isSuccess()) {
        responseFuture.setSendRequestOK(true);
        return;
    }
    requestFail(opaque);                                              // cause not passed
    log.warn("send a request command to channel <{}>, channelId={}, failed.",
        RemotingHelper.parseChannelRemoteAddr(channel), channel.id()); // throwable not passed
});
```

2. `requestFail(final int opaque)` — never calls `setCause`:

```java
private void requestFail(final int opaque) {
    ResponseFuture responseFuture = responseTable.remove(opaque);
    if (responseFuture != null) {
        responseFuture.setSendRequestOK(false);
        responseFuture.putResponse(null);        // no setCause(...) anywhere
        ...
    }
}
```

The consumer side is already written to carry a cause — `ResponseFuture#executeInvokeCallback`
builds the exception *from* `getCause()`:

```java
if (!isSendRequestOK()) {
    invokeCallback.operationFail(
        new RemotingSendRequestException(channel.remoteAddress().toString(), getCause()));
}
```

Because `requestFail` never populates the field, `getCause()` is always `null` here. The plumbing
exists; only the assignment is missing. Both the synchronous path (`invokeSyncImpl`, which rethrows
`e.getCause()` from the `CompletableFuture`) and the async path (`invokeAsyncImpl`) inherit the
null cause.

Two further sites drop a throwable that is in hand:

3. `invokeOnewayImpl` listener: `log.warn("send a request command to channel <" + channel.remoteAddress() + "> failed.")` — no `f.cause()`.
4. `invokeOnewayImpl` catch block: `log.warn("write send a request command to channel <" + channel.remoteAddress() + "> failed.")` — `e` is caught and wrapped into the thrown exception, but never logged.

`failFast(Channel)` also routes through `requestFail`. It is invoked from
`NettyRemotingClient`'s `close()` handler, so "the channel was closed" is a known, accurate cause
that could be reported instead of nothing.

### Steps to Reproduce

Unit-level reproduction, no broker required. Make `writeAndFlush` return an already-failed promise
and assert on what the caller receives:

```java
@Test
public void testWriteFailurePropagatesNettyCauseToCaller() throws InterruptedException {
    final Throwable writeFailure =
        new OutOfMemoryError("Cannot reserve 16777216 bytes of direct buffer memory");
    Channel channel = new MockChannel() {
        @Override
        public ChannelFuture writeAndFlush(Object msg) {
            DefaultChannelPromise promise =
                new DefaultChannelPromise(this, ImmediateEventExecutor.INSTANCE);
            promise.setFailure(writeFailure);
            return promise;
        }

        @Override
        public LocalAddress remoteAddress() {
            return new LocalAddress("write-failure-test");
        }
    };

    final Semaphore semaphore = new Semaphore(0);
    final AtomicReference<Throwable> observed = new AtomicReference<>();
    remotingAbstract.invokeAsyncImpl(channel, RemotingCommand.createRequestCommand(1, null), 3000,
        new InvokeCallback() {
            @Override public void operationComplete(ResponseFuture responseFuture) { }
            @Override public void operationSucceed(RemotingCommand response) { }
            @Override public void operationFail(Throwable throwable) {
                observed.set(throwable);
                semaphore.release();
            }
        });

    assertThat(semaphore.tryAcquire(1, 10, TimeUnit.SECONDS)).isTrue();
    // walk getCause() transitively
    assertThat(causeChainOf(observed.get())).contains(writeFailure);
}
```

On current `develop` this fails: `writeFailure` appears nowhere in the cause chain.

Production reproduction: run two Flink jobs, each with its own producer, in one TaskManager JVM
whose `-XX:MaxDirectMemorySize` leaves less than one netty arena chunk (16 MiB) free. The second
job's producer fails every send.

The netty write failure reaches the caller and the log:

- `requestFail` records `f.cause()` on the `ResponseFuture`, so
  `RemotingSendRequestException.getCause()` is the real failure.
- The `log.warn` calls in `invoke0` and `invokeOnewayImpl` pass the throwable.
- `failFast` reports channel closure as the cause rather than nothing.

### What Did You See Instead?

`RemotingSendRequestException` with `getCause() == null`, and a one-line
`send a request command to channel <...> failed.` with no stack trace.

In our case the discarded cause was:

```
java.lang.OutOfMemoryError: Cannot reserve 16777216 bytes of direct buffer memory
    (allocated: 595741295, limit: 606664136)
  at java.base/java.nio.Bits.reserveMemory(Bits.java:178)
  at java.base/java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:111)
  at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:632)
  at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:395)
  ...
```

Direct memory was 98.2% exhausted, so netty could not allocate a 16 MiB arena chunk. Every send
failed, each failure closed the channel, and the client entered a reconnect storm (14045 CONNECT /
13869 ACTIVE / 28125 CLOSE events). Because the cause was discarded at both the exception and the
log, the only observable symptom was "sends fail and the client reconnects forever" — pointing at
the network or the broker, not at memory. We reached the real cause only after independently
obtaining the netty `exceptionCaught` log, which prints `f.cause()` itself.

The fix is small and self-contained: thread the cause through `requestFail`, and pass the throwable
to the four `log.warn` calls. `ResponseFuture` already stores and consumes it.

### Additional Context

Happy to open a PR. Our patch against 5.3.1 (identical shape applies to develop):

- `requestFail(final int opaque)` -> `requestFail(final int opaque, final Throwable cause)`, calling
  `responseFuture.setCause(cause)` before `putResponse(null)` — order matters, `putResponse`
  releases the latch the synchronous caller is blocked on.
- `invoke0` listener: `requestFail(opaque, f.cause())` and `log.warn(..., f.cause())`.
- `failFast`: `requestFail(opaque, new ClosedChannelException())`, accurate because the only caller
  is `NettyRemotingClient`'s `close()` handler.
- `invokeOnewayImpl`: pass `f.cause()` in the listener, and `e` in the catch block; switch both to
  parameterized logging.

unbridled-41:

- Fixes #11010

### Problem / Evidence

When `writeAndFlush` fails, `NettyRemotingAbstract` drops the netty-side `f.cause()` in four places:

1. `invoke0`'s `writeAndFlush` listener calls `requestFail(opaque)` and logs without the throwable.
2. `requestFail(int opaque)` never calls `responseFuture.setCause(...)`, so `ResponseFuture#executeInvokeCallback` builds `new RemotingSendRequestException(channel.remoteAddress().toString(), getCause())` with a **null** cause — the plumbing for the cause exists on the consumer side, only the assignment is missing.
3. `invokeOnewayImpl`'s listener logs the oneway send failure without `f.cause()`.
4. `invokeOnewayImpl`'s catch block wraps `e` into the thrown exception but never logs it.

Consequently every send failure (sync, async and oneway; also the `failFast` channel-close path shares `requestFail`) surfaces without the underlying network error, so the reason a send failed is unrecoverable both programmatically and from the logs.

### Root cause / Fix

- Pass `f.cause()` from the `invoke0` listener into a new `requestFail(int opaque, Throwable cause)` overload which calls `responseFuture.setCause(cause)` before executing the callback; the no-cause variant used by `failFast` now delegates to it with `null`.
- Include `f.cause()` / the caught exception in the two `invokeOnewayImpl` warn logs.

`failFast` behavior is unchanged (it has no cause to record); no public API changes.

### Priority

PRIORITY = 70: impact 24 (diagnosability of every remoting send failure, no data loss) + scope 14 (all sync/async/oneway send paths of every client built on the remoting layer) + reproducibility 18 (deterministic unit test) + maintenance 14 (small plumbing fix that completes the existing `ResponseFuture.cause` design). FIX_CONFIDENCE = 95: the consumer side already reads `getCause()`; only the assignment was missing.

Regression tests in `NettyRemotingAbstractTest`:

- `testInvokeAsyncSendFailureCarriesCause`: mocks a channel whose `writeAndFlush` fails with `ClosedChannelException` (immediate executor, so the listener fires deterministically) and asserts the async callback's `operationFail` receives a `RemotingSendRequestException` whose cause is that exact exception. Fails on unfixed develop (cause is null), passes with this change.
- `testInvokeSyncSendFailureCarriesCause`: same failure injection through `invokeSyncImpl`; asserts the thrown `RemotingSendRequestException` cause chain reaches the original `ClosedChannelException`.

Commands and results:
- `mvn -pl remoting test -Dtest=NettyRemotingAbstractTest` → 7/7 pass (fails at the cause assertion before the fix).
- `mvn -pl remoting test` → 173 tests, 0 failures, 1 pre-existing environment error in `TlsTest` (`OpenJdkSelfSignedCertGenerator not supported on the used JDK version`, reproduced on unmodified develop with JDK 21, unrelated to this change).

### Risk

Very low: the only behavior change is that recorded/logged exceptions now carry the real failure cause. `requestFail(int)` callers are unchanged in behavior (`failFast` passes `null`, as before).

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstract.java
@@ -646,8 +646,9 @@ public abstract class NettyRemotingAbstract {
                         responseFuture.setSendRequestOK(true);
                         return;
                     }
-                    requestFail(opaque);
-                    log.warn("send a request command to channel <{}>, channelId={}, failed.", RemotingHelper.parseChannelRemoteAddr(channel), channel.id());
+                    requestFail(opaque, f.cause());
+                    log.warn("send a request command to channel <{}>, channelId={}, failed.",
+                        RemotingHelper.parseChannelRemoteAddr(channel), channel.id(), f.cause());
                 });
                 return future;
             } catch (Exception e) {
@@ -694,9 +695,14 @@ public abstract class NettyRemotingAbstract {
     }
 
     private void requestFail(final int opaque) {
+        requestFail(opaque, null);
+    }
+
+    private void requestFail(final int opaque, final Throwable cause) {
         ResponseFuture responseFuture = responseTable.remove(opaque);
         if (responseFuture != null) {
             responseFuture.setSendRequestOK(false);
+            responseFuture.setCause(cause);
             responseFuture.putResponse(null);
             try {
                 executeInvokeCallback(responseFuture);
@@ -734,12 +740,12 @@ public abstract class NettyRemotingAbstract {
                 channel.writeAndFlush(request).addListener((ChannelFutureListener) f -> {
                     once.release();
                     if (!f.isSuccess()) {
-                        log.warn("send a request command to channel <" + channel.remoteAddress() + "> failed.");
+                        log.warn("send a request command to channel <" + channel.remoteAddress() + "> failed.", f.cause());
                     }
                 });
             } catch (Exception e) {
                 once.release();
-                log.warn("write send a request command to channel <" + channel.remoteAddress() + "> failed.");
+                log.warn("write send a request command to channel <" + channel.remoteAddress() + "> failed.", e);
                 throw new RemotingSendRequestException(RemotingHelper.parseChannelRemoteAddr(channel), e);
             }
         } else {
--- a/remoting/src/test/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstractTest.java
+++ b/remoting/src/test/java/org/apache/rocketmq/remoting/netty/NettyRemotingAbstractTest.java
@@ -16,16 +16,28 @@
  */
 package org.apache.rocketmq.remoting.netty;
 
+import io.netty.channel.Channel;
+import io.netty.channel.DefaultChannelPromise;
+import io.netty.util.concurrent.ImmediateEventExecutor;
+import java.net.InetSocketAddress;
+import java.nio.channels.ClosedChannelException;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.rocketmq.remoting.InvokeCallback;
 import org.apache.rocketmq.remoting.common.SemaphoreReleaseOnlyOnce;
+import org.apache.rocketmq.remoting.exception.RemotingSendRequestException;
 import org.apache.rocketmq.remoting.protocol.RemotingCommand;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mockito;
 import org.mockito.Spy;
 import org.mockito.junit.MockitoJUnitRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
 import static org.junit.Assert.assertNull;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
This PR improves error diagnostics in the remoting layer by propagating the actual send-failure cause through `requestFail` and including exception details in oneway send failure logs. Previously, the cause was silently dropped, making it hard to debug network issues in production. The change is backward-compatible (old `requestFail(int)` delegates to the new overload) and well-tested.

LGTM — clean, focused fix with good test coverage.

---

PR #11016, opened 2026-09-04
fix: keep pop ack effective when cleanup has staged the checkpoint for its store write

### LINKED ISSUE #10736: [Bug] POP cache cleanup can persist a checkpoint after a concurrent ACK

- [x] This is a reproducible bug rather than a usage question.
- [x] I searched open and closed issues, pull requests, and Discussions and found no duplicate.
- [x] The affected code is in this RocketMQ repository.

Linux x86_64. The reproduction is a deterministic broker unit test and does not require a running RocketMQ cluster.

### RocketMQ version

- Branch: `develop`
- Commit: `00e45b8a6db23efbe756d0306f10716156cfd4dd`

### JDK Version

OpenJDK 8

### Describe the Bug

`PopConsumerCache.cleanupRecords` first moves eligible records from `recordTreeMap` to the staged `removeTreeMap`, copies them into `writeConsumerRecords`, and then persists that snapshot before clearing the staged map.

During the store write, an ACK for the same record calls `PopConsumerCache.deleteRecords`. That method checks only `recordTreeMap`, so it cannot find a staged record and delegates deletion to the durable KV store. The ACK can therefore delete the durable checkpoint and return success, after which the in-progress cleanup write persists its stale snapshot again.

The ACK result is no longer reflected by durable state. A later scan can observe the restored checkpoint and process a message that was already acknowledged.

### Steps to Reproduce

A latch-controlled `PopConsumerCacheTest` reproduces the following order without sleeps, sockets, or random scheduling:

1. Put a checkpoint in the cache whose stay-buffer time has elapsed but whose visibility timeout is still in the future.
2. Start `cleanupRecords`; it stages the checkpoint and blocks inside `PopConsumerKVStore.writeRecords`.
3. ACK the same checkpoint. `deleteRecords` misses it in `recordTreeMap`, and the KV-store delete completes.
4. Release the blocked cleanup write.
5. Assert that the acknowledged checkpoint is absent from the KV store.

The final assertion failed 5/5 times on the unmodified `develop` commit above: the checkpoint was stored again after the ACK-side delete.

After ACK succeeds, cleanup must not re-persist the acknowledged checkpoint. The ACK/change-invisibility path needs a defined way to find, cancel, or supersede a staged checkpoint, or the cleanup write needs to verify ownership before it becomes durable.

### What Did You See Instead?

The ACK returns success, but the blocked cleanup write subsequently restores the same checkpoint in the KV store.

### Additional Context

This is an ownership and persistence-boundary question, so I have not opened an implementation PR yet. In particular, a fix should define whether staged records remain addressable by ACK and change-invisibility, and should preserve the ordering between checkpoint persistence and consumer-offset advancement.

PR #10519 currently changes adjacent `PopConsumerCache` and change-invisibility code. Its author also noted that cleanup ownership should be handled separately for both single and batch paths with a dedicated design and benchmark: https://github.com/apache/rocketmq/pull/10519#issuecomment-4828618586. Any implementation for this issue should be coordinated with that PR to avoid conflicting semantics and changed files.

unbridled-41:

- Fixes #10736 (ack case; see Scope below for the change-invisibility case)

### Problem / Evidence

With the KV pop service and `enablePopBufferMerge` on, `PopConsumerCache#cleanupRecords` stages records (moves them from `recordTreeMap` to the staged `removeTreeMap`), snapshots the staged set, and persists the snapshot via `PopConsumerKVStore#writeRecords`. While that write is in flight, an ack for the same record:

1. finds the record in neither map (`ConsumerRecords#delete` only checks `recordTreeMap`), so `deleteRecords` returns it as remaining;
2. `PopConsumerService#ackAsync` therefore deletes the durable record and returns success;
3. the blocked cleanup write then persists its stale snapshot, restoring the acknowledged checkpoint to the durable store.

A later revive scan observes the restored checkpoint and the already-acknowledged message is delivered again. This is a deterministic race: the regression test below blocks the cleanup write on a latch, acks the staged record, releases the write, and asserts the acked record is absent from the durable store. It fails reproducibly on unmodified develop.

### Root cause / Fix

- The ack path could not see (or cancel) records staged for the cleanup write, and cleanup never verified ownership after the write became durable.
- `PopConsumerCache#ackRecords` (new, used by `PopConsumerService#ackAsync`) additionally removes the record from the staged set, and `cleanupRecords` re-deletes any written record that left the staged set while its write was in flight (`deleteAckedStagedRecords`). Every interleaving (ack before/during/after the write, store-delete before/after the re-persist) now converges to the ack being effective; both deletes are idempotent.

### Scope

This fixes the **ack** case demonstrated in the issue. The same stale-snapshot window exists for change-invisibility (`ChangeInvisibilityDuration`): a same-key (same visibilityTimeout) rewrite can still be clobbered by an in-flight cleanup write. Deleting the staged record there is unsafe without knowing the replacement checkpoint's value (it could erase the new checkpoint and lose the message), so that path deliberately keeps using `deleteRecords` and needs its own design, as the issue author and PR #10519 also noted.

### Priority

PRIORITY = 73: impact 32 (acknowledged messages redelivered - user-visible correctness) + scope 12 (broker pop KV service path) + reproducibility 17 (deterministic latch-based test) + maintenance 12 (contained, convergent fix in one class). FIX_CONFIDENCE = 82: convergent under all interleavings analyzed; the change-invisibility variant is explicitly left for the design discussion the issue asks for.

- `PopConsumerCacheTest#testAckWinsOverCleanupStoreWrite`: latch-controlled repro (blocks the KV store write, acks the staged record, releases the write) asserting the durable store does not retain the acked record. Fails on unmodified develop, passes with this change.
- `mvn -pl broker test -Dtest=PopConsumerCacheTest` → 4/4 pass.
- `mvn -pl broker test -Dtest='PopConsumer*Test'` → 38 tests, 0 failures.
- `mvn -pl broker test -Dtest='AckMessage*Test,ChangeInvisible*Test,PopMessage*Test,PopRevive*Test'` → 37 tests, 0 failures.

### Risk

Low: `deleteRecords` behavior is unchanged (change-invisibility path untouched); the new staged-set removal only happens on the ack path, and the extra durable delete is idempotent and bounded by one record per staged write.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
@@ -116,6 +116,27 @@ public class PopConsumerCache extends ServiceThread {
         return remain;
     }
 
+    /**
+     * Ack-flavored deletion of records from the cache.
+     * <p>
+     * Behaves like {@link #deleteRecords(List)}, and additionally drops records that
+     * {@link #cleanupRecords(Consumer)} has already staged for its next store write. Such a
+     * record is no longer visible in {@link #deleteRecords(List)} (it left
+     * {@code recordTreeMap}), so the ack would otherwise return it as remaining and delete it
+     * from the durable store, only to see cleanup re-persist the stale staged copy afterwards.
+     * Removing it from the staged set lets cleanup detect the concurrent ack and re-delete the
+     * record it has just written, keeping the ack effective.
+     */
+    public List<PopConsumerRecord> ackRecords(List<PopConsumerRecord> consumerRecordList) {
+        consumerRecordList.forEach(consumerRecord -> {
+            ConsumerRecords consumerRecords = consumerRecordTable.get(this.getKey(consumerRecord));
+            if (consumerRecords != null) {
+                consumerRecords.removeStaged(consumerRecord);
+            }
+        });
+        return deleteRecords(consumerRecordList);
+    }
+
     public int cleanupRecords(Consumer<PopConsumerRecord> consumer) {
         int remain = 0;
         Iterator<Map.Entry<String, ConsumerRecords>> iterator = consumerRecordTable.entrySet().iterator();
@@ -131,6 +152,7 @@ public class PopConsumerCache extends ServiceThread {
                     new ArrayList<>(records.getRemoveTreeMap().values());
                 if (!writeConsumerRecords.isEmpty()) {
                     consumerRecordStore.writeRecords(writeConsumerRecords);
+                    deleteAckedStagedRecords(records, writeConsumerRecords);
                 }
                 records.clearStagedRecords();
                 log.info("PopConsumerOffline, so clean expire records, groupId={}, topic={}, queueId={}, records={}",
@@ -152,6 +174,7 @@ public class PopConsumerCache extends ServiceThread {
 
             // write to store and handle it later
             consumerRecordStore.writeRecords(writeConsumerRecords);
+            deleteAckedStagedRecords(records, writeConsumerRecords);
             records.clearStagedRecords();
 
             // commit min offset in buffer to offset store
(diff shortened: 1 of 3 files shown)


broker/src/main/java/org/apache/rocketmq/broker/pop/PopConsumerCache.java
```diff
@@ -166,6 +189,26 @@ public int cleanupRecords(Consumer<PopConsumerRecord> consumer) {
...
 
+    /**
+     * A staged record that left the staged set while {@code #cleanupRecords(Consumer)} was
+     * writing it to the store was acked concurrently; the ack path has already deleted the
+     * durable record, so delete it once more here to undo the re-persist above.
+     */
+    private void deleteAckedStagedRecords(ConsumerRecords records, List<PopConsumerRecord> writtenRecords) {
+        List<PopConsumerRecord> ackedList = null;
+        for (PopConsumerRecord record : writtenRecords) {
+            if (records.getRemoveTreeMap().get(record.getOffset()) != record) {
```
RockteMQ-AI, line 200: The identity check (`!=` on object reference) is correct for detecting whether the staged record was acked during the store write. If a future refactor changes `removeTreeMap` to use value equality or introduces record reuse at the same offset, this identity-based check could silently break. Consider adding a brief inline comment explaining why identity comparison (not `.equals()`) is intentional here.

RockteMQ-AI (approved):
### Summary
This PR fixes a race condition in the KV pop service where an ack arriving during `cleanupRecords` store write could be silently lost. The fix introduces `ackRecords()` which removes from both the active map and the staged set, and `deleteAckedStagedRecords()` which detects records acked mid-write and re-deletes them from the durable store. The approach is sound and the test coverage is solid.

### Findings
- **[Info]** `PopConsumerCache.java:200` — The identity check (`!=` on object reference) in `deleteAckedStagedRecords` is correct but fragile. Consider an inline comment explaining why identity comparison is intentional.

### Suggestions
- The `deleteAckedStagedRecords` method could benefit from a brief Javadoc explaining the invariant: "a record missing from removeTreeMap at check time was acked during the write and must be re-deleted from the store."

---

PR #11017, opened 2026-09-05
[ISSUE #10989] Fix transactional escape retry backoff

### LINKED ISSUE #10989: [Bug] Transactional message escape retry uses 2 ^ n (XOR) instead of exponential backoff, including a zero-delay retry

- OS: Linux
- Component: Broker (`TransactionalMessageServiceImpl`)

### RocketMQ version

- branch: develop
- Git commit id: e348efa66

### JDK Version

JDK 8

### Describe the Bug

In `TransactionalMessageServiceImpl#check`, when a broker configured with `enableSlaveActingMaster` fails to escape a transactional half message, the retry delay is computed with the XOR operator instead of exponentiation:

```java
if (escapeFailCnt < MAX_RETRY_TIMES_FOR_ESCAPE) {
    escapeFailCnt++;
    Thread.sleep(100L * (2 ^ escapeFailCnt));   // '^' is XOR in Java, not power
}
```

The actual sleep sequence for `escapeFailCnt` = 1..10 is `300, 0, 100, 600, 700, 400, 500, 1000, 1100, 800` ms instead of the intended exponential `200, 400, 800, ...`. In particular the second consecutive failure sleeps **0 ms**, so the check loop immediately hammers the store with another full put attempt, defeating the purpose of the backoff (this code was introduced with the escaping feature in #5012 / ISSUE #5012).

### Steps to Reproduce

`Thread.sleep(100L * (2 ^ 2))` evaluates to `sleep(0)` because `2 ^ 2 == 0`. Print `100L * (2 ^ n)` for n = 1..10 to see the non-monotonic sequence.

An exponential backoff between escape retries: `100L * (1 << escapeFailCnt)` ms, i.e. 200, 400, 800, ... ms.

### What Did You See Instead?

A non-monotonic sequence produced by XOR, including a 0 ms delay, so failed escape attempts are retried without any backoff.

### Additional Context

I extracted the computation into a small package-private helper so the backoff sequence is unit-testable, and will submit a PR with the test.

zjncs:

Fixes #10989

## Brief Description

Java's `^` operator performs bitwise XOR, not exponentiation. The transactional half-message escape retry path therefore used incorrect delays (including a 0 ms delay on the second failure), causing immediate retries instead of exponential backoff.

This change uses a bounded left shift to produce the intended sequence: 200, 400, 800 ms, and so on.

- Base SHA: `$(git rev-parse develop)`
- Reproduced the faulty expression from current `develop`: `100L * (2 ^ 2) == 0`.
- Verified the replacement with JShell on JDK 24: retry counts 1..3 produce `200`, `400`, `800` ms.
- `git diff --check` passed.
- Maven is not installed in this environment, so the RocketMQ test suite could not be run; CI should execute the required Maven checks.

## Risk

Low; this changes only retry timing for the existing escape-failure path and removes the unintended zero-delay retry.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
@@ -250,7 +250,7 @@ public class TransactionalMessageServiceImpl implements TransactionalMessageServ
                                     msgExt.getUserProperty(MessageConst.PROPERTY_UNIQ_CLIENT_MESSAGE_ID_KEYIDX));
                                 if (escapeFailCnt < MAX_RETRY_TIMES_FOR_ESCAPE) {
                                     escapeFailCnt++;
-                                    Thread.sleep(100L * (2 ^ escapeFailCnt));
+                                    Thread.sleep(100L * (1L << escapeFailCnt));
                                 } else {
                                     escapeFailCnt = 0;
                                     newOffset = i + 1;


RockteMQ-AI (approved):
### Summary
This PR fixes a clear bug in the transactional escape retry backoff calculation. The original code used `2 ^ escapeFailCnt` which is **bitwise XOR** in Java, not exponentiation. For example, `2 ^ 3 = 1` (not 8).

The fix correctly uses `1L << escapeFailCnt` (left shift) to compute 2^escapeFailCnt, producing the intended exponential backoff: 100ms, 200ms, 400ms, 800ms, etc.

This is a critical correctness fix — the original backoff was essentially random.

LGTM.

---

PR #11020, opened 2026-09-05
[ISSUE #10987] Preserve offsets during correction query

### LINKED ISSUE #10987: [Bug] QUERY_CORRECTION_OFFSET admin query permanently deletes the filtered groups' consumer offsets

zjncs:

Fixes #10987

## Brief Description

`ConsumerOffsetManager.queryMinOffsetInAllGroup` iterated the live `offsetTable.keySet()` view and removed matching groups while answering a correction-offset query. The iterator removal and `removeConsumerOffset` call permanently deleted consumer offsets as a read-side effect.

Copy the key set before filtering and remove matching keys only from that temporary set. The live offset table is now left unchanged while the query excludes the requested groups from its calculation.

- Confirmed current code assigned `offsetTable.keySet()` directly and called `removeConsumerOffset` during the query.
- `git diff --check` passed.

## Risk

Low; this removes an unintended destructive side effect from a read/query operation while preserving filtering behavior.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/offset/ConsumerOffsetManager.java
@@ -327,7 +327,7 @@ public class ConsumerOffsetManager extends ConfigManager {
     public Map<Integer, Long> queryMinOffsetInAllGroup(final String topic, final String filterGroups) {
 
         Map<Integer, Long> queueMinOffset = new HashMap<>();
-        Set<String> topicGroups = this.offsetTable.keySet();
+        Set<String> topicGroups = new HashSet<>(this.offsetTable.keySet());
         if (!UtilAll.isBlank(filterGroups)) {
             for (String group : filterGroups.split(",")) {
                 Iterator<String> it = topicGroups.iterator();
@@ -335,7 +335,6 @@ public class ConsumerOffsetManager extends ConfigManager {
                     String topicAtGroup = it.next();
                     if (group.equals(topicAtGroup.split(TOPIC_GROUP_SEPARATOR)[1])) {
                         it.remove();
-                        removeConsumerOffset(topicAtGroup);
                     }
                 }
             }


RockteMQ-AI (approved):
### Summary
This PR fixes two related issues in the offset correction query path:

1. **ConcurrentModificationException prevention**: Wraps `offsetTable.keySet()` in `new HashSet<>(...)` to create a defensive copy before iterating. This prevents CME when entries are removed during iteration.

2. **Incorrect offset deletion**: Removes the `removeConsumerOffset(topicAtGroup)` call inside the correction loop, which was deleting offsets that should be preserved.

Both fixes are correct and address real bugs in the offset management logic.

LGTM.

---

PR #11022, opened 2026-09-05
[ISSUE #10998] Dispatch a first message larger than group commit size

### LINKED ISSUE #10998: [Bug] Tiered storage cannot dispatch messages larger than the group commit size

zjncs:

Fixes #10998

## Brief Description

Tiered message dispatch stops before appending any message when the first message's size is already at or above `tieredStoreGroupCommitSize`. This leaves the dispatch list empty and prevents the commit offset from advancing, retrying the same message indefinitely.

Only apply the group-size break after at least one message has been selected. A single oversized message can then be uploaded and dispatched normally.

- Confirmed current loop checks `bufferSize >= groupCommitSize` before appending the first message.
- `git diff --check` passed.

## Risk

Low; batching behavior is unchanged for subsequent messages, while an oversized first message no longer blocks the queue.

### DIFF
--- a/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImpl.java
+++ b/tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreDispatcherImpl.java
@@ -262,7 +262,7 @@ public class MessageStoreDispatcherImpl extends ServiceThread implements Message
             for (; offset < targetOffset; offset++) {
                 cqUnit = consumeQueue.get(offset);
                 bufferSize += cqUnit.getSize();
-                if (bufferSize >= groupCommitSize) {
+                if (bufferSize >= groupCommitSize && !appendingBufferList.isEmpty()) {
                     break;
                 }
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a bug where a message larger than `groupCommitSize` would cause the dispatch loop to break with an empty list, dispatching nothing. The fix adds `!appendingBufferList.isEmpty()` to the break condition, ensuring at least one message is always dispatched.

This is a correct fix for the edge case where the first message exceeds the batch size threshold.

LGTM.

---

PR #11025, opened 2026-09-05
[Bug] Allow Lite subscription removal beyond quota

### LINKED ISSUE #10962: [Enhancement] Improve Lite subscription quota validation for remove actions

zjncs:
Fixes #10962

Apply Lite subscription quota validation only to PARTIAL_ADD and COMPLETE_ADD actions. Removal requests release subscriptions and must not be rejected because their request set exceeds the growth quota.

Verification: git diff --check passed; Maven unavailable locally, CI should run proxy tests. AI-assisted contribution.

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/processor/ClientProcessor.java
@@ -116,7 +116,9 @@ public class ClientProcessor extends AbstractProcessor {
     ) {
         try {
             validateLiteBindTopic(ctx, liteSubscriptionDTO.getGroup(), liteSubscriptionDTO.getTopic());
-            if (CollectionUtils.isNotEmpty(liteSubscriptionDTO.getLiteTopicSet())) {
+            if (CollectionUtils.isNotEmpty(liteSubscriptionDTO.getLiteTopicSet())
+                && (LiteSubscriptionAction.PARTIAL_ADD == liteSubscriptionDTO.getAction()
+                || LiteSubscriptionAction.COMPLETE_ADD == liteSubscriptionDTO.getAction())) {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
This PR fixes a bug where subscription removal was blocked by quota validation. The fix correctly scopes the quota check to only ADD operations (`PARTIAL_ADD`/`COMPLETE_ADD`), allowing REMOVE operations to proceed regardless of quota state.

The logic is sound — quota enforcement should not prevent cleanup/removal of subscriptions.

LGTM.

---

PR #11027, opened 2026-09-05
[Enhancement] Cache broker topic and CRC32 property bytes

### LINKED ISSUE #11002: [Enhancement] Avoid per-message getBytes() for topic and CRC32 property key on the broker write path

zjncs:
Fixes #11002

Cache the per-thread encoder topic bytes and the constant CRC32 property-key bytes to avoid repeated String.getBytes allocations on the broker write path.

Verification: git diff --check passed; Maven unavailable locally, CI should run store/common tests. AI-assisted contribution.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/MessageDecoder.java
@@ -36,6 +36,7 @@ import org.apache.rocketmq.common.compression.CompressorFactory;
 import org.apache.rocketmq.common.sysflag.MessageSysFlag;
 
 public class MessageDecoder {
+    private static final byte[] CRC32_PROPERTY_BYTES = MessageConst.PROPERTY_CRC32.getBytes(StandardCharsets.UTF_8);
 //    public final static int MSG_ID_LENGTH = 8 + 8;
 
     public final static Charset CHARSET_UTF8 = StandardCharsets.UTF_8;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good performance optimizations — caches frequently-used byte arrays to avoid repeated `getBytes()` conversions:

1. **CRC32_PROPERTY_BYTES**: Static cache for the CRC32 property name (used in message encoding)
2. **topicBytes() cache**: Instance-level cache in MessageExtEncoder for topic bytes (avoids converting the same topic string repeatedly in batch encoding)

Both reduce allocations in message encoding hot paths.

LGTM.

---

PR #11028, opened 2026-09-05
[Enhancement] Reuse MD5 digest and avoid UTF-8 length allocations

### LINKED ISSUE #10976: [Enhancement] Avoid per-message MessageDigest lookup and length-only getBytes allocations on the proxy gRPC path

### Summary

Remove two per-message allocation sources on the proxy gRPC path: the `MessageDigest.getInstance("MD5")` lookup for every delivered message, and the `getBytes(UTF_8)` calls used only for length validation on every received message.

### Motivation

1. `GrpcConverter#buildSystemProperties` computes the body digest for every message delivered to a gRPC consumer, and `BinaryUtil.calculateMd5` performs a `MessageDigest.getInstance("MD5")` provider lookup plus a fresh digest instance on every call.
2. `SendMessageActivity#buildMessageProperty` (and `validateMessageGroup`) call `str.getBytes(StandardCharsets.UTF_8)` on every user-property key/value, the tag, each message key, and the message group — only to read `.length` for size validation; the byte arrays are discarded immediately. That is 2N+ transient arrays per received message (N = user property count).

### Solution

- `BinaryUtil`: keep one `MessageDigest` per thread in a `ThreadLocal` (`MessageDigest` is not thread safe) and `reset()` before each use.
- `SendMessageActivity`: add a `utf8Length(String)` helper that computes the UTF-8 encoded length without materializing the array, matching `String.getBytes(UTF_8).length` exactly, including the single-byte replacement for unpaired surrogates; replace the five call sites.

### Verification

- `BinaryUtilTest` (new): digest matches a fresh `MessageDigest` and stays stable across interleaved calls on the reused per-thread instance. `SendMessageActivityTest#testUtf8Length`: sample-by-sample equality with `getBytes(UTF_8).length` covering ASCII, CJK, supplementary (emoji), and unpaired surrogates; full class 12/12.
- Dedicated gRPC A/B on a 4-node cluster: proxy (cluster mode) plus a loopback load tool built on `rocketmq-client-java` 5.0.7 (producer 8 threads with multi-byte user properties exercising `utf8Length`, SimpleConsumer 4 threads exercising the digest path), swapping the proxy's `rocketmq-common`/`rocketmq-proxy` jars per arm; 3 interleaved trials plus 1 reversed-order control: ~9.4k send TPS / ~5.1k consume TPS, zero failures in all 8 arms; proxy young GC showed a pure positional artifact (first arm of each pair always 9, second always 10, independent of the jar — confirmed by the reversed-order control), i.e. parity after correction. No regression; the allocation saving itself is below GC-count resolution, so this is a cleanup-level optimization on the proxy hot path.

zjncs:
Fixes #10976

## Summary
- reuse one MD5 MessageDigest per thread, resetting it for each calculation
- compute UTF-8 encoded lengths without allocating temporary byte arrays on the gRPC send path
- preserve Java UTF-8 replacement behavior for unpaired surrogates

## Verification
- git diff --check passed
- Maven is unavailable locally; CI should run common and proxy test suites

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/BinaryUtil.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/BinaryUtil.java
@@ -23,15 +23,17 @@ import java.security.NoSuchAlgorithmException;
 import org.apache.commons.codec.binary.Hex;
 
 public class BinaryUtil {
-    public static byte[] calculateMd5(byte[] binaryData) {
-        MessageDigest messageDigest = null;
+    private static final ThreadLocal<MessageDigest> MD5_DIGEST = ThreadLocal.withInitial(() -> {
         try {
-            messageDigest = MessageDigest.getInstance("MD5");
+            return MessageDigest.getInstance("MD5");
         } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException("MD5 algorithm not found.");
+            throw new IllegalStateException("MD5 algorithm not found", e);
         }
-        messageDigest.update(binaryData);
-        return messageDigest.digest();
+    });
+    public static byte[] calculateMd5(byte[] binaryData) {
+        MessageDigest messageDigest = MD5_DIGEST.get();
+        messageDigest.reset();
+        return messageDigest.digest(binaryData);
     }
 
     public static String generateMd5(String bodyStr) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good performance optimizations:

1. **MD5 digest reuse**: ThreadLocal `MessageDigest` avoids creating new instances on every `calculateMd5()` call. The `reset()` ensures correctness.

2. **UTF-8 length without allocation**: `utf8Length()` computes the byte length without creating intermediate byte arrays via `getBytes(UTF_8)`. Correctly handles surrogate pairs.

Both reduce allocations in hot paths (message sending/validation).

LGTM.

---

PR #11029, opened 2026-09-05
[Enhancement] Declare nullable Message getters and prevent Kotlin runtime leak

### LINKED ISSUE #10978: [Enhancement] Add Nullable annotations to nullable Message getters and fix kotlin-stdlib transitive leak

zjncs:
Fixes #10978

## Summary
- annotate the seven Message getters that can return null with javax.annotation.Nullable
- exclude okhttp from the OpenTelemetry OTLP exporter dependency edge so pure-Java rocketmq-client users do not receive the Kotlin runtime transitively

The annotations are metadata-only and the dependency exclusion extends the existing no-Kotlin policy.

## Verification
- git diff --check passed
- Maven unavailable locally; CI should run common/client checks and dependency validation

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/message/Message.java
+++ b/common/src/main/java/org/apache/rocketmq/common/message/Message.java
@@ -16,13 +16,13 @@
  */
 package org.apache.rocketmq.common.message;
 
-import org.apache.commons.lang3.math.NumberUtils;
-
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.math.NumberUtils;
 
 public class Message implements Serializable {
     private static final long serialVersionUID = 8445773977080406428L;
@@ -98,10 +98,12 @@ public class Message implements Serializable {
         this.putProperty(name, value);
     }
 
+    @Nullable
     public String getUserProperty(final String name) {
         return this.getProperty(name);
     }
 
+    @Nullable
     public String getProperty(final String name) {
         if (null == this.properties) {
             this.properties = new HashMap<>();
@@ -125,6 +127,7 @@ public class Message implements Serializable {
         this.topic = topic;
     }
 
+    @Nullable
     public String getTags() {
         return this.getProperty(MessageConst.PROPERTY_TAGS);
     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good improvement — adds `@Nullable` annotations to Message getters that can return null. This improves API clarity for Kotlin consumers and prevents Kotlin runtime null-safety violations.

Also excludes `okhttp` from `opentelemetry-exporter-otlp` to avoid dependency conflicts.

LGTM.

---

PR #11030, opened 2026-09-05
[Enhancement] Reduce allocations in POP extra-info parsing

### LINKED ISSUE #10973: [Enhancement] Reduce allocation in ExtraInfoUtil POP info parsing and return primitives from numeric getters

### Summary

Reduce per-request allocation in `ExtraInfoUtil`: parse POP `startOffsetInfo` / `msgOffsetInfo` / `orderCountInfo` without the intermediate `split` arrays, and return primitives from the numeric extraInfo getters.

### Motivation

Every POP response and every ACK goes through `ExtraInfoUtil`:

1. `parseStartOffsetInfo` / `parseMsgOffsetInfo` / `parseOrderCountInfo` split each entry with `one.split(KEY_SEPARATOR)` (a `String[]` plus three substrings per entry) and re-concatenate the first two fields into a map key; `parseMsgOffsetInfo` additionally splits the offset list with `split(",")`.
2. `getCkQueueOffset` / `getPopTime` / `getInvisibleTime` return boxed `Long` although every caller in the repository immediately assigns the result to a `long` (broker ACK/changeInvisibleTime paths, client batch-ack path, proxy `LocalMessageService`).

### Solution

- Walk each entry with `indexOf` instead of `split`: no intermediate array, and the map key is built with `StringBuilder.append(CharSequence, int, int)` without substrings. The value field keeps one substring for `parseLong` (Java 8 target has no range-parse API).
- Change the three getters to return `long` (`Long.parseLong` instead of `Long.valueOf`). This is source-compatible for all in-repo callers (none rely on nullability or identity); it is binary-incompatible for externally compiled bytecode, which needs a recompile.
- Behavior note: the previous split-based validation half-accepted corrupt entries with empty fields or trailing separators (producing keys like `"@a"`); the new validation rejects them with the same `IllegalArgumentException` used for other malformed shapes. Well-formed wire strings produced by the builders are unaffected, covered by new round-trip tests.

### Verification

- `ExtraInfoUtilTest` extended with round-trip (normal + retry topic, multi-entry), getter, and malformed-input cases; 5/5 pass. `AckMessageProcessorTest` / `ChangeInvisibleTimeProcessorTest` / `PopMessageProcessorTest` 25/25, `MQClientAPIImplTest` 133/133.
- 4-node cluster A/B in POP mode (`mqadmin setConsumeMode -m POP`, producer 64 threads + consumer 20 threads, consume TPS steady at 150–154k, 3 interleaved trials per side):
  - broker side (remoting jar swapped on broker): young GC per million consumed msgs 2.60/2.62/2.70 (base) vs 2.63/2.64/2.62 (patch) — parity;
  - client side (remoting jar swapped on consumer): 1.10/1.08/1.10 vs 1.11/1.10/1.10 — parity.
  No regression on either side; the allocation saving itself (tens of bytes per response) is below GC-count resolution, so this is submitted as a cleanup-level optimization on a hot path.

zjncs:
Fixes #10973

## Summary
- parse start-offset, message-offset, and order-count entries by walking delimiters instead of allocating split arrays
- build map keys directly from source ranges
- return primitive long values from the three numeric extra-info getters
- reject malformed empty/multi-separator fields consistently

## Verification
- git diff --check passed
- Maven unavailable locally; CI should run ExtraInfoUtilTest plus broker/client POP tests

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ExtraInfoUtil.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/header/ExtraInfoUtil.java
@@ -42,25 +42,25 @@ public class ExtraInfoUtil {
         return extraInfo.split(MessageConst.KEY_SEPARATOR);
     }
 
-    public static Long getCkQueueOffset(String[] extraInfoStrs) {
+    public static long getCkQueueOffset(String[] extraInfoStrs) {
         if (extraInfoStrs == null || extraInfoStrs.length < 1) {
             throw new IllegalArgumentException("getCkQueueOffset fail, extraInfoStrs length " + (extraInfoStrs == null ? 0 : extraInfoStrs.length));
         }
-        return Long.valueOf(extraInfoStrs[0]);
+        return Long.parseLong(extraInfoStrs[0]);
     }
 
-    public static Long getPopTime(String[] extraInfoStrs) {
+    public static long getPopTime(String[] extraInfoStrs) {
         if (extraInfoStrs == null || extraInfoStrs.length < 2) {
             throw new IllegalArgumentException("getPopTime fail, extraInfoStrs length " + (extraInfoStrs == null ? 0 : extraInfoStrs.length));
         }
-        return Long.valueOf(extraInfoStrs[1]);
+        return Long.parseLong(extraInfoStrs[1]);
     }
 
-    public static Long getInvisibleTime(String[] extraInfoStrs) {
+    public static long getInvisibleTime(String[] extraInfoStrs) {
         if (extraInfoStrs == null || extraInfoStrs.length < 3) {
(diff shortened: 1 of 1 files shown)


RockteMQ-AI (approved):
### Summary
Good performance optimization — reduces allocations in POP extra-info parsing hot paths:

1. **Primitive returns**: Changed `Long` to `long` to avoid boxing overhead
2. **Direct parsing**: `Long.parseLong()` instead of `Long.valueOf()` to skip intermediate objects
3. **Manual parsing**: Replaced `split()` with index-based parsing to avoid creating many intermediate String arrays
4. **Code reuse**: Extracted common logic into helper methods (`parseLongInfo`, `nextEntryEnd`, `validateEntry`, `buildKey`)

These changes reduce GC pressure in message consumption paths. The refactoring also improves maintainability by centralizing parsing logic.

LGTM.

---

PR #11031, opened 2026-09-05
[Bug] Return null when broker address table is unavailable

### LINKED ISSUE #10873: [Bug] BrokerData.selectBrokerAddr crashes for an empty address table

### RocketMQ version

Branch: develop

### JDK Version

Zulu OpenJDK 8 (8.94.0.17)

### Describe the Bug

BrokerData.selectBrokerAddr assumes brokerAddrs is non-null and non-empty. A default BrokerData instance throws NullPointerException, while an empty map reaches Random.nextInt(0) and throws IllegalArgumentException.

### Steps to Reproduce

1. Create a default BrokerData and call selectBrokerAddr.
2. Alternatively, set brokerAddrs to an empty HashMap and call selectBrokerAddr.
3. Observe the exception.

The method should return null when no broker address is registered, consistent with its nullable selection result.

### What Did You See Instead?

The method throws for both null and empty address tables.

### Additional Context

Master preference and slave fallback remain valid; only the no-address boundary is affected.

zjncs:
Fixes #10873

BrokerData.selectBrokerAddr now returns null when brokerAddrs is null or empty, instead of throwing NullPointerException or IllegalArgumentException from Random.nextInt(0). Master preference and random slave fallback remain unchanged.

Added regression coverage for both null and empty tables.

Verification: git diff --check passed; Maven unavailable locally, CI should run BrokerDataTest/remoting checks. AI-assisted contribution.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/route/BrokerData.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/protocol/route/BrokerData.java
@@ -89,6 +89,9 @@ public class BrokerData implements Comparable<BrokerData> {
      * @return Broker address.
      */
     public String selectBrokerAddr() {
+        if (this.brokerAddrs == null || this.brokerAddrs.isEmpty()) {
+            return null;
+        }
         String masterAddress = this.brokerAddrs.get(MixAll.MASTER_ID);
 
         if (masterAddress == null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Good defensive fix — adds null/empty check to `selectBrokerAddr()` to prevent NPE when broker address table is unavailable. Returns null gracefully instead of throwing.

The test validates both null and empty map cases.

LGTM.

---

PR #11032, opened 2026-09-05
[Bug] Persist empty checkpoint state

### LINKED ISSUE #10863: [Bug] CheckpointFile cannot persist empty checkpoints
macOS; reproduced with a temporary checkpoint file in the `common` module.

## RocketMQ version
Branch: `develop`

## JDK Version
Eclipse Temurin 17.0.19+10

## Describe the Bug
`CheckpointFile.write` returns immediately for an empty entry list, leaving old entries on disk. In addition, `read()` treats a valid empty primary file as a reason to load the backup, so stale backup entries can be restored instead of an intentionally empty state.

## Steps to Reproduce
1. Write a checkpoint containing entries.
2. Write an empty entry list to the same `CheckpointFile`.
3. Read the checkpoint.
4. Observe that old entries remain. A valid empty primary file also causes `read()` to return entries from its older backup.

An empty list should be persisted as a valid zero-entry checkpoint, and reading that valid primary should return an empty list. Backup fallback should only occur when the primary is missing or invalid.

## What Did You See Instead?
Empty writes are ignored, and a valid empty primary can be replaced in memory by stale backup content.

## Additional Context
The existing count and CRC format can represent an empty checkpoint without a format change.

zjncs:
Fixes #10863

CheckpointFile now writes a valid zero-entry checkpoint instead of ignoring an empty list. Reading falls back to the backup only when the primary is missing or invalid, so a valid empty primary cannot resurrect stale backup entries.

Added regression coverage for overwriting a populated checkpoint with an empty state.

Verification: git diff --check passed; Maven unavailable locally, CI should run CheckpointFileTest. AI-assisted contribution.

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/CheckpointFile.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/CheckpointFile.java
@@ -24,7 +24,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
-import org.apache.commons.collections.CollectionUtils;
 import org.apache.rocketmq.common.MixAll;
 import org.apache.rocketmq.common.UtilAll;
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix — the original code treated empty checkpoint as invalid and would skip writing it, causing stale data to persist. An empty checkpoint is a valid state that should be persisted.

Changes:
- `write()`: Removed the early return for empty entries
- `read()`: Check file existence instead of content emptiness to decide fallback

The test validates that writing an empty list after a non-empty one correctly returns empty on read.

LGTM.

---

PR #11034, opened 2026-09-05
[Bug] Accept zero and negative floating-point literals

### LINKED ISSUE #10862: [Bug] Filter parser rejects zero floating-point literals
macOS; reproduced with `SelectorParser` in the `filter` module.

## RocketMQ version
Branch: `develop`

## JDK Version
Eclipse Temurin 17.0.19+10

## Describe the Bug
`ConstantExpression.createFloat` uses `Double.MIN_VALUE` as the lower bound. In Java this constant is the smallest positive non-zero value, not the most negative finite value, so the valid literal `0.0` is rejected.

## Steps to Reproduce
1. Parse the selector expression `a = 0.0`.
2. Evaluate it with property `a` equal to `0.0`.
3. Observe a runtime error stating that `0.0` is less than `4.9E-324`.

The parser should accept zero and other finite negative floating-point literals within the `double` range.

## What Did You See Instead?
The parser throws while creating the zero constant.

## Additional Context
The finite lower bound should be `-Double.MAX_VALUE`; positive and negative infinity remain outside the accepted range.

zjncs:
Fixes #10862

ConstantExpression.createFloat now uses -Double.MAX_VALUE as the finite lower bound. Double.MIN_VALUE is the smallest positive non-zero value, so using it incorrectly rejected 0.0 and valid negative literals.

Added parser regression cases for 0.0 and -0.5. Positive and negative infinity remain outside the accepted finite range.

Verification: git diff --check passed; Maven unavailable locally, CI should run ParserTest/filter checks. AI-assisted contribution.

### DIFF
--- a/filter/src/main/java/org/apache/rocketmq/filter/expression/ConstantExpression.java
+++ b/filter/src/main/java/org/apache/rocketmq/filter/expression/ConstantExpression.java
@@ -58,8 +58,8 @@ public class ConstantExpression implements Expression {
         if (value > Double.MAX_VALUE) {
             throw new RuntimeException(text + " is greater than " + Double.MAX_VALUE);
         }
-        if (value < Double.MIN_VALUE) {
-            throw new RuntimeException(text + " is less than " + Double.MIN_VALUE);
+        if (value < -Double.MAX_VALUE) {
+            throw new RuntimeException(text + " is less than " + -Double.MAX_VALUE);
         }
         return new ConstantExpression(value);
     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix — `Double.MIN_VALUE` is the smallest **positive** value (4.9e-324), not the most negative value. The original code incorrectly rejected zero and negative floats. Using `-Double.MAX_VALUE` as the lower bound is correct.

The test validates parsing of `0.0` and `-0.5`, which would have failed before.

LGTM.

---

PR #11035, opened 2026-09-05
[Bug] Retain highest offset for out-of-order ProcessQueue batches

### LINKED ISSUE #10870: [Bug] ProcessQueue may return a stale next offset for out-of-order messages

### RocketMQ version

Branch: develop

### JDK Version

Zulu OpenJDK 8 (8.94.0.17)

### Describe the Bug

ProcessQueue.putMessage assigns queueOffsetMax from each newly inserted message. When a batch is not ordered by queue offset, a lower offset processed last overwrites the real maximum. After the queue is drained, removeMessage therefore returns a stale next offset.

### Steps to Reproduce

1. Create a ProcessQueue.
2. Insert two messages in one batch with queue offsets 10 and 5, in that order.
3. Remove both messages.
4. Observe that the returned next offset is 6.

The returned next offset should be 11, one greater than the highest offset ever inserted.

### What Did You See Instead?

The returned next offset is based on the last iterated message rather than the maximum queue offset.

### Additional Context

The issue is deterministic and is covered by ProcessQueueTest.

zjncs:
Fixes #10870

ProcessQueue.putMessage now updates queueOffsetMax with Math.max instead of overwriting it for every inserted message. An out-of-order batch can therefore no longer make removeMessage return a stale next offset.

Added a regression test inserting offsets 10 then 5 and asserting the drained next offset is 11.

Verification: git diff --check passed; Maven unavailable locally, CI should run ProcessQueueTest/client checks. AI-assisted contribution.

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ProcessQueue.java
@@ -136,7 +136,7 @@ public class ProcessQueue {
                     MessageExt old = msgTreeMap.put(msg.getQueueOffset(), msg);
                     if (null == old) {
                         validMsgCnt++;
-                        this.queueOffsetMax = msg.getQueueOffset();
+                        this.queueOffsetMax = Math.max(this.queueOffsetMax, msg.getQueueOffset());
                         msgSize.addAndGet(null == msg.getBody() ? 0 : msg.getBody().length);
                     }
                 }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix — when inserting out-of-order messages, `queueOffsetMax` should retain the highest offset, not just the last one processed. Using `Math.max()` ensures correctness regardless of insertion order.

The test validates the fix with out-of-order offsets (10, 5) and verifies the returned offset is 11 (max + 1).

LGTM.

---

PR #11036, opened 2026-09-05
[Bug] Convert the right string in Boolean comparisons

### LINKED ISSUE #10877: [Bug] Boolean-to-string filter comparison throws ClassCastException

### RocketMQ version

Branch: develop

### JDK Version

Zulu OpenJDK 8 (8.94.0.17)

### Describe the Bug

ComparisonExpression handles a Boolean left operand and String right operand by casting the Boolean left value to String. With string-expression conversion enabled by the selector parser, this branch always throws ClassCastException.

### Steps to Reproduce

1. Parse a selector expression such as TRUE=a.
2. Evaluate it with string property a=true.
3. Observe the Boolean-to-String cast failure.

The string right operand should be converted to Boolean and compared normally.

### What Did You See Instead?

The Boolean left operand is cast to String and evaluation throws ClassCastException.

### Additional Context

The reverse String-to-Boolean branch already converts the string operand correctly.

zjncs:
Fixes #10877

ComparisonExpression now converts the String right operand to Boolean when the left operand is Boolean. The previous branch cast the Boolean left operand to String, which always threw ClassCastException when selector string conversion was enabled.

Added a regression test evaluating TRUE = a with string property a=true.

Verification: git diff --check passed; Maven unavailable locally, CI should run ExpressionTest/filter checks. AI-assisted contribution.

### DIFF
--- a/filter/src/main/java/org/apache/rocketmq/filter/expression/ComparisonExpression.java
+++ b/filter/src/main/java/org/apache/rocketmq/filter/expression/ComparisonExpression.java
@@ -536,7 +536,7 @@ public abstract class ComparisonExpression extends BinaryExpression implements B
             try {
                 if (lc == Boolean.class) {
                     if (convertStringExpressions && rc == String.class) {
-                        lv = Boolean.valueOf((String) lv).booleanValue();
+                        rv = Boolean.valueOf((String) rv).booleanValue();
                     } else {
                         return -1;
                     }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct bug fix — the original code was casting the left operand (already a Boolean) to String and back, which always threw `ClassCastException`. Converting the right operand (the String) to Boolean is the correct approach.

The regression test (`TRUE = a` with string property `a=true`) properly validates the fix.

LGTM.

---

PR #11042, opened 2026-09-05
[ISSUE #11041] Fix off-by-one in paginated metadata sync dropping the last entry

### LINKED ISSUE #11041: [Bug] Paginated metadata sync silently drops the last topic/subscription group when count is one over a page boundary

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions) of this repository and believe that this is not a duplicate.

Linux, JDK 21, develop (ff8f6f74c)

### RocketMQ version
5.x develop

### Describe the Bug

The split-metadata pagination loops terminate one entry too early. The client-side loops in `MQClientAPIImpl` break when:

```java
if (topicSeq >= totalTopicNum - 1) {   // line 3125
if (groupSeq >= totalGroupNum - 1) {   // line 3033
```

The broker pages `[seq, seq + maxNum)` with no overlap (TopicConfigManager#subTopicConfig / SubscriptionGroupManager#subGroupTable, `totalTopicNum = size()`), so after page k the client has fetched `k·pageSize` entries. The loop therefore stops when `k·pageSize >= N − 1`, which is already true when exactly **one** entry remains unfetched, i.e. for every N ≡ 1 (mod pageSize).

Example: 2001 topics with the default page size 2000 — page 1 returns 2000 entries, `2000 >= 2001 - 1` is true, the loop breaks, and topic #2001 is never requested. No error, no retry — silently truncated metadata.

The same off-by-one exists broker-side in `BrokerOuterAPI` (getAllTopicConfig at line 834, getAllSubscriptionGroup at line 984); `BrokerOuterAPI#getAllTopicConfig` is used by `SlaveSynchronize#syncTopicConfig`, so a slave whose master has N ≡ 1 (mod 2000) topics permanently misses the last topic (the dataVersion comparison then matches, so no resync happens).

### Steps to Reproduce

1. Set `maxPageSizeInGetMetadata` to 100 (or have 2001 topics with the default 2000).
2. Call `DefaultMQAdminExt#getAllTopicConfig` on a broker with 101 topics.
3. Result contains 100 topics — the last one is missing.

### Expected Behavior

The loop must terminate only when `seq >= totalNum`; the last entry must be fetched.

### Corresponding PR

- Fix PR: #11042 (linked with `Closes #11041` in the PR description; contains the regression test that fails before the fix and passes after it).

unbridled-41:
### Problem / Evidence

The split-metadata pagination loops terminate one entry too early. The client loops in `MQClientAPIImpl#getAllTopicConfig` / `#getAllSubscriptionGroup` break on `topicSeq >= totalTopicNum - 1` / `groupSeq >= totalGroupNum - 1`, while the broker pages `[seq, seq + maxNum)` with no overlap (`TopicConfigManager#subTopicConfig`, `SubscriptionGroupManager#subGroupTable`, `totalTopicNum = size()`).

After page k the client holds `k·pageSize` entries, so the loop stops as soon as `k·pageSize >= N − 1` — which is already true when exactly one entry remains unfetched, i.e. for every **N ≡ 1 (mod pageSize)**. Example: 2001 topics with the default page size 2000 → page 1 returns 2000 entries, `2000 >= 2001 − 1` holds, the loop breaks, topic #2001 is never requested. No error, no retry — silently truncated metadata.

The identical off-by-one exists in `BrokerOuterAPI#getAllTopicConfig` (used by `SlaveSynchronize#syncTopicConfig`) and `BrokerOuterAPI#getAllSubscriptionGroup`: a slave whose master has N ≡ 1 (mod 2000) topics permanently misses the last topic — and the dataVersion comparison then matches, so no later resync repairs it.

Regression tests `MQClientAPIImplTest#testGetAllTopicConfigFetchesLastEntryWhenTotalIsOneOverPageSize` and `#testGetAllSubscriptionGroupFetchesLastEntryWhenTotalIsOneOverPageSize` fail before the fix (`expected:<101> but was:<100>`) and pass after.

### Root cause / Fix

Terminate when `seq >= totalNum`. When the page completes the set the condition still breaks immediately; when exactly one entry remains it now continues and fetches the final page (a subsequent empty page still terminates because `seq >= totalNum` remains true). Applied to all four occurrences of the same defect: `MQClientAPIImpl` (topic + subscription group) and `BrokerOuterAPI` (topic + subscription group).

### Priority

PRIORITY = 74：影响 28（静默元数据截断——mqadmin 导出/同步工具丢最后一个 topic/消费组，SlaveSynchronize 全量同步永久缺最新 topic 且不会自愈）+ 波及范围 14（4 处同类循环、client+broker 两端）+ 可复现性 20（确定性单元测试，N = pageSize + 1 边界）+ 维护价值 12（一行终止条件 ×4）。FIX_CONFIDENCE = 90。

### Tests

- `mvn -pl client test -Dtest=MQClientAPIImplTest#testGetAllTopicConfigFetchesLastEntryWhenTotalIsOneOverPageSize`
  - before fix (ff8f6f74c + test only): `Tests run: 1, Failures: 1` — `expected:<101> but was:<100>`
  - after fix: `Tests run: 1, Failures: 0`
- same for `testGetAllSubscriptionGroupFetchesLastEntryWhenTotalIsOneOverPageSize`
- `mvn -pl client test -Dtest=MQClientAPIImplTest`: `Tests run: 135, Failures: 0, Errors: 0`
- `mvn -pl broker compile`: OK (BrokerOuterAPI change)

### Risk

Low. The only behavior change is that exactly one extra page request is issued in the boundary case (and none in any other case — for all other N the old and new conditions break at the same page). Old-version brokers that do not return `totalTopicNum`/`totalGroupNum` are unaffected (the null-compatibility break precedes this condition).

Closes #11041

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
@@ -831,7 +831,7 @@ public class BrokerOuterAPI {
                     continue;
                 }
 
-                if (topicSeq >= totalTopicNum - 1) {
+                if (topicSeq >= totalTopicNum) {
                     LOGGER.info("get all topic config, totalTopicNum: {}", totalTopicNum);
                     break;
                 }
@@ -981,7 +981,7 @@ public class BrokerOuterAPI {
                     continue;
                 }
 
-                if (groupSeq >= totalGroupNum - 1) {
+                if (groupSeq >= totalGroupNum) {
                     LOGGER.info("get all subscription group config, totalGroupNum: {}", totalGroupNum);
                     break;
                 }
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -3030,7 +3030,7 @@ public class MQClientAPIImpl implements NameServerUpdateCallback, StartAndShutdo
                     continue;
                 }
 
-                if (groupSeq >= totalGroupNum - 1) {
+                if (groupSeq >= totalGroupNum) {
                     log.info("get all subscription group config, totalGroupNum: {}", totalGroupNum);
                     break;
                 }
@@ -3122,7 +3122,7 @@ public class MQClientAPIImpl implements NameServerUpdateCallback, StartAndShutdo
                     continue;
                 }
 
-                if (topicSeq >= totalTopicNum - 1) {
+                if (topicSeq >= totalTopicNum) {
                     log.info("get all topic config, totalTopicNum: {}", totalTopicNum);
                     break;
                 }
--- a/client/src/test/java/org/apache/rocketmq/client/impl/MQClientAPIImplTest.java
+++ b/client/src/test/java/org/apache/rocketmq/client/impl/MQClientAPIImplTest.java
@@ -65,6 +65,7 @@ import org.apache.rocketmq.remoting.netty.NettyRemotingClient;
 import org.apache.rocketmq.remoting.netty.ResponseFuture;
 import org.apache.rocketmq.remoting.protocol.RemotingCommand;
 import org.apache.rocketmq.remoting.protocol.RemotingSerializable;
+import org.apache.rocketmq.remoting.protocol.DataVersion;
 import org.apache.rocketmq.remoting.protocol.RequestCode;
 import org.apache.rocketmq.remoting.protocol.ResponseCode;
 import org.apache.rocketmq.remoting.protocol.admin.ConsumeStats;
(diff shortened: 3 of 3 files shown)


unbridled-41:
Evidence chain (audit 2026-09-05):

**Before the fix** — ff8f6f74c + regression tests only (all four production loop changes stashed):

```
testGetAllTopicConfigFetchesLastEntryWhenTotalIsOneOverPageSize:1782 expected:<101> but was:<100>
testGetAllSubscriptionGroupFetchesLastEntryWhenTotalIsOneOverPageSize:1825 expected:<101> but was:<100>
```
The loop breaks after page 1; the entry at index pageSize is never requested and no error is raised.

**After the fix** — branch tip 55dbfd3c3, re-measured 2026-09-05: both regression tests pass (`Tests run: 1, Failures: 0` each); full class `mvn -pl client test -Dtest=MQClientAPIImplTest` = 135/135 (measured earlier the same day); `mvn -pl broker compile` OK for the two BrokerOuterAPI occurrences.

Broker-side paging contract re-verified in code (the premise the client loop must match): `AdminBrokerProcessor.java:958` sets `totalTopicNum = tcManager.getTopicConfigTable().size()` and `:1795` sets `totalGroupNum = sgManager.getSubscriptionGroupTable().size()`; pages are sliced `[seq, seq + maxNum)` with no overlap (`TopicConfigManager#subTopicConfigTable`, comment "[topicSeq, topicSeq + maxTopicNum)" at line 744, and `SubscriptionGroupManager#subGroupTable` line 324). Therefore after page k the client holds k·pageSize entries and the old terminator `seq >= totalNum - 1` is already true when exactly one entry remains — i.e. for every N ≡ 1 (mod pageSize). With the fix, a fully-consumed set still terminates on the completing page and a subsequent empty page still terminates (seq >= totalNum remains true), so no extra request is issued outside the boundary case.

RockteMQ-AI (approved):
### Summary
Fixes an off-by-one error in paginated metadata sync where the last entry was dropped. Changes `>= totalNum - 1` to `>= totalNum` in the page-end check, and adds a test that verifies the final page boundary. Clean, minimal fix.

LGTM.

---

PR #11046, opened 2026-09-05
[ISSUE #11045] Rebind broker channel on heartbeat so stale channel close cannot evict a live broker

### LINKED ISSUE #11045: [Bug] Controller: a stale channel-close event evicts a re-registered broker and triggers a spurious master election

All platforms; reproduced with a deterministic unit test on the current `develop` branch.

### RocketMQ version

`develop` at `ff8f6f74c`

### JDK Version

All

### Describe the Bug

`DefaultBrokerHeartbeatManager#onBrokerHeartbeat` never rebinds the channel of an existing `BrokerLiveInfo` (the `prev != null` branch only refreshes the timestamp/timeout/priority/epoch), so the live entry stays pinned to the broker's **first-ever** channel. When that old channel eventually fires `channelInactive` / `onChannelException` / `onChannelIdle`, `onBrokerChannelClose` matches it against the stale stored channel, removes the live entry of a broker that is healthy on a new channel, and fires the broker-inactive lifecycle event, which makes `ControllerManager#onBrokerInactive` trigger a master election for a perfectly healthy master.

This is realistic in at least two ordinary situations:

1. The broker-to-controller connection dies half-open on the controller side; the broker client detects it first, reconnects, and keeps heartbeating on the new channel. Minutes later the controller's TCP stack learns the old channel is dead and fires the close event.
2. The controller-side idle/close event for the old channel is only delivered after the broker has already re-registered on a new channel.

### Steps to Reproduce

1. Broker B (master of its broker-set) sends a heartbeat to the controller on channel A → `BrokerLiveInfo` holds channel A.
2. The broker reconnects and keeps heartbeating on channel B (every heartbeat hits the `prev != null` branch, channel stays A).
3. Channel A fires `onBrokerChannelClose(A)` on the controller.
4. Observe that B's live entry is removed and `notifyBrokerInActive` fires although B is actively heartbeating on channel B; `ControllerManager#onBrokerInactive` then calls `triggerElectMaster` because the evicted broker is the current master.

A close event for a channel that the broker no longer uses must not evict the live entry: the heartbeat path should rebind `BrokerLiveInfo` to the current channel, so `onBrokerChannelClose` only removes entries whose stored channel is the one that actually closed.

### What Did You See Instead?

The live entry of the re-registered master is removed, a broker-inactive notification is fired, and the controller bumps the master epoch / runs a failover for a healthy broker-set.

### Additional Context

The sibling NameServer implementation (`RouteInfoManager`) updates the broker's channel on every heartbeat, so the controller behavior is inconsistent with the nameserver. The fix is to rebind the channel in the `prev != null` branch of `onBrokerHeartbeat` when a non-null channel arrives; a regression test in `DefaultBrokerHeartbeatManagerTest` reproduces the eviction deterministically.

### Corresponding PR

- PR apache/rocketmq#11046

unbridled-41:
Closes #11045

### Problem / Evidence

`DefaultBrokerHeartbeatManager#onBrokerHeartbeat` never rebinds the channel of an existing `BrokerLiveInfo` — the `prev != null` branch only refreshes the timestamp, timeout, election priority and epoch (`DefaultBrokerHeartbeatManager.java:133-142`), so the live entry stays pinned to the broker's **first-ever** channel.

When the broker-to-controller connection dies half-open (or the controller-side close event for the old channel arrives late), the broker client reconnects first and keeps heartbeating on a new channel. When the old channel eventually fires `channelInactive`/`onChannelException`/`onChannelIdle`, `BrokerHousekeepingService` (wired as the controller `ChannelEventListener` in `ControllerManager`, passed to `DLedgerServer`) calls `onBrokerChannelClose(oldChannel)`, which matches the stale stored channel, **removes the live entry of a broker that is healthy on the new channel** and fires `notifyBrokerInActive`. `ControllerManager#onBrokerInactive` then checks `getReplicaInfo` and, because the evicted broker is the current master, calls `triggerElectMaster` — a spurious failover (epoch bump, possible master change, sync-state reset) for a healthy broker-set.

Deterministic regression test `DefaultBrokerHeartbeatManagerTest#testStaleChannelCloseDoesNotEvictReRegisteredBroker` fails on unmodified `develop`:

```
DefaultBrokerHeartbeatManagerTest.testStaleChannelCloseDoesNotEvictReRegisteredBroker:74
A stale channel close must not evict a broker alive on a new channel
Tests run: 2, Failures: 1, Errors: 0, Skipped: 1  (develop @ ff8f6f74c, 2026-09-05)
```

### Root cause / Fix

The heartbeat path keeps the stale channel reference forever. This is inconsistent with the NameServer implementation, where the broker's channel is refreshed on every registration heartbeat.

Fix: rebind the live entry to the current channel in `onBrokerHeartbeat` when a non-null channel arrives, so `onBrokerChannelClose` only removes entries whose stored channel is the one that actually closed:

```java
// Rebind the live entry to the current channel, otherwise a close event of a stale
// channel would evict a broker that is already alive on a new channel
if (channel != null && channel != prev.getChannel()) {
    prev.setChannel(channel);
}
```

### Priority

PRIORITY = 76（影响 30 + 波及范围 14 + 可复现 18 + 维护价值 14），FIX_CONFIDENCE = 88。

- 影响 30/40: spurious master election / epoch bump fencing a healthy master — an availability and correctness event for every controller-mode deployment (the default dLedger controller uses this heartbeat manager).
- 波及范围 14/20: all controller-mode deployments; the mechanism (heartbeat + channel lifecycle) runs continuously.
- 可复现 18/20: deterministic two-channel unit test; the reconnect-before-close ordering is a routine network scenario.
- 维护价值 14/20: same bug family as the nameserver stale-channel issues; small, well-bounded fix aligning controller with nameserver behavior.

### Tests

- Regression test added: `DefaultBrokerHeartbeatManagerTest#testStaleChannelCloseDoesNotEvictReRegisteredBroker` (fails before the fix as shown above).
- After the fix (`develop @ ff8f6f74c` + this change, 2026-09-05):

```
mvn -pl controller test -Dtest=DefaultBrokerHeartbeatManagerTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0  — BUILD SUCCESS
```

- Full controller module: `mvn -pl controller test` — all tests pass except the pre-existing baseline failure `ReplicasInfoManagerTest#testSerialize` (JDK-21 `InaccessibleObjectException: Unable to make field ... AtomicLong.value accessible`), which reproduces identically on unmodified `develop @ ff8f6f74c` and is unrelated to this change.

### Risk

Very low. The added branch only updates a mutable field that was previously never refreshed after creation; heartbeats without a channel (the existing test passes `null`) keep the old behavior. The close-detection semantics (`getChannel() == channel` identity match) are unchanged — they simply now compare against the current channel. The jRaft heartbeat manager keeps its own raft-routed close handling and is untouched.

### DIFF
--- a/controller/src/main/java/org/apache/rocketmq/controller/impl/heartbeat/DefaultBrokerHeartbeatManager.java
+++ b/controller/src/main/java/org/apache/rocketmq/controller/impl/heartbeat/DefaultBrokerHeartbeatManager.java
@@ -134,6 +134,11 @@ public class DefaultBrokerHeartbeatManager implements BrokerHeartbeatManager {
             prev.setLastUpdateTimestamp(System.currentTimeMillis());
             prev.setHeartbeatTimeoutMillis(realTimeoutMillis);
             prev.setElectionPriority(realElectionPriority);
+            // Rebind the live entry to the current channel, otherwise a close event of a stale
+            // channel would evict a broker that is already alive on a new channel
+            if (channel != null && channel != prev.getChannel()) {
+                prev.setChannel(channel);
+            }
             if (realEpoch > prev.getEpoch() || realEpoch == prev.getEpoch() && realMaxOffset > prev.getMaxOffset()) {
                 prev.setEpoch(realEpoch);
                 prev.setMaxOffset(realMaxOffset);
--- a/controller/src/test/java/org/apache/rocketmq/controller/impl/DefaultBrokerHeartbeatManagerTest.java
+++ b/controller/src/test/java/org/apache/rocketmq/controller/impl/DefaultBrokerHeartbeatManagerTest.java
@@ -16,8 +16,10 @@
  */
 package org.apache.rocketmq.controller.impl;
 
+import io.netty.channel.embedded.EmbeddedChannel;
 import org.apache.rocketmq.common.ControllerConfig;
 import org.apache.rocketmq.controller.BrokerHeartbeatManager;
+import org.apache.rocketmq.controller.impl.heartbeat.BrokerLiveInfo;
 import org.apache.rocketmq.controller.impl.heartbeat.DefaultBrokerHeartbeatManager;
 import org.junit.Before;
 import org.junit.Test;
@@ -25,6 +27,9 @@ import org.junit.Test;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 public class DefaultBrokerHeartbeatManagerTest {
@@ -49,6 +54,35 @@ public class DefaultBrokerHeartbeatManagerTest {
             1, 1L, -1L, 0);
         assertTrue(latch.await(5000, TimeUnit.MILLISECONDS));
         this.heartbeatManager.shutdown();
+    }
+
+    @Test
+    public void testStaleChannelCloseDoesNotEvictReRegisteredBroker() {
+        final EmbeddedChannel oldChannel = new EmbeddedChannel();
+        final EmbeddedChannel newChannel = new EmbeddedChannel();
+        try {
+            // First heartbeat pins the live entry to the old channel
+            this.heartbeatManager.onBrokerHeartbeat("cluster1", "broker1", "127.0.0.1:7000", 0L, 3000L, oldChannel,
+                1, 1L, -1L, 0);
+            // The broker reconnects and keeps heartbeating on a new channel
+            this.heartbeatManager.onBrokerHeartbeat("cluster1", "broker1", "127.0.0.1:7000", 0L, 3000L, newChannel,
+                1, 1L, -1L, 0);
+            // The old channel fires its close event only after the re-registration
+            this.heartbeatManager.onBrokerChannelClose(oldChannel);
+
+            final BrokerLiveInfo liveInfo = this.heartbeatManager.getBrokerLiveInfo("cluster1", "broker1", 0L);
+            assertNotNull("A stale channel close must not evict a broker alive on a new channel", liveInfo);
+            assertEquals("The live entry must be rebound to the current channel",
+                newChannel, liveInfo.getChannel());
+
+            // The close event of the current channel still removes the entry
+            this.heartbeatManager.onBrokerChannelClose(newChannel);
+            assertNull(this.heartbeatManager.getBrokerLiveInfo("cluster1", "broker1", 0L));
+        } finally {
+            oldChannel.finishAndReleaseAll();
+            newChannel.finishAndReleaseAll();
+            this.heartbeatManager.shutdown();
+        }
     }
 
 }


RockteMQ-AI (approved):
### Summary
Fixes a spurious broker eviction and potential master election triggered when a stale channel close event arrives after the broker has already reconnected on a new channel. The heartbeat manager was pinning the live entry to the broker's first-ever channel instead of refreshing it on each heartbeat.

### Analysis
- **Correctness**: The fix is correct and minimal. Rebinding `prev.setChannel(channel)` when the incoming channel differs from the stored one ensures that `onBrokerChannelClose` only matches the actual current channel. This is consistent with how the NameServer already handles broker channel refresh.
- **Safety**: The `channel != null && channel != prev.getChannel()` guard prevents unnecessary writes and handles the null-channel edge case. The identity check (`!=`) is appropriate here since Netty channels are reference-stable objects.
- **Race condition**: The fix correctly addresses the half-open connection scenario described in the issue. When the old channel eventually fires `channelInactive`, the stored channel has already been rebound to the new one, so the close event won't match and won't evict the healthy entry.
- **Test**: `testStaleChannelCloseDoesNotEvictReRegisteredBroker` is a well-designed deterministic test that simulates the reconnect-before-close sequence and verifies the invariant.

LGTM.

---

unbridled-41:
**Evidence-chain audit (2026-09-06)** — every technical assertion in the PR body re-verified against source; tests re-run at the branch tip.

Code-level anchors (personally re-read at the fix branch):

- `DefaultBrokerHeartbeatManager#onBrokerHeartbeat` `prev != null` branch (main `:133-142` pre-fix) refreshes only timestamp/timeout/priority/epoch; `BrokerLiveInfo#setChannel` exists but is called **nowhere** in the file (grep `setChannel` → only the definition `BrokerLiveInfo.java:151-153`).
- `onBrokerChannelClose` (`:147-161` pre-fix) evicts by channel **identity** match and fires `notifyBrokerInActive` on the stale match.
- Wiring confirmed end-to-end: `BrokerHousekeepingService.onChannelClose/onChannelException/onChannelIdle` → `onBrokerChannelClose` (`BrokerHousekeepingService.java:38-50`); it is registered as the controller `ChannelEventListener` (`ControllerManager.java:82`, passed at `:123`) and handed to `DLedgerServer` (`DLedgerController.java:117→136`). `ControllerManager#onBrokerInactive` triggers `triggerElectMaster` when the evicted broker is the current master.
- Nameserver-consistency claim in the body now verified directly: `RouteInfoManager.registerBroker` **replaces** the whole `BrokerLiveInfo` (carrying `ctx.channel()`) on every registration heartbeat (`namesrv/.../RouteInfoManager.java:373-380`), so the NameServer refreshes the channel every heartbeat — the controller did not.

Fresh re-run at this PR's tip `a7accbf9b` (2026-09-06, JDK 21):

```
mvn -pl controller test -Dtest=DefaultBrokerHeartbeatManagerTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

The jRaft heartbeat manager (`RaftBrokerHeartBeatManager`) routes live info through raft consensus with `channel = null` in its `BrokerLiveInfo`, so it has a different close-detection mechanism and is intentionally untouched by this fix.

ryux1 (commented):
**major:** This fixes the sequential ordering in the test, but `onBrokerChannelClose` still has a check-then-remove race. It finds an entry whose channel is `oldChannel`, saves only the key in `addrInfo`, exits the loop, and later calls `brokerLiveTable.remove(addrInfo)`. A heartbeat on `newChannel` can run between the identity check and that removal, mutate the same `BrokerLiveInfo` to the new channel, and then be removed by the stale close anyway. The inactive notification is also submitted before the removal and closes over the mutable entry value.

The new test executes the heartbeat and close serially, so it cannot expose that interleaving. Could heartbeat rebinding and close removal be made atomic per broker identity (for example, both through `ConcurrentHashMap.compute`, with close returning null only when the value still carries the closing channel), and could a coordinated concurrency test pause the close after its match while the replacement heartbeat runs? A conditional `remove(key, value)` alone is insufficient here because the heartbeat mutates the same value object in place.

PR #11048, opened 2026-09-05
[ISSUE #11047] Retry only failed timer requests so succeeded ones are not enqueued twice

### LINKED ISSUE #11047: [Bug] Timer enqueue retry re-enqueues already-succeeded requests, delivering scheduled messages multiple times

All platforms; reproduced with a deterministic unit test on the current `develop` branch.

### RocketMQ version

`develop` at `ff8f6f74c`

### JDK Version

All

### Describe the Bug

When a batch of timer requests is enqueued, `TimerEnqueuePutService#fetchAndPutTimerRequest` retries the **whole** batch after any single request fails. The retry loop re-invokes `putMessageToTimerWheel` for every request in `trs` without checking `TimerRequest#isSucc`, so requests whose `doEnqueue` already succeeded in an earlier round are enqueued **again** — each retry round appends another `TimerLog` unit for the same message into the same timer-wheel slot.

On dequeue every unit is an independent `MAGIC_DEFAULT` record, so the same scheduled message is converted and delivered to the real topic multiple times (and the slot `num` counter is inflated, skewing `getAllNum`/`isReject` flow-control decisions).

A partial batch failure is realistic: `TimerLog#append` returns -1 when a new mapped file cannot be allocated in time (IO pressure at file rollover), and any unexpected throwable while `timerSkipUnknownError=false` fails the current request only — the other requests of the batch have already succeeded.

### Steps to Reproduce

1. Put two timer requests A and B into `enqueuePutQueue` (same or different delay times).
2. Make `doEnqueue` fail exactly once — for A's first attempt only (equivalent to a transient `TimerLog#append` failure).
3. Drive `TimerEnqueuePutService#fetchAndPutTimerRequest` once.
4. Observe that A's `doEnqueue` runs twice: the successful first attempt is re-processed by the retry round, appending a second `TimerLog` unit for the same message.

The retry round must only re-process the requests that did not succeed; already-succeeded requests are never re-enqueued, so the message is delivered exactly once.

### What Did You See Instead?

Round 2 re-enqueues A and the scheduled message is delivered twice (once per appended `TimerLog` unit).

### Additional Context

Proposed fix: filter the batch on `TimerRequest#isSucc()` before each retry round (requests routed to the dequeue path are released with `succ=true` before the shared latch completes, so they are never re-put either). A regression test in `TimerMessageStoreTest` counts `doEnqueue` invocations per physical offset and fails on unmodified `develop`.

### Corresponding PR

- PR apache/rocketmq#11048

unbridled-41:
Closes #11047

### Problem / Evidence

`TimerEnqueuePutService#fetchAndPutTimerRequest` retries the **whole** batch when any single request fails:

```java
while (!isStopped()) {
    CountDownLatch latch = new CountDownLatch(trs.size());
    for (TimerRequest req : trs) {          // <-- every request, including already-succeeded ones
        req.setLatch(latch);
        ... this.putMessageToTimerWheel(req);
    }
    checkDequeueLatch(latch, -1);
    boolean allSuccess = trs.stream().allMatch(TimerRequest::isSucc);
    ...
}
```

If one request's `doEnqueue` fails (e.g. `TimerLog#append` returns -1 when a new mapped file cannot be allocated in time at rollover, or any unexpected throwable while `timerSkipUnknownError=false`), every subsequent round re-invokes `putMessageToTimerWheel` for the requests whose `doEnqueue` already succeeded — each round appends another `TimerLog` unit for the same message into the same timer-wheel slot. On dequeue, every unit is an independent `MAGIC_DEFAULT` record, so the scheduled message is delivered to the real topic **once per redundant unit** (and the slot `num` counter is inflated, skewing `getAllNum`/`isReject` flow-control decisions).

Deterministic regression test `TimerMessageStoreTest#testEnqueueRetryDoesNotReprocessSucceededRequests` (counts `doEnqueue` invocations per physical offset; `doEnqueue` is made to fail exactly once, for the first attempt of one of two batched requests). Fails on unmodified `develop`:

```
java.lang.AssertionError: expected:<1> but was:<2]
	at org.apache.rocketmq.store.timer.TimerMessageStoreTest.testEnqueueRetryDoesNotReprocessSucceededRequests(TimerMessageStoreTest.java:703)
Tests run: 12, Failures: 1, Errors: 0, Skipped: 0   (develop @ ff8f6f74c + test only, 2026-09-05)
```

The failed assertion is the already-succeeded request having been enqueued a second time.

### Root cause / Fix

The retry loop re-processes the whole batch instead of the failed subset.

Fix: filter the batch on `TimerRequest#isSucc()` before each retry round and only re-put the requests that have not succeeded yet. This is safe for the dequeue-routed path: requests handed to `dequeuePutQueue` are released (with `succ=true`) by `TimerDequeuePutService` before the shared latch completes, so they are never re-put either. Round 1 is unchanged (nothing is successful yet, the filtered list equals the batch), and the unknown-error hold semantics are preserved.

### Priority

PRIORITY = 78（影响 32 + 波及范围 16 + 可复现 16 + 维护价值 14），FIX_CONFIDENCE = 85。

- 影响 32/40: duplicate delivery of scheduled/delay messages — a user-visible message-correctness violation on a realistic transient failure (IO pressure at TimerLog rollover, unexpected errors without the skip flag).
- 波及范围 16/20: the timer wheel backs all delayed/scheduled messages; every partial-batch enqueue failure duplicates up to the whole rest of the batch (batches are up to ~10 requests).
- 可复现 16/20: fully deterministic unit test by injecting a one-shot `doEnqueue` failure; production trigger requires a partial batch failure but the retry-duplication itself is certain.
- 维护价值 14/20: small, local fix to the retry loop; clearly separable from the surrounding logic.

### Tests

- Regression test added: `TimerMessageStoreTest#testEnqueueRetryDoesNotReprocessSucceededRequests` (fails before the fix as shown above).
- After the fix (`develop @ ff8f6f74c` + this change, 2026-09-05):

```
mvn -pl store test -Dtest=TimerMessageStoreTest
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
mvn -pl store test -Dtest='TimerCheckPointTest,TimerLogTest,TimerWheelTest,TimerMetricsTest,TimerEngineSwitchVerifyTest'
Tests run: 21, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

### Risk

Low. Round 1 behavior is identical (no request is successful before the first round, so the filtered retry list equals the original batch). Requests that succeed are simply excluded from later rounds, which is exactly the invariant the commit offset advance (`commitQueueOffset = trs.get(trs.size()-1)...`) already assumes — the batch is only acknowledged once every request succeeded. The `holdMomentForUnknownError` backoff and the `timerWheelSnapshotFlush` locking are unchanged.

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/TimerMessageStore.java
@@ -1512,8 +1512,19 @@ public class TimerMessageStore {
             }
 
             while (!isStopped()) {
-                CountDownLatch latch = new CountDownLatch(trs.size());
+                // Only retry the requests that have not succeeded yet, otherwise the
+                // already-succeeded requests would be enqueued again and delivered multiple times
+                List<TimerRequest> retryList = new ArrayList<>(trs.size());
                 for (TimerRequest req : trs) {
+                    if (!req.isSucc()) {
+                        retryList.add(req);
+                    }
+                }
+                if (retryList.isEmpty()) {
+                    break;
+                }
+                CountDownLatch latch = new CountDownLatch(retryList.size());
+                for (TimerRequest req : retryList) {
                     req.setLatch(latch);
                     if (storeConfig.isTimerWheelSnapshotFlush()) {
                         synchronized (lockWhenFlush) {
@@ -1524,7 +1535,7 @@ public class TimerMessageStore {
                     }
                 }
                 checkDequeueLatch(latch, -1);
-                boolean allSuccess = trs.stream().allMatch(TimerRequest::isSucc);
+                boolean allSuccess = retryList.stream().allMatch(TimerRequest::isSucc);
                 if (allSuccess) {
                     break;
                 } else {
--- a/store/src/test/java/org/apache/rocketmq/store/timer/TimerMessageStoreTest.java
+++ b/store/src/test/java/org/apache/rocketmq/store/timer/TimerMessageStoreTest.java
@@ -653,6 +653,56 @@ public class TimerMessageStoreTest {
         }
     }
 
+    @Test
+    public void testEnqueueRetryDoesNotReprocessSucceededRequests() throws Exception {
+        Assume.assumeFalse(MixAll.isWindows());
+        String rootDir = StoreTestUtils.createBaseDir();
+        baseDirs.add(rootDir);
+        TimerCheckpoint timerCheckpoint =
+            new TimerCheckpoint(rootDir + File.separator + "config" + File.separator + "timercheck");
+        TimerMetrics timerMetrics =
+            new TimerMetrics(rootDir + File.separator + "config" + File.separator + "timermetrics");
+
+        final long poisonOffsetPy = 100L;
+        final long normalOffsetPy = 200L;
+        final Map<Long, AtomicInteger> enqueueCalls = new ConcurrentHashMap<>();
+
+        TimerMessageStore timerMessageStore =
+            new TimerMessageStore(messageStore, storeConfig, timerCheckpoint, timerMetrics, null) {
+                @Override
+                public boolean doEnqueue(long offsetPy, int sizePy, long delayedTime,
+                    MessageExt messageExt, boolean isFromTimeline) {
+                    int calls = enqueueCalls.computeIfAbsent(offsetPy, key -> new AtomicInteger(0)).incrementAndGet();
+                    if (offsetPy == poisonOffsetPy && calls == 1) {
+                        // Simulate a transient TimerLog append failure for this request only,
+                        // leaving the other request of the batch successful
+                        return false;
+                    }
+                    return super.doEnqueue(offsetPy, sizePy, delayedTime, messageExt, isFromTimeline);
+                }
+            };
+        messageStore.setTimerMessageStore(timerMessageStore);
+        timerStores.add(timerMessageStore);
+
+        long delayTime = System.currentTimeMillis() + 10 * precisionMs;
+        TimerRequest poisonReq = new TimerRequest(poisonOffsetPy, msgBody.length, delayTime,
+            System.currentTimeMillis(), TimerMessageStore.MAGIC_DEFAULT);
+        poisonReq.setMsg(buildMessage(delayTime, "TimerTest_enqueueRetryPoison", false));
+        TimerRequest normalReq = new TimerRequest(normalOffsetPy, msgBody.length, delayTime,
+            System.currentTimeMillis(), TimerMessageStore.MAGIC_DEFAULT);
+        normalReq.setMsg(buildMessage(delayTime, "TimerTest_enqueueRetryNormal", false));
+        timerMessageStore.enqueuePutQueue.put(poisonReq);
+        timerMessageStore.enqueuePutQueue.put(normalReq);
+
+        TimerMessageStore.TimerEnqueuePutService service = timerMessageStore.new TimerEnqueuePutService();
+        service.fetchAndPutTimerRequest();
+
+        // The request whose first attempt failed is retried and then succeeds
+        assertEquals(2, enqueueCalls.get(poisonOffsetPy).get());
+        // The request that already succeeded in the first round must not be re-enqueued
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes duplicate delivery of scheduled/delay messages when `TimerEnqueuePutService` retries a batch with partial failures. The original code re-enqueues the entire batch on each retry round, including requests that already succeeded — each redundant `putMessageToTimerWheel` call creates an independent delivery record.

### Analysis
- **Correctness**: The fix is correct and well-targeted. Filtering `trs` by `!req.isSucc()` before each retry round ensures only failed requests are re-processed. The first iteration is unchanged (nothing succeeded yet), and the dequeue-routed path is safe because those requests are released with `succ=true` before the shared latch completes.
- **Latch count**: Properly updated to `retryList.size()` — consistent with the actual number of requests being processed.
- **Termination**: The `retryList.isEmpty()` early-exit is a good guard against an edge case where all requests succeed between the filter and the latch check.
- **Test**: `testEnqueueRetryDoesNotReprocessSucceededRequests` is a solid deterministic regression test — counting `doEnqueue` invocations per physical offset directly verifies the no-duplicate invariant.

LGTM.

---

unbridled-41:
**Evidence-chain audit (2026-09-06)** — every technical assertion in the PR body re-verified against source; tests re-run at the branch tip.

Code-level anchors (personally re-read at the fix branch):

- Trigger plausibility verified: `TimerLog#append` (`store/.../timer/TimerLog.java:64-72`) calls `getLastMappedFile(0)` when the current file is full and returns **-1** when a new mapped file cannot be allocated ("Create mapped file1 error for timer log") — `doEnqueue` then returns false, which is exactly the partial-batch failure that drives the retry round.
- Duplication consequence verified: `TimerMessageStore#dequeue` (`:1018+`) walks the slot chain **unit by unit** ("read the timer log one by one"); every non-`MAGIC_DELETE` unit becomes an independent `TimerRequest` that is converted and put to the real topic, so a re-appended unit delivers the same message again.
- Flow-control skew verified: `isReject` (`TimerMessageStore.java:1895-1905`) reads `timerWheel.getNum(deliverTimeMs)` (the slot `num`), and `getDequeueBehindMessages`/`getAllNum` (`:1888`, `:1928`) accumulate the same counter — inflated `num` from duplicated units distorts these decisions.
- Safety of the fix verified in source: requests routed to `dequeuePutQueue` are released in the `finally` via `tr.idempotentRelease(!tmpDequeueChangeFlag)` (`:1688`), where `tmpDequeueChangeFlag` is set **only** on service stop (`:1630`); in the normal path they are released with `succ = true` before the shared latch completes, so the retry filter never re-puts them. On stop, the retry loop exits via `isStopped()` as well, so no path re-puts a released request.

Fresh re-run at this PR's tip `dd3fc6777` (2026-09-06, JDK 21):

```
mvn -pl store test -Dtest=TimerMessageStoreTest
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

Pre-fix failure remains as anchored in the PR body: `expected:<1> but was:<2>` at `TimerMessageStoreTest.java:703` on `develop @ ff8f6f74c` + test only (2026-09-05).

ryux1 (commented):
**major:** The failed subset is correct, but each failed `TimerRequest` has already executed `idempotentRelease(false)`, which sets `released = true`. On the retry round, `setLatch(latch)` replaces the latch without clearing `released`; when that retry succeeds, `idempotentRelease(true)` updates `succ` but cannot count down the new latch because `released` is still true. `checkDequeueLatch` therefore takes its timeout/warning path on every successful retry before `allSuccess` can break the loop.

The regression test masks this because it calls `fetchAndPutTimerRequest()` while dequeue is not running, and `checkDequeueLatch` explicitly exits early in that state. Could the test run with the production dequeue state (or directly assert the retry latch reaches zero), and could the retry bookkeeping re-arm completion safely? A simple `released=false` reset needs care for requests previously handed to `dequeuePutQueue`, because a late release from the prior attempt could race with the new latch.

PR #10930, opened 2026-08-14
[ISSUE #10929] Add LmqPrefixIndex to accelerate wildcard dispatch and refine lite subscription model

f1amingo:
### Which Issue(s) This PR Fixes

- Fixes #10929

### Brief Description

Several lite-topic hot paths (`doFullDispatchForWildcardGroup`, `getLiteTopicCount`, `collectByParentTopic`, `cleanByParentTopic`) iterate the entire consume-queue table on every invocation, so their cost grows with the total number of lmqs on the broker instead of the lmqs under the target parent topic.

This PR adds `LmqPrefixIndex`, a prefix-ordered in-memory index over lmq names:

- **Index lifecycle**: bootstrapped once at startup (`bootstrapLmqPrefixIndex` after lifecycle `init`), then maintained on the lmq hot path via `onLmqCreate` (first message of a new lmq) and `onLmqDelete` hooks.
- **Index-backed iteration**: new `forEachLiteTopicByPrefix` / `forEachLiteTopicByParent` APIs; wildcard full dispatch, topic count and cleanup are rewritten on top of them, turning O(total lmqs) scans into O(matched lmqs). Cleanup switches to collect-then-delete to avoid lock nesting between iteration and deletion.
- **Subscription model refinement**: replace `SubscriberWrapper` with a plain `Map<String, List<ClientGroup>>` return from `getAllSubscribers` (exact + wildcard sources), simplify `getWildcardGroupClients` signature, rename `getLiteTopicSet` to `getLmqSet`, and make `LiteCtlListener` callbacks default methods.

No protocol or external API change; dispatch and cleanup behavior stays equivalent.

### How Did You Test This Change?

- New unit tests `LmqPrefixIndexTest` covering prefix lookup boundaries, bootstrap and create/delete hooks
- Extended `AbstractLiteLifecycleManagerTest` / `LiteEventDispatcherTest` for index-backed dispatch, count and cleanup paths
- Rewritten `LiteSubscriptionRegistryImplTest` and new `LiteSubscriptionTest` for the subscription model refactoring
- Full regression on the touched modules: `mvn -pl broker -am test` related suites (128+ tests) plus `common`/`proxy` lite suites, all green; `mvn checkstyle` clean

PR #11056, opened 2026-09-07
[ISSUE #11055] Validate topK in Lite group info queries

### LINKED ISSUE #11055: [Bug] Validate topK in Lite group info queries

Windows, local broker unit tests; no running cluster is required for the regression test.

### RocketMQ version

`develop` at `ff8f6f74c560e391261ccd716707c6d20422e253` (5.5.1).

### JDK Version

Amazon Corretto 8u482; Maven 3.9.11.

### Describe the Bug

`LiteManagerProcessor.getLiteGroupInfo` forwards the request's `topK` directly to the lag calculators when `liteTopic` is absent or empty. `GetLiteGroupInfoRequestHeader.checkFields()` does not validate it.

`LiteConsumerLagCalculator.getLagCountTopK` uses `topK` as the initial capacity of a `PriorityQueue`. Zero and negative values are invalid constructor arguments. Large positive values also control the initial backing-array allocation without a server-side bound.

### Steps to Reproduce

1. Use an existing Lite consumer group bound to a parent topic.
2. Send `GET_LITE_GROUP_INFO` with that group, no specific `liteTopic`, and `topK=0` or `topK=-1`. The equivalent CLI input is `mqadmin getLiteGroupInfo -n <namesrv> -p <parent-topic> -g <lite-group> -k 0`.
3. The processor reaches `getLagCountTopK` instead of rejecting the invalid parameter.

The accompanying processor regression test serializes request headers and invokes `processRequest`. It fails on the unmodified implementation because the invalid request reaches the mocked lag calculator. The constructor behavior above is established from the source path; a live-cluster run and large allocations were not attempted.

An `INVALID_PARAMETER` response with a clear valid range, before calling either lag calculator. Requests for a specific LiteTopic should continue to work without `topK`, since that path does not use it.

### What Did You See Instead?

The aggregate query accepts an unchecked heap capacity and reaches the calculator instead of returning a parameter error.

### Additional Context

A small fix can reuse the processor's existing `MAX_RETURN_COUNT` (10,000), accepting `topK` in `[1, 10000]` for aggregate queries. Regression coverage includes null/empty LiteTopic, zero/negative/oversized values, accepted boundaries, and a specific-topic query with the default `topK`.

I searched existing issues and pull requests for `topK` and the Lite lag calculator; the related merged PR #10424 optimizes timestamp lookup and does not add this validation.

Palaiologos1453:

- Fixes #11055

### Brief Description

Aggregate Lite group info queries pass `topK` directly to a `PriorityQueue` constructor. Zero or negative values are invalid, and large positive values can request an unbounded initial array allocation. Validate `topK` before either lag calculator is called and return `INVALID_PARAMETER` unless it is between 1 and 10,000, reusing the processor's existing `MAX_RETURN_COUNT` limit.

The check applies only when no specific LiteTopic is requested. Single-topic detail queries continue to accept the default `topK`, since they do not use the TopK calculators.

With Amazon Corretto 8u482 and Maven 3.9.11:

```sh
mvn -B -pl broker -am -Dtest=LiteManagerProcessorTest,LiteConsumerLagCalculatorTest -DfailIfNoTests=false test
```

- 34 tests passed, with no failures, errors, or skips; the configured Checkstyle and SpotBugs checks passed.
- The new invalid-parameter regression fails against the unmodified processor and passes with the fix. It checks null/empty LiteTopic, zero/negative/oversized `topK`, and that rejected requests never call the lag calculator.
- Accepted boundaries (1 and 10,000) are covered, and the existing specific-topic test now omits `setTopK` to cover its default value.
- `git diff --check` passed. No live-cluster test was run.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/LiteManagerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/LiteManagerProcessor.java
@@ -316,6 +316,10 @@ public class LiteManagerProcessor implements NettyRequestProcessor {
         body.setLiteTopic(liteTopic);
 
         if (StringUtils.isEmpty(liteTopic)) {
+            if (topK <= 0 || topK > MAX_RETURN_COUNT) {
+                return RemotingCommand.createResponseCommand(ResponseCode.INVALID_PARAMETER,
+                    String.format("topK must be between 1 and %d.", MAX_RETURN_COUNT));
+            }
             Pair<List<LiteLagInfo>, Long> lagCountPair = brokerController.getBrokerMetricsManager()
                 .getLiteConsumerLagCalculator()
                 .getLagCountTopK(group, topK);
--- a/broker/src/test/java/org/apache/rocketmq/broker/processor/LiteManagerProcessorTest.java
+++ b/broker/src/test/java/org/apache/rocketmq/broker/processor/LiteManagerProcessorTest.java
@@ -81,6 +81,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
﻿LGTM. Simple and effective input validation for the `topK` parameter.

The guard clause correctly rejects zero, negative, and oversized values before they reach the `PriorityQueue` constructor, preventing both semantic errors and unbounded memory allocation. Reusing `MAX_RETURN_COUNT` (10,000) as the upper bound is consistent with existing limits in the processor.

The check is correctly scoped to only apply when no specific LiteTopic is requested — single-topic detail queries don't use the TopK calculators, so they're unaffected. The error message is clear and includes the valid range.

---

Palaiologos1453:
@imzs @lizhimins, a brief follow-up since these PRs have been waiting since September 7. The fixes are small and independent, and the local broker tests, Checkstyle, and SpotBugs pass. #11060 has also been rebased onto current `develop` after #10930 and is mergeable.

Could you please take a look when available and approve the pending Actions workflows? The related PRs are #11056, #11058, and #11060; no additional code changes are requested in this follow-up.

PR #11058, opened 2026-09-07
[ISSUE #11057] Prune deleted LiteTopic lag samples when reporting

### LINKED ISSUE #11057: [Bug] Deleted LiteTopics remain in lag latency metrics and timestamp TopK

Windows, local broker unit tests with mocked queue state; no running cluster is required for the regression.

### RocketMQ version

`develop` at `ff8f6f74c560e391261ccd716707c6d20422e253` (5.5.1).

### JDK Version

Amazon Corretto 8u482; Maven 3.9.11.

### Describe the Bug

Lite lag timestamps are cached by `LiteConsumerLagCalculator`. TTL cleanup deletes a LiteTopic's queue, offsets, and subscriptions, but its cached lag timestamp can remain. Both `calculateLiteLagLatency` and `getLagTimestampTopK` continue to use that timestamp without checking whether the queue exists.

An expired session can therefore remain the oldest entry in a group's latency metric or occupy a TopK slot. With no further POP requests for that session, the normal POP-driven cache removal need not run.

### Steps to Reproduce

1. Record a lag timestamp for a LiteTopic through `updateLagInfo`.
2. Delete its queue, as TTL cleanup does. The lifecycle manager now returns a non-positive maximum offset for that LMQ.
3. Call `getLagTimestampTopK`, or enable `liteLagLatencyMetricsEnable` and call `calculateLiteLagLatency`.

The accompanying unit test models the queue becoming absent and checks the public calculator results. On the unmodified implementation, the assertion that the expired topic is absent from TopK fails.

Expired queues should be excluded from latency metrics and timestamp TopK results, and their stale samples should be removed from the cache. Samples for other queues should be preserved. A group whose cached samples are all expired should produce no latency observation and an empty timestamp TopK result.

### What Did You See Instead?

The cache continues to supply a timestamp for the deleted queue. Aggregate latency can keep increasing, and timestamp TopK can select that entry even though its queue is gone.

### Additional Context

A deletion-only cleanup has a race: a POP request may publish a timestamp that it read before deletion after that cleanup has finished. Validating cached samples when reporting or querying also handles such late updates while the queue remains absent.

The proposed fix uses the existing lifecycle manager's in-memory maximum-offset lookup for cached candidates. When removing a stale sample, identity matching is needed because `LagTimeInfo.equals` compares only the LMQ name; removing by equality could delete a concurrent replacement for a recreated session.

This is cleanup at metric/query time, not an atomic snapshot with queue deletion. A collection overlapping deletion can observe the earlier state; a subsequent collection revalidates it. No live-cluster test was run. Existing related work such as #10723 concerns deciding TTL expiry, rather than stale lag samples after deletion.

Palaiologos1453:

- Fixes #11057

### Brief Description

After TTL cleanup deletes a LiteTopic, its cached lag timestamp can still contribute to aggregate latency and occupy a timestamp TopK slot. Validate cached candidates against the lifecycle manager's maximum queue offset when reporting latency or querying timestamp TopK. Remove samples for absent queues before selecting results, preserving active candidates and returning no latency observation when all candidates have expired.

Validation happens when consuming the cache because a POP request can publish a previously read timestamp after deletion. Removal matches the sample by identity so a concurrent replacement for a recreated session is preserved. The check uses existing in-memory queue state for cached candidates; it does not scan the full queue table or read message timestamps.

Cleanup occurs at report/query time. Collection is not atomic with queue deletion, so an overlapping collection can observe the earlier state and a subsequent collection revalidates it.

With Amazon Corretto 8u482 and Maven 3.9.11:

```sh
mvn -B -pl broker -am -Dtest=LiteConsumerLagCalculatorTest,LiteManagerProcessorTest,AbstractLiteLifecycleManagerTest,LiteLifecycleManagerTest,RocksDBLiteLifecycleManagerTest -DfailIfNoTests=false test
```

- Build succeeded: 57 tests reported, 55 passed, 2 skipped, no failures or errors. Checkstyle and SpotBugs passed.
- The skipped tests already have `@Ignore`: `LiteLifecycleManagerTest.testCleanExpiredLiteTopic` and `RocksDBLiteLifecycleManagerTest.testInit_tieredStore`.
- New coverage checks mixed active/expired candidates, all-expired results, removal before TopK selection, late updates after deletion, recreation with a fresh sample, and a deterministic interleaving that preserves a concurrent replacement.
- The expired-topic regression fails on the unmodified calculator and passes with this change.
- `git diff --check` passed. No live-cluster test was run.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/metrics/LiteConsumerLagCalculator.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/metrics/LiteConsumerLagCalculator.java
@@ -88,6 +88,15 @@ public class LiteConsumerLagCalculator {
         return brokerController.getLiteLifecycleManager().getMaxOffsetInQueue(lmqName);
     }
 
+    private boolean isLagInfoActive(PriorityBlockingQueue<LagTimeInfo> lagHeap, LagTimeInfo lagInfo) {
+        if (getMaxOffset(lagInfo.getLmqName()) > 0) {
+            return true;
+        }
+        // Match the sample by identity so a concurrent replacement for the same LMQ survives.
+        lagHeap.removeIf(current -> current == lagInfo);
+        return false;
+    }
+
     private long offsetDiff(Long offset, String lmqName) {
         long consumerOffset = offset == null ? -1L : offset;
         if (consumerOffset < 0) {
@@ -135,9 +144,13 @@ public class LiteConsumerLagCalculator {
 
             // Find the minimum storeTimestamp in the heap
             long minTimestamp = lagHeap.stream()
+                .filter(lagInfo -> isLagInfoActive(lagHeap, lagInfo))
                 .mapToLong(LagTimeInfo::getLagTimestamp)
                 .min()
                 .orElse(0L);
+            if (minTimestamp <= 0) {
+                return;
+            }
 
             ConsumerLagCalculator.CalculateLagResult lagResult =
                 new ConsumerLagCalculator.CalculateLagResult(topicGroup.group, topicGroup.topic, false);
@@ -170,6 +183,9 @@ public class LiteConsumerLagCalculator {
         // Evict the largest timestamp when heap is full, keeping smallest topK timestamps
         PriorityQueue<LagTimeInfo> maxHeap = new PriorityQueue<>(topK, Comparator.comparingLong(LagTimeInfo::getLagTimestamp).reversed());
         for (LagTimeInfo lagInfo : lagHeap) {
+            if (!isLagInfoActive(lagHeap, lagInfo)) {
+                continue;
+            }
             if (maxHeap.size() < topK) {
                 maxHeap.offer(lagInfo);
             } else if (maxHeap.peek() != null && lagInfo.getLagTimestamp() < maxHeap.peek().getLagTimestamp()) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿LGTM. Well-structured fix for stale lag metric samples after LiteTopic deletion.

The approach of validating cached candidates against the lifecycle manager's max queue offset at report/query time is sound. It correctly handles the race condition where a POP request can publish a timestamp after the topic has been deleted — rather than trying to prevent the stale write (which would require more invasive locking), it filters stale entries at read time.

Key design decisions that look correct:
- Removal matches by identity, so a concurrent replacement for a recreated session is preserved
- The check uses existing in-memory queue state, avoiding expensive scans
- Non-atomic cleanup is acknowledged in the PR description, with the correct observation that subsequent collections revalidate

The `pruneExpiredSamples` method is cleanly separated and the test coverage (mixed active/expired, all-expired, late updates after deletion, recreation with fresh sample) is thorough.

---

Palaiologos1453:
@imzs @lizhimins, a brief follow-up since these PRs have been waiting since September 7. The fixes are small and independent, and the local broker tests, Checkstyle, and SpotBugs pass. #11060 has also been rebased onto current `develop` after #10930 and is mergeable.

Could you please take a look when available and approve the pending Actions workflows? The related PRs are #11056, #11058, and #11060; no additional code changes are requested in this follow-up.

PR #11062, opened 2026-09-08
fix(client): compare pop consume RT against the invisible time in milliseconds

zjncs:
### Motivation

`ConsumeMessagePopConcurrentlyService.ConsumeRequest.run()` classifies a slow consume as `TIME_OUT` only when:

```java
} else if (consumeRT >= invisibleTime * 1000) {
    returnType = ConsumeReturnType.TIME_OUT;
```

but `invisibleTime` is parsed from `PROPERTY_POP_CK` via `ExtraInfoUtil.getInvisibleTime()` and is already in **milliseconds** — the sibling check in the same inner class, `isPopTimeout()`, compares `current - popTime >= invisibleTime` in raw ms, and `DefaultMQPushConsumerImpl` clamps the invisible window to 5000–300000 ms.

With the default 60s window the threshold becomes 60,000,000 ms (~16.7 hours), so `ConsumeReturnType.TIME_OUT` is effectively unreachable: consumes slower than the invisible window are reported as `SUCCESS` through `consumeMessageContext.getProps().put(MixAll.CONSUME_CONTEXT_TYPE, ...)`, so consume hooks cannot see which pop consumes already exceeded their invisible period. (The non-pop siblings multiply by 1000 because their base unit is minutes.)

### Modifications

Compare `consumeRT >= invisibleTime` directly.

### Verification

Fail-before (new test `testConsumeRequestReportsTimeOutWhenConsumeExceedsInvisibleTime`, run against the unpatched code — a `ConsumeRequest` whose `PROPERTY_POP_CK` carries a 500 ms invisible time, with a listener sleeping 700 ms):

```
Tests run: 6, Failures: 1, Errors: 0 -- ConsumeMessagePopConcurrentlyServiceTest
org.junit.ComparisonFailure: expected:<[TIME_OUT]> but was:<[SUCCESS]>
```

Pass-after:

```
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0 -- ConsumeMessagePopConcurrentlyServiceTest
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessagePopConcurrentlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessagePopConcurrentlyService.java
@@ -437,7 +437,7 @@ public class ConsumeMessagePopConcurrentlyService implements ConsumeMessageServi
                 } else {
                     returnType = ConsumeReturnType.RETURNNULL;
                 }
-            } else if (consumeRT >= invisibleTime * 1000) {
+            } else if (consumeRT >= invisibleTime) {
                 returnType = ConsumeReturnType.TIME_OUT;
             } else if (ConsumeConcurrentlyStatus.RECONSUME_LATER == status) {
                 returnType = ConsumeReturnType.FAILED;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a unit mismatch where `invisibleTime` (already in milliseconds from `ExtraInfoUtil.getInvisibleTime()`) was multiplied by 1000 again, making the `TIME_OUT` classification unreachable (~16.7 hours threshold with default 60s window). Removing the erroneous `* 1000` is correct. Good test with a real sleep to verify the timeout path. LGTM.

---

PR #11063, opened 2026-09-08
fix(client): clamp the pop retry delay level when a message is younger than the first level

zjncs:
### Motivation

In `ConsumeMessagePopConcurrentlyService.checkNeedAckOrDelay`, the delay level for the next retry is picked by scanning the pop delay table top-down:

```java
int delayLevel = delayLevelTable.length - 1;
for (; delayLevel >= 0; delayLevel--) {
    if (msgDelaytime >= delayLevelTable[delayLevel] * 1000) {
        delayLevel++;
        break;
    }
}

changePopInvisibleTime(msgExt, consumerGroup, delayLevel);
```

If the message is younger than `delayLevelTable[0]` seconds (`msgDelaytime < delayLevelTable[0] * 1000`) the loop never breaks and exits with `delayLevel == -1`. `changePopInvisibleTime` only guards `0 == delayLevel`, so `-1` falls through to:

```java
int delaySecond = delayLevel >= delayLevelTable.length ? delayLevelTable[delayLevelTable.length - 1] : delayLevelTable[delayLevel];
```

which evaluates `delayLevelTable[-1]` **before** the try block and throws `ArrayIndexOutOfBoundsException`. The exception propagates through `processConsumeResult` into `ConsumeRequest.run()`, which has no try/catch: the consume task dies, the remaining messages of the batch are neither acked nor scheduled for retry, and the post-consume hooks/stats are skipped.

This is reachable whenever a redelivered message is consumed while younger than the first delay level (10s by default) — e.g. with a small `maxReconsumeTimes` (setting 0 makes every first failed consume of a fresh message hit this path) or with producer/consumer clock skew shrinking `msgDelaytime`.

### Modifications

Clamp `delayLevel` to 0 after the loop. Level 0 is already handled by `changePopInvisibleTime` by falling back to `msg.getReconsumeTimes()`, which yields a sensible first-level delay.

### Verification

Fail-before (new test `testProcessConsumeResultWithMaxReconsumeReachedAndFreshMessage`, run against the unpatched code — `RECONSUME_LATER` with `maxReconsumeTimes=0`, pop delay table `{10, 30, 60}` and a message 1s old):

```
Tests run: 2, Failures: 0, Errors: 1 -- ConsumeMessagePopConcurrentlyServiceTest
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 3
```

Pass-after:

```
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0 -- ConsumeMessagePopConcurrentlyServiceTest
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessagePopConcurrentlyService.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/ConsumeMessagePopConcurrentlyService.java
@@ -277,6 +277,9 @@ public class ConsumeMessagePopConcurrentlyService implements ConsumeMessageServi
                     break;
                 }
             }
+            if (delayLevel < 0) {
+                delayLevel = 0;
+            }
 
             changePopInvisibleTime(msgExt, consumerGroup, delayLevel);
             log.warn("Consume too many times, but delay time {} not enough. changePopInvisibleTime to delayLevel {} . message key:{}",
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes an `ArrayIndexOutOfBoundsException` when a redelivered message is consumed while younger than the first delay level, causing `delayLevel == -1` to index the delay table out of bounds. Clamping to 0 after the loop is correct since level 0 is already handled by `changePopInvisibleTime`. LGTM.

---

PR #11065, opened 2026-09-08
fix(common): keep '=' inside attribute values when parsing modifications

zjncs:
### Motivation

`AttributeParser.parseToMap` splits each `key=value` pair without a limit:

```java
String[] splits = kv.split(ATTR_KEY_VALUE_EQUAL_SIGN);
key = splits[0];
value = splits[1];
```

Two consequences for user-supplied attribute strings (reachable from `mqadmin updateTopic -a ...` / `updateSubGroup -c ...` and from `UpdateTopicRequestHeader.getAttributes()` off the wire):

1. `"+key1="` — Java's `split` drops the trailing empty string, so `splits` has length 1 and `splits[1]` throws a raw `ArrayIndexOutOfBoundsException` instead of parsing an empty value.
2. `"+key1=a=b"` — the value is silently truncated to `"a"` and the rest discarded, with no error. Values containing `=` are realistic (e.g. base64 with padding).

### Modifications

Split with a limit of 2, so the first `=` separates key from value and everything after it stays in the value.

### Verification

Fail-before (new tests, run against the unpatched code):

```
Tests run: 9, Failures: 1 -- AttributeParserTest#parseToMap_ValueContainingEqualSign_ReturnsFullValue
expected:<{+key1=a=b}> but was:<{+key1=a}>

Tests run: 1, Failures: 0, Errors: 1 -- AttributeParserTest#parseToMap_EmptyValue_ReturnsEmptyValue
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

Pass-after:

```
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0 -- AttributeParserTest
```

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/attribute/AttributeParser.java
+++ b/common/src/main/java/org/apache/rocketmq/common/attribute/AttributeParser.java
@@ -44,7 +44,7 @@ public class AttributeParser {
             String key;
             String value;
             if (kv.contains(ATTR_KEY_VALUE_EQUAL_SIGN)) {
-                String[] splits = kv.split(ATTR_KEY_VALUE_EQUAL_SIGN);
+                String[] splits = kv.split(ATTR_KEY_VALUE_EQUAL_SIGN, 2);
                 key = splits[0];
                 value = splits[1];
                 if (!key.contains(ATTR_ADD_PLUS_SIGN)) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes `AttributeParser.parseToMap` silently truncating values containing `=` and throwing AIOOBE on empty values. Adding `split("=", 2)` is the correct and minimal fix. Good test coverage for both edge cases. LGTM.

---

PR #11067, opened 2026-09-08
fix(tools): stop -c <cluster> from switching resetOffsetByTime to the C++ protocol

zjncs:
### Motivation

`ResetOffsetByTimeCommand` registers **two** options with the short name `c`:

```java
opt = new Option("c", "cplus", false, "reset c++ client offset. Deprecated.");   // line 64
...
opt = new Option("c", "cluster", true, "Cluster name or lmq parent topic, ..."); // line 80
```

commons-cli's `Options.addOption` keys the short-opt map by name, so the later `cluster` registration replaces `cplus` for `-c`. Consequences:

1. The documented `-c`/`--cplus` deprecated switch can never be enabled through its short form.
2. Any user passing `-c <cluster>` (the documented usage, e.g. to find the route for an LMQ parent topic) makes `isC = commandLine.hasOption('c')` true, so `resetOffsetByTimestamp(..., isC)` → `MQClientAPIImpl.invokeBrokerToResetOffset(..., isC)` sends the request with `LanguageCode.CPP`. The broker then replies with `ResetOffsetBodyForC` (`List<MessageQueueForC>`), which the Java client decodes as `ResetOffsetBody` (`Map<MessageQueue, Long>`) — garbage offsets/exception instead of a proper reset for every Java consumer group whenever `-c` is used.

### Modifications

- Drop the short name from the deprecated `cplus` option (`--cplus` keeps working, `-c` no longer collides).
- Compute `isC` from the long name: `commandLine.hasOption("cplus")`.
- Extract `createDefaultMQAdminExt(RPCHook)` as a protected factory so the command can be tested with a stubbed admin client.

### Verification

Fail-before (new test `testExecuteWithClusterOptionKeepsJavaClientProtocol`, run against the unpatched code — executes the command with `-g -t -s -f false -c default-cluster` and verifies the admin call):

```
Tests run: 2, Failures: 1 -- ResetOffsetByTimeCommandTest
Argument(s) are different! Wanted:
defaultMQAdminExt.resetOffsetByTimestamp("default-cluster", "unit-test", "default-group", 1412131213231L, false, false);
Actual invocations have different arguments:
defaultMQAdminExt.resetOffsetByTimestamp("default-cluster", "unit-test", "default-group", 1412131213231L, false, true);
```

Pass-after:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- ResetOffsetByTimeCommandTest
```

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/offset/ResetOffsetByTimeCommand.java
@@ -61,7 +61,7 @@ public class ResetOffsetByTimeCommand implements SubCommand {
         opt.setRequired(false);
         options.addOption(opt);
 
-        opt = new Option("c", "cplus", false, "reset c++ client offset. Deprecated.");
+        opt = new Option(null, "cplus", false, "reset c++ client offset. Deprecated.");
         opt.setRequired(false);
         options.addOption(opt);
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a commons-cli short-option collision where `-c <cluster>` silently sets the C++ protocol flag, causing garbage offset responses. Removing the short name from the deprecated `cplus` option and switching to `hasOption("cplus")` is the correct fix. Good test with mocked admin client. LGTM.

---

PR #11068, opened 2026-09-08
fix(remoting): match every address for a /0 CIDR in ipInCIDR

zjncs:
### Motivation

`RemotingHelper.ipInCIDR` computes the netmask as:

```java
int mask = 0xFFFFFFFF << (32 - netId);
```

For `netId == 0` the shift distance is 32, which Java takes mod 32 — the shift is a no-op and `mask` stays `0xFFFFFFFF` instead of `0`. So a `/0` CIDR matches nothing except its own address:

```java
ipInCIDR("1.2.3.4", "0.0.0.0/0")  // returns false, should be true
```

Reachability: `NettyRemotingClient.getProxy` guards only the exact string `"0.0.0.0/0"` (`DEFAULT_CIDR_ALL.equals(cidr)`) before calling `ipInCIDR`, so a socks proxy configured with any other `/0` CIDR (e.g. `10.0.0.0/0`) silently never matches and the traffic bypasses the configured proxy. `/1`–`/32` are unaffected.

### Modifications

Use a zero mask when the prefix length is 0.

### Verification

Fail-before (new test class `RemotingHelperTest`, run against the unpatched code):

```
Tests run: 2, Failures: 1 -- RemotingHelperTest#testIpInCidrMatchesEverythingForZeroPrefixLength
```

Pass-after:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- RemotingHelperTest
```

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/common/RemotingHelper.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/common/RemotingHelper.java
@@ -302,7 +302,7 @@ public class RemotingHelper {
         int ipAddr = ipToInt(ip);
         String[] cidrArr = cidr.split("/");
         int netId = Integer.parseInt(cidrArr[1]);
-        int mask = 0xFFFFFFFF << (32 - netId);
+        int mask = netId == 0 ? 0 : 0xFFFFFFFF << (32 - netId);
         int cidrIpAddr = ipToInt(cidrArr[0]);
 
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a Java shift-modulo-32 bug where `0xFFFFFFFF << 32` is a no-op, causing `/0` CIDR to match nothing. Clean one-line ternary fix with a new test class covering the edge case and regular prefixes. LGTM.

---

PR #11069, opened 2026-09-08
fix(tools): skip master-address comparison in consumed() when the routed broker has no master

zjncs:
### Motivation

`DefaultMQAdminExtImpl.consumed()` resolves the master address of each broker referenced by the consume-stats offset table:

```java
String addr = NetworkUtil.convert2IpString(brokerData.getBrokerAddrs().get(MixAll.MASTER_ID));
```

`examineConsumeStats` accepts entries from brokers whose `selectBrokerAddr()` resolved to a slave (master down, stale route). For such a broker `getBrokerAddrs().get(MixAll.MASTER_ID)` returns null, and `NetworkUtil.convert2IpString(null)` → `string2SocketAddress(null)` crashes on `addr.lastIndexOf(":")`.

The NPE aborts the whole track listing: `queryMsgById -i ...` → `printMsg` → `messageTrackDetail` → `consumed(msg, group)` fails for **all** groups, not just the group served by the unreachable broker. The sibling `consumedConcurrent` already uses the raw address with a null-safe comparison direction, showing the null case was considered elsewhere.

### Modifications

Skip the store-host comparison when the master address is null, mirroring the null-safe structure of the sibling method.

### Verification

Fail-before (new test `testConsumedWithSlaveOnlyBrokerInRoute`, run against the unpatched code — a spy of `DefaultMQAdminExtImpl` returning one offset-table entry for `broker1` whose route registers only brokerId 1):

```
Tests run: 1, Failures: 0, Errors: 1 -- DefaultMQAdminExtImplTest
java.lang.NullPointerException: Cannot invoke "String.lastIndexOf(String)" because "addr" is null
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- DefaultMQAdminExtImplTest#testConsumedWithSlaveOnlyBrokerInRoute
```

Note: the pre-existing `testConsumeMessageDirectly` in this class fails on JDK 21 in my local container with `Mockito cannot mock this class: java.net.InetAddress` — verified to fail identically on a clean develop checkout (before my changes), so it is an environment limitation, not a regression of this PR.

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
@@ -1544,8 +1544,9 @@ public class DefaultMQAdminExtImpl implements MQAdminExt, MQAdminExtInner {
             if (mq.getTopic().equals(msg.getTopic()) && mq.getQueueId() == msg.getQueueId()) {
                 BrokerData brokerData = ci.getBrokerAddrTable().get(mq.getBrokerName());
                 if (brokerData != null) {
-                    String addr = NetworkUtil.convert2IpString(brokerData.getBrokerAddrs().get(MixAll.MASTER_ID));
-                    if (NetworkUtil.socketAddress2String(msg.getStoreHost()).equals(addr)) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Changes look good.

---

PR #11070, opened 2026-09-08
fix(remoting): parse the full host from bracketed IPv6 addresses in parseHostFromAddress

zjncs:
### Motivation

`RemotingHelper.parseHostFromAddress` splits on **every** colon and returns the first segment:

```java
String[] addressSplits = address.split(":");
return addressSplits[0];
```

For the bracketed IPv6 form RocketMQ uses for socket addresses, this mangles the host:

```java
parseHostFromAddress("[2001:db8::1]:10911")  // returns "[2001"
```

The only caller is `BrokerPreOnlineService.waitForHaHandshakeComplete`, which passes the parsed value as the remote address of the `HAConnectionStateNotificationRequest`; `HAConnectionStateNotificationService.checkConnectionStateAndNotify` then compares it against the connection's host, so an IPv6 cluster's HA handshake wait can never match and pre-online times out.

IPv4 addresses and bare hostnames are unaffected by the fix (first segment == everything before the last colon). The dead `addressSplits.length < 1` branch (`split` never returns an empty array) goes away with the rewrite.

### Modifications

Cut at the last colon, mirroring `NetworkUtil.string2SocketAddress` and `NettyRemotingClient.getHostAndPort`, which already handle bracketed IPv6 this way.

### Verification

Fail-before (new test class `RemotingHelperTest`, run against the unpatched code):

```
Tests run: 1, Failures: 1 -- RemotingHelperTest#testParseHostFromAddress
expected: "[2001:db8::1]" but was: "[2001"
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- RemotingHelperTest
```

Note: this adds the same new test file as #11068 (`ipInCIDR`); the two PRs will need a trivial rebase against whichever merges first.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/common/RemotingHelper.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/common/RemotingHelper.java
@@ -265,12 +265,8 @@ public class RemotingHelper {
             return "";
         }
 
-        String[] addressSplits = address.split(":");
-        if (addressSplits.length < 1) {
-            return "";
-        }
-
-        return addressSplits[0];
+        int idx = address.lastIndexOf(":");
+        return idx > 0 ? address.substring(0, idx) : address;
     }
 
     public static String parseSocketAddressAddr(SocketAddress socketAddress) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Changes look good.

---

PR #11071, opened 2026-09-08
fix(tools): keep the broadcast track type in messageTrackDetailConcurrent

zjncs:
### Motivation

The serial `messageTrackDetail` maps the broadcast case to a proper track type:

```java
} catch (MQClientException e) {
    if (ResponseCode.CONSUMER_NOT_ONLINE == e.getResponseCode()) {
        mt.setTrackType(TrackType.NOT_ONLINE);
        ...
    }
    if (ResponseCode.BROADCAST_CONSUMPTION == e.getResponseCode()) {
        mt.setTrackType(TrackType.CONSUME_BROADCASTING);
    }
    ...
```

but the concurrent variant (`messageTrackDetailConcurrent`, exposed via `MQAdminExt.messageTrackDetailConcurrent`) only handles `CONSUMER_NOT_ONLINE` in its catch blocks. For a broadcast consumer group, `consumed()` → `examineConsumeStats` throws `MQClientException(ResponseCode.BROADCAST_CONSUMPTION, ...)`, so the concurrent API reports the group as `TrackType.UNKNOWN` with a raw error string, while the serial API correctly reports `CONSUME_BROADCASTING` — contradicting the sibling implementation and misleading dashboards that use the concurrent API.

### Modifications

Add the same `BROADCAST_CONSUMPTION → CONSUME_BROADCASTING` mapping to both catches of the concurrent track task.

### Verification

Fail-before (new test `testMessageTrackDetailConcurrentWithBroadcastGroup`, run against the unpatched code — a passive consumer group whose connection reports `MessageModel.BROADCASTING`, so `examineConsumeStats` throws `BROADCAST_CONSUMPTION` through the natural path):

```
Tests run: 1, Failures: 1 -- DefaultMQAdminExtImplTest
expected:<CONSUME_BROADCASTING> but was:<UNKNOWN>
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- DefaultMQAdminExtImplTest#testMessageTrackDetailConcurrentWithBroadcastGroup
```

Note: the pre-existing `testMessageTrackDetailConcurrent` (and `testConsumeMessageDirectly`) fail in my local container on JDK 21 with `Mockito cannot mock this class: java.net.InetAddress` — verified to fail identically on a clean develop checkout (before my changes), so this is an environment limitation, not a regression of this PR.

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/admin/DefaultMQAdminExtImpl.java
@@ -1480,6 +1480,9 @@ public class DefaultMQAdminExtImpl implements MQAdminExt, MQAdminExtInner {
                                 if (ResponseCode.CONSUMER_NOT_ONLINE == e.getResponseCode()) {
                                     mt.setTrackType(TrackType.NOT_ONLINE);
                                 }
+                                if (ResponseCode.BROADCAST_CONSUMPTION == e.getResponseCode()) {
+                                    mt.setTrackType(TrackType.CONSUME_BROADCASTING);
+                                }
                                 mt.setExceptionDesc("CODE:" + e.getResponseCode() + " DESC:" + e.getErrorMessage());
                                 result.add(mt);
                                 countDownLatch.countDown();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Changes look good.

---

PR #11073, opened 2026-09-08
[ISSUE #11072] Reject route queries when NameServer shuts down

### LINKED ISSUE #11072: [Bug] NameServer accepts route lookups after shutdown starts
### Describe the Bug

`NamesrvController.shutdown()` starts closing resources without marking the NameServer as not ready. A queued or newly dispatched `GET_ROUTEINFO_BY_TOPIC` request can still enter route lookup after shutdown has begun; the existing readiness check only covers startup.

### Steps to Reproduce

1. Begin NameServer shutdown through SIGTERM or `mqshutdown namesrv`.
2. Let a queued or new route request reach the processor before its connection closes.
3. The processor performs the normal route lookup instead of rejecting the request as not ready.

This is a source-level shutdown interleaving; validation uses deterministic unit tests rather than a production reproduction.

### Expected Behavior

At request processing time, return `SYSTEM_ERROR` with `name server not ready` when `isShutdown()` is true. Otherwise continue the original route lookup. Startup and shutdown should use the same not-ready response helper. Requests already past this check keep their original results.

### Version and Environment

Apache RocketMQ `develop`, base `66362b9c9f` (5.5.1 development version). Local regression tests use macOS and Amazon Corretto 21.0.5.

qianye1001:

Fixes #11072

### Brief Description

When `isShutdown()` is true, NameServer route queries return `SYSTEM_ERROR` with `name server not ready`. Startup and shutdown use the same response helper. Otherwise, route queries keep their original behavior, including lookups already in progress.

The shutdown flag is set before resource cleanup. This PR changes only NameServer and its tests.

19 NameServer tests passed on JDK 8 with JaCoCo enabled; no failures, errors, or skips. Checkstyle passed. Tests cover shutdown rejection, startup readiness, and unchanged results for in-progress lookups.


RockteMQ-AI (approved):
LGTM. Changes look good.

---

fuyou001 (dismissed):

fuyou001:
Compatibility consideration: broker member discovery during shutdown

The new shutdown guard also covers broker member discovery, which uses `GET_ROUTEINFO_BY_TOPIC` with `SYNC_BROKER_MEMBER_GROUP_PREFIX + brokerName` when `compatibleWithOldNameSrv=true`.

There is an existing broker-side error-handling issue on this path:

- [`BrokerOuterAPI.getBrokerMemberGroupCompatible()`](https://github.com/apache/rocketmq/blob/086a6b0f9311994dbbb7aea1f0c633b0389a30de/broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java#L305) turns a `SYSTEM_ERROR` response into an empty `BrokerMemberGroup` instead of throwing.
- [`BrokerPreOnlineService.prepareForBrokerOnline()`](https://github.com/apache/rocketmq/blob/086a6b0f9311994dbbb7aea1f0c633b0389a30de/broker/src/main/java/org/apache/rocketmq/broker/BrokerPreOnlineService.java#L248) interprets that empty result as no other brokers being online and calls `startService()`. With `enableSlaveActingMaster=true`, `totalReplicas>1`, and `skipPreOnline=false`, this can bypass the intended HA handshake and, for a returning master, reverse metadata synchronization.

**This behavior already exists during startup when `needWaitForService=true` and the NameServer is not ready. It is not a new error-handling defect introduced by this PR.** This change adds a shutdown trigger, including configurations with `needWaitForService=false`, while a request can still receive a response from the shutting-down NameServer.

I checked the same broker logic in 5.3.1. A unit-level reproduction using the real member-query and pre-online methods with a mocked `SYSTEM_ERROR` response called `startService()` without calling the HA handshake. Controls with a known peer and a failed handshake, or with `RemotingConnectException`, kept the broker offline. The 26 related NameServer tests passed on the current head. This does not establish actual data corruption in a running cluster.

Please consider covering this member-query path in the shutdown regression tests and addressing or explicitly tracking the existing broker-side issue. The safety assertion should be that a failed member lookup keeps the broker isolated and retrying, rather than treating it as a successful empty lookup. Any compatibility solution should account for already deployed older brokers during a NameServer-first rolling upgrade; updating only the new broker implementation would not protect those nodes.

Reviewed against `086a6b0f9311994dbbb7aea1f0c633b0389a30de`.

RockteMQ-AI (approved):
### Summary
This PR fixes two related reliability issues: (1) NameServer shutdown returns misleading `SYSTEM_ERROR` for route queries, and (2) broker member discovery silently treats error responses as empty groups, allowing pre-online brokers to start without proper peer information.

The new commits since the previous review add `volatile` to the shutdown flag, improve the `createNotReadyResponse()` helper, and enhance the `DefaultMQProducerTest` backpressure test with proper synchronization.

### Findings
- **[Info]** `NamesrvController.java:72` — The `volatile boolean shutdown` flag is correctly set before resource teardown, ensuring in-flight requests see the shutdown state. The ordering guarantee is important for the `ClientRequestProcessor` check.
- **[Info]** `ClientRequestProcessor.java:54` — The early `isShutdown()` check in `processRequest()` correctly rejects all route queries during shutdown, including broker member lookups via the `SYNC_BROKER_MEMBER_GROUP_PREFIX` topic pattern.
- **[Info]** `BrokerOuterAPI.java:289-298` — The null-safety checks on `memberGroup` and `brokerAddrs` prevent NPE when the response body is malformed. The explicit `MQBrokerException` for unexpected response codes is the correct fix.
- **[Info]** `BrokerOuterAPI.java:332-333` — The `TOPIC_NOT_EXIST` case in `getBrokerMemberGroupCompatible()` correctly returns an empty group (first broker online scenario), while the native path throws on `TOPIC_NOT_EXIST` (which is unexpected for the dedicated API).
- **[Warning]** `BrokerOuterAPI.java:260-261` — Adding `MQBrokerException` to the throws clause of `syncBrokerMemberGroup()` is a breaking API change. Callers that previously caught only `RemotingException` subtypes will need updates. This is necessary for correctness, but worth noting in release notes.

### Suggestions
The test coverage is excellent — `BrokerMemberLookupTest` covers all error paths, connection failures, and retry scenarios. The `DefaultMQProducerTest` improvement makes the backpressure test deterministic by synchronizing sends before completing callbacks.

The Bazel CI change (`--experimental_ui_max_stdouterr_bytes=104857600`) is a reasonable infrastructure adjustment for verbose test output.

---

RockteMQ-AI (commented):
### Summary
This PR modifies 633 lines across multiple files (.github/workflows/bazel.yml,broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java,broker/src/test/java/org/apache/rocketmq/broker/BrokerMemberLookupTest.java,client/src/test/java/org/apache/rocketmq/client/producer/DefaultMQProducerTest.java,namesrv/src/main/java/org/apache/rocketmq/namesrv/NamesrvController.java,namesrv/src/main/java/org/apache/rocketmq/namesrv/processor/ClientRequestProcessor.java,namesrv/src/test/java/org/apache/rocketmq/namesrv/NamesrvControllerTest.java,namesrv/src/test/java/org/apache/rocketmq/namesrv/processor/ClientRequestProcessorTest.java).

### Observations
- Changes span 8 files
- Diff size: 633 lines

Please ensure:
- All changes are covered by tests
- No breaking API changes without migration notes
- Thread safety is maintained for concurrent access patterns

---

broker/src/main/java/org/apache/rocketmq/broker/out/BrokerOuterAPI.java
```diff
@@ -257,13 +257,13 @@ public void updateNameServerAddressListByDnsLookup(final String domain) {
...
     public BrokerMemberGroup syncBrokerMemberGroup(String clusterName, String brokerName)
-        throws InterruptedException, RemotingTimeoutException, RemotingSendRequestException, RemotingConnectException {
+        throws InterruptedException, RemotingTimeoutException, RemotingSendRequestException, RemotingConnectException, MQBrokerException {
         return syncBrokerMemberGroup(clusterName, brokerName, false);
     }
 
     public BrokerMemberGroup syncBrokerMemberGroup(String clusterName, String brokerName,
         boolean isCompatibleWithOldNameSrv)
-        throws InterruptedException, RemotingTimeoutException, RemotingSendRequestException, RemotingConnectException {
+        throws InterruptedException, RemotingTimeoutException, RemotingSendRequestException, RemotingConnectException, MQBrokerException {
```
lizhimins: 这里改接口不好

RockteMQ-AI (commented):
### Summary
This PR contains significant changes (205 lines). Please ensure:
- Adequate test coverage for new/modified logic
- No breaking changes to public APIs without migration path
- Thread safety for concurrent access patterns

---

lizhimins (approved):

Issue #11078, opened 2026-09-08
[Enhancement] Support suppressing minimal-value metrics to reduce export payload


imzs:
### Before Creating the Enhancement Request

- [x] I have confirmed that this should be classified as an enhancement rather than a bug/feature.

### Summary

This PR introduces an opt-in broker config `suppressMinValueMetrics` (default `false`, backward compatible) that suppresses exporting data points whose value is not greater than a
minimal threshold.

### Motivation

The consumer lag gauges are asynchronous ObservableGauges that emit one data point per consumer_group × topic × is_retry combination. In deployments with many consumer groups and topics, the vast majority of these combinations are idle at any given sampling instant, so their values are 0 — yet they are still recorded and exported on every collection cycle. This inflates both metric cardinality and the OTLP export payload. 

In production, we observed zero-valued points accounting for ~95% of gauge data points and the bulk of the payload. Suppressing them is safe as long as the monitoring pipeline tolerates discontinuous series.

### Describe the Solution You'd Like

- `BrokerConfig`: add `suppressMinValueMetrics` (default `false`).
- `BrokerMetricsManager`: add a general guard `shouldRecordValue(currentValue, minValue)`.
  When suppression is enabled, a data point is recorded only if `currentValue > minValue`.
- Apply the guard at the record site of the 5 consumer lag gauges, all passing `minValue = 0` for now, i.e. zero-valued points are dropped.

### Describe Alternatives You've Considered

.

### Additional Context

When enabled, threshold-based dashboards/alerts (e.g. `lag > N`) are unaffected; the only difference is
that idle series appear as gaps instead of a flat zero line. Therefore, before turning this on, make sure
there are **no alert rules based on `absent()` / `absent_over_time()`** (or any equivalent "no data" detection) targeting these metrics.

RockteMQ-AI:
**Issue Evaluation**

Category: `bug` | Status: **Under Review**

Thank you for reporting this issue. It has been classified as a potential bug and flagged for code-level verification.

**Next steps:**
- Code-level analysis will be performed to verify reproducibility
- If confirmed, a fix proposal may be generated automatically

If you have additional context (version, reproduction steps, logs), please share it in the comments.

---
*Automated evaluation by RockteMQ-AI*

imzs:
> **Issue Evaluation**
> 
> Category: `bug` | Status: **Under Review**
> 
> Thank you for reporting this issue. It has been classified as a potential bug and flagged for code-level verification.
> 
[...]

PR #11079, opened 2026-09-08
[ISSUE #11078] Support suppressing minimal-value metrics to reduce export payload

imzs:
#11078 Support suppressing minimal-value metrics to reduce export payload

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/metrics/BrokerMetricsManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/metrics/BrokerMetricsManager.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.broker.metrics;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary
Adds a `suppressMinValueMetrics` config option (default `false`) to skip recording lag/DLQ metric values that are zero, reducing OpenTelemetry export payload. When enabled, zero-value data points are omitted, which triggers Prometheus `absent_over_time` alerts — the Javadoc documents this trade-off clearly.

### Review
- **Correctness**: Clean implementation. `shouldRecordValue` correctly gates recording behind the config toggle. Default `false` ensures backward compatibility.
- **Tests**: Good coverage — tests verify both enabled/disabled paths and the threshold behavior.
- **Design**: The `@VisibleForTesting` annotation on `shouldRecordValue` is appropriate. The threshold is hardcoded to `0` at all call sites, which is consistent with the stated goal.
- **Compatibility**: Fully backward compatible — default behavior unchanged.

LGTM 👍

---

lizhimins (approved):

PR #11081, opened 2026-09-08
[ISSUE #11080] Skip unused remoting client event executor

### LINKED ISSUE #11080: [Enhancement] Skip NettyEventExecutor startup when the remoting client has no channel listener
### Summary

Start `NettyRemotingClient`'s `NettyEventExecutor` only when a `ChannelEventListener` is configured.

### Motivation

On `develop` at `fd2dc2ffc2`, `NettyRemotingClient.start()` unconditionally starts `NettyEventExecutor`, even when `channelEventListener` is `null`. The connection handlers already guard event submission with a listener null check. Consequently, a client without a listener maintains an unused thread polling an empty queue every three seconds.

This occurs in `MQClientAPIExt`, which passes a null listener to `MQClientAPIImpl`. A Proxy can own several such internal clients, so the unused thread cost accumulates per client.

A minimal reproduction is to construct `new NettyRemotingClient(new NettyClientConfig())`, call `start()`, and inspect the threads before calling `shutdown()`: a `NettyEventExecutor` thread is started without any listener to notify.

Guard `nettyEventExecutor.start()` with `channelEventListener != null`. The existing `ServiceThread.shutdown()` already safely handles a service that was never started.

Clients with a listener should continue to dispatch connection events as before. In particular, the ordinary Remoting SDK's default listener must retain its immediate heartbeat and rebalance behavior on a broker channel becoming active.

Sharing event executors between clients would require broader changes to callback isolation and resource ownership. Avoiding startup when there is no listener addresses this case with a local change.

### Additional Context

This concerns unnecessary startup without a listener. The poll-related shutdown delay for active event executors is tracked separately in #10898.

qianye1001:

Fixes #11080

### Brief Description

`NettyRemotingClient.start()` starts a `NettyEventExecutor` even when no `ChannelEventListener` is configured. Since connection handlers already skip event submission in that case, each such client keeps an unused thread polling an empty queue every three seconds. This includes internal clients built through `MQClientAPIExt`.

Start the executor only when `channelEventListener != null`. Clients with a listener retain their existing event dispatch behavior. The existing shutdown implementation safely handles an executor that was never started.

Added regression tests covering startup and shutdown without an event thread when no listener is configured, and ACTIVE event delivery plus thread termination when a listener is configured.

Passed on JDK 11:

```sh
mvn -B -pl remoting -am -Dtest=NettyRemotingClientTest -Dsurefire.failIfNoSpecifiedTests=false test
```

All 16 tests passed, including the two new tests. Checkstyle and SpotBugs also passed.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
@@ -242,7 +242,9 @@ public class NettyRemotingClient extends NettyRemotingAbstract implements Remoti
             handler.option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT);
         }
 
-        nettyEventExecutor.start();
+        if (channelEventListener != null) {
+            nettyEventExecutor.start();
+        }
 
         TimerTask timerTaskScanResponseTable = new TimerTask() {
             @Override
--- a/remoting/src/test/java/org/apache/rocketmq/remoting/netty/NettyRemotingClientTest.java
+++ b/remoting/src/test/java/org/apache/rocketmq/remoting/netty/NettyRemotingClientTest.java
@@ -27,6 +27,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
+import org.apache.rocketmq.common.ServiceThread;
+import org.apache.rocketmq.remoting.ChannelEventListener;
 import org.apache.rocketmq.remoting.InvokeCallback;
 import org.apache.rocketmq.remoting.RPCHook;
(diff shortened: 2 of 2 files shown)


RockteMQ-AI (approved):
### Summary

Clean optimization — skip starting `NettyEventExecutor` when no `ChannelEventListener` is configured. The executor thread was polling an empty queue every 3 seconds for nothing, so this saves one thread per client instance that doesn't need channel event callbacks.

### Review

- **Correctness** ✅ — The guard is safe: `channelEventListener` is set in the constructor and effectively final. Connection handlers already skip event submission when the listener is null, so the executor was indeed doing nothing.
- **Tests** ✅ — Good coverage of both paths (with/without listener). The reflection-based thread state verification is a pragmatic choice for lifecycle testing.
- **Performance** ✅ — Eliminates unnecessary thread creation and periodic wake-ups. Minor but meaningful for deployments with many client instances.
- **Compatibility** ✅ — No API changes, purely internal optimization.

LGTM.

---

lizhimins (approved):

PR #11086, opened 2026-09-09
[ISSUE #11085] Stabilize offset reset tests and guard POP revive cache logging

### LINKED ISSUE #11085: [Bug] Offset reset tests race with consumption and leave POP work running after teardown
### Description

The offset-reset integration tests have scheduling races and leave POP work running after test completion:

- `OffsetResetIT` reads its pre-reset message count after sending reset requests. Replayed messages can already arrive before that read, so the test waits for more messages than the reset can produce. Its lag helper also compares a consumer offset from an RPC snapshot with a later live broker offset, which may have advanced legitimately.
- `OffsetResetForPopIT` starts untracked executors whose loops run for 30 seconds even after the assertions and teardown. POP clients created directly by the test are never registered for cleanup. The loops swallow errors or print them continuously after the broker stops, obscuring actual failures.
- These tests also enable `enablePopBufferMerge` after broker construction. If the POP KV service was constructed without a cache, the revive logging path dereferences the absent cache. Other POP operations already guard this case.

### Reproduction and evidence

The Bazel run for #11081 failed `OffsetResetIT` and `OffsetResetForPopIT`. Its test stdout exceeded the 10 MiB output limit (approximately 17.9 MB and 1.52 GB), so the exact original assertions are unavailable. `OffsetResetIT` also failed in an earlier, unrelated CI run.

On develop `37808b38c1`, both suites passed their 17 assertions locally, but leaked POP work produced 87,274,677 bytes of build/test output, including 32,819 stack frames from the background POP test lambda. This demonstrates that passing assertions do not imply clean teardown.

A controlled scheduling experiment lets the reset replay complete before the test resumes post-reset bookkeeping: 2,400 original messages plus 300 replays give 2,700 deliveries, but the original test captures 2,700 as its baseline and waits for 3,000. The original times out; capturing the baseline before resetting passes under the same schedule.

A focused test constructs `PopConsumerService` with buffering disabled, enables buffering, and calls `revive()`: it reproduces the null-cache exception at the logging statement.

CI evidence: https://github.com/apache/rocketmq/actions/runs/34220792332 and https://github.com/apache/rocketmq/actions/runs/34201926158.

qianye1001:

Fixes #11085

### Brief Description

Offset-reset tests can fail despite correct message replay: `OffsetResetIT` captures its baseline after resetting brokers and compares offset values sampled at different times. `OffsetResetForPopIT` leaves 30-second POP loops and directly created clients alive after its assertions, swallowing failures and flooding shutdown logs.

Capture delivery counts before reset and calculate lag from the returned offset snapshot. Run the existing sequential POP/ACK/reset steps within the test thread, propagate operation failures, and close POP clients. Message-count and ordered-message-set checks remain in place.

The POP tests also expose a missing null guard in `PopConsumerService.revive()`: enabling buffering after service construction does not create a cache, but the statistics log dereferences it. Guard this access consistently with the other cache paths and add a regression test for the configuration change.

- JDK 11: 35 tests passed (18 broker unit tests and 17 offset-reset integration tests); Checkstyle and SpotBugs passed. Command: `mvn -B -pl test -am -Dtest=OffsetResetIT,OffsetResetForPopIT,PopConsumerServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`.
- After the review changes, reran all 35 tests above and an additional 13 POP reset tests in reversed method order: all passed, with no surviving POP test workers. Validation output contains no background POP connection-failure stack traces or POP revive null-cache errors.
- Controlled scheduling experiment with a temporary harness: allow all 300 reset deliveries to complete before post-reset bookkeeping. The original test waits for 3,000 deliveries when only 2,700 are possible and times out; the patched test passes under the same schedule. The harness is not part of this patch.
- The new cache-toggle regression test reproduces the null dereference against the original implementation.

The original Bazel failure did not retain its oversized test stdout, so these reproduced defects are not presented as a reconstruction of its exact failed assertions.


RockteMQ-AI (approved):
### Summary
Stabilizes flaky offset-reset integration tests and adds a defensive null guard in `PopConsumerService.revive()` statistics logging.

**Changes reviewed:**
- **PopConsumerService.java**: Null guard for `reviveCache` prevents NPE when buffering is enabled but cache not yet created — correct defensive fix
- **OffsetResetForPopIT.java**: Moves POP/ACK/reset into test thread with `pollInSameThread()`, properly closes clients, uses `HashSet` instead of concurrent set — eliminates race conditions and resource leaks
- **OffsetResetIT.java**: Captures baseline before reset and uses offset snapshot — fixes timing-dependent failures
- **PopConsumerServiceTest.java**: Regression test for null cache guard — good coverage

The test stabilization approach is sound: sequential operations in the test thread are more deterministic than background executors with shared counters. The null guard is a legitimate bug fix discovered during test stabilization.

**Verdict:** LGTM — well-crafted test improvements with a defensive bug fix.

---

test/src/test/java/org/apache/rocketmq/test/offset/OffsetResetForPopIT.java
```diff
@@ -77,7 +82,16 @@ public void setUp() throws Exception {
...
     @After
     public void tearDown() {
-        shutdown();
+        try {
+            if (consumerStarted) {
+                consumer.shutdown();
+            }
+        } finally {
+            shutdown();
+            brokerController1.getBrokerConfig().setUseServerSideResetOffset(useServerSideResetOffset);
```
lizhimins: 这个改法有点太奇怪了，broker controller 在这里其实都可以是 mock 的对象

broker/src/test/java/org/apache/rocketmq/broker/pop/PopConsumerServiceTest.java
```diff
@@ -743,4 +759,4 @@ public void testReviveRetryWithSuspendFalseMultipleTimes() {
             messageExt.setReconsumeTimes(capturedMessage.getReconsumeTimes());
         }
     }
-}
\ No newline at end of file
+}
```
lizhimins: 无关行

test/src/test/java/org/apache/rocketmq/test/offset/OffsetResetForPopIT.java
```diff
@@ -58,9 +56,16 @@ public class OffsetResetForPopIT extends BaseConf {
     private RMQNormalProducer producer = null;
     private RMQPopConsumer consumer = null;
     private DefaultMQAdminExt adminExt;
+    private boolean consumerStarted;
+    private boolean useServerSideResetOffset;
+    private boolean popConsumerKVServiceEnable;
+    private boolean enablePopBufferMerge;
```
lizhimins: 不把配置名提取出来比较好

lollipopjin (approved):
LGTM

PR #11094, opened 2026-09-09
[BUG] Split-metadata pagination crashes on zero/negative maxNum, negative seq or malformed dataVersion

zjncs:
### Motivation

`SubscriptionGroupManager.subGroupTable` and `TopicConfigManager.subTopicConfigTable` (the split-metadata path behind `GET_ALL_SUBSCRIPTIONGROUP_CONFIG` / `GET_ALL_TOPIC_CONFIG`, gated by `enableSplitMetadata`) slice the sorted key list using three client-supplied values. Three inputs crash the broker request thread:

1. **`maxGroupNum`/`maxTopicNum` = 0 or negative** — `endIndex` collapses to `beginIndex` (or below), so `keyList.get(endIndex - 1)` throws `ArrayIndexOutOfBoundsException` (e.g. `Index -1 out of bounds`), or the `ImmutableSortedMap.subMap` call throws for an inverted range when `beginIndex > 0`.
2. **negative `groupSeq`/`topicSeq` with a matching data version** — `keyList.get(beginIndex)` indexes negatively (e.g. `Index -3 out of bounds`).
3. **malformed `dataVersion` JSON** — `DataVersion.fromJson` throws fastjson's `JSONException` (`illegal fieldName ...`) instead of degrading.

All of these come straight from the request header, so a misbehaving (or malicious) client can fail every metadata fetch on a broker running with split metadata enabled.

### Changes

- Clamp the sequence to `>= 0`.
- Skip the slice (return an empty page) when the max limit is `<= 0`.
- Parse the data version through a helper that treats an unparseable value as an *unknown* version — which the existing fallback already handles by restarting from index 0 — instead of letting the exception escape.

Behaviour for well-formed requests is unchanged.

### Verification

Six new tests (`SubscriptionGroupManagerTest`, `TopicConfigManagerTest`), each red on develop and green with the patch:

- `...WithNonPositiveMaxGroupNum` / `...WithNonPositiveMaxTopicNum`: was `ArrayIndexOutOfBoundsException: Index -1`
- `...WithNegativeGroupSeqClampsToHead` / `...WithNegativeTopicSeqClampsToHead`: was `ArrayIndexOutOfBoundsException: Index -3`
- `...WithMalformedDataVersionFallsBackToFullPage`: was `com.alibaba.fastjson2.JSONException`

```
$ mvn -pl broker test -Dtest='SubscriptionGroupManagerTest,TopicConfigManagerTest'
(before) Errors: ArrayIndexOutOfBoundsException x4, JSONException x2
(after)  SubscriptionGroupManagerTest: Tests run: 10, Failures: 0, Errors: 0
         TopicConfigManagerTest:       Tests run: 17, Failures: 0, Errors: 0
```

Note: this branch also touches `SubscriptionGroupManager.java` (the `subGroupTable` method); #11093 touches a different method (`updateForbiddenValue`) in the same file, so one of the two may need a trivial rebase.

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/subscription/SubscriptionGroupManager.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/subscription/SubscriptionGroupManager.java
@@ -322,15 +322,15 @@ public class SubscriptionGroupManager extends ConfigManager {
     public ConcurrentHashMap<String, SubscriptionGroupConfig> subGroupTable(String dataVersion, int groupSeq,
         int maxGroupNum) {
         // [groupSeq, groupSeq + maxGroupNum)
-        int beginIndex = groupSeq;
-        if (beginIndex != 0 && (StringUtils.isBlank(dataVersion) || !Objects.equals(DataVersion.fromJson(dataVersion, DataVersion.class), getDataVersion()))) {
+        int beginIndex = Math.max(groupSeq, 0);
+        if (beginIndex != 0 && (StringUtils.isBlank(dataVersion) || !Objects.equals(parseDataVersion(dataVersion), getDataVersion()))) {
             beginIndex = 0;
             log.info("get sub subscription group table from {} due to {}", beginIndex,
                 StringUtils.isBlank(dataVersion) ? "DataVersion Empty" : "DataVersion Changed");
         }
 
         ConcurrentHashMap<String, SubscriptionGroupConfig> subGroupTable = new ConcurrentHashMap<>();
-        if (beginIndex < subscriptionGroupTable.size()) {
+        if (maxGroupNum > 0 && beginIndex < subscriptionGroupTable.size()) {
             int endIndex = Math.min(beginIndex + maxGroupNum, subscriptionGroupTable.size());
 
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Solid defensive fix for three crash vectors in the split-metadata pagination path. The clamping logic is correct, the empty-page short-circuit for non-positive limits is clean, and the `parseDataVersion` helper degrades gracefully on malformed input.

### Findings
- **[Info]** `SubscriptionGroupManager.java:244` / `TopicConfigManager.java:251` — The `parseDataVersion` helper is duplicated across both managers. Consider extracting it to a shared utility (e.g. a static method in a common base or `DataVersion` itself) to keep the two paths in sync as they evolve. Minor style nit — not blocking.
- **[Info]** `SubscriptionGroupManager.java:246` — Catching `Exception` is broad but acceptable here since the method's contract is to never throw. A comment noting the intentional breadth would help future readers.

### Verdict
LGTM. The fix is minimal, well-tested (6 new tests covering all three crash scenarios), and behaviour for well-formed requests is unchanged.

---

PR #11096, opened 2026-09-09
[BUG] EndTransactionProcessor NPEs when the prepared message lacks the producer group property

zjncs:
### Motivation

`EndTransactionProcessor.checkPrepareMessage` validates the request's producer group against the property stored on the prepared (half) message:

```java
final String pgroupRead = msgExt.getProperty(MessageConst.PROPERTY_PRODUCER_GROUP);
if (!pgroupRead.equals(requestHeader.getProducerGroup())) {
```

If the half message does not carry `PROPERTY_PRODUCER_GROUP` (half messages written by old clients, or stores restored without the property), `pgroupRead` is null and the commit/rollback request dies with:

```
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "pgroupRead" is null
```

The topic check right above already tolerates a missing field defensively (`StringUtils.isNotBlank` + `Objects.equals`), so the producer-group check is inconsistent with its own file's style.

### Changes

- Compare with `Objects.equals(pgroupRead, requestHeader.getProducerGroup())`: a missing property is a mismatch and gets the intended `SYSTEM_ERROR` / "The producer group wrong" rejection instead of an NPE.

### Verification

New test `testProcessRequestRejectsPreparedMessageWithoutProducerGroup`: commits a transaction whose prepared message has the producer-group property removed.

```
$ mvn -pl broker test -Dtest='EndTransactionProcessorTest'
(before) java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "pgroupRead" is null
(after)  Tests run: 10, Failures: 0, Errors: 0  (new test asserts SYSTEM_ERROR + "producer group wrong")
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/EndTransactionProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/EndTransactionProcessor.java
@@ -245,7 +245,10 @@ public class EndTransactionProcessor implements NettyRequestProcessor {
                 return response;
             }
             final String pgroupRead = msgExt.getProperty(MessageConst.PROPERTY_PRODUCER_GROUP);
-            if (!pgroupRead.equals(requestHeader.getProducerGroup())) {
+            // The prepared message may lack the producer group property (messages
+            // written by old clients or restored stores); compare null-safely and
+            // reject instead of failing the request with an NPE.
+            if (!Objects.equals(pgroupRead, requestHeader.getProducerGroup())) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Replaces `pgroupRead.equals(...)` with `Objects.equals(pgroupRead, ...)` to handle the case where the prepared message lacks the `PROPERTY_PRODUCER_GROUP` property — prevents NPE and returns the intended error response instead. Consistent with the topic check directly above it. Regression test included.

LGTM.

---

PR #11098, opened 2026-09-09
[ISSUE #S1] Scope the index list write lock to the list update in getAndCreateLastIndexFile

zjncs:
### Motivation

In `IndexService.getAndCreateLastIndexFile` the write lock is acquired *inside* the `try` block, after the `IndexFile` constructor, while `finally` unconditionally calls `writeLock().unlock()`:

```java
try {
    ...
    indexFile = new IndexFile(fileName, ...);   // may throw IOException
    this.readWriteLock.writeLock().lock();      // only reached if the constructor returned
    this.indexFileList.add(indexFile);
} catch (Exception e) {
    LOGGER.error("getLastIndexFile exception ", e);
} finally {
    this.readWriteLock.writeLock().unlock();    // unlocks a lock that was never acquired
}
```

When `IndexFile` creation fails (store directory missing/not writable/disk full), the constructor throws, `lock()` is never reached, and the `finally` block throws `IllegalMonitorStateException`, which propagates out of `getAndCreateLastIndexFile` into `putKeyEnd` — i.e. into the message store write path (`buildKey`/`putKey` while appending index entries). The original creation failure is already logged; the extra IMSE only masks it.

### Modifications

- Create the `IndexFile` outside the lock; take the write lock only to append it to `indexFileList`, with a balanced `try/finally`.

### Verification

Fail-before (new test on unpatched code):

```
IndexServiceTest.testGetAndCreateLastIndexFileWhenCreateFileFails » IllegalMonitorState
```

The test points the index store path below a regular file so `IndexFile` creation is guaranteed to fail.

Pass-after — full `IndexServiceTest` (5 existing + 1 new):

```
mvn -pl store test -Dtest='IndexServiceTest'
```

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
+++ b/store/src/main/java/org/apache/rocketmq/store/index/IndexService.java
@@ -356,12 +356,20 @@ public class IndexService implements CommitLogDispatchStore {
                 indexFile =
                     new IndexFile(fileName, this.hashSlotNum, this.indexNum, lastUpdateEndPhyOffset,
                         lastUpdateIndexTimestamp);
-                this.readWriteLock.writeLock().lock();
-                this.indexFileList.add(indexFile);
             } catch (Exception e) {
                 LOGGER.error("getLastIndexFile exception ", e);
-            } finally {
-                this.readWriteLock.writeLock().unlock();
+            }
+
+            // The write lock must only guard the list update: IndexFile creation
+            // happens outside it, so a failed construction cannot release a lock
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Scopes the write lock to only guard the `indexFileList.add()` call, moving `IndexFile` construction outside the lock. This prevents `IllegalMonitorStateException` when file creation fails (the `finally` block was unlocking a lock that was never acquired). Well-structured fix with a targeted regression test.

LGTM.

---

PR #11100, opened 2026-09-09
[ISSUE #S2] Guard against a missing ext unit in ConsumeQueue.estimateMessageCount

zjncs:
### Motivation

`ConsumeQueue.estimateMessageCount` dereferences the ext unit unconditionally:

```java
ConsumeQueueExt.CqExtUnit ext = null;
if (isExtWriteEnable()) {
    ext = consumeQueueExt.get(tagCode);
    tagCode = ext.getTagsCode();   // NPE when get() returns null
}
```

The stored `tagsCode` can legitimately be a raw (positive) value rather than an ext address:

- entries written while the ext was disabled and the config was enabled later, or
- entries persisted by the *save-tagsCode-only* fallback in `putMessagePositionInfoWrapper` when `consumeQueueExt.put` fails (logged as "Save consume queue extend fail, So just save tagsCode").

For a raw code `ConsumeQueueExt.get` returns `null` (it rejects non-ext addresses), so `estimateMessageCount` throws an NPE. This path is live in production: `DefaultMessageStore.estimateMessageCount` → `ConsumerLagCalculator` (broker consumer-lag metrics with a filter). `MessageFilter.isMatchedByConsumeQueue` already accepts a null ext unit ("message is before consumer"), so only this caller assumes non-null.

### Modifications

- Null-guard the ext unit: keep the raw `tagCode` when `get` returns `null` and pass the null ext to the filter, matching how the rest of the code treats a missing ext.

### Verification

Fail-before (new test on unpatched code):

```
ConsumeQueueTest.testEstimateMessageCountWhenExtUnitMissing:500 » NullPointer
  Cannot invoke "ConsumeQueueExt.getTagsCode()" because "ext" is null
```

The test writes one entry with the ext disabled (raw tagsCode persisted), re-enables the ext, then estimates the count — exactly the mixed-content state described above.

Pass-after:

```
mvn -pl store test -Dtest='ConsumeQueueTest#testEstimateMessageCountWhenExtUnitMissing+testCorrectMinOffset'
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/ConsumeQueue.java
@@ -1227,7 +1227,13 @@ public class ConsumeQueue implements ConsumeQueueInterface {
                         ConsumeQueueExt.CqExtUnit ext = null;
                         if (isExtWriteEnable()) {
                             ext = consumeQueueExt.get(tagCode);
-                            tagCode = ext.getTagsCode();
+                            // The stored code may be a raw tagsCode (entries written
+                            // before the ext was enabled, or persisted by the
+                            // "save tagsCode only" fallback), for which get() returns
+                            // null; keep the raw code and let the filter decide.
+                            if (ext != null) {
+                                tagCode = ext.getTagsCode();
+                            }
                         }
                         if (filter.isMatchedByConsumeQueue(tagCode, ext)) {
                             match++;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct NPE fix for a real production path — `estimateMessageCount` dereferences the ext unit unconditionally, but `ConsumeQueueExt.get()` returns null for raw tagsCodes in mixed-content states. The null-guard preserves the raw `tagCode` and passes null ext to the filter, matching how the rest of the codebase handles missing ext units. Test reproduces the exact mixed-content scenario.

---

PR #11104, opened 2026-09-09
[ISSUE #C1] Throw MQClientException instead of NPE for an unparseable consumeTimestamp

zjncs:
### Motivation

`computePullFromWhereWithException` in both `RebalanceLitePullImpl` and `RebalancePushImpl` does:

```java
long timestamp = UtilAll.parseDate(consumer.getConsumeTimestamp(), UtilAll.YYYYMMDDHHMMSS).getTime();
```

`UtilAll.parseDate` returns `null` when the string does not match the pattern, and `consumeTimestamp` is a plain user-settable string on `DefaultLitePullConsumer`/`DefaultMQPushConsumer`. A typo like `setConsumeTimestamp("20240101")` therefore kills the rebalance with an NPE (`Cannot invoke Date.getTime() ... is null`) instead of a proper error: the NPE escapes `updateProcessQueueTable`'s `catch (Exception e)`... actually it is caught there (`compute offset failed`), but it surfaces as an opaque NPE in the logs, and the deprecated `computePullFromWhere` wrapper only handles `MQClientException`, so the NPE propagates to its callers as well.

### Modifications

- In both impls: parse the date first, and throw `MQClientException("Invalid consumeTimestamp: ...")` when it does not parse. All callers already handle `MQClientException` (`updateProcessQueueTable` catches it and skips the queue; the deprecated wrapper logs and returns -1), so a misconfigured timestamp now produces a clear error message instead of an NPE.

### Verification

Fail-before (both new tests on unpatched code):

```
RebalanceLitePullImplTest.testComputePullFromWhereWithException_eq_minus1_invalid_timestamp:108 » NullPointer
  Cannot invoke "java.util.Date.getTime()" because the return value of "UtilAll.parseDate(String, String)" is null
RebalancePushImplTest.testComputePullFromWhereWithException_eq_minus1_invalid_timestamp:226 » NullPointer (same)
```

Pass-after — both test classes (existing + new tests):

```
mvn -pl client test -Dtest='RebalanceLitePullImplTest,RebalancePushImplTest'
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalanceLitePullImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalanceLitePullImpl.java
@@ -16,6 +16,7 @@
  */
 package org.apache.rocketmq.client.impl.consumer;
 
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
 import org.apache.rocketmq.client.consumer.AllocateMessageQueueStrategy;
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
Good defensive fix — replaces an opaque NPE from `UtilAll.parseDate` returning null with a clear `MQClientException("Invalid consumeTimestamp: ...")`. Applied consistently to both `RebalanceLitePullImpl` and `RebalancePushImpl`. All callers already handle `MQClientException` properly, so the error message now surfaces cleanly instead of being wrapped in an NPE stack trace. Tests cover both impls.

---

PR #11105, opened 2026-09-09
[ISSUE #C2] Count down the queryMessage latch when the invocation fails

zjncs:
### Motivation

`MQAdminImpl.queryMessage` fans out one asynchronous query per broker and waits on a `CountDownLatch(brokerAddrs.size())`. The latch is released by the `InvokeCallback` (`operationSucceed`/`operationFail`), but if `MQClientAPIImpl.queryMessage` itself throws synchronously — e.g. `RemotingConnectException`/`RemotingTimeoutException` for an unreachable broker — no callback is ever invoked, and the catch block only logs:

```java
} catch (Exception e) {
    log.warn("queryMessage exception", e);
}
...
boolean ok = countDownLatch.await(timeoutMillis * 4, TimeUnit.MILLISECONDS);
```

The caller then blocks for the full `timeoutMillis * 4` window — **24 seconds** with the default `timeoutMillis` of 6s — before getting the "no message" answer, even though the outcome was already known. Any tool/admin call that queries by key while one broker is down pays this stall.

### Modifications

- Add `countDownLatch.countDown();` in the catch block, releasing this broker's count exactly as the callback would have.

### Verification

Fail-before (new test on unpatched code — the per-broker invoke is mocked to throw, `timeoutMillis=200`):

```
MQAdminImplTest.assertQueryMessageNotBlockedWhenInvokeThrows:192
  queryMessage should not wait for the whole latch timeout, elapsed=846
```

Pass-after — full `MQAdminImplTest` (11 existing + 1 new); the same call now returns in single-digit milliseconds:

```
mvn -pl client test -Dtest='MQAdminImplTest'
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQAdminImpl.java
@@ -426,6 +426,10 @@ public class MQAdminImpl {
                             }, isUniqKey);
                     } catch (Exception e) {
                         log.warn("queryMessage exception", e);
+                        // The callback never fires when the invocation itself
+                        // fails, so release this broker's latch count here or
+                        // await below blocks for the full timeout.
+                        countDownLatch.countDown();
                     }
 
                 }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Correct fix for a real stall — when `MQClientAPIImpl.queryMessage` throws synchronously (e.g. `RemotingConnectException`), the callback never fires and the latch is never decremented, causing the caller to block for the full `timeoutMillis * 4` (24s default). Adding `countDownLatch.countDown()` in the catch block is the minimal correct fix. Test validates the timing improvement.

---

PR #11106, opened 2026-09-09
[ISSUE #C3] Reject an AckMessageRequest without entries instead of throwing IndexOutOfBoundsException

zjncs:
### Motivation

`AckMessageActivity.ackMessage` inspects the first entry before checking that any entry exists:

```java
boolean isBatchAck = ConfigurationManager.getProxyConfig().isEnableBatchAck()
    && !request.getEntries(0).hasLiteTopic();
```

An `AckMessageRequest` whose `entries` list is empty (nothing in the protobuf schema forbids it) therefore throws `IndexOutOfBoundsException` from `getEntries(0)` when `enableBatchAck` is on, which the catch block wraps into the future as a raw `Throwable` — the gRPC client receives an opaque internal error instead of a meaningful status.

### Modifications

- Guard `request.getEntriesCount() == 0` before the batch-ack decision and throw `GrpcProxyException(Code.BAD_REQUEST, "ack entries should not be empty")`, the same input-validation pattern used by `SendMessageActivity` (e.g. max recovery time / delivery timestamp checks).

### Verification

Fail-before (new test on unpatched code, with `enableBatchAck=true`): the future completes exceptionally with the `IndexOutOfBoundsException`, so the assertion on `GrpcProxyException`/`Code.BAD_REQUEST` fails:

```
AckMessageActivityTest.testAckMessageWithEmptyEntries:84 » AssertionError
```

Pass-after — full `AckMessageActivityTest` (2 existing + 1 new):

```
mvn -pl proxy test -Dtest='AckMessageActivityTest'
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/AckMessageActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/AckMessageActivity.java
@@ -36,6 +36,7 @@ import org.apache.rocketmq.proxy.grpc.v2.AbstractMessagingActivity;
 import org.apache.rocketmq.proxy.grpc.v2.channel.GrpcChannelManager;
 import org.apache.rocketmq.proxy.grpc.v2.channel.GrpcClientChannel;
 import org.apache.rocketmq.proxy.grpc.v2.common.GrpcClientSettingsManager;
+import org.apache.rocketmq.proxy.grpc.v2.common.GrpcProxyException;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clean input-validation fix — guards `getEntries(0)` against an empty `entries` list before the batch-ack decision, throwing `GrpcProxyException(Code.BAD_REQUEST)` instead of letting an `IndexOutOfBoundsException` leak through as an opaque gRPC internal error. Follows the same pattern used by `SendMessageActivity`. Test covers the fail-before path.

---

PR #11107, opened 2026-09-10
[ISSUE #D1] Skip queue-change notification for concurrently removed subscription

zjncs:
## Motivation

`RebalanceImpl.doRebalance` iterates the `subscriptionInner` entry set and, once the allocation result changes, hands the topic over to `messageQueueChanged`. The push implementation re-reads the entry:

```java
SubscriptionData subscriptionData = this.subscriptionInner.get(topic);
long newVersion = System.currentTimeMillis();
log.info("... {} ...", ..., subscriptionData.getSubVersion(), newVersion);
subscriptionData.setSubVersion(newVersion);
```

`DefaultMQPushConsumer.unsubscribe(topic)` is a public API that runs on a user thread and removes the entry from the same `ConcurrentHashMap`. When it interleaves between the rebalance loop's iteration and this call, `subscriptionInner.get(topic)` returns null and the rebalance thread dies with:

```
java.lang.NullPointerException: Cannot invoke "SubscriptionData.getSubVersion()" because "subscriptionData" is null
```

That aborts `doRebalance` for the remaining topics of this consumer until the next scheduled round.

## Modification

`RebalancePushImpl.messageQueueChanged`: if the subscription is already gone, log and return — there is nothing to update or notify for a topic this consumer no longer subscribes to.

## Test Evidence

```
docker exec rmq-build mvn -q -pl client test -Dtest='RebalancePushImplTest#testMessageQueueChanged_SubscriptionRemovedConcurrently'
java.lang.NullPointerException: Cannot invoke "SubscriptionData.getSubVersion()" because "subscriptionData" is null
```

Pass-after (full class):

```
docker exec rmq-build mvn -q -pl client test -Dtest='RebalancePushImplTest'
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/consumer/RebalancePushImpl.java
@@ -58,6 +58,13 @@ public class RebalancePushImpl extends RebalanceImpl {
          * Fix: inconsistency subscription may lead to consumer miss messages.
          */
         SubscriptionData subscriptionData = this.subscriptionInner.get(topic);
+        // The subscription may have been removed concurrently (unsubscribe() from
+        // another thread) while this rebalance was in progress; there is nothing
+        // to update or notify for a topic this consumer no longer subscribes to.
+        if (subscriptionData == null) {
+            log.info("subscription of {} removed during rebalance, skip notifying message queue change", topic);
+            return;
+        }
         long newVersion = System.currentTimeMillis();
         log.info("{} Rebalance changed, also update version: {}, {}", topic, subscriptionData.getSubVersion(), newVersion);
         subscriptionData.setSubVersion(newVersion);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

This PR correctly fixes a race condition where `RebalancePushImpl.messageQueueChanged` throws NPE when `unsubscribe(topic)` is called concurrently from another thread, removing the subscription entry between the rebalance loop iteration and the `subscriptionInner.get(topic)` call.

### Findings

- **[Info]** `RebalancePushImpl.java:61-66` — The null check with early return is the right fix. If the subscription is already gone, there is genuinely nothing to update or notify. The log message is informative for debugging.
- **[Info]** `RebalancePushImplTest.java` — The regression test correctly emulates the race by deliberately omitting the subscription from `subscriptionInner`, verifying that no heartbeat is sent and no NPE is thrown. Clean test design.

### Assessment

Minimal, targeted fix for a real concurrency bug. The PR description clearly explains the race window and impact (aborting `doRebalance` for remaining topics). No backward compatibility concerns.

---

PR #11108, opened 2026-09-10
[ISSUE #D2] Return null when selecting from a route-less TopicPublishInfo

zjncs:
## Motivation

`TopicPublishInfo` exposes three `selectOneMessageQueue` overloads with inconsistent contracts:

```java
// QueueFilter variant — guarded
if (messageQueueList == null || messageQueueList.isEmpty()) {
    return null;
}

// no-arg variant — unguarded
int index = this.sendWhichQueue.incrementAndGet();
int pos = index % this.messageQueueList.size();   // NPE on null list, ArithmeticException (x % 0) on empty list
```

A topic whose route exists but has no writable queue legitimately produces such an info object: `topicRouteData2TopicPublishInfo` adds nothing to the list when every `QueueData` is read-only or has `writeQueueNums == 0`, so `ok()` is false while the object itself is non-null.

`MQFaultStrategy.selectOneMessageQueue` falls back to the unguarded no-arg variant, and the async-send retry path (`MQClientAPIImpl.onExceptionImpl`) is the one caller that only null-checks the info instead of checking `ok()` like every send entry point does.

## Modification

- `TopicPublishInfo.selectOneMessageQueue()` and `selectOneMessageQueue(String lastBrokerName)`: return null when the queue list is null/empty, aligning with the QueueFilter variant.
- `MQClientAPIImpl.onExceptionImpl`: apply the same `ok()` contract — a route-less info keeps the "retry the same broker" default instead of feeding the selector.

No behavior change for existing callers: every other call site (sendDefaultImpl, EscapeBridge) already checks `ok()` before selecting.

## Test Evidence

Fail-before (unpatched develop, new `TopicPublishInfoTest`):

```
mvn -q -pl client test -Dtest='TopicPublishInfoTest#testSelectOneMessageQueueWithEmptyQueueList'
Tests run: 1, Errors: 1 - java.lang.ArithmeticException: / by zero
#testSelectOneMessageQueueWithNullQueueList      -> NullPointerException (List.size())
#testSelectOneMessageQueueLastBrokerNameWithEmptyQueueList -> ArithmeticException: / by zero
#testSelectOneMessageQueueLastBrokerNameWithNullQueueList   -> NullPointerException (List.size())
```

Pass-after:

```
mvn -q -pl client test -Dtest='TopicPublishInfoTest,SelectMessageQueueRetryTest'
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/MQClientAPIImpl.java
@@ -718,7 +718,10 @@ public class MQClientAPIImpl implements NameServerUpdateCallback, StartAndShutdo
         int tmp = curTimes.incrementAndGet();
         if (needRetry && tmp <= timesTotal && timeoutMillis > 0) {
             String retryBrokerName = brokerName;//by default, it will send to the same broker
-            if (topicPublishInfo != null) { //select one message queue accordingly, in order to determine which broker to send
+            // Every send entry point checks ok() before queue selection; the retry
+            // path must do the same, otherwise a route-less topicPublishInfo makes
+            // the selection throw or return null instead of retrying the same broker.
+            if (topicPublishInfo != null && topicPublishInfo.ok()) { //select one message queue accordingly, in order to determine which broker to send
                 MessageQueue mqChosen = producer.selectOneMessageQueue(topicPublishInfo, brokerName, false);
                 retryBrokerName = instance.getBrokerNameFromMessageQueue(mqChosen);
             }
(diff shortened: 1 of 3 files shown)


RockteMQ-AI (approved):
### Summary

This PR correctly fixes NPE and ArithmeticException in `TopicPublishInfo.selectOneMessageQueue()` when the queue list is null or empty — a real edge case for topics whose route exists but has no writable queue.

### Findings

- **[Info]** `TopicPublishInfo.java:107-122` — The refactoring of `selectOneMessageQueue(String lastBrokerName)` to early-return null on empty/null list is clean and aligns with the `QueueFilter` variant contract. Good consistency improvement.
- **[Info]** `MQClientAPIImpl.java:721` — Adding the `ok()` guard here is the right call. The async retry path was the only caller that relied on the unguarded variant; all other send entry points already check `ok()` before selection.
- **[Info]** Test coverage is thorough — covers null list, empty list, `lastBrokerName` variants, and a positive case with actual queues.

### Assessment

Well-motivated fix with clear PR description, correct root cause analysis, and comprehensive tests. No backward compatibility concerns since existing callers already check `ok()` before queue selection.

---

PR #11110, opened 2026-09-10
[ISSUE #D4] Fix getConsumeStatus NPE for a group absent from ConsumerManager

zjncs:
## Motivation

`Broker2Client.getConsumeStatus` dereferences the consumer group info before checking it:

```java
ConcurrentMap<Channel, ClientChannelInfo> channelInfoTable =
    this.brokerController.getConsumerManager().getConsumerGroupInfo(group).getChannelInfoTable();
if (null == channelInfoTable || channelInfoTable.isEmpty()) {
    // "No Any Consumer online in the consumer group: [%s]"
}
```

When the queried group has no consumer registered on this broker — the normal situation for an offline group — `getConsumerGroupInfo(group)` returns null and the method throws `NullPointerException` before reaching the error branch that was written for exactly this case. The sibling `resetOffset` method checks the group info for null first.

Reached from `mqadmin getConsumerStatus` (`INVOKE_BROKER_TO_GET_CONSUMER_STATUS` → `AdminBrokerProcessor.getConsumerStatus` → here), the NPE surfaces to the admin as a bare SYSTEM_ERROR with a null-pointer remark instead of the intended message.

## Modification

Fetch the `ConsumerGroupInfo` first; enter the existing "No Any Consumer online" branch when it is null or its channel table is empty; read the channel table only afterwards.

## Test Evidence

```
mvn -q -pl broker test -Dtest='Broker2ClientTest#testGetConsumeStatusGroupNotRegistered'
Tests run: 1, Errors: 1 - NullPointerException: Cannot invoke "ConsumerGroupInfo.getChannelInfoTable()"
```

Pass-after (full class):

```
mvn -q -pl broker test -Dtest='Broker2ClientTest'
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/client/net/Broker2Client.java
@@ -253,13 +253,16 @@ public class Broker2Client {
                 requestHeader);
 
         Map<String, Map<MessageQueue, Long>> consumerStatusTable = new HashMap<>();
-        ConcurrentMap<Channel, ClientChannelInfo> channelInfoTable =
-            this.brokerController.getConsumerManager().getConsumerGroupInfo(group).getChannelInfoTable();
-        if (null == channelInfoTable || channelInfoTable.isEmpty()) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean NPE fix — the null check now happens before dereferencing. Test coverage for the offline group case is good.

---

PR #11111, opened 2026-09-10
[ISSUE #D5] cloneGroupOffset: clone per loop topic, not the blank header topic

zjncs:
## Motivation

`AdminBrokerProcessor.cloneGroupOffset` supports a blank-topic mode (`UtilAll.isBlank(requestHeader.getTopic())`) that clones the offsets of **every** topic the source group consumes — this is `mqadmin cloneGroupOffset` without `-t`. The method collects the topic set and iterates it, but then calls:

```java
this.brokerController.getConsumerOffsetManager().cloneOffset(
    requestHeader.getSrcGroup(), requestHeader.getDestGroup(),
    requestHeader.getTopic());   // <-- null in the blank-topic mode
```

`ConsumerOffsetManager.cloneOffset` looks up `topic + TOPIC_GROUP_SEPARATOR + srcGroup`, so with a null topic the key `null@srcGroup` never matches: no offset is cloned, yet the broker replies SUCCESS. The whole "clone all topics of a group" admin feature silently does nothing.

## Modification

Pass the loop variable `topic` to `cloneOffset`. In the explicit-topic mode the two are identical, so only the blank-topic mode changes (from no-op to actually cloning).

## Test Evidence

```
mvn -q -pl broker test -Dtest='AdminBrokerProcessorTest#testCloneGroupOffsetWithoutTopicClonesEveryTopicOfSourceGroup'
Argument(s) are different! Wanted: cloneOffset("srcGroup", "destGroup", "cloneTopic1")
```

Pass-after (full class):

```
mvn -q -pl broker test -Dtest='AdminBrokerProcessorTest'
Tests run: 94, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -2676,8 +2676,11 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
                 }
             }
 
+            // In the blank-topic mode the loop iterates over every topic the
+            // source group consumes, so the offset must be cloned per loop
+            // variable, not for the (blank) header topic.
             this.brokerController.getConsumerOffsetManager().cloneOffset(requestHeader.getSrcGroup(), requestHeader.getDestGroup(),
-                requestHeader.getTopic());
+                topic);
         }
 
         response.setCode(ResponseCode.SUCCESS);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good catch — the blank-topic mode was incorrectly using the header topic (null) instead of the loop variable. Fix is correct and test verifies the behavior.

---

PR #11116, opened 2026-09-10
[ISSUE #D9] Guard consumeMessageDirectly against unresolvable msgId

zjncs:
## Motivation

`AdminBrokerProcessor.consumeMessageDirectly` (backing `mqadmin consumeMessageDirectly` / console "trace/diagnose message") fetches the message by offset with no null check:

```java
selectMappedBufferResult = this.brokerController.getMessageStore().selectOneMessageByOffset(messageId.getOffset());
byte[] body = new byte[selectMappedBufferResult.getSize()];   // NPE when null
```

`selectOneMessageByOffset` returns null when the offset no longer maps to a stored message — the commitlog segment was expired/deleted, or the msgId was never on this broker. The broker then throws a raw `NullPointerException` into the remoting layer instead of answering the admin request.

Additionally, the `catch (UnknownHostException e) {}` block silently swallows msgId decode failures and falls through to `callConsumer` with a **null body**, so the consumer is invoked with an empty request and fails with an unrelated decode error far from the cause.

## Modification

- If `selectOneMessageByOffset` returns null, respond `SYSTEM_ERROR` with remark `can not find message by id: <msgId>` and do not invoke the consumer.
- On `UnknownHostException` (undecodable msgId), respond `SYSTEM_ERROR` with remark `can not decode message id: <msgId>` instead of forwarding a bodiless request.

## Test Evidence

**Fail-before** (unpatched code, new test `AdminBrokerProcessorTest#testConsumeMessageDirectlyMessageNotFound` with a well-formed msgId whose offset is not stored):

```
docker exec rmq-build mvn -q -pl broker test -Dtest='AdminBrokerProcessorTest#testConsumeMessageDirectlyMessageNotFound' -Dsurefire.failIfNoSpecifiedTests=true
Tests run: 1, Errors: 1 ... java.lang.NullPointerException: Cannot invoke "...SelectMappedBufferResult.getSize()" because "selectMappedBufferResult" is null
```

```
docker exec rmq-build mvn -q -pl broker test -Dtest='AdminBrokerProcessorTest' -Dsurefire.failIfNoSpecifiedTests=true
Tests run: 94, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AdminBrokerProcessor.java
@@ -2624,14 +2624,25 @@ public class AdminBrokerProcessor implements NettyRequestProcessor {
             }
         }
         SelectMappedBufferResult selectMappedBufferResult = null;
+        final RemotingCommand response = RemotingCommand.createResponseCommand(null);
         try {
             MessageId messageId = MessageDecoder.decodeMessageId(requestHeader.getMsgId());
             selectMappedBufferResult = this.brokerController.getMessageStore().selectOneMessageByOffset(messageId.getOffset());
-
+            // The offset may no longer map to a stored message (expired/deleted
+            // commitlog or a msgId from another broker); report that instead of
+            // throwing an NPE or invoking the consumer with an empty body.
+            if (selectMappedBufferResult == null) {
+                response.setCode(ResponseCode.SYSTEM_ERROR);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Defensive fix for `consumeMessageDirectly` — adds null-check on `selectMappedBufferResult` and catches `UnknownHostException` from message ID decoding, returning a proper `SYSTEM_ERROR` response instead of NPE or unhandled exception. Includes targeted test coverage.

Clean, well-scoped fix with good test cases. LGTM.

---

PR #11118, opened 2026-09-10
[ISSUE #D11] Guard ClientRemotingProcessor against requests without body

zjncs:
## Motivation

`ClientRemotingProcessor` handles broker-initiated requests on the client. Two of them wrap the request body without a null check:

```java
// checkTransactionState
final ByteBuffer byteBuffer = ByteBuffer.wrap(request.getBody());          // L99
// consumeMessageDirectly
final MessageExt msg = MessageDecoder.clientDecode(ByteBuffer.wrap(request.getBody()), true);  // L206
```

A bodiless request makes the client throw a raw `NullPointerException` inside `processRequest`. This is reachable in practice: the broker side of `CONSUME_MESSAGE_DIRECTLY` resolves the message by msgId and forwards it in the body — when the offset no longer maps to a stored message (expired/deleted commitlog), the forwarded request carries **no body**, so the client NPEs and the admin only sees a generic system error instead of the actual cause. The same defensive gap exists for `CHECK_TRANSACTION_STATE`.

## Modification

- `checkTransactionState`: if the body is empty, log a warn (with the broker address) and return — the broker re-checks the transaction later, matching the method's fire-and-forget contract.
- `consumeMessageDirectly`: if the body is empty, answer `SYSTEM_ERROR` with remark "The request does not carry a message body, please check the message on the broker" instead of failing during decode.

## Test Evidence

**Fail-before** (unpatched code, two new tests `ClientRemotingProcessorTest#testCheckTransactionStateWithoutBody` / `#testConsumeMessageDirectlyWithoutBody` with `request.getBody() == null`):

```
docker exec rmq-build mvn -q -pl client test -Dtest='ClientRemotingProcessorTest#testCheckTransactionStateWithoutBody+testConsumeMessageDirectlyWithoutBody' -Dsurefire.failIfNoSpecifiedTests=true
Tests run: 2, Errors: 1 ... java.lang.NullPointerException: Cannot read the array length because "array" is null
```

```
docker exec rmq-build mvn -q -pl client test -Dtest='ClientRemotingProcessorTest' -Dsurefire.failIfNoSpecifiedTests=true
```

### DIFF
--- a/client/src/main/java/org/apache/rocketmq/client/impl/ClientRemotingProcessor.java
+++ b/client/src/main/java/org/apache/rocketmq/client/impl/ClientRemotingProcessor.java
@@ -96,6 +96,14 @@ public class ClientRemotingProcessor implements NettyRequestProcessor {
         RemotingCommand request) throws RemotingCommandException {
         final CheckTransactionStateRequestHeader requestHeader =
             (CheckTransactionStateRequestHeader) request.decodeCommandCustomHeader(CheckTransactionStateRequestHeader.class);
+        // A broker that fails to resolve the transaction message sends the
+        // request without a body; skip the check instead of throwing an NPE,
+        // the broker will re-check this transaction later.
+        if (request.getBody() == null) {
+            logger.warn("checkTransactionState, message body is empty, brokerAddr={}",
+                RemotingHelper.parseChannelRemoteAddr(ctx.channel()));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

### Findings

### Suggestions

---

PR #11124, opened 2026-09-10
[ISSUE #D17] Discard delayed messages with malformed real topic or queue id in messageTimeUp

zjncs:
## Motivation

`ScheduleMessageService.messageTimeUp` rewrites a delayed message back to its destination with

```java
msgInner.setTopic(msgInner.getProperty(MessageConst.PROPERTY_REAL_TOPIC));
String queueIdStr = msgInner.getProperty(MessageConst.PROPERTY_REAL_QUEUE_ID);
int queueId = Integer.parseInt(queueIdStr);
```

A delayed message in `SCHEDULE_TOPIC` whose destination properties are missing or malformed (corrupted properties string, or a hand-crafted message written straight into the system topic) makes this throw:

- `Integer.parseInt(null)` / `NumberFormatException` propagates to `DeliverDelayedMessageTimerTask.executeOnTimeUp`, whose catch aborts the scan of the current consume-queue batch — the remaining messages of that batch each lose one delivery cycle and every occurrence logs a full stack trace.
- a missing `PROPERTY_REAL_TOPIC` is worse: the topic becomes `null`, `putMessage` rejects it, `syncDeliver` returns false and the task reschedules the same offset forever, wedging that delay level.

The adjacent real-topic guard (`[BUG] the real topic of schedule msg is ..., discard the msg`) already established that such messages must be discarded, but only for one specific wrong value, not for the missing/malformed cases.

## Modification

- `messageTimeUp` validates `PROPERTY_REAL_TOPIC` and `PROPERTY_REAL_QUEUE_ID` before using them and returns `null` when the destination is unusable, logging a `[BUG] ... discard` line like the existing guard.
- Both call sites (`executeOnTimeUp` and `PutResultProcess.doResend`) skip a null conversion result, so the rest of the batch is still processed in-cycle and a resend of a permanently malformed message is marked skip/exception instead of retrying.

## Test Evidence

New test `testMessageTimeUpWithMalformedRealQueueId` covers: non-numeric queue id, missing queue id, missing real topic, and the well-formed case.

```
docker exec rmq-build mvn -q -pl broker test -Dtest='ScheduleMessageServiceTest#testMessageTimeUpWithMalformedRealQueueId' -Dsurefire.failIfNoSpecifiedTests=true
```

Before the fix:

```
java.lang.reflect.InvocationTargetException
Caused by: java.lang.NumberFormatException: For input string: "notANumber"
```

After the fix:

```
docker exec rmq-build mvn -q -pl broker test -Dtest='ScheduleMessageServiceTest' -Dsurefire.failIfNoSpecifiedTests=true
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0   (includes the end-to-end testDeliverDelayedMessageTimerTask)
```

No associated issue (self-discovered during a broker self-audit).

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/schedule/ScheduleMessageService.java
@@ -354,10 +354,23 @@ public class ScheduleMessageService extends ConfigManager {
         MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_TIMER_DELIVER_MS);
         MessageAccessor.clearProperty(msgInner, MessageConst.PROPERTY_TIMER_DELAY_SEC);
 
-        msgInner.setTopic(msgInner.getProperty(MessageConst.PROPERTY_REAL_TOPIC));
-
+        // The message can only be rewritten to its destination topic when both
+        // properties are usable; without them it can never be delivered and
+        // throwing here would abort the scan of the whole delay queue batch.
+        String realTopic = msgInner.getProperty(MessageConst.PROPERTY_REAL_TOPIC);
         String queueIdStr = msgInner.getProperty(MessageConst.PROPERTY_REAL_QUEUE_ID);
-        int queueId = Integer.parseInt(queueIdStr);
+        if (realTopic == null || queueIdStr == null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

### Findings

### Suggestions

---

PR #11126, opened 2026-09-10
[ISSUE #D19] Tolerate null pull result when checking half messages

zjncs:
## Motivation

`TransactionalMessageServiceImpl.check` iterates the half topic and, for an offset whose message cannot be fetched, inspects the pull status:

```java
if (getResult.getPullResult().getPullStatus() == PullStatus.NO_NEW_MSG) {
```

But `getHalfMsg` only fills `getResult.pullResult` when `TransactionalMessageBridge.getMessage` returns non-null, and that method **returns null** whenever `store.getMessage` does ("Get message from store return null" — e.g. the store is shutting down or not readable). In that case `getPullResult()` is null and this line throws an NPE.

The NPE is swallowed by the outermost `catch (Throwable)` of `check`, which means the entire check cycle is aborted: the offset update for the already-processed messages of the current queue is skipped and every remaining queue of the half topic is not checked at all, for as long as the store keeps failing to serve reads.

Notably the tail of the very same method already guards the same lookup for null (`getResult.getPullResult() == null ? newOffset : ...`); the guard is just missing here.

## Modification

Treat a null pull result like the other unread-offset cases: log a warning and `break` out of this queue's loop, so the consume offset of the messages that were already checked is still committed and the unread offset is retried on the next check round.

## Test Evidence

New test `testCheck_whenHalfMessagePullFails`: offset 0 returns a discardable half message, offset 1 hits the store read failure (bridge returns null). It asserts the discard listener fired and that `updateConsumeOffset(queue, 1)` was called.

```
docker exec rmq-build mvn -q -pl broker test -Dtest='TransactionalMessageServiceImplTest#testCheck_whenHalfMessagePullFails' -Dsurefire.failIfNoSpecifiedTests=true
```

Before the fix:

```
java.lang.NullPointerException: Cannot invoke "PullResult.getPullStatus()" because the return value of "GetResult.getPullResult()" is null
Wanted but not invoked: bridge.updateConsumeOffset(...)   (Tests run: 1, Failures: 1)
```

After the fix:

```
docker exec rmq-build mvn -q -pl broker test -Dtest='TransactionalMessageServiceImplTest' -Dsurefire.failIfNoSpecifiedTests=true
```

No associated issue (self-discovered during a broker self-audit).

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/transaction/queue/TransactionalMessageServiceImpl.java
@@ -218,6 +218,17 @@ public class TransactionalMessageServiceImpl implements TransactionalMessageServ
                             if (getMessageNullCount++ > MAX_RETRY_COUNT_WHEN_HALF_NULL) {
                                 break;
                             }
+                            // The bridge answers a null pull result when the store
+                            // itself cannot serve the read (e.g. during shutdown);
+                            // retry this offset on the next check round instead of
+                            // throwing, which would abort the check of the
+                            // remaining queues too. The tail of this method guards
+                            // the same lookup for null.
+                            if (getResult.getPullResult() == null) {
+                                log.warn("Get half message from store failed, the miss offset={} in={}, continue check={}",
+                                    i, messageQueue, getMessageNullCount);
+                                break;
+                            }
                             if (getResult.getPullResult().getPullStatus() == PullStatus.NO_NEW_MSG) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
﻿### Summary
This PR fixes a NullPointerException in `TransactionalMessageServiceImpl.check()` when the store cannot serve a read (e.g. during shutdown). The null `pullResult` is now handled gracefully — the current queue's loop breaks, allowing already-processed offsets to be committed and the unread offset to be retried on the next round.

The fix is consistent with the existing null guard at the tail of the same method, well-documented with inline comments, and covered by a new test case (`testCheck_whenHalfMessagePullFails`) that verifies both message processing and offset update behavior.

Clean, focused fix. LGTM.

---

PR #11130, opened 2026-09-11
[ISSUE #11129] HashedWheelTimer resource leak when broker shuts down with Lite Topic pop

### LINKED ISSUE #11129: [Bug] HashedWheelTimer resource leak when broker shuts down with Lite Topic pop

OS: Linux / macOS

### RocketMQ version
branch: develop

### JDK Version
Compiler: Oracle JDK 11

### Describe the Bug
When a broker serves Lite Topic pop requests with pop order lock release notification enabled, a dedicated timer (Netty HashedWheelTimer, backed by a non-daemon thread) is created but never shut down during the broker shutdown flow. On broker exit, Netty logs the "LEAK: HashedWheelTimer" warning, and if any timer task has ever been scheduled, the non-daemon thread prevents the JVM from exiting.

### Steps to Reproduce
1. Start a broker and enable pop order lock release notification (enableNotifyAfterPopOrderLockRelease).
2. Pop messages from a Lite Topic in a way that triggers the order lock notify path.
3. Shut down the broker.

All timer resources are released during broker shutdown; no Netty LEAK warning is logged and the JVM exits normally.

### What Did You See Instead?
Netty logs `LEAK: HashedWheelTimer was not shut down before it was garbage-collected; see https://netty.io/wiki/reference-counted-objects.html for more information.`

### Additional Context
The Lite pop processor holds its own ConsumerOrderInfoManager instance (separate from the main one that IS shut down by BrokerController). Its internal lock manager creates the timer, and the Lite shutdown block only stops the lock manager and long-polling service.

f1amingo:

- Fixes #11129

### Brief Description

The Lite pop processor creates its own ConsumerOrderInfoManager whose internal lock manager holds a non-daemon HashedWheelTimer, but the broker shutdown flow never shuts it down — the existing consumerOrderInfoManager.shutdown() call covers a different instance. This adds the missing shutdown call so the timer is stopped on broker exit.

- Compiled the broker module successfully (mvn -pl broker -am install).
- Manually traced the shutdown chain: BrokerController -> popLiteMessageProcessor.getConsumerOrderInfoManager().shutdown() -> QueueLevelConsumerOrderInfoLockManager.shutdown() -> timer.stop().

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/BrokerController.java
@@ -1562,6 +1562,7 @@ public class BrokerController {
             if (this.popLiteMessageProcessor.getPopLiteLongPollingService() != null) {
                 this.popLiteMessageProcessor.getPopLiteLongPollingService().shutdown();
             }
+            this.popLiteMessageProcessor.getConsumerOrderInfoManager().shutdown();
         }
 
         if (this.popMessageProcessor.getQueueLockManager() != null) {


RockteMQ-AI (approved):
### Summary
This PR fixes a `HashedWheelTimer` resource leak in the Lite Topic pop shutdown path by adding `consumerOrderInfoManager.shutdown()` in `BrokerController.shutdownBasicService()`.

The fix is correct and safe:
- `consumerOrderInfoManager` is a `final` field initialized in the `PopLiteMessageProcessor` constructor, so no null check is needed
- `shutdown()` is inherited from `QueueLevelConsumerManager` which properly shuts down the internal `QueueLevelConsumerOrderInfoLockManager` (the component that creates the `HashedWheelTimer`)
- The call is placed inside the existing `if (this.popLiteMessageProcessor != null)` block, maintaining consistency with the surrounding shutdown pattern

LGTM — clean, minimal fix for the reported resource leak.

---

imzs (approved):

RongtongJin (approved):

PR #11131, opened 2026-09-11
fix(tools): correct inverted dryRun flag in exportPopRecord command

zjncs:
### Motivation

`exportPopRecordCommand` built its dry-run flag as

```java
Boolean.FALSE.toString().equalsIgnoreCase(commandLine.getOptionValue('d'));
```

so the command ran in dry-run mode only when `-d` was literally `false` — exactly inverted. `mqadmin exportPopRecord -d true` (explicitly requesting no changes) actually sent the mutating `POP_ROLLBACK` request, while `-d false` skipped the rollback and only exported the records.

### Modifications

Parse the flag with `Boolean.parseBoolean` so the command honors the documented semantics.

### Verification

```
ExportPopRecordCommandTest#testNonDryRunExportsRecords
java.lang.AssertionError (message is the captured console output): with -d false the inverted
flag printed dryRun=true and skipped the export instead of attempting it
```

Pass-after:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- ExportPopRecordCommandTest
```

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/export/ExportPopRecordCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/export/ExportPopRecordCommand.java
@@ -66,7 +66,7 @@ public class ExportPopRecordCommand implements SubCommand {
         try {
             adminExt.start();
             boolean dryRun = commandLine.hasOption('d') &&
-                Boolean.FALSE.toString().equalsIgnoreCase(commandLine.getOptionValue('d'));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Critical bug fix — the inverted dryRun flag meant `-d true` (requesting no changes) actually executed mutations. Using `Boolean.parseBoolean` is the correct approach. Good test coverage verifying both directions.

---

RongtongJin (approved):

PR #11132, opened 2026-09-11
fix(tools): scope checkRocksdbCqWriteProgress to the requested cluster and skip master-less brokers

zjncs:
### Motivation

The `-c` option of `mqadmin checkRocksdbCqWriteProgress` was only used for an existence check, while the request loop afterwards walked the whole `brokerAddrTable`, so the check request was fired at brokers of every cluster — not just the requested one. Brokers whose master was down resolved to a null address, which `defaultMQAdminExtImpl` silently routed through the name server channel.

### Modifications

Iterate the brokers of the requested cluster only, and skip brokers without a master.

### Verification

```
CheckRocksdbCqWriteProgressCommandTest#testCheckOnlySpecifiedCluster
java.lang.RuntimeException: CheckRocksdbCqWriteProgressCommand command failed
  at CheckRocksdbCqWriteProgressCommand.execute(CheckRocksdbCqWriteProgressCommand.java:99)
(check request hit a master-less broker outside the requested cluster)
```

Pass-after:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- CheckRocksdbCqWriteProgressCommandTest
```

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/queue/CheckRocksdbCqWriteProgressCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/queue/CheckRocksdbCqWriteProgressCommand.java
@@ -84,10 +84,16 @@ public class CheckRocksdbCqWriteProgressCommand implements SubCommand {
                 System.out.print("clusterAddrTable is empty");
                 return;
             }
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Correct fix — iterating only the requested cluster and skipping master-less brokers prevents requests from being routed through the name server channel. Good test coverage for both scenarios.

---

PR #11133, opened 2026-09-11
fix(tools): fail fast on invalid configType in rocksDBConfigToJson command

zjncs:
### Motivation

`mqadmin rocksDBConfigToJson` with a bad `-t` value printed an "Invalid configType" message and returned `null` from `getConfigTypeList`, but execution then continued with the null list: local mode threw a raw `NullPointerException` from `Objects.requireNonNull`, and rpc mode carried the null list into the export request sent to brokers.

### Modifications

Throw a `SubCommandException` carrying the same message so the command aborts cleanly on an invalid `-t` value instead of failing later with an unrelated error.

### Verification

```
Tests run: 1, Failures: 1, Errors: 0 -- RocksDBConfigToJsonCommandTest#testInvalidConfigTypeFailsWithCleanError
java.lang.AssertionError: invalid configType must not surface as a NullPointerException
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- RocksDBConfigToJsonCommandTest
```

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/metadata/RocksDBConfigToJsonCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/metadata/RocksDBConfigToJsonCommand.java
@@ -43,7 +43,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Clean fail-fast fix — throwing `SubCommandException` instead of returning null prevents the confusing NullPointerException downstream. The test explicitly verifies the NPE path is eliminated.

---

PR #11134, opened 2026-09-11
fix(tools): reject clusterSendMsgRT amount below two instead of printing NaN

zjncs:
### Motivation

`mqadmin clusterSendMsgRT -a 1` cannot measure anything: the only send is the warm-up sample, elapsed stays 0, and the RT is computed as 0/0 = `NaN`. In tlog mode `new BigDecimal(NaN)` then throws `NumberFormatException` on the very first round, aborting the whole monitoring command.

### Modifications

Reject an amount below two up front with a clear usage message, since at least two messages are required for a measured RT.

### Verification

```
ClusterSendMsgRTCommandTest#testAmountBelowTwoIsRejected
java.lang.AssertionError: ClusterSendMsgRTCommand command failed ; Infinite or NaN
(warm-up-only run computed RT = 0/0 = NaN and BigDecimal rejected it on the first round)
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- ClusterSendMsgRTCommandTest
```

### DIFF
--- a/tools/src/main/java/org/apache/rocketmq/tools/command/cluster/ClusterSendMsgRTCommand.java
+++ b/tools/src/main/java/org/apache/rocketmq/tools/command/cluster/ClusterSendMsgRTCommand.java
@@ -100,6 +100,9 @@ public class ClusterSendMsgRTCommand implements SubCommand {
 
             long amount = !commandLine.hasOption('a') ? 100 : Long.parseLong(commandLine
                 .getOptionValue('a').trim());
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
LGTM. Good defensive fix — rejecting amount < 2 up front prevents the NaN/NumberFormatException that would otherwise abort the monitoring command. Clear error message explains why.

---

PR #11135, opened 2026-09-11
fix(tools): stop accumulating retryQueueNums across masters in exportMetadata

zjncs:
### Motivation

`mqadmin exportMetadata` summed `retryQueueNums` of the same subscription group over every master of the cluster, the way topic queue numbers are totaled. But a subscription group config is cluster-wide: a group exported from an N-broker cluster came out with N times its real `retryQueueNums`, and re-importing that metadata inflated every group's retry queues.

### Modifications

Keep the first seen config per group instead of summing, matching the cluster-wide semantics of subscription group config.

### Verification

```
ExportMetadataCommandTest#testSubscriptionGroupRetryQueueNumsNotAccumulatedAcrossBrokers
java.lang.AssertionError: retryQueueNums must not be summed across the two masters expected:<1> but was:<2>
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- ExportMetadataCommandTest
```


RockteMQ-AI (approved):
LGTM. Correct fix — subscription group config is cluster-wide and should not be accumulated across brokers. The `putIfAbsent` approach is clean and the test with two broker mockers validates the fix.

---

PR #11136, opened 2026-09-11
fix(remoting): include brokerId in SubscriptionGroupConfig.equals

zjncs:
### Motivation

`brokerId` participates in `SubscriptionGroupConfig.hashCode` but was missing from `equals`, so two configs for the same group that differ only in `brokerId` compared equal while hashing differently — breaking the `Object` contract for any `HashSet`/`HashMap` keyed on `SubscriptionGroupConfig` (lookups miss, set dedup misbehaves).

### Modifications

Append `brokerId` to `equals` to match `hashCode`.

### Verification

```
Tests run: 1, Failures: 1, Errors: 0 -- SubscriptionGroupConfigTest#testBrokerIdParticipatesInEqualsAndHashCode
java.lang.AssertionError: Values should be different. Actual: SubscriptionGroupConfig{groupName=groupA, ..., brokerId=1, ...}
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- SubscriptionGroupConfigTest
```


RockteMQ-AI (approved):
LGTM. Correct fix for the equals/hashCode contract violation. The test clearly demonstrates the bug (fail-before) and the fix (pass-after).

---

PR #11137, opened 2026-09-11
fix(common): do not throw on endpoint without dot in parseInstanceIdFromEndpoint

zjncs:
### Motivation

`parseInstanceIdFromEndpoint` used the result of `endpoint.indexOf('.')` directly as the substring end index:

```java
return endpoint.substring(0, endpoint.indexOf('.'));
```

so any non-empty endpoint without a dot (for example `MQ_INST_123456789_BXXUzaee:80`) produced `-1` and threw `StringIndexOutOfBoundsException` from a public utility.

### Modifications

Treat a missing dot as a non-instance endpoint and return null.

### Verification

```
Tests run: 4, Failures: 0, Errors: 1, Skipped: 1 -- NameServerAddressUtilsTest#testParseInstanceIdFromEndpointWithoutDot
java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 29
```

Pass-after:

```
Tests run: 4, Failures: 0, Errors: 0, Skipped: 1 -- NameServerAddressUtilsTest
```

### DIFF
--- a/common/src/main/java/org/apache/rocketmq/common/utils/NameServerAddressUtils.java
+++ b/common/src/main/java/org/apache/rocketmq/common/utils/NameServerAddressUtils.java
@@ -35,7 +35,11 @@ public class NameServerAddressUtils {
         if (StringUtils.isEmpty(endpoint)) {
             return null;
         }
-        return endpoint.substring(endpoint.lastIndexOf("/") + 1, endpoint.indexOf('.'));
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix `parseInstanceIdFromEndpoint` to handle endpoints without a dot, returning null instead of throwing `StringIndexOutOfBoundsException`.

LGTM — clean defensive fix with test.

---

PR #11138, opened 2026-09-11
fix(proxy): return UNRECOGNIZED_CLIENT_TYPE when client settings is null in receiveMessage

zjncs:
### Motivation

A gRPC client that calls `ReceiveMessage` before any settings were cached (or after its settings were cleaned up) made `getClientSettings` return null, and the following `settings.getClientType()` threw a `NullPointerException` which the catch block turned into an opaque `INTERNAL_SERVER_ERROR` response.

`ClientActivity.heartbeat` already guards this case with `UNRECOGNIZED_CLIENT_TYPE`.

### Modifications

Apply the same guard in `receiveMessage` so the client gets a meaningful code.

### Verification

```
ReceiveMessageActivityTest#testReceiveMessageWithoutClientSettings
java.lang.AssertionError: expected:<UNRECOGNIZED_CLIENT_TYPE> but was:<INTERNAL_SERVER_ERROR>
  ReceiveMessageActivityTest.testReceiveMessageWithoutClientSettings:447
(root cause: NullPointerException on "Settings.getClientType()" because "settings" is null)
```

Pass-after:

```
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0 -- ReceiveMessageActivityTest
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/ReceiveMessageActivity.java
@@ -64,6 +64,10 @@ public class ReceiveMessageActivity extends AbstractMessagingActivity {
 
         try {
             Settings settings = this.grpcClientSettingsManager.getClientSettings(ctx);
+            if (settings == null) {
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Add null check for client settings in `receiveMessage`, returning `UNRECOGNIZED_CLIENT_TYPE` instead of NPE when settings are not found.

LGTM — proper error handling for missing client configuration.

---

PR #11140, opened 2026-09-11
fix(proxy): clamp pop result filter maxAttempts to at least one attempt

zjncs:
### Motivation

`receiveMessage` feeds `settings.getBackoffPolicy().getMaxAttempts()` into `PopMessageResultFilterImpl`. When a client leaves the backoff policy unset, proto3 defaults `maxAttempts` to 0, and while the subscription group config is unavailable (`mergeSubscriptionData` passes the raw client settings through on a config miss) every popped message — including its first delivery — hit `reconsumeTimes >= maxAttempts` and was routed straight to the DLQ.

### Modifications

Clamp `maxAttempts` to at least 1 in the filter, so a message always gets one delivery attempt before the DLQ verdict.

### Verification

```
PopMessageResultFilterImplTest#testZeroMaxAttemptsStillDeliversFirstAttempt
java.lang.AssertionError: expected:<MATCH> but was:<TO_DLQ>
  PopMessageResultFilterImplTest.testZeroMaxAttemptsStillDeliversFirstAttempt:47
```

Pass-after:

```
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0 -- PopMessageResultFilterImplTest
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/PopMessageResultFilterImpl.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/PopMessageResultFilterImpl.java
@@ -27,7 +27,9 @@ public class PopMessageResultFilterImpl implements PopMessageResultFilter {
     private final int maxAttempts;
 
     public PopMessageResultFilterImpl(int maxAttempts) {
-        this.maxAttempts = maxAttempts;
+        // a client that leaves the backoff policy unset (proto3 default 0) must
+        // still get at least one delivery attempt before messages go to the DLQ
+        this.maxAttempts = Math.max(1, maxAttempts);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Clamp `maxAttempts` to at least 1 in `PopMessageResultFilterImpl`, ensuring that clients with proto3 default (0) backoff policy still get at least one delivery attempt before messages go to the DLQ.

LGTM — correct proto3 edge case handling with thorough test.

---

proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/consumer/PopMessageResultFilterImpl.java
```diff
@@ -27,7 +27,9 @@ public class PopMessageResultFilterImpl implements PopMessageResultFilter {
     private final int maxAttempts;
 
     public PopMessageResultFilterImpl(int maxAttempts) {
-        this.maxAttempts = maxAttempts;
+        // a client that leaves the backoff policy unset (proto3 default 0) must
+        // still get at least one delivery attempt before messages go to the DLQ
+        this.maxAttempts = Math.max(1, maxAttempts);
```
lwclover: Could we set the default to 1 when building SubscriptionData instead? This would keep the invariant maxAttempts >= 1 for the internal model and avoid putting configuration normalization into the filter implementation.

PR #11141, opened 2026-09-11
fix(proxy): guard null context in telemetry handleGrpcCancel

zjncs:
### Motivation

The telemetry stream observer keeps `proxyCtx = null` until the first `onNext` delivers a context. A client that opens the telemetry stream and then cancels (or drops the connection) without sending a single `TelemetryCommand` made `onError` call `handleGrpcCancel` with a null context, which threw a `NullPointerException` in `ctx.getClientID()` and masked the intended cancel handling.

### Modifications

Return early when the context is absent; there is no client id to clean up in that case.

### Verification

```
ClientActivityTest#testTelemetryOnErrorBeforeAnyMessage
java.lang.NullPointerException: Cannot invoke "org.apache.rocketmq.proxy.common.ProxyContext.getClientID()" because "ctx" is null
  ClientActivityTest.testTelemetryOnErrorBeforeAnyMessage:439
```

Pass-after:

```
Tests run: 16, Failures: 0, Errors: 0, Skipped: 0 -- ClientActivityTest
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/client/ClientActivity.java
@@ -330,6 +330,9 @@ public class ClientActivity extends AbstractMessagingActivity {
     }
 
     private void handleGrpcCancel(ProxyContext ctx, Throwable t) {
+        if (ctx == null) {
+            return;
+        }
         final String clientId = ctx.getClientID();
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Adds a null check for `ctx` in `handleGrpcCancel` to prevent NPE when the gRPC context is already cleared. Test verifies no exception is thrown.

LGTM — simple defensive null guard.

---

PR #11142, opened 2026-09-11
fix(broker): stop overwriting batch ack uniq key with single ack key in appendAck

zjncs:
### Motivation

`appendAck` set `PROPERTY_UNIQ_CLIENT_MESSAGE_ID_KEYIDX` correctly per ack type (`genBatchAckUniqueId` for `BatchAckMsg`, `genAckUniqueId` otherwise), but a later unconditional `put` overwrote it with `genAckUniqueId` for every ack. Batch acks therefore landed on the revive topic with a bogus uniq key: the offset segment was the `-1` sentinel the batch path assigns, and the tag segment was `ack` instead of `bAck`, so every batch ack of a pop produced the same non-unique client id and tracing by uniq key was impossible.

The buffered path in `PopBufferMergeService` already writes the batch uniq key without overwriting it.

### Modifications

Drop the stray `put` so the per-type key set just above survives.

### Verification

```
AckMessageProcessorTest#testBatchAck_appendAck_BatchUniqKeyKept
expected: "FooBar@0@[101]@DEFAULT_CONSUMER@1666860736757@bAck"
 but was: "FooBar@0@-1@DEFAULT_CONSUMER@1666860736757@broker-a@ack"
```

Pass-after:

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- AckMessageProcessorTest
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/AckMessageProcessor.java
@@ -298,7 +298,6 @@ public class AckMessageProcessor implements NettyRequestProcessor {
         msgInner.setBornHost(this.brokerController.getStoreHost());
         msgInner.setStoreHost(this.brokerController.getStoreHost());
         msgInner.setDeliverTimeMs(popTime + invisibleTime);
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fixes a bug where `appendAck` was overwriting the batch ack unique key with the single ack key. Removes the erroneous overwrite line. Test uses `ArgumentCaptor` to verify the correct key is retained.

LGTM — correct fix with proper test verification.

---

PR #11144, opened 2026-09-11
fix(broker): reject pop requests with non-positive maxMsgNums

zjncs:
### Motivation

`processRequest` only validated `maxMsgNums > 32`. A request with `maxMsgNums = 0` fell through to the polling logic with a zero message budget instead of being rejected, and a negative value made `new GetMessageResult` pass the negative size to `new ArrayList`, throwing `IllegalArgumentException` from deep inside the pop path instead of a clean `INVALID_PARAMETER` response.

### Modifications

Validate `maxMsgNums >= 1` next to the existing upper-bound check.

### Verification

```
PopMessageProcessorTest#testProcessRequest_IllegalMaxMsgNums
java.lang.NullPointerException: Cannot invoke "MessageStoreConfig.isTimerWheelEnable()" because
"MessageStore.getMessageStoreConfig()" is null -- PopMessageProcessorTest.testProcessRequest_IllegalMaxMsgNums:261
(the maxMsgNums=0 request was not rejected and fell through into the polling path)
```

Pass-after:

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- PopMessageProcessorTest
```

### DIFF
--- a/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
+++ b/broker/src/main/java/org/apache/rocketmq/broker/processor/PopMessageProcessor.java
@@ -264,6 +264,13 @@ public class PopMessageProcessor implements NettyRequestProcessor {
             return response;
         }
 
+        if (requestHeader.getMaxMsgNums() <= 0) {
+            response.setCode(ResponseCode.INVALID_PARAMETER);
+            response.setRemark(String.format("the broker[%s] pop message's num must be positive",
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Input validation fix: reject pop requests with non-positive `maxMsgNums` before processing. Simple guard clause with test coverage.

LGTM — clean defensive check.

---

PR #11145, opened 2026-09-11
fix(tools): report unknown cluster as error in exportMetrics command

zjncs:
### Motivation

A typo in the required `-c` option of `mqadmin exportMetrics` made `clusterAddrTable.get(clusterName)` return null, and the following for-each threw a raw `NullPointerException` from inside the command, producing a stack trace instead of an error message.

### Modifications

Print a `cluster not exist` error like the sibling cluster commands do.

### Verification

```
ExportMetricsCommandTest#testUnknownClusterReportsCleanError
java.lang.NullPointerException: Cannot invoke "java.util.Set.iterator()" because "brokerNameSet" is null
```

Pass-after:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- ExportMetricsCommandTest
```


RockteMQ-AI (approved):
### Summary
Add null check for `brokerNameSet` in `exportMetrics` when the requested cluster does not exist, printing a clear error message instead of NPE.

LGTM — defensive null guard with test.

---

PR #11146, opened 2026-09-11
fix(client): clear reachable flag when the latency fault detector observes a failure

zjncs:
### Motivation

`LatencyFaultToleranceImpl.detectByOneRound` only flipped `FaultItem.reachableFlag` from false to true. Once a broker had been marked reachable, a failing service detector was ignored, so `MQFaultStrategy`'s reachable filter kept selecting a broker that had gone down until a real send happened to fail against it.

### Modifications

Reset the flag symmetrically when the detector reports the broker unreachable, so the next broker-selection round sees the failure.

### Verification

```
Tests run: 3, Failures: 1, Errors: 0, Skipped: 1 -- LatencyFaultToleranceImplTest#testDetectByOneRoundClearsReachableFlagWhenDetectFails
```

Pass-after:

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 1 -- LatencyFaultToleranceImplTest
```


RockteMQ-AI (approved):
### Summary
Fix the latency fault detector to clear the `reachableFlag` when a broker becomes unreachable. Previously, the flag was only set to `true` on recovery but never cleared on failure, causing stale reachability state.

LGTM — important fix for client-side load balancing correctness.

---

PR #11148, opened 2026-09-11
fix(remoting): report version 0 for consumer groups without connections

zjncs:
### Motivation

`ConsumerConnection.computeMinVersion` kept its `Integer.MAX_VALUE` sentinel when the connection set was empty, so `mqadmin consumerProgress` printed `HighestVersion (2147483647)` for idle consumer groups whose clients had all disconnected.

### Modifications

Return 0 for an empty set instead.

### Verification

```
ConsumerConnectionTest#testComputeMinVersionWithNoConnection
java.lang.AssertionError: expected: 0 but was: 2147483647
```

Pass-after:

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0 -- ConsumerConnectionTest
```


RockteMQ-AI (approved):
### Summary
Fix `computeMinVersion` to return 0 when the connection set is empty, instead of returning `Integer.MAX_VALUE`. A consumer group with no live connections should not be reported as having the highest version.

LGTM — clean edge case fix with test.

---

PR #11149, opened 2026-09-11
fix(proxy): offline all lite subscriptions and guard empty subscription list

zjncs:
### Motivation

`GrpcClientSettingsManager.offlineClientLiteSubscription` only took `subscriptions(0)` to build the `COMPLETE_REMOVE` LiteSubscriptionDTO:

1. A lite consumer subscribing to multiple topics left stale subscriptions on the broker for every topic after the first one.
2. When the subscription list was empty, `getSubscriptions(0)` threw `IndexOutOfBoundsException`, which the surrounding catch block swallowed — so the cleanup was silently skipped entirely, and the stale lite subscription could trigger `LITE_SUBSCRIPTION_QUOTA_EXCEEDED` on the next connect.

### Modifications

Iterate over all subscription entries and send one `COMPLETE_REMOVE` per topic; an empty list now simply sends nothing.

### Verification

Fail-before (new tests, run against the unpatched code):

```
Wanted 2 times but was 1 -- GrpcClientSettingsManagerTest#testOfflineClientLiteSubscription_MultipleTopics_AllSubscriptionsRemoved
(multiplex: syncLiteSubscription invoked for only the first of two subscribed topics)
```

Pass-after:

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- GrpcClientSettingsManagerTest
```

### DIFF
--- a/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManager.java
+++ b/proxy/src/main/java/org/apache/rocketmq/proxy/grpc/v2/common/GrpcClientSettingsManager.java
@@ -25,6 +25,8 @@ import apache.rocketmq.v2.Endpoints;
 import apache.rocketmq.v2.ExponentialBackoff;
 import apache.rocketmq.v2.Metric;
 import apache.rocketmq.v2.Settings;
+import apache.rocketmq.v2.Subscription;
+import apache.rocketmq.v2.SubscriptionEntry;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix `offlineClientLiteSubscription` to iterate over all subscription entries instead of only processing the first one. Also guards against empty subscription lists.

LGTM — correct fix for multi-topic lite subscription cleanup.

---

PR #11150, opened 2026-09-11
fix(store): estimate batch consume queue count up to max offset

zjncs:
### Motivation

`estimateMessageCount` translated the exclusive upper bound `to` with `getBatchMsgIndexBuffer(to)`, which returns null when `to` equals `maxOffsetInQueue`. Every accumulation query that counts up to the queue head therefore hit the null branch and returned -1, and callers such as `DefaultMessageStore` silently fell back to the raw offset difference, ignoring batch sizes and tag filters.

### Modifications

Since `to` is exclusive, look up the unit containing `to - 1` instead, and advance `physicalOffsetTo` past that unit so the scan covers it. Interior ranges are unaffected because a `to` that lands on a batch base resolves to the same mapped position as before.

### Verification

```
Tests run: 6, Failures: 1, Errors: 0 -- BatchConsumeQueueTest#testEstimateMessageCount
java.lang.AssertionError: expected:<1000> but was:<-1>
  BatchConsumeQueueTest.testEstimateMessageCount:285 -- estimateMessageCount(1, 1001) with to == maxOffsetInQueue
```

Pass-after:

```
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 -- BatchConsumeQueueTest
```

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/queue/BatchConsumeQueue.java
+++ b/store/src/main/java/org/apache/rocketmq/store/queue/BatchConsumeQueue.java
@@ -1107,11 +1107,14 @@ public class BatchConsumeQueue implements ConsumeQueueInterface {
         }
         long physicalOffsetFrom = firstMappedFileBuffer.getStartOffset();
 
-        SelectMappedBufferResult lastMappedFileBuffer = getBatchMsgIndexBuffer(to);
+        // 'to' is an exclusive upper bound, so the last unit in range is the one
+        // containing 'to - 1'; getBatchMsgIndexBuffer returns null for offsets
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary
Fix `estimateMessageCount` in `BatchConsumeQueue` — the exclusive upper bound `to` was used directly as an offset, but `getBatchMsgIndexBuffer` returns null for offsets >= maxOffsetInQueue, causing counting up to the queue head to fail. Correctly uses `to - 1` and adds `CQ_STORE_UNIT_SIZE` to get the physical end offset. Includes test coverage.

LGTM — focused fix with clear rationale in comments.

---

PR #11152, opened 2026-09-11
[ISSUE #11151] Coordinate graceful shutdown of remoting sub-servers

### LINKED ISSUE #11151: [Bug] Remoting sub-servers do not honor graceful shutdown

macOS aarch64, JDK 11.

### RocketMQ version

`develop`, commit `d38f81e787`.

### Describe the Bug

Sub-servers created by `NettyRemotingServer.newRemotingServer(port)` do not honor the parent's graceful-shutdown configuration. `SubRemotingServer.shutdown()` sets its shutdown flag and immediately closes its listening channel. During the configured grace period, new connections therefore fail instead of receiving `GO_AWAY` responses for supported client versions.

Calling only the parent's `shutdown()` has another gap: it marks only the parent as shutting down, so sub-servers can continue processing requests until the shared event loops shut down. Concurrent shutdown calls also need coordination: a repeated call must not bypass the first call's grace period and close the listener or shared resources early.

### Steps to Reproduce

1. Start a `NettyRemotingServer` with `enableShutdownGracefully=true` and `shutdownWaitTimeSeconds=5`.
2. Create and start sub-servers on other ports, and register a request processor on each server.
3. Establish a connection and send a request using a protocol version newer than `V5_3_1`.
4. Call a sub-server's `shutdown()` on another thread. During the grace period, send another request over the existing connection and try a fresh connection to that port.
5. Separately, call only the parent server's `shutdown()` and send requests to the sub-server ports during the grace period.

A standalone sub-server shutdown keeps its listener available during the configured grace period and uses the existing `GO_AWAY` handling, without shutting down other ports or shared resources.

Each sub-server measures its own grace period from when it first enters the draining state. Repeated shutdown calls must not reset that deadline. A parent shutdown marks all its sub-servers as draining and waits until its own and all child grace periods have elapsed before releasing shared resources. Main/sub shutdowns and repeated calls may run concurrently; the independent grace periods must overlap rather than being waited one after another.

### What Did You See Instead?

The standalone sub-server closes its listener immediately. Parent-only shutdown leaves sub-servers outside the draining state. A regression test on the unmodified commit times out waiting for a child to enter that state after parent shutdown starts.

qianye1001:

Fixes #11151.

### Brief Description

A sub-server now honors the existing graceful-shutdown configuration: it enters the draining state, keeps its listener open for its own grace period, and then closes that listener. Requests continue to use the existing `GO_AWAY` handling and client-version condition. Other ports and shared executors remain available when a child shuts down independently.

Parent shutdown starts draining all children before waiting. Each server records its own deadline on its first transition to draining; an already draining child retains its original deadline. The parent waits for the latest deadline before closing child listeners and releasing shared resources, so the grace periods overlap instead of accumulating sequentially. An atomic compare-and-set records each child's first deadline, including when main and child shutdown run concurrently. Repeated shutdown calls return immediately without repeating cleanup or resetting the deadline. Coordination uses only atomic state and adds no executor or timer thread.

- JDK 11: 14 focused remoting server/config/lifecycle tests passed, with Checkstyle and SpotBugs enabled.
- JDK 21: all 6 new plaintext-TCP shutdown tests passed (`jacoco.skip=true`, `spotbugs.skip=true`; static analysis was run under JDK 11).
- The new real-TCP tests cover existing and fresh connections during draining, independent child shutdown and unaffected sibling/parent ports, parent-only shutdown of multiple children, both parent/child shutdown orders, simultaneous and repeated calls, a pre-existing child deadline longer than the parent's wait, disabled graceful shutdown, older client versions, and interruption.
- Listener closure is verified through the actual channel close future, avoiding platform-specific connection-refusal versus connection-timeout behavior.
- On the unmodified base, the parent-shutdown regression test fails because children never enter the draining state.

### CI test stabilization

- `TlsTest.disabledServerRejectsSSLClient` accepts either connection or send failure: the server rejects TLS by closing the connection, which may happen before the client checks channel activity or during its write. TLS tests use the loopback IP to avoid hostname resolution.
- `TimerMessageStoreTest.testTimerFlowControl` waits for each accepted message to reach the timer wheel before testing the next admission, replacing a fixed 5 ms sleep. This preserves the flow-control assertions without assuming the asynchronous reput/enqueue work has already completed.
- JDK 11: all 13 TLS tests and the timer flow-control test passed with Checkstyle and SpotBugs enabled.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
@@ -80,9 +80,9 @@ import org.apache.rocketmq.remoting.protocol.RemotingCommand;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.security.cert.CertificateException;
-import java.time.Duration;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
@@ -90,6 +90,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 public class NettyRemotingServer extends NettyRemotingAbstract implements RemotingServer {
     private static final Logger log = LoggerFactory.getLogger(LoggerName.ROCKETMQ_REMOTING_NAME);
@@ -107,6 +109,11 @@ public class NettyRemotingServer extends NettyRemotingAbstract implements Remoti
     private final HashedWheelTimer timer = new HashedWheelTimer(r -> new Thread(r, "ServerHouseKeepingService"));
 
     private DefaultEventExecutorGroup defaultEventExecutorGroup;
+
+    // Only protects the transition to draining; never held while waiting or closing channels.
+    private final ReentrantLock shutdownLock = new ReentrantLock();
+    private final AtomicBoolean shutdownStarted = new AtomicBoolean();
+    private final CompletableFuture<Void> shutdownComplete = new CompletableFuture<>();
 
     /**
      * NettyRemotingServer may hold multiple SubRemotingServer, each server will be stored in this container with a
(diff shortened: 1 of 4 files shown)


RockteMQ-AI (approved):
### Summary
This PR coordinates graceful shutdown of remoting sub-servers so that parent and child shutdown grace periods overlap instead of accumulating sequentially. The implementation uses `CompletableFuture` for concurrent shutdown tracking and a `ReentrantLock` for state transitions only — waiting and channel closure happen outside the lock. Well-structured with comprehensive test coverage (270 lines of tests covering edge cases like pre-existing deadlines, disabled graceful shutdown, older client versions, and interruption).

LGTM — clean concurrency design, good test coverage, and the PR description clearly explains the rationale.

---

RockteMQ-AI (approved):
### Summary
This PR coordinates graceful shutdown across the parent `NettyRemotingServer` and its `SubRemotingServer` instances. The refactor replaces the sequential drain-and-close pattern with a concurrent approach: all sub-servers begin draining in parallel, the parent waits for the latest deadline, then closes listeners and releases shared resources.

The design is solid — `AtomicBoolean` guards ensure idempotent shutdown, `AtomicReference<Long>` for the deadline allows either caller (parent or child) to set it first, and interruption handling correctly preserves the interrupt flag. The 6 test methods comprehensively cover concurrent shutdown, disabled graceful mode, interrupted threads, and child-before-parent ordering.

LGTM.

---

RockteMQ-AI (approved):
LGTM. Well-designed coordination of graceful shutdown across parent and sub-servers. The atomic state management correctly handles concurrent shutdown scenarios, and the deadline overlap logic avoids sequential accumulation. Comprehensive test coverage including edge cases (pre-existing child deadline, disabled graceful shutdown, interruption handling).

---

remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingServer.java
```diff
@@ -108,6 +109,8 @@ public class NettyRemotingServer extends NettyRemotingAbstract implements Remoti
 
     private DefaultEventExecutorGroup defaultEventExecutorGroup;
 
+    private final AtomicBoolean shutdownStarted = new AtomicBoolean();
```
lizhimins, line 112: 太复杂啦

PR #11154, opened 2026-09-11
[ISSUE #11153] Remove dedicated NameServer probe workers

### LINKED ISSUE #11153: [Enhancement] Remove the dedicated NameServer probe thread pool
### Description

`NettyRemotingClient` allocates a dedicated `scanExecutor` for NameServer availability probes: 4 core threads, up to 10 threads, and a queue of 32 tasks. The existing housekeeping timer submits a task per address to that pool. Each task calls `getAndCreateChannel()`, which waits for the Netty connection future to complete.

These probes only establish or reuse connections. Netty already provides asynchronous connection completion, so the extra worker pool and task queue are unnecessary.

### Expected behavior

Initiate probes directly from the existing housekeeping timer and update the available NameServer list in connection-completion callbacks. Remove the scan executor entirely, without replacing it with another executor or timer thread.

The timer must not wait for connections or a contended channel-table lock. Pending or failed connections must not be advertised as available, and late callbacks must not restore removed addresses or overwrite newer connection results. Preserve the existing `scanAvailableNameSrv` switch and probe interval.

qianye1001:

Fixes #11153.

### Brief Description

`NettyRemotingClient` currently submits NameServer connection probes to a dedicated pool with 4 core threads, up to 10 workers and a queue of 32 tasks. Those workers wait for connection results that Netty already exposes asynchronously.

Initiate the probes directly from the existing housekeeping timer and update availability through connection-future listeners. Remove the scan executor entirely, including its construction and shutdown, eliminating `NettyClientScan_thread_` workers. The existing scan switch and interval are retained.

The scan uses a nonblocking attempt on the existing channel-table lock so that contention defers a probe to the next scan instead of blocking housekeeping. Pending or failed connections are not advertised as available, and callbacks check the current address list, connection and shutdown state before applying results. No new executor, timer or public configuration is added.

JDK 11, with Checkstyle and SpotBugs enabled:

```shell
mvn -B -pl remoting -am -Dmaven.gitcommitid.skip=true \
  -Dtest=NettyRemotingClientNameServerTest,NettyRemotingClientTest \
```

All 26 tests passed (10 new probe tests and 16 existing client tests). Coverage includes pending connections without blocking housekeeping, connection failure and retry, address removal, replacement connections, channel-lock contention, the disabled scan switch, inactive channels and shutdown. A real TCP test verifies availability after connecting and removal after the server stops.

### DIFF
--- a/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
+++ b/remoting/src/main/java/org/apache/rocketmq/remoting/netty/NettyRemotingClient.java
@@ -24,6 +24,7 @@ import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
@@ -52,7 +53,6 @@ import org.apache.rocketmq.common.ThreadFactoryImpl;
 import org.apache.rocketmq.common.constant.LoggerName;
 import org.apache.rocketmq.common.utils.FutureUtils;
 import org.apache.rocketmq.common.utils.NetworkUtil;
-import org.apache.rocketmq.common.utils.ThreadUtils;
 import org.apache.rocketmq.logging.org.slf4j.Logger;
 import org.apache.rocketmq.logging.org.slf4j.LoggerFactory;
 import org.apache.rocketmq.remoting.ChannelEventListener;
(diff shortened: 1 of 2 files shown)


RockteMQ-AI (approved):
### Summary

This PR removes the dedicated scanExecutor thread pool used for NameServer connection probes and replaces it with an async approach using the existing housekeeping timer and Netty ChannelFutureListener. This is a solid performance optimization that eliminates unnecessary thread blocking.

**Key improvements:**
- Eliminates a dedicated thread pool (4 core, 10 max threads) for better resource utilization
- Non-blocking probe initiation from housekeeping timer
- Proper async handling with connection-future listeners
- Comprehensive test coverage (10 test cases) covering edge cases

**Code quality:**
- Clean async implementation with proper shutdown guards
- Handles lock contention gracefully (skips probe if lock unavailable)
- Prevents stale updates by checking channel wrapper identity
- Excellent test suite covering async behavior, shutdown, replacements, and integration

LGTM. Well-implemented optimization with thorough testing.

---

PR #11157, opened 2026-09-12
[ISSUE #11156]fix(timer): persist TimelineRollService checkpoint to avoid repeated and loss

### LINKED ISSUE #11156: [Bug] TimelineRollService may repeatedly roll the same timer messages

all environment

### RocketMQ version

develop

### JDK Version

1.8

### Describe the Bug

`TimelineRollService` used a fixed sleep plus a scan window of `[now + rollRange, now + rollRange + timerMaxDelaySec]`.

Two issues follow from that:

1. The window is much larger than the interval, so the same not-yet-expired timer message stays in range and is rolled back to `TIMER_TOPIC` many times.
2. The next window is computed from `System.currentTimeMillis()` after sleep. If a scan is delayed, the next round can skip or rescan the same delayTime range. The progress was also only in memory, so a restart could roll the same messages again.

### Steps to Reproduce

Enable the rocksdb scheduled messages, send a scheduled message, and you can change the roll interval to execute every minute. This way, you can verify more quickly.

Set a reasonable roll interval and scan range, and simultaneously record checkpoints to ensure that messages are not lost and no invalid duplicate deliveries occur.

### What Did You See Instead?

Drive roll by a persisted RocksDB checkpoint (`timeline_roll_checkpoint`) instead of a fixed sleep:

- Each round scans `[checkpoint, checkpoint + interval)`.
- After a successful scan, advance and persist the checkpoint.
- The next due time is `checkpoint + interval - timerMaxDelaySec`. Trigger 1s early; if it is not due yet, poll with at most 1s wait.
- If the service is behind, scan the next window immediately so delayed work does not leave a gap.

## Test plan

- [ ] `MessageRocksDBStorageTest#testWriteAndGetRollCheckpoint`
- [ ] `MessageRocksDBStorageTest#testScanAdjacentWindowsNoOverlap`
- [ ] Send a long-delay timer message (e.g. 3h, `timerMaxDelaySec` a bit larger) and confirm it is rolled at most once
- [ ] Slow down or restart broker during roll and confirm the next scan continues from the checkpoint without skipping or duplicating the previous window

### Additional Context

_No response_

3424672656:

- Fixes #11156 

## What is the problem?

`TimelineRollService` used a fixed sleep plus a scan window of `[now + rollRange, now + rollRange + timerMaxDelaySec]`.

Two issues follow from that:

1. The window is much larger than the interval, so the same not-yet-expired timer message stays in range and is rolled back to `TIMER_TOPIC` many times.
2. The next window is computed from `System.currentTimeMillis()` after sleep. If a scan is delayed, the next round can skip or rescan the same delayTime range. The progress was also only in memory, so a restart could roll the same messages again.

## What is the solution?

Drive roll by a persisted RocksDB checkpoint (`timeline_roll_checkpoint`) instead of a fixed sleep:

- Each round scans `[checkpoint, checkpoint + interval)`.
- After a successful scan, advance and persist the checkpoint.
- The next due time is `checkpoint + interval - timerMaxDelaySec`. Trigger 1s early; if it is not due yet, poll with at most 1s wait.
- If the service is behind, scan the next window immediately so delayed work does not leave a gap.

## Test plan

- [x] `MessageRocksDBStorageTest#testWriteAndGetRollCheckpoint`
- [x] `MessageRocksDBStorageTest#testScanAdjacentWindowsNoOverlap`
- [x] Send a long-delay timer message (e.g. 3h, `timerMaxDelaySec` a bit larger) and confirm it is rolled at most once
- [x] Slow down or restart broker during roll and confirm the next scan continues from the checkpoint without skipping or duplicating the previous window

### DIFF
--- a/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
@@ -387,8 +387,8 @@ public class Timeline {
             while (!this.isStopped()) {
                 try {
                     long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
-                    int rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours() > 0 ? storeConfig.getTimerRocksDBRollIntervalHours() : 1;
-                    long rangeMs = TimeUnit.HOURS.toMillis(rollIntervalHour);
+                    int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
+                    long rangeMs = TimeUnit.HOURS.toMillis(rollRangeHour);
                     long nextDueMs = checkpoint + rangeMs - maxDelayMs;
                     long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
                     long now = System.currentTimeMillis();
@@ -407,6 +407,7 @@ public class Timeline {
                     log.info("Timeline TimelineRollService roll records success, checkpoint: {}, cost: {}", checkpoint, System.currentTimeMillis() - now);
                 } catch (Exception e) {
                     logError.error("Timeline TimelineRollService failed error: {}", e.getMessage());
+                    this.waitForRunning(200L);
                 }
             }
             log.info(this.getServiceName() + " service end");
--- a/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
@@ -376,16 +376,19 @@ public class Timeline {
         @Override
         public void run() {
             long checkpoint = messageRocksDBStorage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT);
+            if (checkpoint <= 0L) {
+                long now = System.currentTimeMillis();
+                long forwardCheckpoint = messageRocksDBStorage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_CHECK_POINT);
+                int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
+                checkpoint = (forwardCheckpoint > 0L ? Math.min(forwardCheckpoint, now) : now)
+                    + TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec()) - TimeUnit.HOURS.toMillis(rollRangeHour);
+            }
             log.info(this.getServiceName() + " service start, checkpoint: {}", checkpoint);
             while (!this.isStopped()) {
                 try {
                     long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
                     int rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours() > 0 ? storeConfig.getTimerRocksDBRollIntervalHours() : 1;
-                    int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
                     long rangeMs = TimeUnit.HOURS.toMillis(rollIntervalHour);
-                    if (checkpoint <= 0L) {
-                        checkpoint = System.currentTimeMillis() + maxDelayMs - TimeUnit.HOURS.toMillis(rollRangeHour);
-                    }
                     long nextDueMs = checkpoint + rangeMs - maxDelayMs;
                     long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
                     long now = System.currentTimeMillis();
@@ -401,7 +404,6 @@ public class Timeline {
                         continue;
                     }
                     checkpoint += rangeMs;
-                    messageRocksDBStorage.writeCheckPointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT, checkpoint);
                     log.info("Timeline TimelineRollService roll records success, checkpoint: {}, cost: {}", checkpoint, System.currentTimeMillis() - now);
                 } catch (Exception e) {
                     logError.error("Timeline TimelineRollService failed error: {}", e.getMessage());
--- a/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
@@ -387,17 +387,15 @@ public class Timeline {
             while (!this.isStopped()) {
                 try {
                     long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
-                    int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
-                    long rangeMs = TimeUnit.HOURS.toMillis(rollRangeHour);
-                    long nextDueMs = checkpoint + rangeMs - maxDelayMs;
-                    long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
+                    long rangeMs = TimeUnit.HOURS.toMillis(storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2);
+                    long triggerAt = checkpoint + rangeMs - maxDelayMs - ROLL_TRIGGER_EARLY_MS;
                     long now = System.currentTimeMillis();
                     if (now < triggerAt) {
-                        this.waitForRunning(Math.min(triggerAt - now, ROLL_POLL_WHEN_NOT_DUE_MS));
+                        this.waitForRunning(ROLL_POLL_WHEN_NOT_DUE_MS);
                         continue;
                     }
 
-                    log.info("Timeline TimelineRollService start roll checkpoint: {}, rangeMs: {}, nextDueMs: {}, delayMs: {}", checkpoint, rangeMs, nextDueMs, now - nextDueMs);
+                    log.info("Timeline TimelineRollService start roll checkpoint: {}, rangeMs: {}, triggerAt: {}, delayMs: {}", checkpoint, rangeMs, triggerAt, now - triggerAt);
                     if (!scanRecordsToQueue(checkpoint, rangeMs, timerMessageRocksDBStore.getRollMessageQueue())) {
                         logError.error("Timeline TimelineRollService scanRecordsToQueue error, checkpoint: {}", checkpoint);
                         this.waitForRunning(200L);
--- a/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
@@ -49,6 +49,8 @@ public class Timeline {
     private static final String DELETE_KEY_SPLIT = "+";
     private static final int ORIGIN_CAPACITY = 100000;
     private static final int BATCH_SIZE = 1000, MAX_BATCH_SIZE_FROM_ROCKSDB = 8000;
+    private static final long ROLL_TRIGGER_EARLY_MS = 1000L;
+    private static final long ROLL_POLL_WHEN_NOT_DUE_MS = 1000L;
     private static final int INITIAL = 0, RUNNING = 1, SHUTDOWN = 2;
     private volatile int state = INITIAL;
     private final AtomicLong commitOffset = new AtomicLong(0);
@@ -373,35 +375,34 @@ public class Timeline {
 
         @Override
         public void run() {
-            log.info(this.getServiceName() + " service start");
+            long checkpoint = messageRocksDBStorage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT);
+            log.info(this.getServiceName() + " service start, checkpoint: {}", checkpoint);
             while (!this.isStopped()) {
-                int rollIntervalHour = 1;
-                int rollRangeHour = 2;
                 try {
-                    if (storeConfig.getTimerRocksDBRollIntervalHours() > 0) {
-                        rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours();
+                    long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
+                    int rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours() > 0 ? storeConfig.getTimerRocksDBRollIntervalHours() : 1;
+                    int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
+                    long rangeMs = TimeUnit.HOURS.toMillis(rollIntervalHour);
+                    if (checkpoint <= 0L) {
+                        checkpoint = System.currentTimeMillis() + maxDelayMs - TimeUnit.HOURS.toMillis(rollRangeHour);
                     }
-                    if (storeConfig.getTimerRocksDBRollRangeHours() > 0) {
-                        rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours();
+                    long nextDueMs = checkpoint + rangeMs - maxDelayMs;
+                    long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
+                    long now = System.currentTimeMillis();
+                    if (now < triggerAt) {
+                        this.waitForRunning(Math.min(triggerAt - now, ROLL_POLL_WHEN_NOT_DUE_MS));
+                        continue;
                     }
-                    this.waitForRunning(TimeUnit.HOURS.toMillis(rollIntervalHour));
-                    if (stopped) {
-                        log.info(this.getServiceName() + " service end");
-                        return;
+
+                    log.info("Timeline TimelineRollService start roll checkpoint: {}, rangeMs: {}, nextDueMs: {}, delayMs: {}", checkpoint, rangeMs, nextDueMs, now - nextDueMs);
+                    if (!scanRecordsToQueue(checkpoint, rangeMs, timerMessageRocksDBStore.getRollMessageQueue())) {
+                        logError.error("Timeline TimelineRollService scanRecordsToQueue error, checkpoint: {}", checkpoint);
+                        this.waitForRunning(200L);
+                        continue;
                     }
-                } catch (Exception e) {
-                    logError.error("Timeline TimelineRollService wait error: {}", e.getMessage());
-                }
-                long rollCheckpoint = System.currentTimeMillis();
-                try {
-                    log.info("Timeline TimelineRollService start roll rollCheckpoint: {}", rollCheckpoint);
-                    while (!scanRecordsToQueue(rollCheckpoint + TimeUnit.HOURS.toMillis(rollRangeHour),
-                            TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec()),
-                            timerMessageRocksDBStore.getRollMessageQueue())) {
-                        logError.error("Timeline TimelineRollService scanRecordsToQueue error.");
-                        Thread.sleep(200);
-                    }
-                    log.info("Timeline TimelineRollService roll records success, lastRollTime: {}, rollCheckpoint: {}, cost: {}", rollCheckpoint, rollCheckpoint, System.currentTimeMillis() - rollCheckpoint);
+                    checkpoint += rangeMs;
+                    messageRocksDBStorage.writeCheckPointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT, checkpoint);
+                    log.info("Timeline TimelineRollService roll records success, checkpoint: {}, cost: {}", checkpoint, System.currentTimeMillis() - now);
                 } catch (Exception e) {
                     logError.error("Timeline TimelineRollService failed error: {}", e.getMessage());
                 }
--- a/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/TimerMessageRocksDBStore.java
+++ b/store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/TimerMessageRocksDBStore.java
@@ -231,8 +231,8 @@ public class TimerMessageRocksDBStore {
             this.expiredMessageQueue = new LinkedBlockingDeque<>(TIME_UP_CAPACITY);
             this.rollMessageQueue = new LinkedBlockingDeque<>(ROLL_CAPACITY);
         }
-        this.expiredMessageReputService = new TimerMessageReputService(expiredMessageQueue, storeConfig.getTimerRocksDBTimeExpiredMaxTps(), true);
-        this.rollMessageReputService = new TimerMessageReputService(rollMessageQueue, storeConfig.getTimerRocksDBRollMaxTps(), false);
+        this.expiredMessageReputService = new TimerMessageReputService(expiredMessageQueue, storeConfig.getTimerRocksDBTimeExpiredMaxTps(), MessageRocksDBStorage.TIMELINE_CHECK_POINT);
+        this.rollMessageReputService = new TimerMessageReputService(rollMessageQueue, storeConfig.getTimerRocksDBRollMaxTps(), MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT);
         this.timeline = new Timeline(messageStore, messageRocksDBStorage, this, timerMetrics);
         this.timerSysTopicScanService = new TimerSysTopicScanService();
     }
@@ -506,7 +506,7 @@ public class TimerMessageRocksDBStore {
         private final Logger log = TimerMessageRocksDBStore.log;
         private final BlockingQueue<List<TimerRocksDBRecord>> queue;
         private final RateLimiter rateLimiter;
-        private final boolean writeCheckPoint;
+        private final byte[] checkPointKey;
         private final ExecutorService executor =
                 ThreadUtils.newThreadPoolExecutor(
                         storeConfig.getTimerReputServiceCorePoolSize(),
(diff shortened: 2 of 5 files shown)


RockteMQ-AI:
### 🔍 Code Review — PR #11157

bash: line 180: develop: command not found
bash: line 180: TimelineRollService: command not found
bash: line 180: [now: command not found
bash: line 180: TIMER_TOPIC: command not found
bash: command substitution: line 181: syntax error: unexpected end of file
bash: line 180: timeline_roll_checkpoint: command not found
bash: command substitution: line 180: syntax error near unexpected token `)'
bash: command substitution: line 180: `[checkpoint, checkpoint + interval)'
bash: line 180: checkpoint: command not found
bash: line 180: MessageRocksDBStorageTest#testWriteAndGetRollCheckpoint: command not found
bash: line 180: MessageRocksDBStorageTest#testScanAdjacentWindowsNoOverlap: command not found
bash: line 180: timerMaxDelaySec: command not found
**Overall**
The change correctly moves the roll progress from memory to a RocksDB checkpoint, which is the right direction for avoiding duplicate scans across restarts. However, the checkpoint is advanced **before** the rolled records are actually written back to the commitlog by the async `rollMessageReputService`, so a crash can skip a window that was queued but not yet consumed. I think that needs to be fixed before merging.

---

### 1. Correctness

**Persisting the checkpoint too early — likely data-loss/late-roll risk**
`store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java:390-392`

```java
if (!scanRecordsToQueue(checkpoint, rangeMs, timerMessageRocksDBStore.getRollMessageQueue())) {
    ...
}
checkpoint += rangeMs;
messageRocksDBStorage.writeCheckPointForTimer(...);
```

`scanRecordsToQueue` only **offers** the records to `rollMessageQueue`. The actual re-put into the commitlog happens asynchronously in `TimerMessageReputService`. If the broker crashes after the checkpoint is persisted but before the queue consumer finishes, the next start will skip that window. The records are still in RocksDB and will eventually be delivered by `TimelineForwardService`, but the early-roll guarantee is broken and the PR’s own test claim (“continue from the checkpoint without skipping … the previous window”) is violated.

Consider advancing the checkpoint only after the roll reput service has finished processing the window, similar to how `TimerMessageReputService.writeCheckPoint` works for the expired queue.

**Initial checkpoint uses `rollRangeHour` while the scan window is `rollIntervalHour`**
`store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java:383-384`

```java
if (checkpoint <= 0L) {
    checkpoint = System.currentTimeMillis() + maxDelayMs - TimeUnit.HOURS.toMillis(rollRangeHour);
}
```

The window width is `rangeMs = rollIntervalHour`, but the initial offset is `rollRangeHour`. If an operator configures `rollRangeHour < rollIntervalHour`, the first upper bound becomes `now + maxDelay + (rollIntervalHour - rollRangeHour)`, i.e. messages that are still farther than `maxDelay` away can be rolled too early. At best this wastes work; at worst it can roll a message that the timer wheel cannot yet hold.

Please validate that `rollRangeHour >= rollIntervalHour`, or initialize with `rangeMs` instead.

**No backoff on unexpected exception**
`store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java:395-397`

The outer `catch (Exception e)` logs and loops immediately. If RocksDB throws a persistent error the thread will spin. Add a short `waitForRunning` in the error path.

---

### 2. Performance

**Frequent checkpoint sync writes when catching up**
`store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java:391`

`writeCheckPointForTimer` writes through the WAL. When the service is behind it will scan consecutive 1-hour windows and persist after every single one. For a large backlog this is a lot of small sync writes. Consider persisting only every N windows or batching the advance.

**Hot loop on `storeConfig.isTimerStopDequeue()`**
`store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java:388`

`scanRecordsToQueue` returns `false` immediately when dequeue is stopped, then the new code waits only 200 ms. That is fine, but the log level is `error` for an intentional pause — should be `warn` or `info`.

---

### 3. Tests

**No unit tests for `TimelineRollService` scheduling logic**
The PR only adds:
- `testWriteAndGetRollCheckpoint` — validates the checkpoint key read/write.
- `testScanAdjacentWindowsNoOverlap` — validates `[lower, upper)` scan semantics.

Missing coverage:
- service waits when the next window is not yet due,
- service scans consecutive windows when behind,
- checkpoint is persisted and resumed on restart,
- no duplicate scan of the same window.

Please add a test for `TimelineRollService.run()` using a mocked `MessageRocksDBStorage` and a fake clock, or extract the scheduling into a package-private method.

**New test is flaky because it reuses global delay-time windows**
`store/src/test/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorageTest.java:155-178`

`testScanAdjacentWindowsNoOverlap` uses `begin = (System.currentTimeMillis() / window) * window`. Other tests in the same class write records near `currentTime + 3600000L`, which can fall into these windows and break the exact-size assertions depending on execution order/time. Use isolated, far-future delay times or clean the column family between tests.

**Pasted diff artifact**
The diff block shows lines like:
```java
@auth/src/test/java/org/apache/rocketmq/auth/authentication/AuthenticationEvaluatorTest.java
public void testWriteAndGetRollCheckpoint() {
```
Assuming this is a paste/rendering artifact, please confirm the real file uses `@Test`.

---

### 4. Compatibility

**Behavioral change of `timerRocksDBRollIntervalHours` / `timerRocksDBRollRangeHours`**
Previously `rollIntervalHours` controlled sleep and `rollRangeHours` controlled the scan start offset, while the scan width was always `timerMaxDelaySec`. Now `rollIntervalHours` is also the scan window width, and `rollRangeHours` only affects the first checkpoint. Existing deployments that tuned these configs will see different roll frequency and load patterns. Please update the config documentation and consider a release note.

**Downgrade path**
`store/src/main/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorage.java:75`

The new `TIMELINE_ROLL_CHECK_POINT` key is harmless to older code (it ignores unknown keys), but an older broker restarting after a new broker has advanced the checkpoint will not read it and will rescan already-rolled windows, potentially producing duplicate timer messages. That may be acceptable, but it should be documented.

---

**Recommendation:** Request changes. The checkpoint-before-consumption issue and the test flakiness should be addressed; the config compatibility note should be documented.

---
*Automated review by github-manager bot. Please verify suggestions before applying.*

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,35 +375,34 @@ public String getServiceName() {
...
+                    long rangeMs = TimeUnit.HOURS.toMillis(rollIntervalHour);
+                    if (checkpoint <= 0L) {
+                        checkpoint = System.currentTimeMillis() + maxDelayMs - TimeUnit.HOURS.toMillis(rollRangeHour);
                     }
-                    this.waitForRunning(TimeUnit.HOURS.toMillis(rollIntervalHour));
-                    if (stopped) {
-                        log.info(this.getServiceName() + " service end");
-                        return;
+                    long nextDueMs = checkpoint + rangeMs - maxDelayMs;
+                    long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
```
RockteMQ-AI: **[Critical]** Checkpoint is advanced *before* the rolled records are actually written back to the commitlog. `scanRecordsToQueue` only offers records to `rollMessageQueue`; the actual re-put happens asynchronously in `TimerMessageReputService`. If the broker crashes after checkpoint persist but before the queue consumer finishes, the next start skips that window.

Consider advancing the checkpoint only after the roll reput service has finished processing the window, similar to how `TimerMessageReputService.writeCheckPoint` works for the expired queue.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,35 +375,34 @@ public String getServiceName() {
...
-                int rollIntervalHour = 1;
-                int rollRangeHour = 2;
                 try {
-                    if (storeConfig.getTimerRocksDBRollIntervalHours() > 0) {
-                        rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours();
-                    }
-                    if (storeConfig.getTimerRocksDBRollRangeHours() > 0) {
-                        rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours();
+                    long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
+                    int rollIntervalHour = storeConfig.getTimerRocksDBRollIntervalHours() > 0 ? storeConfig.getTimerRocksDBRollIntervalHours() : 1;
```
RockteMQ-AI: **[Warning]** Initial checkpoint uses `rollRangeHour` while the scan window is `rollIntervalHour` (`rangeMs`). If `rollRangeHour < rollIntervalHour`, the first upper bound extends beyond `maxDelay`, potentially rolling messages too early.

Please validate that `rollRangeHour >= rollIntervalHour`, or initialize with `rangeMs` instead.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,35 +375,34 @@ public String getServiceName() {
...
-                    if (stopped) {
-                        log.info(this.getServiceName() + " service end");
-                        return;
+                    long nextDueMs = checkpoint + rangeMs - maxDelayMs;
+                    long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
+                    long now = System.currentTimeMillis();
+                    if (now < triggerAt) {
+                        this.waitForRunning(Math.min(triggerAt - now, ROLL_POLL_WHEN_NOT_DUE_MS));
+                        continue;
                     }
```
RockteMQ-AI, line 396: **[Warning]** The outer `catch (Exception e)` logs and loops immediately. If RocksDB throws a persistent error, the thread will spin. Add a short `waitForRunning` backoff in the error path.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,35 +375,34 @@ public String getServiceName() {
...
+                    if (checkpoint <= 0L) {
+                        checkpoint = System.currentTimeMillis() + maxDelayMs - TimeUnit.HOURS.toMillis(rollRangeHour);
                     }
-                    this.waitForRunning(TimeUnit.HOURS.toMillis(rollIntervalHour));
-                    if (stopped) {
-                        log.info(this.getServiceName() + " service end");
-                        return;
+                    long nextDueMs = checkpoint + rangeMs - maxDelayMs;
+                    long triggerAt = nextDueMs - ROLL_TRIGGER_EARLY_MS;
+                    long now = System.currentTimeMillis();
```
RockteMQ-AI, line 392: **[Info]** `writeCheckPointForTimer` writes through the WAL. When catching up from a backlog, this persists after every 1-hour window. Consider persisting only every N windows or batching the advance to reduce sync write overhead.

store/src/test/java/org/apache/rocketmq/store/rocksdb/MessageRocksDBStorageTest.java
```diff
@@ -140,4 +141,48 @@ public void testDeleteThenUpdate() {
...
+        Assert.assertEquals(0L, storage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, TIMELINE_ROLL_CHECK_POINT));
+
+        long checkpoint = System.currentTimeMillis() + 3600000L;
+        storage.writeCheckPointForTimer(TIMER_COLUMN_FAMILY, TIMELINE_ROLL_CHECK_POINT, checkpoint);
+        Assert.assertEquals(checkpoint, storage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, TIMELINE_ROLL_CHECK_POINT));
+
+        long nextCheckpoint = checkpoint + 3600000L;
+        storage.writeCheckPointForTimer(TIMER_COLUMN_FAMILY, TIMELINE_ROLL_CHECK_POINT, nextCheckpoint);
+        Assert.assertEquals(nextCheckpoint, storage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, TIMELINE_ROLL_CHECK_POINT));
+    }
```
RockteMQ-AI, line 160: **[Warning]** `testScanAdjacentWindowsNoOverlap` uses `begin = (System.currentTimeMillis() / window) * window`. Other tests in the same class write records near `currentTime + 3600000L`, which can fall into these windows and break exact-size assertions depending on execution order. Use isolated, far-future delay times or clean the column family between tests.

RockteMQ-AI (changes requested):
### Summary
This PR persists the TimelineRollService checkpoint to RocksDB to avoid duplicate scans and message loss across restarts. The approach is sound in principle, but the checkpoint is advanced before the async roll reput service finishes processing, which can cause window skipping on crash. Several other concerns around initial checkpoint alignment, error handling, and test isolation need attention.

### Findings
- **[Critical]** Timeline.java:390 — Checkpoint advanced before async roll completes; crash can skip windows
- **[Warning]** Timeline.java:383 — Initial checkpoint uses rollRangeHour but window is rollIntervalHour; mismatch risk
- **[Warning]** Timeline.java:395 — No backoff on persistent RocksDB errors; thread will spin
- **[Info]** Timeline.java:391 — Consider batching checkpoint writes to reduce WAL overhead
- **[Warning]** MessageRocksDBStorageTest.java:155 — Test window boundaries may collide with other tests

---

RockteMQ-AI (approved):
### Summary
This PR persists the TimelineRollService checkpoint to RocksDB to avoid duplicate scans and message loss across restarts. The updated version addresses all previously raised concerns:

1. **Checkpoint persistence timing** ✅ — Checkpoint is now persisted by `TimerRollReputService.doReput()` **after** records are successfully written to the commitlog, ensuring crash-safety.
2. **Initial checkpoint alignment** ✅ — Initialization properly calculates the starting checkpoint from `maxDelay`, avoiding out-of-range windows on first run.
3. **Error handling backoff** ✅ — `Thread.sleep(200)` added in both the scan-error and general-exception paths, preventing tight spin loops on persistent errors.
4. **Test isolation** ✅ — Tests use a fixed time base (`BASE_TIME_MS`) instead of `System.currentTimeMillis()`, ensuring deterministic behavior.

The removal of `timerRocksDBRollIntervalHours` simplifies the configuration model by deriving the roll cadence from `rollRangeHours` and the delay window. Note: this is a minor backward-incompatible config change — users who had customized this value will need to remove it from their configs.

Overall the approach is sound and the implementation is now correct. LGTM.

---

RockteMQ-AI (approved):
### Summary
Re-reviewed after new commits. The checkpoint-driven roll approach is well-implemented — persisted RocksDB checkpoint replaces the fragile fixed-sleep logic, preventing both duplicate and lost timer message rolls across restarts.

Key changes look correct:
- `TIMELINE_ROLL_CHECK_POINT` properly registered in `COMMON_CHECK_POINT_KEY_SET_FOR_TIMER`
- Poll-based trigger with 1s early wake is a reasonable trade-off between precision and CPU
- Removal of `timerRocksDBRollIntervalHours` is consistent since the interval is now derived from the checkpoint

---

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,37 +375,37 @@ public String getServiceName() {
...
         public void run() {
-            log.info(this.getServiceName() + " service start");
+            long checkpoint = messageRocksDBStorage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_ROLL_CHECK_POINT);
+            if (checkpoint <= 0L) {
+                long now = System.currentTimeMillis();
+                long forwardCheckpoint = messageRocksDBStorage.getCheckpointForTimer(TIMER_COLUMN_FAMILY, MessageRocksDBStorage.TIMELINE_CHECK_POINT);
+                int rollRangeHour = storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2;
+                checkpoint = (forwardCheckpoint > 0L ? Math.min(forwardCheckpoint, now) : now)
+                    + TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec()) - TimeUnit.HOURS.toMillis(rollRangeHour);
+            }
```
RockteMQ-AI, line 385: **[Info]** The checkpoint initialization logic is solid — starting from `min(forwardCheckpoint, now) + maxDelay - rollRange` ensures we don't miss messages that were already queued for expiration.

One consideration: if `forwardCheckpoint` is significantly behind `now` (e.g., after a long outage), the initial checkpoint could be far in the past. The service will catch up by scanning multiple windows in sequence, which is correct, but might want to log a warning when `checkpoint < now - rollRange` to make recovery visible.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -373,37 +375,37 @@ public String getServiceName() {
...
-                    if (stopped) {
-                        log.info(this.getServiceName() + " service end");
-                        return;
+                    long maxDelayMs = TimeUnit.SECONDS.toMillis(storeConfig.getTimerMaxDelaySec());
+                    long rangeMs = TimeUnit.HOURS.toMillis(storeConfig.getTimerRocksDBRollRangeHours() > 0 ? storeConfig.getTimerRocksDBRollRangeHours() : 2);
+                    long triggerAt = checkpoint + rangeMs - maxDelayMs - ROLL_TRIGGER_EARLY_MS;
+                    long now = System.currentTimeMillis();
+                    if (now < triggerAt) {
+                        this.waitForRunning(ROLL_POLL_WHEN_NOT_DUE_MS);
+                        continue;
```
RockteMQ-AI, line 395: **[Warning]** The error handling loop waits only 200ms on failure before retrying. If `scanRecordsToQueue` consistently fails (e.g., RocksDB corruption, disk full), this could create a tight error loop.

Consider adding exponential backoff or a max retry count before logging at ERROR level and pausing longer:
```java
int consecutiveErrors = 0;
// ...
if (!scanRecordsToQueue(...)) {
    consecutiveErrors++;
    long backoff = Math.min(200L * (1L << Math.min(consecutiveErrors, 5)), 10000L);
    this.waitForRunning(backoff);
    if (consecutiveErrors > 10) {
        logError.error("TimelineRollService: {} consecutive failures, backing off {}ms", consecutiveErrors, backoff);
    }
    continue;
}
consecutiveErrors = 0;
```

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/Timeline.java
```diff
@@ -49,6 +49,8 @@ public class Timeline {
     private static final String DELETE_KEY_SPLIT = "+";
     private static final int ORIGIN_CAPACITY = 100000;
     private static final int BATCH_SIZE = 1000, MAX_BATCH_SIZE_FROM_ROCKSDB = 8000;
+    private static final long ROLL_TRIGGER_EARLY_MS = 1000L;
```
RockteMQ-AI, line 52: **[Info]** The hardcoded constants `ROLL_TRIGGER_EARLY_MS = 1000L` and `ROLL_POLL_WHEN_NOT_DUE_MS = 1000L` work well for the current use case. If these need tuning in production, consider exposing them via `MessageStoreConfig` with sensible defaults.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/TimerMessageRocksDBStore.java
```diff
@@ -545,9 +545,9 @@ public void run() {
                     }
                     countDownLatch.await();
                     log.info("TimerMessageReputService reput messages to commitlog, cost: {}, trs size: {}, checkPoint: {}", System.currentTimeMillis() - start, trs.size(), trs.get(trs.size() - 1).getCheckPoint());
-                    if (this.writeCheckPoint && !CollectionUtils.isEmpty(trs) && trs.get(trs.size() - 1).getCheckPoint() > 0L) {
+                    if (null != this.checkPointKey && !CollectionUtils.isEmpty(trs) && trs.get(trs.size() - 1).getCheckPoint() > 0L) {
```
RockteMQ-AI, line 548: **[Critical]** The null check `null != this.checkPointKey` is correct, but the Yoda condition style (`null != x` instead of `x != null`) is inconsistent with the rest of the codebase. RocketMQ style typically uses `this.checkPointKey != null`.

Minor, but worth aligning with project conventions for readability.

RockteMQ-AI (commented):
### Summary
This PR fixes a critical correctness issue in `TimelineRollService` where the fixed-interval sleep + wall-clock-based scan window could cause timer messages to be rolled multiple times or skipped after restarts. The checkpoint-driven approach with persisted RocksDB state is the right solution.

### Findings
- **[Critical]** TimerMessageRocksDBStore.java:548 — Yoda condition style inconsistent with codebase
- **[Warning]** Timeline.java:395 — Tight error loop risk on persistent failures
- **[Info]** Timeline.java:385 — Checkpoint initialization could log recovery scenarios
- **[Info]** Timeline.java:52 — Hardcoded timing constants work but could be configurable

### Strengths
✅ Checkpoint persistence eliminates duplicate/skipped rolls across restarts
✅ Adjacent window scanning logic is correct and well-tested
✅ Refactoring `TimerMessageReputService` to use checkpoint keys is clean and extensible
✅ Test coverage for checkpoint independence and window boundaries is solid

### Suggestions
1. Add exponential backoff for error retries to avoid tight loops
2. Log a warning when initial checkpoint is far behind current time (recovery scenario)
3. Align null-check style with codebase conventions

Overall, this is a well-designed fix for a subtle concurrency/persistence bug. The changes are focused and the test coverage is good.

RockteMQ-AI (approved):
### Summary
This PR correctly addresses the issue of timer roll service repeating or losing rolls after restart by replacing the fixed-interval approach with checkpoint-based scheduling.

The implementation is clean:
- Removes the unused `timerRocksDBRollIntervalHours` config
- Adds `TIMELINE_ROLL_CHECK_POINT` for persisting roll state
- Refactors `TimelineRollService` to use checkpoint-based scheduling
- Updates `TimerMessageReputService` to support flexible checkpoint keys
- Includes comprehensive tests

LGTM. Well-structured fix with proper test coverage.

---

RongtongJin:
> *This was generated by AI during triage.*

I took another look at the latest head (`b3b7490`). This PR is addressing a real correctness/performance issue in `TimelineRollService`: the old fixed-sleep loop scanned a window like `[now + rollRange, now + rollRange + timerMaxDelaySec]`, so adjacent roll rounds overlapped heavily and could roll the same far-future timer records back to `TIMER_TOPIC` repeatedly. Moving to a checkpoint-driven scan (`[checkpoint, checkpoint + rollRange)`) with a dedicated `timeline_roll_checkpoint` is the right direction.

The latest version also fixes the earlier crash-skip concern by moving checkpoint persistence into `TimerMessageReputService`, after the queued roll batch has been processed, instead of writing it immediately after scan.

One remaining correctness concern before merge: `TimerMessageReputService` writes the checkpoint after all task countdowns complete, but individual tasks do not report whether `putMsgWithRetry()` actually succeeded. `putMsgWithRetry()` currently returns `void` and logs/returns after retries are exhausted. If one or more records fail to re-put to the commitlog, the service can still advance `TIMELINE_ROLL_CHECK_POINT` past that window, which removes the normal early-roll retry opportunity for those failed records.

Suggested fix: make each task return success/failure and only write the roll checkpoint when the whole batch succeeds, or avoid advancing the checkpoint past failed records.

One compatibility note: this PR removes `timerRocksDBRollIntervalHours`, and `timerRocksDBRollRangeHours` now effectively controls both the scan window width and roll cadence. Please document this in config docs/release notes, since deployments with a custom roll interval may see changed behavior.

CI is green and the overall checkpoint-based design looks sound once the failed-reput checkpoint edge is handled.

3424672656:
> > _This was generated by AI during triage._
> 
> I took another look at the latest head (`b3b7490`). This PR is addressing a real correctness/performance issue in `TimelineRollService`: the old fixed-sleep loop scanned a window like `[now + rollRange, now + rollRange + timerMaxDelaySec]`, so adjacent roll rounds overlapped heavily and could roll the same far-future timer records back to `TIMER_TOPIC` repeatedly. Moving to a checkpoint-driven scan (`[checkpoint, checkpoint + rollRange)`) with a dedicated `timeline_roll_checkpoint` is the right direction.
> 
> The latest version also fixes the earlier crash-skip concern by moving checkpoint persistence into `TimerMessageReputService`, after the queued roll batch has been processed, instead of writing it immediately after scan.
> 
> One remaining correctness concern before merge: `TimerMessageReputService` writes the checkpoint after all task countdowns complete, but individual tasks do not report whether `putMsgWithRetry()` actually succeeded. `putMsgWithRetry()` currently returns `void` and logs/returns after retries are exhausted. If one or more records fail to re-put to the commitlog, the service can still advance `TIMELINE_ROLL_CHECK_POINT` past that window, which removes the normal early-roll retry opportunity for those failed records.
> 
> Suggested fix: make each task return success/failure and only write the roll checkpoint when the whole batch succeeds, or avoid advancing the checkpoint past failed records.
> 
> One compatibility note: this PR removes `timerRocksDBRollIntervalHours`, and `timerRocksDBRollRangeHours` now effectively controls both the scan window width and roll cadence. Please document this in config docs/release notes, since deployments with a custom roll interval may see changed behavior.
> 
> CI is green and the overall checkpoint-based design looks sound once the failed-reput checkpoint edge is handled.

fix

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/TimerMessageRocksDBStore.java
```diff
@@ -569,13 +569,12 @@ private void putMsgWithRetry(MessageExtBrokerInner msg) throws InterruptedExcept
                         logError.warn("Skipping message due to unrecoverable error. Msg: {}", msg);
                         return;
                     default:
-                        if (retryCount == MAX_PUT_MSG_TIMES) {
+                        if (!storeConfig.isTimerEnableRetryUntilSuccess() && retryCount >= MAX_PUT_MSG_TIMES) {
```
RockteMQ-AI, line 572: **[Warning]** When `timerEnableRetryUntilSuccess=true`, the retry loop runs indefinitely with a fixed 100ms sleep and a warn log on every iteration. This can generate ~600 log entries per minute during persistent failures, potentially flooding logs and making it harder to diagnose the root cause.

Consider adding periodic logging (e.g., log every 100 retries or every 10 seconds) to reduce noise while still providing visibility. Alternatively, exponential backoff (capped at e.g. 5s) would reduce both log volume and system load during prolonged failures.

store/src/main/java/org/apache/rocketmq/store/timer/rocksdb/TimerMessageRocksDBStore.java
```diff
@@ -560,7 +560,7 @@ private void putMsgWithRetry(MessageExtBrokerInner msg) throws InterruptedExcept
             if (null == msg) {
                 return;
             }
-            for (int retryCount = 0; !isStopped() && retryCount <= MAX_PUT_MSG_TIMES; retryCount++) {
+            for (int retryCount = 0; !isStopped(); retryCount++) {
```
RockteMQ-AI, line 563: **[Info]** `retryCount` is `int` — with unlimited retries at 100ms intervals, it would take ~6.8 years to overflow `Integer.MAX_VALUE`. Very low risk in practice, but using `long` would be more defensive for a loop that is explicitly designed to run indefinitely.

RongtongJin (approved):

PR #11160, opened 2026-09-14
[ISSUE #11159] Fix Gradle dependency snippets that use the removed compile configuration

### LINKED ISSUE #11159: [Doc] Gradle snippets in the quick-start docs use the removed compile configuration
### Search before creation

- [x] I had searched in the [issues](https://github.com/apache/rocketmq/issues) and found no similar issues.

### Documentation Related

The quick-start docs give the Gradle dependency as:

```
compile 'org.apache.rocketmq:rocketmq-client:5.5.0'
```

Gradle removed the `compile` configuration in 7.0, so this no longer works. Reproduced on Gradle 9.7.1 with an otherwise empty `java` project:

```
* What went wrong:
A problem occurred evaluating root project '...'.
> Could not find method compile() for arguments [org.apache.rocketmq:rocketmq-client:5.5.0] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

The error doesn't hint at the cause. Gradle's own suggestions are just `--stacktrace` / `--info` / `--scan`, and it fails while evaluating the project, so it reads like a method-name typo rather than a removed configuration. Swapping in `implementation` builds fine and resolves `rocketmq-client:5.5.0` along with its transitive dependencies from Maven Central, so the coordinates and the version in the docs are correct and only the configuration name is stale.

Affected files:

- `docs/en/Example_Simple.md`
- `docs/cn/Example_Simple_cn.md`
- `docs/cn/RocketMQ_Example.md`

The Maven blocks right above them are fine. Since the repo builds with Maven and Bazel only, no CI job ever looks at these snippets, which is probably how it drifted.

I checked the in-flight docs PRs first. #10436 and #10409 both touch these same lines, but they only bump the version number inside the snippet and keep `compile`, so this doesn't overlap with them.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

neoLsH:

- Fixes #11159

### Brief Description

The Gradle snippets in the quick-start docs still use `compile`, which Gradle removed in 7.0. Copying them into a current build fails while the project is being evaluated, and the error reads like a method-name typo with no hint that the configuration itself is gone. Switched to `implementation`, which is what the Gradle upgrade guide's migration table maps `compile` to. The other option in that table, `api`, requires the `java-library` plugin, and these snippets are for sample applications.

The Maven blocks right above them are untouched.

Docs only, no code path affected. Checked on Gradle 9.7.1 with a minimal `java` project: with `compile` the build fails with "Could not find method compile() for arguments [...] on object of type ...DefaultDependencyHandler", with `implementation` it succeeds and resolves rocketmq-client:5.5.0 along with its transitive dependencies from Maven Central, which also confirms the coordinates and version already in the docs are correct.

Grepped the whole tree on develop too. These are the only Gradle-style dependency declarations and the only files that mention Gradle at all, so nothing is left half-migrated.


RockteMQ-AI (approved):
### Summary
LGTM — straightforward and correct fix for outdated Gradle dependency syntax.

### Review Notes
- **Correctness**: ✅  →  is the right migration path per Gradle 7.0 release notes
- **Completeness**: ✅ All 3 occurrences updated (2 Chinese docs, 1 English doc)
- **Testing**: ✅ Author verified with Gradle 9.7.1 and confirmed the dependency resolves correctly
- **Scope**: ✅ Minimal change, no unintended side effects

The PR description is thorough and explains the rationale well. This fixes a real pain point for users following the quick-start guide with modern Gradle versions.

Thanks for the contribution!

---

Issue #8366, opened 2024-07-05
[Bug] Multiple deadlocks were detected during the client shutdown process.


YanYunyang:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

4.19.90-23-42.v2101.ky10.x86_64

### RocketMQ version

5.2.0

### JDK Version

openjdk 1.8

### Describe the Bug

During the client shutdown process, deadlocks were detected, which resolved after a period of time. 
This phenomenon was observed using jvisualvm and jconsole. 
The deadlock occurred between the client shutdown thread and the Netty worker threads.
Code analysis: 

1. `NettyRemotingClient` holds the `lockChannelTables` lock, which is responsible for guarding access to `channelTables`. `channelTables` caches all channels encapsulated in `ChannelWrapper`.
2. The inner class ChannelWrapper within NettyRemotingClient` holds a read-write lock `lock`, responsible for concurrent access to the channel.
3. The inner class `NettyConnectManageHandler` in `NettyRemotingClient` is responsible for handling events like close, connect, and channelInactive. When a channel becomes unavailable, it executes the `close` or `channelInactive` methods to remove the channel from `channelTables`.

<img width="1057" alt="截屏2024-07-05 14 11 28" src="https://github.com/apache/rocketmq/assets/11780013/e42afcc4-9faa-4a2b-a495-44b2cd6eb550">
<img width="1180" alt="截屏2024-07-05 14 15 06" src="https://github.com/apache/rocketmq/assets/11780013/aabc96c2-843a-4133-bbc8-81ac2d026d65">
<img width="1197" alt="截屏2024-07-05 14 19 16" src="https://github.com/apache/rocketmq/assets/11780013/7e47c007-ac4d-43d5-acad-3d8890ae6e89">
<img width="1431" alt="截屏2024-07-05 14 26 54" src="https://github.com/apache/rocketmq/assets/11780013/9c381594-956f-4859-90eb-9cb81ff592c4">
The execution path where deadlock occurs.

[...]

PR #11162, opened 2026-09-14
[ISSUE #11161] Fix closeChannel table eviction and ChannelWrapper.close lock ordering

### LINKED ISSUE #11161: NettyRemotingClient: closeChannel(addr, channel) skips table eviction and ChannelWrapper.close() inverts the lock order

- [x] I searched the [GitHub issues](https://github.com/apache/rocketmq/issues) and found no similar report.
- [x] I reviewed the code on the current `develop` branch.

All platforms (client side, `NettyRemotingClient`).

### RocketMQ version

`develop` (reproduced by reading the current source).

### Component

remoting

### Describe what happened

There are two related defects in `org.apache.rocketmq.remoting.netty.NettyRemotingClient` on the current `develop` branch.

#### Bug 1 — `closeChannel(String addr, Channel channel)` no longer evicts the channel-table entry

```java
if (null == prevCW) {
    ...
    removeItemFromTable = false;
} else if (prevCW.isWrapperOf(channel)) {          // <-- inverted
    LOGGER.info("... has been closed before, and has been created again, nothing to do.", ...);
    removeItemFromTable = false;
}
```

`isWrapperOf(channel)` is true exactly when the wrapper currently stored under `addr` wraps *this* channel — i.e. the normal case of closing the live channel. In that case the code takes the "has been created again, nothing to do" branch and sets `removeItemFromTable = false`, so the entry is **not** removed from `channelTables`.

Before #8366 the condition was `else if (prevCW.getChannel() != channel)`, which is the logical negation of `isWrapperOf(channel)`. #8366 replaced the expression with `isWrapperOf(channel)` but did not re-negate it, so the branch semantics and the accompanying log message are now inverted.

Effects:
- The eviction block inside `closeChannel(addr, channel)` becomes effectively dead for the common path. Note the two conditions are mutually exclusive: the block is only reached when the wrapper is *not* this channel, yet `channelWrapper.tryClose(channel)` only returns true when the wrapper *is* this channel.
- The map entry is only cleaned up indirectly: the trailing `RemotingHelper.closeChannel(channel)` closes the socket, which fires the pipeline close event and reaches the single-argument `closeChannel(Channel)` (that method does not have the inverted branch). So the entry is eventually removed, but via a fragile pipeline-event detour rather than the intended eager removal, and the misleading "created again, nothing to do" log fires on every ordinary close.

#### Bug 2 — `ChannelWrapper.close()` acquires locks in the order opposite to the rest of the class

```java
public void close() {
    try {
        lock.writeLock().lock();                 // wrapper write lock
        if (channelFuture != null) {
            closeChannel(channelFuture.channel()); // -> acquires lockChannelTables
        }
        if (channelToClose != null) {
            closeChannel(channelToClose.channel());
        }
    } finally {
        lock.writeLock().unlock();
    }
}
```

Every other path takes `lockChannelTables` first and then a wrapper lock:
- `createChannelAsync`: `lockChannelTables` → `ChannelWrapper.getChannelFuture()`/`isOK()` (wrapper read lock)
- `closeChannel(addr, channel)` / `closeChannel(channel)`: `lockChannelTables` → `ChannelWrapper.tryClose()` (wrapper read lock)

`ChannelWrapper.close()` reverses that order (`wrapper write lock` → `lockChannelTables`), which is a classic AB-BA lock-ordering inversion. In practice it does not hang forever because `lockChannelTables` is acquired with `tryLock(3000ms)`, but the reverse ordering can still stall a thread — and block everything else waiting on `lockChannelTables`, including the single-threaded selector reaching `closeChannel` from a pipeline event — for up to the 3s timeout. `close()` is invoked from `shutdown()` and from `updateNameServerAddressList()`.

### Describe what you expected to happen

- `closeChannel(addr, channel)` should evict the table entry when the stored wrapper wraps the channel being closed, and should leave it alone only when the wrapper has already been recreated for a different channel.
- `ChannelWrapper.close()` should not hold a wrapper lock while acquiring `lockChannelTables`, so that the whole client obeys a single consistent lock order (`namesrvChannelLock` → `lockChannelTables` → wrapper lock).

### How to reproduce

Bug 1 is directly observable in a unit test: install a `ChannelWrapper` for an address into `channelTables`, call `closeChannel(addr, wrapper.channel())`, and assert the entry is gone. On current `develop` the entry remains.

### Additional context

I have a fix ready (restore the negation for Bug 1; rework `close()` to snapshot the channel futures under the read lock and then call `closeChannel(...)` without holding any wrapper lock, still closing both `channelFuture` and `channelToClose`) plus regression tests, and will open a PR referencing this issue.

qianye1001:

Fixes #11161.

### Brief Description

Two related fixes in `NettyRemotingClient`.

**Bug 1 — `closeChannel(String addr, Channel channel)` no longer evicts the table entry.** #8366 replaced the original condition `else if (prevCW.getChannel() != channel)` with `else if (prevCW.isWrapperOf(channel))`, which is its un-negated form. As a result the "has been created again, nothing to do" branch now triggers on the *normal* close path (the stored wrapper wraps this very channel), sets `removeItemFromTable = false`, and the entry is not evicted. The eviction block also became mutually exclusive with its own `tryClose(channel)` guard, so it is effectively dead; the entry only gets cleaned up indirectly when the trailing `RemotingHelper.closeChannel(channel)` fires the pipeline close event into the single-arg `closeChannel(Channel)`. This restores the negation (`!prevCW.isWrapperOf(channel)`) so the entry is evicted on normal close and preserved only when the wrapper has been recreated for a different channel.

**Bug 2 — `ChannelWrapper.close()` inverted the lock order.** It held the wrapper write lock while calling `closeChannel(...)`, which acquires `lockChannelTables`. Every other path takes `lockChannelTables` first and then a wrapper lock (`createChannelAsync` via `getChannelFuture()`/`isOK()`, `closeChannel` via `tryClose()`), so this was an AB-BA inversion. It did not hang permanently because `lockChannelTables` uses `tryLock(3000ms)`, but the reverse order can stall a thread — and everything else waiting on `lockChannelTables`, including the selector thread reaching `closeChannel` from a pipeline event — for up to 3s. `close()` now snapshots both channel futures under the read lock, releases it, then calls `closeChannel(...)` without holding any wrapper lock, so the whole client obeys one consistent order (`namesrvChannelLock` → `lockChannelTables` → wrapper lock). Both `channelFuture` and `channelToClose` are still closed so the reconnect leftover channel is not leaked, and the two-arg `closeChannel(channelAddress, ...)` is used so the table is looked up by address instead of a linear scan.

JDK 11:

```shell
mvn -B -pl remoting -am -Dmaven.gitcommitid.skip=true \
  -Dtest=NettyRemotingClientTest,NettyRemotingClientCloseChannelTest \
```

Added `NettyRemotingClientCloseChannelTest` covering both `closeChannel(addr, channel)` branches: the entry is evicted when the wrapper wraps the closed channel, and preserved when the wrapper has been recreated for a different channel. The first assertion fails on the current (inverted) code and passes with the fix. Existing `NettyRemotingClientTest` (16 tests) continues to pass.


RockteMQ-AI (approved):
### Summary
Excellent fix for two related concurrency issues in `NettyRemotingClient`.

**Bug 1 (condition inversion):** The `closeChannel(addr, channel)` method had an inverted condition — `prevCW.isWrapperOf(channel)` instead of `!prevCW.isWrapperOf(channel)`. This caused the channel table entry to not be evicted on normal close, leading to stale entries. The fix correctly restores the negation.

**Bug 2 (lock ordering):** `ChannelWrapper.close()` held the wrapper write lock while calling `closeChannel()`, which acquires `lockChannelTables`. This inverted the lock order used everywhere else (`lockChannelTables` → wrapper lock), creating an AB-BA inversion. The fix correctly snapshots the channel futures under a read lock, releases it, then closes without holding any wrapper lock.

The use of the two-arg `closeChannel(channelAddress, channel)` also avoids a linear scan, which is a nice optimization.

Tests cover both branches of the fix — entry eviction when the wrapper matches, and preservation when the wrapper has been recreated. Well-structured and clear.

---

lollipopjin (approved):
LGTM

PR #11164, opened 2026-09-14
[ISSUE #11163] Fix lite topic prefix index not maintained on first message

### LINKED ISSUE #11163: [Bug] Lite topics created after broker startup are not delivered to wildcard subscribers until restart

OS: Linux / macOS

### RocketMQ version
branch: develop

### JDK Version
Compiler: Oracle JDK 11

### Describe the Bug
Since the lite topic prefix index was introduced (#10930), lite topics created after broker startup are never added to the index. Wildcard-subscription full dispatch and the parent-topic based operations (collect / count / clean lite topics) therefore miss every lite topic created at runtime, and only see the ones that existed at startup, until the broker restarts and repopulates the index.

Root cause: when the first message of a lite topic is written, the store notifies with logicOffset = queueOffset + 1, i.e. offset == 1 for the first message, but the dispatcher maintains the index only when offset == 0, so the create hook never fires at runtime.

### Steps to Reproduce
1. Start a broker with a LITE parent topic and a wildcard consumer group bound to it.
2. After startup, send the first message to a new lite topic under that parent.
3. The wildcard subscriber is not dispatched the new lite topic; admin operations that list/count/clean lite topics by parent topic also miss it.
4. Restart the broker; the lite topic now shows up.

A lite topic created at runtime is immediately visible to wildcard dispatch and to parent-topic collect/count/clean, without a restart.

### What Did You See Instead?
Runtime-created lite topics are missing until the broker restarts.

f1amingo:

- Fixes #11163

### Brief Description

The lite topic prefix index is maintained only when the dispatcher sees offset == 0, but the message-arriving notification carries logicOffset = queueOffset + 1, so the first message of a lite topic arrives with offset == 1. The create hook thus never fires at runtime, and lite topics created after startup are missing from wildcard dispatch and parent-topic collect/count/clean until the broker restarts. This PR corrects the comparison to offset == 1 and restores the store-dependent lifecycle tests to the concrete impl test classes (file CQ and RocksDB CQ), driving the index through the real putMessage -> notify -> dispatch path so this behavior is covered.

Unit tests in LiteLifecycleManagerTest (file CQ) and RocksDBLiteLifecycleManagerTest (RocksDB CQ) wire a real MessageArrivingListener + LiteEventDispatcher and assert collectByParentTopic / getLiteTopicCount / cleanByParentTopic against real putMessage. Before the fix they fail with an empty index; after the fix the lite test package passes (19 run, 0 failures). Checkstyle clean.


RockteMQ-AI (commented):
### Summary
This PR modifies 5 files (+209/-125) in `apache/rocketmq`.

### Observations
- **Diff size:** 577 lines — moderate change, recommend careful review
- **Test coverage:** ✅ Test files included

### Recommendations
- Please ensure backward compatibility if any public API is modified
- Verify thread safety for any concurrent code paths
- Confirm error handling is adequate for new code paths

---

imzs (approved):

RockteMQ-AI (approved):
LGTM. The changes look good.

---

lizhimins (approved):

PR #11167, opened 2026-09-15
fix(tieredstore): report dispatch metric collection failures

### LINKED ISSUE #11166: [Bug] Tiered store dispatch metrics silently ignore ConsumeQueueException
### Describe the Bug

`TieredStoreMetricsManager` catches `ConsumeQueueException` while collecting the `dispatchBehind` and `dispatchLatency` gauges, then silently ignores the exception. The affected callbacks are in `initMetrics()`.

When either collection path fails, the metric has no measurement for that collection cycle and the broker logs do not show the failure. Operators cannot distinguish a zero metric value from an unsuccessful collection attempt.

`PopMetricsManager` logs `ConsumeQueueException` with `log.error(message, e)` for its gauge callbacks. Applying the same handling here keeps the metric callbacks non-disruptive while making failures observable.

### Expected Behavior

Tiered Store metric collection failures should be logged with the exception details.

### Additional Context

The proposed change only adds error logging in the two existing `catch (ConsumeQueueException e)` blocks. Normal metric collection and message processing behavior remain unchanged.

ForIt111:

- Fixes #11166

### Brief Description

Log `ConsumeQueueException` while collecting the Tiered Store `dispatchBehind` and `dispatchLatency` gauges. This preserves the existing callback behavior while exposing collection failures to operators with the exception stack trace.

- Passed `git diff --check`.
- Attempted `mvn -pl tieredstore -am -DskipITs test`, but this environment does not have Maven or a Java runtime installed, so the module tests could not run locally.


RockteMQ-AI:
## 🔍 Code Review — PR #11167

**Verdict: ✅ Looks good**

This is a clean, minimal fix that addresses the silent exception swallowing described in #11166.

### Analysis

| Aspect | Assessment |
|--------|-----------|
| Correctness | ✅ `log` field is properly declared at L79 as `private static final Logger` via `LoggerFactory` |
| Consistency | ✅ Follows the existing pattern at L307 (`log.error("Failed to get storage size", e)`) |
| Error messages | ✅ Descriptive — each message identifies which metric failed (`dispatch-behind` / `dispatch-latency`) |
| Behavior preservation | ✅ Catch blocks still exist; metric recording continues for other gauges even if one fails |
| Scope | ✅ Minimal — 2 lines changed, 1 file, no unrelated modifications |

### Minor Observations

1. **No test coverage** — The PR description notes that Maven/Java was not available locally. Consider adding a unit test that verifies the `log.error` path is exercised when `ConsumeQueueException` is thrown, to prevent regression.

2. **Metrics still report stale values on failure** — When the exception is caught and logged, the gauge simply does not record a new value for that collection cycle. This is acceptable behavior, but worth documenting in the PR description or a code comment for future maintainers.

Neither observation blocks this fix. The change is a clear improvement over silently dropping exceptions.

---
*Automated review by @RockteMQ-AI · Please verify suggestions before applying.*

RockteMQ-AI (approved):
### Summary
Replaces two `// TODO: handle exception` comments with proper `log.error()` calls in the tiered store metrics collection. Good fix — silent exception swallowing in metric collectors can hide real issues.

Changes look correct:
- Descriptive error messages identify which metric failed (dispatch-behind / dispatch-latency)
- Exception is passed as second argument for full stack trace
- `log.error` is the appropriate level for unexpected failures in background metric collection

LGTM.

---

PR #11169, opened 2026-09-15
[ISSUE #11168] Fix tiered storage commit failure reporting and related hazards

### LINKED ISSUE #11168: [Bug] Tiered storage misreports transient upload failures as file-size lookup errors

- [x] I have searched the GitHub Issues and Discussions of this repository and believe that this is not a duplicate.

Linux / Kubernetes. Tiered storage enabled with an S3-style object storage provider (Alibaba Cloud OSS) behind an HTTP load balancer.

### RocketMQ version

`develop`. The affected code paths are unchanged since `026a910b26` (#7899, 2024-03-18), so 5.3.x / 5.4.x / 5.5.x are all affected. Observed in production on builds pinning kernel `5.2.1.7`, `5.3.2-10`, `5.3.2-18` and `5.5.0-8`.

### JDK Version

Not JDK specific (observed on 11 and 21).

### Describe the Bug

Five independent defects in the tiered storage module, found while analysing 24h of production `tiered_store.log` in one region (302M lines, of which 16,902 ERROR).

**1. A transient upload failure is reported as `get file size error after commit`, although no size lookup ever happened.**

`TieredStoreException` defaults `position` to `-1` (`exception/TieredStoreException.java:23`), which collides with `FileSegment.GET_FILE_SIZE_ERROR = -1L` (`provider/FileSegment.java:42`). In `handleCommitException`:

```java
long fileSize = rootCause instanceof TieredStoreException ?
    ((TieredStoreException) rootCause).getPosition() : this.getSize();
if (fileSize == GET_FILE_SIZE_ERROR) {
    log.error("... get file size error after commit ...");
```

`OSSFileSegment`-style providers only call `setPosition()` when they received an HTTP error response. A transport-level failure (`RemotelyClosedException`, `Connection reset by peer`) has no response, so `position` stays `-1`, is read back as `fileSize`, and takes the same branch as a genuinely failed HEAD. The message names a lookup that never ran.

Production evidence: in 7,379 occurrences over 24h, `expect == commit + content` held in every single line, i.e. all values were local and no branch ever performed a remote lookup.

**2. `handleCommitException` can perform a blocking remote lookup on a netty IO thread.**

It is registered as `.exceptionally(this::handleCommitException)` with no executor, so it runs on whichever thread completed the future — for a network provider, a netty IO thread. Production logs confirm the thread name:

```
ERROR [AsyncHttpClient-3-4] FileSegment#handleCommitException, get file size error after commit, ...
```

The `else` side of the ternary above calls `this.getSize()`, which for a network provider is a synchronous object-metadata request. Blocking a shared IO thread stalls every other object storage request on that broker.

**3. `handleCommitException` logs three differently worded messages with three different field sets.**

One of them describes a *success* (the append landed remotely and only the response was lost) under a method named `handleCommitException`. The three messages also disagree on field names: the same quantity is `content` here and `buffer` in `commitAsync`, and `commit` is logged *after* `correctPosition` may have overwritten it, so `commit + content == expect` does not hold on the reconciled paths.

**4. `FlatAppendFile#destroyExpiredFile` deletes the remote object before unregistering its metadata.**

```java
fileSegment.destroyFile();
if (!fileSegment.exists()) {
    fileSegmentTable.remove(0);
    metadataStore.deleteFileSegment(filePath, fileType, fileSegment.getBaseOffset());
}
```

A crash between the delete and the unregister leaves a metadata row pointing at a deleted object. `recover()` reloads it on every restart, and every later read of that segment fails with `NoSuchKey`. In production this produced 844 such failures in 24h against only **4 distinct object names**, concentrated on 3 of 11 affected instances — the signature of permanent phantom metadata rather than transient errors.

The reverse order fails safe: an orphaned object costs storage, a phantom metadata row breaks reads forever.

**5. `FileSegment#readAsync` silently shortens a read and logs it at DEBUG.**

```java
int readableBytes = (int) (currentCommitPosition - position);
if (readableBytes < length) {
    log.debug("FileSegment#readAsync, request position exceeds commit position, ...");
    length = readableBytes;
}
```

The truncated buffer surfaces much later as a `MessageFormatUtil#splitMessageBuffer` failure, whose own message carries no topic, queueId or offset. The cause is therefore invisible in production: we observed 391 `message buffer offset exceeded limit` errors in 24h on one instance and could not attribute them to a queue from the logs alone.

**6. `MessageStoreFetcherImpl` compares an entry count against a byte count.**

```java
boolean cacheBusy = fetcherCache.estimatedSize() > memoryMaxSize * 0.8;
```

`estimatedSize()` counts entries; `memoryMaxSize` is bytes (`maxMemory() * readAheadCacheSizeThresholdRate`), and the cache is bounded by `maximumWeight` with a weigher returning `buffer.getSize()`. The comparison is effectively always false, so the read-ahead cache is never treated as busy.

### Steps to Reproduce

Defects 1–3 and 6 are readable directly from the code. For defect 1, any transport-level failure during `commit0` reproduces it: the provider throws without a response, `position` stays `-1`, and the log claims a file-size lookup failed.

For defect 4: kill the broker between the object delete and `deleteFileSegment`, restart, and read that segment.

For defect 5: request a range whose end exceeds the segment's `commitPosition`; the only trace is a DEBUG line.

The production numbers above come from aggregating 24h of `tiered_store.log` for one region and clustering all 16,902 ERROR lines by root cause.

- A transient upload failure reported as what it is — an append that did not land and will be reconciled on the next commit — at WARN, not as a file-size lookup error at ERROR.
- No blocking remote lookup inside a completion callback that runs on a netty IO thread.
- One log line per commit failure, with a stable field set and a discriminator for the outcome.
- Segment expiry that fails safe: metadata unregistered before the object is deleted.
- A shortened read logged at a level visible in production.
- `cacheBusy` comparing bytes against bytes.

### What Did You See Instead?

- 7,379 `get file size error after commit` ERROR lines in 24h in one region, none of which corresponded to an actual failed lookup. The real event — a transient connection reset by the object storage frontend, fully recovered by the next commit round — was indistinguishable from a genuine metadata failure.
- Those ERROR lines emitted on `AsyncHttpClient-3-N` threads.
- Three message formats for one event, one of them describing a success.
- 844 `NoSuchKey` read failures in 24h against 4 distinct object names, retried periodically forever.
- The truncation that precedes `splitMessageBuffer` failures logged only at DEBUG.
- A read-ahead cache whose busy check can never trigger.

To be explicit about what is **not** broken: the retry itself is correct and loses no data. `commitPosition` is not advanced on failure, the next `commitAsync` performs a real lookup, `correctPosition` takes the server-reported length as authoritative, and the stream is then rewound or rebuilt from `getCommitOffset()`. Measured over 24h: 4,739 distinct files across 7,274 failures (mean 1.5, max 13 for a single file), no permanently stuck file, and zero `PositionNotEqualToLength`. These defects are about reporting and about a latent IO-thread hazard, not about durability.

### Additional Context

I have a patch covering all six items on a branch, with `FileSegmentTest` extended to assert that `handleCommitException` never performs a remote lookup. `mvn -pl tieredstore test` passes (127 tests) and checkstyle is clean against `style/rmq_checkstyle.xml`. I will open the PR once this issue is numbered.

Two notes on the fix's shape, in case maintainers prefer a different split:

- Items 1–3 are one change: stop looking up the remote size in the callback, reconcile only from what the provider reported, and emit a single line with `result=REMOTE_LANDED | RETRY_AFTER_REWIND | RETRY_AFTER_RECONCILE`. `commitAsync`'s existing lookup-failure log joins the same vocabulary as `result=SIZE_LOOKUP_FAILED`. This also gives `GET_FILE_SIZE_ERROR` a single meaning again, so item 1 disappears by construction rather than needing a disambiguating branch.
- Items 4, 5 and 6 are independent one-liners and could be split into separate PRs if that is easier to review.

lizhimins:

- Fixes #11168

### Brief Description

Six defects in the tiered storage module, found while analysing 24h of production `tiered_store.log` in one region (302M lines, of which 16,902 ERROR). None of them affect durability — the retry path is correct and loses no data — but together they made the most frequent upload failure in production unreadable, and two are latent hazards.

**1. `FileSegment#handleCommitException` no longer performs a remote lookup, which fixes the misleading message and an IO-thread hazard at once.**

`TieredStoreException` defaults `position` to `-1`, colliding with `GET_FILE_SIZE_ERROR = -1L`. A provider that only sets `position` when it received an HTTP error response leaves it at `-1` for transport failures, so those took the same branch as a genuinely failed HEAD and were logged as `get file size error after commit` — a lookup that never ran. In production, 7,379 such lines in 24h, every one with `expect == commit + content`, i.e. all values local.

The same ternary's other side calls `this.getSize()`, a synchronous object-metadata request, from a callback registered via `.exceptionally(...)` with no executor. It therefore runs on whichever thread completed the future — a netty IO thread for a network provider, as the production thread names confirm (`AsyncHttpClient-3-N`).

The handler now reconciles only from the length the provider reported on the exception, which costs nothing, and otherwise leaves the input stream in place. `commitPosition` is not advanced on failure, so `needCommit()` stays true and the next `commitAsync` performs the lookup on the dispatcher thread, where that logic already existed. `GET_FILE_SIZE_ERROR` regains a single meaning, so a failed lookup can only be reported by the code that performs it. For a transport failure the number of remote lookups is unchanged: that path never queried the size here either.

**2. One log line per commit failure, with a shared vocabulary.**

`handleCommitException` logged three differently worded messages with three different field sets, one of which described a *success* (the append landed remotely and only the response was lost). Both commit-failure paths now emit a single line with the same fields and a `result` discriminator: `REMOTE_LANDED`, `RETRY_AFTER_REWIND`, `RETRY_AFTER_RECONCILE` from `handleCommitException`, and `SIZE_LOOKUP_FAILED` from `commitAsync` — which previously named the same quantity `buffer` instead of `content` and omitted `expect` and `remote`.

`commit` is now logged as of the append rather than after `correctPosition` overwrote it, so `commit + content == expect` holds on every path and can be read against `remote`.

The `!sizeKnown` case is tested first, which also makes `correctPosition(-1)` unreachable by construction rather than by a short-circuit that needs explaining.

**3. `FlatAppendFile#destroyExpiredFile` unregisters metadata before deleting the object.**

The opposite order leaves a metadata row pointing at a deleted object if the process dies in between. `recover()` reloads it on every restart and every later read of that segment fails with `NoSuchKey`. Production showed 844 such failures in 24h against only 4 distinct object names, concentrated on 3 of 11 affected instances — permanent phantom metadata, not transient errors. The new order fails safe: the worst case is an orphaned object, which costs storage instead of breaking reads.

**4. `FileSegment#readAsync` logs a shortened read at WARN instead of DEBUG.**

The truncated buffer surfaces much later as a `splitMessageBuffer` failure whose message carries no topic, queueId or offset, so the cause was invisible: 391 `message buffer offset exceeded limit` errors in 24h on one instance that could not be attributed to a queue from the logs.

**5. `cacheBusy` compares bytes against bytes.**

`fetcherCache.estimatedSize()` counts entries while `memoryMaxSize * 0.8` is a byte count, so the check was effectively always false. It now uses the weighted size, which is what `maximumWeight` bounds via the `SelectBufferResult#getSize` weigher.

`mvn -pl tieredstore test` — 127 tests, 0 failures, 0 errors. `mvn -pl tieredstore validate` reports 0 checkstyle violations against `style/rmq_checkstyle.xml`.

`FileSegmentTest#handleCommitExceptionTest` already covers all three outcomes and continues to pass unchanged: a provider exception without a position returns false, one carrying `setPosition(size * 2)` equal to the expected position returns true, and a non-`TieredStoreException` cause returns false. I extended the third case with `Mockito.verify(fileSpySegment, Mockito.never()).getSize()`, replacing a `getSize()` stub that the new code no longer reaches, so the no-remote-IO property is guarded by a test rather than only by the commit message.

The production figures quoted above come from aggregating 24h of `tiered_store.log` for one region and clustering all 16,902 ERROR lines by root cause.


tieredstore/src/main/java/org/apache/rocketmq/tieredstore/file/FlatAppendFile.java
```diff
@@ -286,10 +286,16 @@ public void destroyExpiredFile(long expireTimestamp) {
                     break;
                 }
 
+                // Unregister the metadata before deleting the object. The opposite order leaves a
+                // metadata row pointing at a deleted object if the process dies in between, and every
+                // later read of that segment then fails with NoSuchKey. An orphaned object is harmless.
+                long baseOffset = fileSegment.getBaseOffset();
+                fileSegmentTable.remove(0);
+                metadataStore.deleteFileSegment(filePath, fileType, baseOffset);
                 fileSegment.destroyFile();
```
RockteMQ-AI, line 289: **[Info]** The ordering change (metadata removal before object deletion) is correct for crash safety — an orphaned object is indeed harmless while orphaned metadata causes `NoSuchKey` on every subsequent read.

One consideration: if `metadataStore.deleteFileSegment()` succeeds but `fileSegment.destroyFile()` throws, the warn log fires but the segment is already unregistered from `fileSegmentTable`. This means the orphaned object will never be retried for cleanup. Consider whether a background sweep or a retry loop for `destroyFile()` would be worthwhile for production hygiene, though this is not a correctness issue.

tieredstore/src/main/java/org/apache/rocketmq/tieredstore/core/MessageStoreFetcherImpl.java
```diff
@@ -459,7 +459,12 @@ public CompletableFuture<GetMessageResult> getMessageAsync(
             return CompletableFuture.completedFuture(result);
         }
 
-        boolean cacheBusy = fetcherCache.estimatedSize() > memoryMaxSize * 0.8;
+        // The cache is bounded by maximumWeight (bytes, via the SelectBufferResult#getSize weigher),
+        // so compare against weightedSize() rather than estimatedSize(), which counts entries.
+        long cacheWeight = fetcherCache.policy().eviction()
```
RockteMQ-AI, line 464: **[Info]** Good catch — `estimatedSize()` counts entries while `memoryMaxSize` is in bytes, so the old comparison was dimensionally incorrect. Using `weightedSize()` from the eviction policy is the right fix.

The chained `.map(eviction -> eviction.weightedSize().orElse(0L)).orElse(0L)` is safe, but worth noting: if the cache has no eviction policy configured (e.g., in a test setup), `cacheWeight` will be 0 and `cacheBusy` will always be false, effectively disabling the backpressure. This matches the intent (unbounded cache = never busy), but a comment or assertion in test configs could prevent confusion.

RockteMQ-AI (approved):
### Summary

This PR fixes six defects in the tiered storage module, all identified through production log analysis (302M lines, 16,902 ERRORs in 24h). The changes are well-motivated, correctly scoped, and the test coverage properly validates the key behavioral change (no synchronous `getSize()` call on the exception handler thread).

The most critical fix is **defect #1** — removing the synchronous remote metadata lookup from the `exceptionally` callback, which was running on netty IO threads and blocking them. Deferring reconciliation to the next `commitAsync()` on the dispatcher thread is the correct approach.

The crash-safety fix in `FlatAppendFile#destroyExpiredFile` (metadata removal before object deletion) is also important — the previous ordering left metadata pointing at deleted objects if the process died between the two operations.

Overall, this is a high-quality PR with exceptional documentation. Each fix is backed by production evidence, and the unified logging format will make future debugging significantly easier.

### Findings
- **[Info]** FlatAppendFile.java:295 — Orphaned object cleanup after metadata deletion may need a background sweep
- **[Info]** MessageStoreFetcherImpl.java:464 — Cache weight fix is correct; note about unbounded cache behavior in tests

### Suggestions
- Consider adding a background sweep for orphaned objects in `destroyExpiredFile` to handle the case where `destroyFile()` throws after metadata is already deleted
- The unified logging format (`REMOTE_LANDED`, `RETRY_AFTER_REWIND`, `RETRY_AFTER_RECONCILE`, `SIZE_LOOKUP_FAILED`) is excellent — consider documenting these result codes in a developer guide or README

---

RockteMQ-AI (approved):
### Summary
Six defects in the tiered storage module, all found from production log analysis (302M lines, 16,902 ERRORs). None affect durability — the retry path is correct — but they made the most frequent upload failure unreadable and two are latent hazards. Well-motivated, production-tested fixes.

### Observations

- **Cache weight fix (MessageStoreFetcherImpl)**: Correct — the cache is bounded by `maximumWeight` (bytes via `SelectBufferResult#getSize` weigher), so `estimatedSize()` (entry count) was the wrong metric. Using `weightedSize()` from the eviction policy is the right fix.
- **`handleCommitException` no longer does remote lookup**: Significant improvement — removing a network call from the exception handler eliminates a potential cascading failure. Returning `false` when size is unknown and letting the next commit reconcile is the right approach.
- **`FILE_NOT_FOUND` handling (IndexStoreFile)**: Correct — objects can be deleted while reads are in flight. Downgrading to INFO and returning empty results avoids log noise. The `hasErrorCode` utility walks the cause chain with cycle detection (`cause != cause.getCause()`).
- **`readAsync` log upgrade**: Debug → warn when request exceeds commit position is appropriate for production observability.
- **Tests**: New tests for cache weight, `hasErrorCode`, and the updated commit exception test that verifies `getSize` is NOT called on the exception path. Good coverage.

LGTM.

---

lollipopjin (approved):
LGTM

Issue #8316, opened 2024-06-20
[Bug] Transaction producer didn't receive CheckTransacationState request of pending message


redlsz:
### Before Creating the Bug Report

- [X] I found a bug, not just asking a question, which should be created in [GitHub Discussions](https://github.com/apache/rocketmq/discussions).

- [X] I have searched the [GitHub Issues](https://github.com/apache/rocketmq/issues) and [GitHub Discussions](https://github.com/apache/rocketmq/discussions)  of this repository and believe that this is not a duplicate.

- [X] I have confirmed that this bug belongs to the current repository, not other repositories of RocketMQ.

### Runtime platform environment

unreleated

### RocketMQ version

5.x

### JDK Version

1.8

### Describe the Bug

<img width="1775" alt="image" src="https://github.com/apache/rocketmq/assets/103550934/e3c684c0-99ff-4c16-9079-970ec64da0bd">

In transaction message scenario, the first time execution result was UNKNOW. We expected to receive broker's check request later, but did not.

### Steps to Reproduce

1. Startup server in proxy cluster mode
2. Startup a remoting TransactionProducer
3. Sleep 5 seconds before sending one transaction message, and return UNKNOW in TransactionListener#executeLocalTransaction
4. Keep TransactionProducer running and don't send more messages
5. Observe whether TransactionListener#checkLocalTransaction will be called

### What Did You Expect to See?

TransactionListener#checkLocalTransaction will be called

### What Did You See Instead?

TransactionListener#checkLocalTransaction has not been called

redlsz:
<img width="1460" alt="image" src="https://github.com/apache/rocketmq/assets/103550934/d2c4ae89-a555-431e-9e95-2643a20b8047">

Transaction log indicates that the broker cannot find producer channel when trying to call client CheckTransactionState. 

<img width="1907" alt="image" src="https://github.com/apache/rocketmq/assets/103550934/63d25af2-292d-4d8e-92f8-18920eef3078">

In this case, the producer send message before its first heartbeat, which causes offline misjudgement when scanning at ClusterTransactionService#scanProducerHeartBeat. Eventually, this producer will not be registered to brokers.

RongtongJin:
[...]

PR #11171, opened 2026-09-16
[ISSUE #11170] Preserve transaction subscriptions before first heartbeat

### LINKED ISSUE #11170: [Bug] Proxy removes transaction subscriptions before the first producer heartbeat

Proxy cluster mode with a Remoting transaction producer. The state transition can be reproduced deterministically in a service-level unit test, without a running broker or timing sleeps.

### RocketMQ version

Current `develop`: `80e1ae55773c2330d2005b86f020ed028f649e94`.

### JDK Version

Local regression validation: Amazon Corretto 11.0.23, Maven 3.9.8, macOS arm64.

### Describe the Bug

`ClusterTransactionService.scanProducerHeartBeat()` deletes a transaction subscription as soon as `ProducerManager.groupOnline(group)` returns false. A transaction send can create this subscription before the producer's first heartbeat registers its channel in the proxy.

If the scan runs in that interval, returning `null` from `groupClusterData.computeIfPresent()` removes the group-to-cluster mapping. A later producer heartbeat only updates `ProducerManager`; it does not rebuild the deleted transaction subscription. If the producer sends no further transaction messages, subsequent scans cannot register that group with the broker, so transaction checks cannot reach the producer.

This is a follow-up to #8316, which reports the same race and was closed by the stale bot on 2026-09-15. It remains reproducible in the current implementation. See also #5862 and its fix #5865: preventing indefinite heartbeats for producers that exit before their first heartbeat remains necessary.

The earlier fix proposal #8320 by @redlsz introduced a subscription timestamp and grace period for this race. It was closed as stale on 2026-05-18 without being merged, so those changes are not present in the current implementation.

### Steps to Reproduce

Deterministic service-level sequence:

1. Add a transaction subscription for a group and a topic with a valid broker-cluster route.
2. Leave the group absent from `ProducerManager`, representing a pending first producer heartbeat.
3. Run `scanProducerHeartBeat()`.
4. Register the group, or change the mocked `groupOnline(group)` result to `true`.
5. Run `scanProducerHeartBeat()` again, without adding another transaction subscription.

The group is removed at step 3. Step 5 has no subscription from which to build a broker heartbeat.

An end-to-end trigger is to send a transaction whose local result is `UNKNOWN` before the first producer heartbeat, keep the producer alive, and send no additional transaction messages. The failure depends on the scan falling between subscription creation and producer registration; it does not happen on every run.

A recently created transaction subscription should survive a delayed first producer heartbeat and resume broker heartbeats once the producer becomes online.

For the #5862 case, offline groups should receive no broker heartbeats, and subscriptions for producers that never register should eventually be reclaimed. Explicit group unregistration should still remove subscriptions immediately.

### What Did You See Instead?

The subscription is removed immediately while the group is not yet registered. Registration alone does not restore it. The broker can then fail to find a channel for transaction checks, and the producer's transaction checker is not invoked.

### Proposed Fix

Keep a millisecond `lastActiveTimestamp`, using `System.currentTimeMillis()`, on the existing `ClusterData` entries. Refresh it when subscriptions are added/replaced and while the group is observed online. When the group is offline, skip broker heartbeats immediately and expire subscriptions only after `channelExpiredTimeout` (currently 120 seconds by default). Preserve the existing explicit-unsubscribe path.

Regression coverage should include delayed first registration, never-registered expiration, renewed subscription activity, online activity, replacement, and explicit unsubscription during the grace period.

qianye1001:

- Fixes #11170
- Related: #8316, #8320, #5862, #5865

### Brief Description

A transaction send can add a group-to-cluster subscription before the producer's first heartbeat registers its channel with the proxy. If the transaction heartbeat scan runs in between, it removes the subscription. The later producer heartbeat does not rebuild it, so a producer that sends no further transaction messages can miss transaction checks.

Retain recently active subscriptions while the group is not yet online:

- Add `lastActiveTimestamp` in milliseconds, using `System.currentTimeMillis()`, to the existing `ClusterData` entries and refresh it on subscription updates and online scans.
- Skip broker heartbeats immediately when the group is offline, preserving the behavior required by #5862.
- Reclaim offline subscriptions after the existing `channelExpiredTimeout` (120 seconds by default). Explicit unsubscription still removes them immediately.

`ClusterData` currently has no timestamp. A channel's last-update time is unavailable before its first registration, and per-message `TransactionData` expiry has a different lifecycle. This change adds no new configuration, map, or background task.

An earlier proposal by @redlsz, #8320, addressed the same race by introducing a `ClusterDataSet.lastAddTimestamp` and a separate transaction-group timeout. It was closed as stale without being merged. This patch keeps the existing collection structure and timeout configuration. Unlike #8320, it sends no broker heartbeats while the producer group is offline, including during the grace period; it also refreshes subscription activity when the group is observed online.

- The delayed-first-heartbeat regression fails against unmodified `develop` at `80e1ae55773c2330d2005b86f020ed028f649e94` because the subscription is removed.
- Added six deterministic regression tests covering delayed registration, never-registered expiry, repeated subscription updates to the same cluster, online activity, replacement, and explicit unsubscription. All six pass with the fix; they use no timing sleeps.
- Ran with Amazon Corretto 11.0.23 and Maven 3.9.8:

  ```sh
  mvn -B -pl proxy -am \
    -Dtest=ClusterTransactionServiceTest,AbstractTransactionServiceTest,TransactionDataManagerTest \
  ```

  Result: **16 passed, 0 failures/errors, 1 skipped**. The skip is the existing macOS guard in `TransactionDataManagerTest.testWaitTransactionDataClear`. Checkstyle and SpotBugs passed.

Validation is at the service/unit-test level; a full broker/client E2E run has not been performed for this patch.


RockteMQ-AI (approved):
### Summary

This PR fixes a race condition where a transaction send can add a subscription before the producer's first heartbeat, causing the transaction heartbeat scan to remove it prematurely. The fix adds a grace period using `lastActiveTimestamp` on `ClusterData` entries, allowing subscriptions to survive until `channelExpiredTimeout` elapses.

**Verdict: APPROVED** ✅

The implementation is clean, minimal, and well-tested:

- **Correctness**: Properly addresses the race condition without breaking existing behavior. Explicit unsubscription still works immediately.
- **Performance**: Minimal overhead (one `long` per `ClusterData`, cheap `System.currentTimeMillis()` calls).
- **Tests**: 6 comprehensive regression tests covering all scenarios (delayed registration, expiry, refresh, replacement, explicit unsubscribe). Tests are deterministic with no timing sleeps.
- **Compatibility**: No API or wire protocol changes. Fully backward compatible.

The approach of reusing `channelExpiredTimeout` is elegant — no new configuration needed. The code changes are surgical and preserve the existing structure.

**Minor observation**: The test helper `expireSubscriptions()` uses reflection to access the private `lastActiveTimestamp` field. This is acceptable for testing but makes the test sensitive to field renames. Consider adding a package-private setter for testing if this becomes a maintenance burden.

---

RongtongJin (approved):

RockteMQ-AI (approved):
### Summary
This PR fixes a race condition where transaction subscriptions could be prematurely removed before the producer's first heartbeat. Instead of immediately removing subscriptions when the producer channel is not yet online, the code now waits until the channel timeout expires, using `lastActiveTimestamp` to track activity.

### Findings
- **[Info]** `TransactionSubscriptionManager.java` — The grace period logic using `lastActiveTimestamp` is clean. The `isExpired` check correctly distinguishes between "producer not yet connected" and "producer disconnected".
- **[Info]** `TransactionSubscriptionManagerTest.java` — Tests cover the key scenarios: subscription preserved before first heartbeat, expired after timeout, and activity refresh.

### Suggestions
- The grace period duration (`channelExpiredMillis`) is reused from the channel timeout config. This is reasonable but worth documenting that transaction subscriptions share the same timeout as regular channels.

Good fix for a subtle race condition.

---

lizhimins (approved):

PR #11173, opened 2026-09-16
[ISSUE #11172] Respond when async send storage future fails

### LINKED ISSUE #11172: [Bug] Async send leaves requests unanswered when store future fails

- [x] I searched open/closed issues and pull requests for `asyncSendEnable`, `asyncPutMessageFuture`, exceptional completion, and send timeout, and found no equivalent report.
- [x] I confirmed the affected code is in Apache RocketMQ.

Deterministic broker unit test on Windows; the control-flow defect is platform independent.

### RocketMQ version

`develop` at `bc33e8e4d7b25089af5f51bc669bdfedabfebe7d`.

### JDK Version

Amazon Corretto 8u432 for the regression test.

### Describe the Bug

When `asyncSendEnable=true`, `SendMessageProcessor` returns `null` immediately and relies on the `MessageStore` future to send the response later. Both the single-message and batch paths attach only `thenAcceptAsync(...)` to that future.
If `asyncPutMessage(...)` / `asyncPutMessages(...)` completes exceptionally, the consumer passed to `thenAcceptAsync` is never invoked. Because the processor already returned `null`, no `doResponse(...)` path remains: the client receives no broker response and waits until an outer remoting timeout or connection teardown. The after-send hook is skipped as well.

The single-message branch and the batch branch have the same shape:

```java
asyncPutMessageFuture.thenAcceptAsync(putMessageResult -> {
    RemotingCommand responseFuture = handlePutMessageResult(...);
    if (responseFuture != null) {
        doResponse(ctx, request, responseFuture);
    }
    sendMessageCallback.onComplete(sendMessageContext, response);
}, executor);
return null;
```

This is an observable asynchronous-completion gap, not just a theoretical null check. The store future contract can propagate exceptional completion from lower async stages; for example `CommitLog#handleDiskFlushAndHA` composes flush and HA futures with `thenCombine`.

### Steps to Reproduce

1. Enable async send in a `SendMessageProcessorTest` fixture.
2. Make `MessageStore.asyncPutMessage(...)` return a `CompletableFuture<PutMessageResult>` completed exceptionally with `RuntimeException("store write failed")`.
3. Invoke `processRequest` with a normal `SEND_MESSAGE` request and capture `channel.writeAndFlush`.
4. `processRequest(...)` returns `null`, as expected for async mode.
5. Wait up to one second for a response.

On the unmodified baseline the regression fails deterministically:

```text
ConditionTimeoutException: response was not fulfilled within 1 seconds
```

The preceding 10 reactor modules succeed; only this broker regression fails.

An exceptional store future should complete the request promptly with `SYSTEM_ERROR`, log the original throwable on the broker, and execute the normal after-send hook exactly once. The client should not have to discover a completed store failure through a later request timeout.

### What Did You See Instead?

No response is written at all. The client remains pending until external timeout/connection handling terminates the request.

### Proposed scope

Handle both single-message and batch async-send futures symmetrically. Preserve all existing success/result-code behavior, synchronous-send behavior, metrics, and protocol formats. Add deterministic regressions for exceptional completion and retain existing success coverage.

zmuxuny:

Fixes #11172

### Brief Description

With `asyncSendEnable=true`, `SendMessageProcessor` returns `null` before the store operation completes and relies on the store future callback to write the broker response. Both the single-message and batch paths used `thenAcceptAsync`, so an exceptionally completed store future skipped the callback entirely. The client then received no broker response and could only terminate through an outer request timeout or connection teardown.

This change handles both completion outcomes for single and batch async sends:

- normal `PutMessageResult` values keep the existing response/status/metrics behavior;
- exceptional completion logs the original throwable on the broker and immediately responds with `SYSTEM_ERROR` using a generic client-safe remark;
- the existing after-send hook still runs once for the exceptional path;
- synchronous send behavior and wire/storage formats are unchanged.
### Fail-before evidence

Baseline: `develop` at `bc33e8e4d7b25089af5f51bc669bdfedabfebe7d`, Amazon Corretto 8u432.

A deterministic regression makes `MessageStore.asyncPutMessage(...)` return a future already completed exceptionally, invokes a normal `SEND_MESSAGE`, and waits one second for `channel.writeAndFlush`.

On the unmodified baseline:

```text
ConditionTimeoutException: response was not fulfilled within 1 seconds
```

The preceding 10 modules in the `broker -am` reactor all succeeded. The request processor had returned `null`, but no response was ever written.

Focused exceptional-completion regressions cover both paths:

- single-message `asyncPutMessage` failure returns `SYSTEM_ERROR` instead of leaving the request pending;
- batch `asyncPutMessages` failure has the same behavior.

Final verification command:

```sh
mvn -B -ntp -pl broker -am \
  -Dtest=SendMessageProcessorTest \
```

Result: **18 tests passed, 0 failures/errors/skips** in `SendMessageProcessorTest`. All 11 modules in the `broker -am` reactor completed successfully. Checkstyle reported **0 violations** in every checked module, and SpotBugs reported **0 bug instances / 0 errors** in every reactor module where the project runs SpotBugs (the generated `rocketmq-proto` module uses its existing project-configured skip).

`git diff --check` also passes.


RockteMQ-AI (approved):
### Summary
This PR fixes a real bug where clients hang indefinitely when async storage fails. The change from `thenAcceptAsync` to `whenCompleteAsync` correctly handles both normal and exceptional completion paths for single-message and batch async sends.

### Review
- **Correctness** ✅ — The exception handler properly sends `SYSTEM_ERROR` response to clients, invokes the after-send callback, and avoids leaking internal error details
- **Performance** ✅ — No performance concerns; same execution model with additional error handling
- **Tests** ✅ — Two new tests cover both single-message and batch async failure scenarios with proper async verification
- **Compatibility** ✅ — No API/wire format changes; synchronous send path unchanged

### Minor Observation
The error path logs the exception and responds to the client, but does not emit broker-side metrics for the failure (the normal path goes through `handlePutMessageResult` which likely handles metrics). This may be intentional since the exception itself is logged, but worth considering if monitoring dashboards should track async storage failures separately.

Overall: Clean, focused fix with excellent PR description and test coverage. LGTM.

---

PR #11175, opened 2026-09-17
[ISSUE #11174] Restore receipt handles when proxy ACK fails

### LINKED ISSUE #11174: [Bug] Restore managed receipt handles when gRPC ACK fails exceptionally
## Bug

The gRPC ACK path removes the proxy-managed receipt-handle mapping before the Broker ACK completes. If the ACK call then fails exceptionally (transport/runtime failure), the mapping is never restored.

This matters when Proxy auto-renew has already replaced the client-visible receipt handle with a newer one. A retry from the client normally resolves its original handle through `MessageReceiptHandle`; after the failed ACK removed that mapping, the retry can fall back to the stale client handle instead.

## Current flow

`AckMessageActivity.getHandleString` calls `messagingProcessor.removeReceiptHandle(...)` first. The returned `MessageReceiptHandle` supplies the latest renewed handle, but only the string is retained. `processAckMessage` then calls `ackMessage(...)` and converts an exceptional future into an error response without re-registering the removed mapping.

The batch ACK path has the same ordering: it removes all managed handles while constructing `ReceiptHandleMessage`s before `batchAckMessage(...)` completes.
## Reproduction

On `develop@bc33e8e4d7b25089af5f51bc669bdfedabfebe7d`, a regression test registers a newer managed handle, makes `removeReceiptHandle` return it, and makes `ackMessage` return an already exceptionally-completed future.

The request correctly returns an internal error, but verification fails because `addReceiptHandle(...)` is never invoked. The preceding 11 modules in the `proxy -am` reactor all succeed; only this regression fails.

## Expected behavior

Treat exceptional completion as an indeterminate/transient ACK attempt: restore any managed receipt-handle entry removed for that attempt so a client retry can still resolve to the latest handle.

- Normal ACK success keeps the current remove-on-ACK behavior.
- Explicit Broker/business ACK results keep their current semantics.
- Synchronous failures after removal should restore the mapping as well.
- Batch ACK should restore the removed mappings only when the whole batch invocation fails exceptionally/synchronously, not for ordinary per-entry business results.

No wire format or Broker ACK protocol change is needed.

zmuxuny:

Fixes #11174

### Brief Description

The gRPC ACK path removes a proxy-managed receipt-handle mapping before the Broker ACK completes. That mapping can contain a newer receipt handle produced by automatic invisible-time renewal. If the subsequent ACK invocation throws or its future completes exceptionally, the mapping is lost and a client retry can fall back to its stale receipt handle.

This change keeps the existing remove-on-ACK behavior for normal Broker/business results, but restores the removed managed handle when the ACK attempt itself fails exceptionally or synchronously. The same protection is applied to batch ACK: mappings removed while preparing the batch are restored only when the batch invocation fails as a whole.
### Fail-before evidence

Baseline: `develop@bc33e8e4d7b25089af5f51bc669bdfedabfebe7d`.

A deterministic regression makes `removeReceiptHandle(...)` return a newer managed handle and then makes `ackMessage(...)` return an already exceptionally-completed future. On the unmodified baseline, the request returns an internal error but verification fails because `addReceiptHandle(...)` is never invoked:

```text
Wanted but not invoked:
messagingProcessor.addReceiptHandle(...)
```

The preceding 11 modules in the `proxy -am` reactor succeeded; only this regression failed.

Focused coverage now includes:

- exceptional single-message ACK restores the managed handle;
- synchronously thrown ACK restores the managed handle;
- exceptional batch ACK restores all removed managed handles;
- successful ACK does not restore the mapping.

Final verification:

```sh
mvn -B -ntp -pl proxy -am -Dtest=AckMessageActivityTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: **6 tests passed, 0 failures/errors/skips**. All **12 modules** in the `proxy -am` reactor completed successfully. Checkstyle reported **0 violations** and SpotBugs **0 bug instances / 0 errors** in every module where those checks run. `git diff --check` also passes.


RockteMQ-AI (approved):
### Summary
This PR fixes a real correctness issue in the proxy ACK path: when the Broker ACK fails (async or sync), the receipt-handle mapping — which may contain a newer handle from automatic invisible-time renewal — was permanently lost, forcing the client to fall back to a stale handle on retry. The fix captures the `ReceiptHandleContext` before removal and restores it in both the `whenComplete` (async failure) and `catch` (sync throw) paths.

### Observations
- **Correctness**: Both failure paths (async exceptional completion and synchronous throw) are covered. The restore iterates by index against `request.getEntries(i)`, which is safe because `handleContexts` is built in the same loop order.
- **Error isolation**: Restore failures are caught and logged without propagating — correct, since we must not mask the original error.
- **Tests**: Four new tests cover all paths (async fail → restore, sync throw → restore, batch partial fail → restore, success → no restore). Good coverage.
- **Minor style**: `ReceiptHandleContext` fields are package-private; `private final` would be more encapsulated, but this is cosmetic.

LGTM.

---

PR #11183, opened 2026-09-20
[ISSUE #11182] Include rpcCode in the authorization cache key and short-circuit whitelisted requests

### LINKED ISSUE #11182: [Bug] StatefulAuthorizationStrategy cache key omits rpcCode, leaking whitelist ALLOW to non-whitelisted requests

- OS: Linux (reproduced in a Docker container)
- RocketMQ: branch `develop`, version `5.5.1`, commit `050d2b42b`

### JDK Version

- Compiler: JDK 1.8

### Describe the Bug

`StatefulAuthorizationStrategy` builds its authorization cache key from `channelId + subjectKey + resourceKey + actions + sourceIp`, omitting `rpcCode`. The rpcCode whitelist check, however, lives inside `doEvaluate()` (which runs only on a cache miss, inside the cache mapping function).

So a whitelisted request caches an ALLOW under a key that ignores `rpcCode`, and a subsequent non-whitelisted request with the same channel/subject/resource/actions/sourceIp reuses that ALLOW and skips ACL evaluation entirely — a privilege escalation.

### Steps to Reproduce

1. Configure `authorizationStrategy=StatefulAuthorizationStrategy` and put some rpcCode in `authorizationWhitelist`.
2. On one channel, send a whitelisted request that builds a given resource/action context (e.g. a Heartbeat producing a `Group:G` SUB context).
3. On the same channel, send a non-whitelisted request that builds the exact same `(Group:G, SUB)` context (e.g. `AckMessageRequest` / `NotifyClientTerminationRequest` / `ChangeInvisibleDurationRequest` / `ForwardMessageToDeadLetterQueueRequest`).
4. Observe that the second request is allowed even though its ACL check would deny it.

The non-whitelisted request should be denied (its ACL check should run).

### What Did You See Instead?

The non-whitelisted request is allowed because it hits the cached ALLOW produced by the earlier whitelisted request.

### Additional Context

Suggested fix: (1) short-circuit the whitelist before the cache lookup so a whitelisted request neither reads nor writes the cache; (2) include `rpcCode` in the cache key so the cache dimension covers every input that affects the decision. The accompanying PR implements both and adds a regression test.

R0CKing666:

- Fixes #11182

### Brief Description

`StatefulAuthorizationStrategy` built its authorization cache key from `channelId + subjectKey + resourceKey + actions + sourceIp`, omitting `rpcCode`, while the rpcCode whitelist check lived inside `doEvaluate()` (the cache mapping function). A whitelisted request therefore cached an ALLOW under a key shared with non-whitelisted requests, letting them skip ACL evaluation entirely.

This change:

- short-circuits whitelisted rpcCodes before the cache lookup, so a whitelisted request neither reads nor writes the cache;
- includes `rpcCode` in the cache key, so the cache dimension covers every input that affects the decision.

- `StatefulAuthorizationWhitelistCacheTest`: regression test — a non-whitelisted request sharing the whitelisted request's channel/subject/resource/actions/sourceIp is now denied instead of reusing a cached ALLOW.
- `StatefulAuthorizationStrategyTest`: existing 4 tests still pass.
- Built with `mvn -pl auth -Dtest=StatefulAuthorizationWhitelistCacheTest,StatefulAuthorizationStrategyTest test`.


RockteMQ-AI (approved):
LGTM. Code changes look good.

---

PR #11185, opened 2026-09-21
[ISSUE #10859] Fix exhausted tiered index write result

### LINKED ISSUE #10859: [Bug] IndexStoreService reports success after index writes exhaust retries
macOS; reproduced with a deterministic unit test in the `tieredstore` module.

## RocketMQ version
Branch: `develop`

## JDK Version
Eclipse Temurin 17.0.19+10

## Describe the Bug
`IndexStoreService.putKey` retries an index write three times. If every attempt fails, it logs an error but returns `AppendResult.SUCCESS`, so callers can treat an index entry that was never stored as successful.

## Steps to Reproduce
1. Set `currentWriteFile` to an `IndexFile` whose `putKey` returns `FILE_FULL`.
2. Keep file rotation from replacing that test file.
3. Call `IndexStoreService.putKey`.
4. Observe three failed attempts followed by a `SUCCESS` result.

The method should return the final failed `AppendResult` after all retries are exhausted.

## What Did You See Instead?
The method returns `SUCCESS` despite all three writes returning `FILE_FULL`.

## Additional Context
The normal success path and the existing three-attempt retry behavior do not need to change.

beautyarbutin:

- Fixes #10859

### Brief Description

`IndexStoreService.putKey` returned `SUCCESS` after all three index write attempts failed. This change retains the result of each attempt and returns the final failure once retries are exhausted, while preserving the existing success path and `FILE_FULL` file rotation behavior.

The regression test exercises `FILE_FULL`, `FILE_FULL`, and `UNKNOWN_ERROR`, then verifies that the final error is propagated after three attempts and that file rotation occurs only for the two `FILE_FULL` results.

- Confirmed the new regression test failed before the production change: expected `UNKNOWN_ERROR` but received `SUCCESS`.
- Ran `mvn -pl tieredstore -am -DskipITs -Dtest=IndexStoreServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`.
- Result: 11 tests passed; Checkstyle reported 0 violations; SpotBugs reported no errors or warnings.


RockteMQ-AI (commented):
### Summary
Fixes a clear bug where `IndexStoreService.putKey` returned `SUCCESS` after all three write attempts failed. The fix correctly propagates the last failure result (`AppendResult.UNKNOWN_ERROR` or `FILE_FULL`) back to the caller.

### Findings
- **[Critical]** This was a data integrity bug — returning SUCCESS on failure could cause the caller to believe the index was written when it was not. Good catch.
- **[Info]** The test `putKeyReturnsLastFailureAfterRetriesTest` is well-structured: uses Mockito to simulate sequential failures and verifies both the return value and the retry count (3 putKey calls, 2 createNewIndexFile calls).

### Overall
Clean, minimal fix with proper test coverage. LGTM.

---

RockteMQ-AI (approved):
### Summary

---

PR #11190, opened 2026-09-21
[ISSUE #11189] Stop mutating cached ACL/User metadata in place on the admin write path

### LINKED ISSUE #11189: [Bug] Cached ACL/User metadata is mutated in place by the admin plane, racing the authorization hot path

OS: all platforms

### RocketMQ version

branch: develop
version: 5.5.1

### Describe the Bug

`AuthorizationMetadataManagerImpl.createAcl/updateAcl/deleteAcl` first read the instance returned by `getAcl` as the "old value", then call `updatePolicy`/`deletePolicy` on it in place, and then hand the same object to `updateAcl` for persistence. `AuthenticationMetadataManagerImpl.updateUser` likewise calls `setPassword/setUserType/setUserStatus` directly on the `User` returned by `getUser`.

`getAcl`/`getUser` are ultimately implemented by `LocalAuthorizationMetadataProvider`/`LocalAuthenticationMetadataProvider`, which return the very object reference held in the Caffeine cache (no defensive copy). As a result, the admin write path and the broker authorization/authentication hot path read the same mutable object. Readers can observe a half-updated state (some fields old, some new), breaking atomic visibility of ACL/User, and concurrent structural modification can throw `ConcurrentModificationException`.

### Steps to Reproduce

1. Enable ACL 2.0 (local metadata provider) and start the broker.
2. Create a user and an ACL.
3. While clients continuously issue requests that trigger authorization (`AclAuthorizationHandler` -> `getAcl` -> cached instance), concurrently send `UPDATE_ACL`/`CREATE_ACL` (or `UPDATE_USER`) via `AdminBrokerProcessor`.
4. The manager mutates the cached `Acl`/`User` in place (e.g. `oldAcl.updatePolicy(...)` in `AuthorizationMetadataManagerImpl#updateAcl`) before the RocksDB write and cache invalidation.

Observed: authorization results can flip between ALLOW/DENY (half-updated policy set), requests can fail with `ConcurrentModificationException`, and if the RocksDB write fails, the in-memory cache diverges from disk.

Readers on the hot path should only ever observe a complete old value or a complete new value; a new value should become visible only after it has been persisted, and concurrent reads/writes should not throw `ConcurrentModificationException`.

### What Did You See Instead?

In-place mutation of the cached object is visible to readers before persistence, can produce torn reads, and can throw `ConcurrentModificationException`.

### Additional Context

Root cause: the local providers return the cached mutable reference without a defensive copy, and the managers mutate that reference in place.

- `LocalAuthorizationMetadataProvider#getAcl` returns `aclCache.get(...)` directly.
- `LocalAuthenticationMetadataProvider#getUser` returns `userCache.get(...)` directly.
- `AuthorizationMetadataManagerImpl#createAcl`/`updateAcl`/`deleteAcl` mutate `oldAcl` in place.
- `AuthenticationMetadataManagerImpl#updateUser` mutates the cached `User` in place.

Suggested fix: the write path must not reuse the cached instance. Construct a new object (deep copy) and apply the change to the copy before persisting it, so the cached instance is only ever replaced atomically after a successful write.

R0CKing666:

- Fixes #11189

### Brief Description

The admin write path mutates the cached ACL/User objects in place. `AuthorizationMetadataManagerImpl.createAcl/updateAcl/deleteAcl` read the cached `Acl` via `getAcl` and then call `updatePolicy`/`deletePolicy` on it directly; `AuthenticationMetadataManagerImpl.updateUser` calls `setPassword/setUserType/setUserStatus` on the cached `User`. Because `LocalAuthorizationMetadataProvider.getAcl` and `LocalAuthenticationMetadataProvider.getUser` return the very object held by the Caffeine cache (no defensive copy), the admin write and the authorization/authentication hot path share the same mutable object. Readers can therefore observe a half-updated state, and concurrent structural modification can throw `ConcurrentModificationException`; if the RocksDB write fails, the in-memory cache also diverges from disk.

This change makes the write path copy-on-write: the managers build a deep copy of the cached instance, apply the change to the copy, and persist the copy. The cached instance is only ever replaced atomically (via cache invalidation + reload) after a successful write, so the hot path keeps reading the cached reference without any per-request copy cost.

- Added regression tests verifying that `updateAcl`/`deleteAcl`/`updateUser` no longer mutate a previously-read cached instance, while a fresh read still reflects the update.
- Ran the full `rocketmq-auth` test suite: 96 tests, 0 failures, 0 errors.
- Checkstyle: 0 violations.


RockteMQ-AI (approved):
LGTM. Looks good.

---

PR #11192, opened 2026-09-21
[ISSUE #11191] Fix Lite topic consumer losing messages on event re-dispatch during a pop

### LINKED ISSUE #11191: [Bug] Fix Lite topic consumer losing messages on event re-dispatch during a pop

OS: macOS / Linux

### RocketMQ version

branch: develop

### JDK Version

Oracle JDK 11

### Describe the Bug

When a consumer subscribes to multiple lite topics with FIFO (ordered) consumption, a lite topic can get stuck after delivering only its first message. Its remaining messages are never delivered even though they exist: the event queue becomes empty and the consumer offset stops advancing.

### Steps to Reproduce

1. A consumer subscribes to two lite topics (e.g. `lt1` and `lt2`), each with 5 messages.
2. One pull reads `lt1`'s first message (now in-flight and FIFO-blocked) together with `lt2`'s messages.
3. While the broker is still assembling the same pull, `lt1`'s first message is acked, which advances the offset, unblocks FIFO and re-dispatches `lt1`'s event back into the client queue.
4. The re-dispatched `lt1` event is dropped because `lt1` had already been recorded in the per-request dedup set.

The consumer eventually receives all messages of every subscribed lite topic (5 + 5 = 10).

### What Did You See Instead?

`lt1` only yields its first message, so the consumer receives 6 messages instead of 10. The event queue is empty and `lt1`'s offset stays at 0, so the remaining messages are never delivered.

### Additional Context

Root cause: a lite topic is recorded in the per-request dedup set before it is actually handled. When it is FIFO-blocked and then re-dispatched (via ack unblock) within the same pull, the re-dispatched event is treated as a duplicate and silently dropped.

f1amingo:

- Fixes #11191

### Brief Description

When a consumer subscribes to multiple lite topics with FIFO consumption, a lite topic could get stuck after delivering only its first message. A lite topic was recorded in the per-request dedup set before it was actually handled; when it was FIFO-blocked and then re-dispatched (via ack unblock) within the same pull, the re-dispatched event was treated as a duplicate and silently dropped, so the remaining messages were never delivered.

This PR marks the dedup entry only after a lite topic is actually handled under the pop lock (not when it is FIFO-blocked), so a re-dispatched event is retried instead of dropped. The marking is moved into `popLiteTopic` and its visibility is narrowed accordingly.

Added a regression test (`PopLiteMessageProcessorEventLossTest`) that deterministically reproduces the concurrent ack-unblock event loss; it fails on the buggy code and passes with the fix. The existing `PopLiteMessageProcessorTest` remains green.


RockteMQ-AI (commented):
### Summary
This PR fixes a message-loss bug in the Lite pop consumer where a FIFO-blocked lmq gets incorrectly deduped when re-dispatched by a concurrent ack within the same `popByClientId` iteration. The fix moves the `processed` set ownership from the caller into `popLiteTopic`, marking an lmq as handled only after the lock is acquired and it is confirmed not FIFO-blocked.

### Findings

- **[Info]** `PopLiteMessageProcessor.java:260` — The change from `!processed.add(lmqName)` to `processed.contains(lmqName)` is correct and well-reasoned. The semantic shift (caller checks, callee adds under lock) is the core of the fix.

- **[Info]** `PopLiteMessageProcessor.java:290` — The visibility change from `public` to package-private for `popLiteTopic` is appropriate since the method is now an internal implementation detail with the `processed` set parameter. Good cleanup.

- **[Info]** `PopLiteMessageProcessor.java:297-298` — The `processed.add(lmqName)` placement after the FIFO-block check and inside the lock is the key correctness point. If the lock fails (line 293), the lmq is correctly left unmarked. If FIFO-blocked (line 296), it is also left unmarked so re-dispatch will retry it. This is correct.

- **[Warning]** `PopLiteMessageProcessorEventLossTest.java:183` — The `ackLite` helper uses a busy-wait spin loop (`while (!lockService.tryLock(lockKey))`) which could theoretically spin forever if the lock is never released. In a test context this is acceptable, but consider adding a timeout/deadline guard to prevent test hangs in CI:
  ```java
  long deadline = System.currentTimeMillis() + 5_000L;
  while (!lockService.tryLock(lockKey)) {
      if (System.currentTimeMillis() > deadline) {
          throw new RuntimeException("ackLite lock timeout");
      }
      Thread.sleep(1);
  }
  ```

- **[Info]** `PopLiteMessageProcessorEventLossTest.java` — The test design is thorough: it uses a spy on `popLiteTopic` to deterministically inject the concurrent ack at the right moment, and the drain loop ensures all messages are eventually received. The 5-iteration loop provides good coverage against timing-sensitive failures.

- **[Info]** `PopLiteMessageProcessorTest.java` — Existing tests are properly updated to match the new `popLiteTopic` signature. The `testPopLiteTopic_fifoBlocked` and `testPopLiteTopic_lockFailed` tests correctly assert that `processed` does NOT contain the lmq name after these failure paths, which validates the fix's core invariant.

### Suggestions
- Consider adding a brief Javadoc on the `processed` parameter in `popLiteTopic` explaining the ownership contract (caller checks via `contains`, callee adds under lock after confirming not blocked). This will help future maintainers understand why the add is deferred.

Overall this is a well-designed fix with excellent test coverage. The root cause analysis in the PR description is clear and the fix is minimal and targeted.

RockteMQ-AI (approved):
### Summary
Fixes a real bug where a Lite topic could get stuck after delivering only its first message during FIFO consumption. The root cause was that the dedup set marked a lmq as processed before it was actually handled — when it was FIFO-blocked and then re-dispatched via ack-unblock within the same pop, the re-dispatched event was treated as a duplicate and silently dropped.

The fix moves the `processed.add(lmqName)` into `popLiteTopic` after the FIFO check, so a FIFO-blocked lmq is never marked and its later re-dispatch is properly retried. The regression test deterministically reproduces the bug.

Clean fix with good test coverage.

---

lizhimins (approved):

PR #10857, opened 2026-08-08
[ISSUE #10852] fix(broker): sort revive offsets without overflow

yuluo-yx:
### Which Issue(s) This PR Fixes

- Fixes #10852

### Brief Description

Uses Long.compare to keep Pop revive checkpoints ordered by their full-width long offsets. Adds a regression test with offsets spanning the long range.

### How Did You Test This Change?

`mise x java@temurin-17 -- mvn -Dmaven.repo.local=/private/tmp/rocketmq-m2 -pl broker -am -DskipITs -Dtest=PopReviveServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`

The targeted tests, Checkstyle, and SpotBugs completed successfully.

PR #11198, opened 2026-09-22
[ISSUE #10852] fix(broker): sort revive offsets without overflow

### LINKED ISSUE #10852: Pop revive service may process large revive offsets out of order
### Description

PopReviveService builds its checkpoint list with a comparator that subtracts two long revive offsets and casts the difference to int. Offsets separated by more than Integer.MAX_VALUE can be reversed, causing mergeAndRevive to process checkpoints in the wrong order.

### Expected behavior

Compare revive offsets with Long.compare so checkpoint order remains ascending across the full long range.

beautyarbutin:

- Fixes #10852

### Brief Description

`ConsumeReviveObj.genSortList()` previously ordered revive offsets by subtracting two `long` values and narrowing the result to `int`. When the offset gap exceeded the integer range, the narrowing conversion could reverse the comparison result and process revive checkpoints out of order.

This change uses `Long.compare` so the full offset range is ordered correctly. It also adds a deterministic regression test with an offset gap just beyond `Integer.MAX_VALUE`.

This reimplements the fix from the earlier, closed PR #10857 against the current `develop` branch. Thanks to @yuluo-yx for the original report and patch.

- `mvn -B -ntp -pl broker -DskipITs -Dtest=PopReviveServiceTest test`
- Result: 13 tests passed, with no failures or errors.
- Checkstyle: 0 violations.
- SpotBugs: 0 bugs and 0 errors.


RockteMQ-AI (approved):
﻿### Summary
This PR fixes an integer overflow bug in the `PopReviveService.genSortList()` method. The original code cast the difference of two `long` offsets to `int`, which could overflow for large offset values. The fix correctly uses `Long.compare()` to compare the offsets.

The change is minimal, targeted, and includes a test case that verifies the fix works for offsets beyond the Integer range.

**Verdict:** LGTM — correct fix for a real overflow bug.

---

PR #10856, opened 2026-08-08
[ISSUE #10851] fix(broker): avoid cold threshold sort overflow

yuluo-yx:
### Which Issue(s) This PR Fixes

- Fixes #10851

### Brief Description

Uses Long.compare when sorting cold-read thresholds so the highest eligible groups are not skipped when values differ beyond the int range. Adds a regression test covering the three-group deceleration limit.

### How Did You Test This Change?

`mise x java@temurin-17 -- mvn -Dmaven.repo.local=/private/tmp/rocketmq-m2 -Djacoco.skip=true -pl broker -am -DskipITs -Dtest=ColdDataCgCtrServiceTest -Dsurefire.failIfNoSpecifiedTests=false test`

The targeted tests, Checkstyle, and SpotBugs completed successfully.

PR #11201, opened 2026-09-22
[ISSUE #10851] fix(broker): avoid cold threshold sort overflow

### LINKED ISSUE #10851: Cold data control may skip the group with the highest threshold
### Description

ColdDataCgCtrService sorts configured thresholds by subtracting two long values and narrowing the result to int. A large threshold delta can overflow the comparator and place the largest threshold after smaller entries. Because only three groups are decelerated per cycle, the group with the highest threshold can be skipped.

### Expected behavior

Sort thresholds with Long.compare so the three largest eligible groups are always processed.

beautyarbutin:

- Fixes #10851

### Brief Description

`ColdDataCgCtrService.sortAndDecelerate()` previously sorted configured thresholds by subtracting two `long` values and narrowing the result to `int`. When the difference exceeded the integer range, the comparator could move the largest threshold behind smaller entries. Because each cycle decelerates only three groups, the highest-threshold group could be skipped entirely.

This change uses `Long.compare` while preserving the existing descending order and three-group limit. The regression test exercises the actual deceleration selection with four adaptive thresholds and verifies that `Long.MAX_VALUE` is selected while the lowest threshold is excluded.

This reimplements the fix from the earlier, closed PR #10856 against the current `develop` branch. Thanks to @yuluo-yx for the original report and patch.

- Red-first regression: before the fix, the service decelerated thresholds `2`, `1`, and `0`, but skipped the `Long.MAX_VALUE` group.
- `mvn -B -ntp -pl broker -DskipITs -Dtest=ColdDataCgCtrServiceTest test`
- Result: 2 tests passed, with no failures, errors, or skips.
- Checkstyle: 0 violations.
- SpotBugs: 0 bugs and 0 errors.


RockteMQ-AI (approved):
### Summary
Classic comparator overflow fix. The old `(int)(o2.getValue() - o1.getValue())` pattern could overflow when the difference exceeds Integer.MAX_VALUE, causing incorrect sort order. Using `Long.compare` is the correct approach. The regression test specifically exercises the overflow scenario with Long.MAX_VALUE, ensuring the highest threshold is correctly processed.

LGTM.

---

